PR #4, opened 2024-10-17
Fix longest common substring implementation

rlrs:
This should fix https://github.com/unslothai/unsloth/issues/1015

The issue was that the existing LCS implementation would find a leading comma, leading it to call `int('')`.
Since I was performing this fix anyways, I took the liberty of implementing it a bit cleaner. Instead of finding the LCS of a string repr of a list of tokens, we just find the least common sublist of tokens directly. The string/int conversions are not necessary anymore.

I also added a very simple sanity test at the bottom of the file.

### DIFF
--- a/unsloth_zoo/dataset_utils.py
+++ b/unsloth_zoo/dataset_utils.py
@@ -1,6 +1,4 @@
 # Unsloth Zoo
-# Copyright (C) 2024-present the Unsloth AI team. All rights reserved.
-
 # This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


danielhanchen:
Thanks a lot - will take a look!!

danielhanchen:
@rlrs Thanks a lot! This definitely is much better than my previous impl :) I noticed that adding `substring = [int(x) for x in substring if x.isdigit()]` would have solved the issue, but your impl definitely doesn't need to perform in the "string" space (ie via the longest common substring).

I left the old impl just in case I might have to refer back to it!

PR #5, opened 2024-10-18
Fix for save_steps, save_strategy, and resume_from_checkpoint

TheTsar8756:
Currently this seems to fix unsloth_train so that it can save checkpoints and use save_strategy. A function to allow resume_from_checkpoint is implemented in the code but doesn't work. After testing it on my machine it saves the checkpoints but the training_state json is empty.

Also thank you so much for making Unsloth!


TheTsar8756:
Since the gradient accumulation fix was added into huggingface transformers there is no need for this pr.

PR #18, opened 2024-11-24
Update type checking in vision_utils.py

coding-famer:
`|` syntax is only supported in python 3.10 or later, but unsloth now still support python 3.9. 
So this PR changes it to traditional type checking.


danielhanchen:
Oh thanks! Did not notice this :)

PR #25, opened 2024-12-16
Add imports for base64 and BytesIO in vision_utils.py

whoisltd:
at function `fetch_image` you guys code support read image from bytes, base64 but not import lib to load images.
```python
    elif image.startswith("data:image"):
        if "base64," in image:
            _, base64_data = image.split("base64,", 1)
            data = base64.b64decode(base64_data)
            image_obj = Image.open(BytesIO(data))
```


danielhanchen:
@whoisltd Thanks so much!

PR #31, opened 2024-12-30
upload .gitignore

developer0hye:
We should make sure adding a .gitignore file for clean project.


danielhanchen:
main already has a .gitignore with __pycache__/ in it, so I do not think this is needed any more. Sorry for the slow review, and thanks for putting it together.

danielhanchen:
main already has a .gitignore with `__pycache__/` on the first line, so I do not think this is needed any more. Sorry for the slow review, and thanks for putting it together.

PR #49, opened 2025-02-19
Add VLM support to train_on_responses_only

AdrianoWeid:
Added `is_vlm` and `image_token_id` parameters to `train_on_responses_only` to support Vision-Language Models. When `is_vlm=True`, image tokens are preserved in the response part while masking instructions with `-100`.

### DIFF
--- a/unsloth_zoo/dataset_utils.py
+++ b/unsloth_zoo/dataset_utils.py
@@ -175,10 +175,12 @@ def train_on_responses_only(
     trainer,
     instruction_part = None,
     response_part    = None,
+    is_vlm           = False,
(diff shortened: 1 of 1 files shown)


danielhanchen:
Thanks for the PR! Apologies on the delay - is it possible to auto detect if it's an image model via the `trainer`?

danielhanchen:
Also would it be possible for a screenshot of whether this is functional and also some basic test code thanks - no need for pytest, such a standalone function so I can run it! I'm assuming this is for the bounty?

coding-famer:
I don't think any VL dataset has image tokens in the response part.

PR #50, opened 2025-02-20
VLMs image resizing

Bezdarnost:
Added a convenient option to set the maximum number of pixels per image during training. You can do something like this:

```
from unsloth import is_bf16_supported
from unsloth.trainer import UnslothVisionDataCollator
from trl import SFTTrainer, SFTConfig

FastVisionModel.for_training(model)  # Enable for training!

trainer = SFTTrainer(
    model=model,
    tokenizer=tokenizer,
    data_collator=UnslothVisionDataCollator(
        model,  # Must use!
        tokenizer, 
        max_image_size=16384 * 28 * 28  # Reduce if OOM occurs
    ),
    train_dataset=converted_dataset,
    ...

```
This helps prevent OOM issues when dealing with large images in the dataset.

### DIFF
--- a/unsloth_zoo/vision_utils.py
+++ b/unsloth_zoo/vision_utils.py
@@ -161,7 +161,6 @@ def fetch_image(
             max_pixels=max_pixels,
         )
     image = image.resize((resized_width, resized_height))
-    print(image.width)
     return image
 pass
 
--- a/unsloth_zoo/vision_utils.py
+++ b/unsloth_zoo/vision_utils.py
@@ -114,7 +114,11 @@ def smart_resize(
 pass
 
 
-def fetch_image(ele: dict[Union[Tuple[str, str], Image.Image]], size_factor: int = IMAGE_FACTOR) -> Image.Image:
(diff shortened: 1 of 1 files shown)


danielhanchen:
Hi @Bezdarnost, thanks for the contribution. Closing this as superseded.

`UnslothVisionDataCollator` already accepts a `resize` argument in current `main` (see `unsloth_zoo/vision_utils.py` `__init__`), which supports `"min"`, `"max"`, an `int`, or a `(h, w)` tuple, plus a `resize_dimension` selector. That covers the max-image-size cap your PR was adding, and it has been the canonical API since `unslothai/unsloth#4287`.

The user-facing knob is now exposed in Unsloth Studio via `unslothai/unsloth#5743` ("Studio: expose image size setting in training UI"), which wires the existing `resize` kwarg through the Studio UI / YAML / API.

If there is a specific case the existing `resize` doesn't cover (e.g. preserving aspect via smart_resize's grid alignment), please open a focused issue and we'll look at it.

PR #52, opened 2025-02-20
Added Image Resizing

MayankYadav39:


danielhanchen:
Hi @MayankYadav39, thanks for the contribution. Closing this as superseded.

`UnslothVisionDataCollator` already accepts a `resize` argument in current `main` (see `unsloth_zoo/vision_utils.py` `__init__`), which supports `"min"`, `"max"`, an `int`, or a `(h, w)` tuple, plus a `resize_dimension` selector. That covers the max-image-size cap your PR was adding via a parallel `max_image_size` parameter, and the canonical name has been `resize` since `unslothai/unsloth#4287`.

The user-facing knob is now exposed in Unsloth Studio via `unslothai/unsloth#5743` ("Studio: expose image size setting in training UI"), which wires the existing `resize` kwarg through the Studio UI / YAML / API.

If you ran into a case the existing `resize` doesn't cover (e.g. the extra PIL resize step your PR added after smart_resize), please open a focused issue and we'll look at it.

PR #54, opened 2025-02-21
Update vision_utils.py

NinoRisteski:
my try for VLMs Data Collator

### DIFF
--- a/unsloth_zoo/vision_utils.py
+++ b/unsloth_zoo/vision_utils.py
@@ -255,41 +255,53 @@ class UnslothVisionDataCollator:
         self.processor = processor
         self.formatting_func = formatting_func
         return
-    pass
 
     def __call__(self, examples):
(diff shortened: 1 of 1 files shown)


danielhanchen:
main now handles text-only and mixed batches in the collator in unsloth_zoo/vision_utils.py, where images are only passed to the processor when a batch actually has any, so I do not think this is needed any more. Sorry for the slow review, and thanks for putting it together.

danielhanchen:
The collator in vision_utils.py has since been rewritten and now only passes images when a batch actually has them, so I do not think this is needed any more. Sorry for the very slow review, and thanks for putting it together.

danielhanchen:
Main now handles text-only batches in the vision collator in unsloth_zoo/vision_utils.py, where images are only passed to the processor when there are some, so I do not think this is needed any more. Sorry for the slow review, and thanks for putting it together.

PR #55, opened 2025-02-21
update: Update VLMs Data Collator

rs545837:
Addressing first part of this issue [#1559](https://github.com/unslothai/unsloth/issues/1559)

Make text & image mixing work efficiently -so some inputs can be text only. Must work on Qwen, Llama, Pixtral.

Works with all the models mentioned above.

### DIFF
--- a/unsloth_zoo/vision_utils.py
+++ b/unsloth_zoo/vision_utils.py
@@ -262,6 +262,7 @@ class UnslothVisionDataCollator:
         # The issue is batch = self.processor( forces tensors to be returned and not None.
         texts  = []
         images = []
+        has_images = False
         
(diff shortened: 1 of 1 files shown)


danielhanchen:
Main now handles mixed text and image batches in unsloth_zoo/vision_utils.py, including the guarded pixel_values cast, so I do not think this is needed any more. Sorry for the slow review, and thanks for putting it together.

danielhanchen:
main now handles this in unsloth_zoo/vision_utils.py, where images are only passed to the processor when present and the pixel_values cast is guarded, so I do not think this is needed any more. Sorry for the slow review, and thanks for putting it together.

PR #56, opened 2025-02-22
fixed UnslothVisionDataCollator

KareemMusleh:
I've tested this on text only text-only-and-image-mix text+image


danielhanchen:
main now does this in unsloth_zoo/vision_utils.py, where empty images are skipped before the processor call, so I do not think this is needed any more. Sorry for the slow review, and thanks for putting it together.

danielhanchen:
main now skips empty images in the collator in unsloth_zoo/vision_utils.py, where it only appends when the example has images and only passes images to the processor when the batch has any, so I do not think this is needed any more. Sorry for the slow review, and thanks for putting it together.

PR #57, opened 2025-02-24
Support `image_url` with the `url` field

mehmetoguzderin:
First of all, thank you very much for `unsloth` and the utilities in `unsloth-zoo` as they help a lot.

Increasingly, JSONL or otherwise datasets have a `url` field rather than simply providing an `image_url` directly as a URL (probably due to tutorials related to OpenAI API) for vision. This PR minimally changes the type annotation and the code to support when `image_url` is a dictionary rather than a URL string.

I tested this with various data, including WebP images encoded as base64.

Reference tutorial with many such examples: https://platform.openai.com/docs/guides/vision

Thank you for your consideration! If there are any changes that would benefit from adjustment, please let me know, and I can update the PR. Kind regards. - Oguz

### DIFF
--- a/unsloth_zoo/vision_utils.py
+++ b/unsloth_zoo/vision_utils.py
@@ -114,11 +114,15 @@ def smart_resize(
 pass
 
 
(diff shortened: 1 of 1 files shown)


danielhanchen:
Yep this looks good! Thanks!

PR #59, opened 2025-03-01
Add vlm support in `train_on_responses_only`

patel-zeel:
Edited `train_on_responses_only` function to support VLMs.

Test Colab: https://colab.research.google.com/drive/18PkC1dYETTdbGzBj6GwUW5cv3T3NmGYf?usp=sharing

Main changes:
- Vision model detection by checking the following VLM-specific keywards in names of `trainer.model.named_modules()`
    - `visual`: Qwen2-VL models
    - `vision_tower`: Llava/Pixtral (Mistral AI) models
    - `vision_model`: Llama models
- `trainer.train_dataset` is a list of messages. Remove a content entry if "role" is "user" and "type" is "text".

Input Example:
```py
{'messages': [{'role': 'user',
   'content': [{'type': 'text',
     'text': 'Write the LaTeX representation for this image.'},
    {'type': 'image',
     'image': <PIL.PngImagePlugin.PngImageFile image mode=RGB size=160x40>}]},
  {'role': 'assistant',
   'content': [{'type': 'text',
     'text': '{ \\frac { N } { M } } \\in { \\bf Z } , { \\frac { M } { P } } \\in { \\bf Z } , { \\frac { P } { Q } } \\in { \\bf Z }'}]}]}
```

Output Example:
```py
{'messages': [{'role': 'user',
   'content': [{'type': 'image',
     'image': <PIL.PngImagePlugin.PngImageFile image mode=RGB size=160x40>}]},
  {'role': 'assistant',
   'content': [{'type': 'text',
     'text': '{ \\frac { N } { M } } \\in { \\bf Z } , { \\frac { M } { P } } \\in { \\bf Z } , { \\frac { P } { Q } } \\in { \\bf Z }'}]}]}
```

### DIFF
--- a/unsloth_zoo/dataset_utils.py
+++ b/unsloth_zoo/dataset_utils.py
@@ -179,8 +179,8 @@ def train_on_responses_only(
     """
     Trains only on responses and not on the instruction by masking out
     the labels with -100 for the instruction part for Language models
-    and by removing the "text" key from message if "role" is "user" for
-    Vision models.
+    and by removing content entries from messages with "role": "user" 
+    and "type": "text" for Vision models.
     """
     # All Unsloth Zoo code licensed under LGPLv3
     
--- a/unsloth_zoo/dataset_utils.py
+++ b/unsloth_zoo/dataset_utils.py
@@ -178,12 +178,28 @@ def train_on_responses_only(
 ):
     """
     Trains only on responses and not on the instruction by masking out
-    the labels with -100 for the instruction part.
+    the labels with -100 for the instruction part for Language models
+    and by removing the "text" key from message if "role" is "user" for
+    Vision models.
     """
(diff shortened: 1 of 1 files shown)


patel-zeel:
Sorry, I realized my solution is wrong because my conceptual understanding of "training on responses only" was wrong. Now, my updated understanding is that,
1. If no collator is passed, the default collator `DataCollatorForLanguageModeling` used in SFTTrainer creates "labels" identical to "input_ids".
2. As soon as the trainer is instantiated, data is tokenized, and `input_ids` are created.
3. `train_on_responses_only` function post-processes `trainer.train_dataset` and `trainer.eval_dataset` to mask out tokens for "instruction_part" because we want LLM to reconstruct answers correctly and ignore reconstruction loss for the questions.

I see that #60 is on the right track.

PR #60, opened 2025-03-01
Fix VLM train_on_response_only

coding-famer:
Make `train_on_response_only` compatible with VLM. Fix unslothai/unsloth#1396

In VLM, tokenization occurs within the `data_collator` rather than in `SFTTrainer._prepare_dataset`. To ensure consistency in the `train_on_response_only` interface, I modified the trainer’s `data_collator` to handle label construction.

### DIFF
--- a/unsloth_zoo/dataset_utils.py
+++ b/unsloth_zoo/dataset_utils.py
@@ -331,7 +331,7 @@ def train_on_responses_only(
                 batch = self.collator(examples)
                 batch["labels"] = self.modifier_fn(batch)["labels"]
                 return batch
-        if hasattr(trainer, "data_collator") and hasattr(trainer.data_collator, "collator"):
+        if hasattr(trainer, "data_collator") and not hasattr(trainer.data_collator, "collator"):
             trainer.data_collator = UnslothResponseOnlyCollator(trainer.data_collator, _train_on_responses_only)
         else:
             pass
--- a/unsloth_zoo/dataset_utils.py
+++ b/unsloth_zoo/dataset_utils.py
@@ -317,7 +317,7 @@ def train_on_responses_only(
             old_collator = trainer.data_collator
             def __call__(self, examples):
                 batch = self.old_collator(examples)
-                batch["labels"] = _train_on_responses_only(batch)
+                batch["labels"] = _train_on_responses_only(batch)["labels"]
                 return batch
         trainer.data_collator = data_collator()
 
--- a/unsloth_zoo/dataset_utils.py
+++ b/unsloth_zoo/dataset_utils.py
@@ -301,16 +301,6 @@ def train_on_responses_only(
     from multiprocessing import cpu_count
     num_proc = cpu_count()
 
-    # Edit data collator if not DataCollatorForSeq2Seq or UnslothVisionDataCollator, then apply _train_on_responses_only
-    from transformers import DataCollatorForSeq2Seq
-    from unsloth_zoo.vision_utils import UnslothVisionDataCollator
-    if hasattr(trainer, "data_collator"):
-        if not isinstance(trainer.data_collator, DataCollatorForSeq2Seq) and not hasattr(trainer.data_collator, "tokenizer"):
-            trainer.data_collator = DataCollatorForSeq2Seq(tokenizer = tokenizer)
-        elif not isinstance(trainer.data_collator, UnslothVisionDataCollator) and hasattr(trainer.data_collator, "tokenizer"):
-            trainer.data_collator = UnslothVisionDataCollator(model = trainer.model, tokenizer = tokenizer)
-        else:
-            pass
 
     if has_tokenized:
         if hasattr(trainer, "train_dataset") and trainer.train_dataset is not None:
@@ -326,8 +316,14 @@ def train_on_responses_only(
                 trainer.eval_dataset = trainer.eval_dataset.map(_train_on_responses_only, batched = True, num_proc = num_proc)
             pass
         pass
+
+        # Edit data collator as well if not DataCollatorForSeq2Seq
+        from transformers import DataCollatorForSeq2Seq
+        if hasattr(trainer, "data_collator") and \
+            not isinstance(trainer.data_collator, DataCollatorForSeq2Seq):
+            trainer.data_collator = DataCollatorForSeq2Seq(tokenizer = tokenizer)
     else:
-        class CustomDataCollator:
+        class UnslothResponseOnlyCollator:
             def __init__(self, collator, modifier_fn):
                 self.collator = collator
                 self.modifier_fn = modifier_fn
@@ -335,8 +331,8 @@ def train_on_responses_only(
                 batch = self.collator(examples)
                 batch["labels"] = self.modifier_fn(batch)["labels"]
                 return batch
-        if hasattr(trainer, "data_collator") and not isinstance(trainer.data_collator, CustomDataCollator):
-            trainer.data_collator = CustomDataCollator(trainer.data_collator, _train_on_responses_only)
+        if hasattr(trainer, "data_collator") and not isinstance(trainer.data_collator, UnslothResponseOnlyCollator):
+            trainer.data_collator = UnslothResponseOnlyCollator(trainer.data_collator, _train_on_responses_only)
         else:
             pass
     # Check if all labels randomnly got masked to nothing - maybe wrong chat template?
--- a/unsloth_zoo/dataset_utils.py
+++ b/unsloth_zoo/dataset_utils.py
@@ -182,6 +182,7 @@ def train_on_responses_only(
     """
     # All Unsloth Zoo code licensed under LGPLv3
     tokenizer = trainer.processing_class if hasattr(trainer, "processing_class") else trainer.tokenizer
+    has_tokenized = not trainer.args.dataset_kwargs["skip_prepare_dataset"]
     
     if  not hasattr(tokenizer, "_unsloth_input_part") or \
         not hasattr(tokenizer, "_unsloth_output_part"):
@@ -218,6 +219,8 @@ def train_on_responses_only(
         all_labels = []
 
         for input_ids in input_ids_:
+            if not has_tokenized:
+                input_ids = input_ids.tolist()
             n = len(input_ids)
             labels = [-100] * n
             n_minus_1 = n - 1
@@ -285,26 +288,39 @@ def train_on_responses_only(
             pass
             all_labels.append(labels)
         pass
+        if not has_tokenized:
+            import torch
+            all_labels = torch.tensor(all_labels)
         return { "labels" : all_labels }
     pass
 
-    if hasattr(trainer, "train_dataset") and trainer.train_dataset is not None:
-        trainer.train_dataset = trainer.train_dataset.map(_train_on_responses_only, batched = True)
-    pass
-    
-    if hasattr(trainer, "eval_dataset")  and trainer.eval_dataset  is not None:
-        # Eval datasets could be a dict!
-        if type(trainer.eval_dataset) is dict:
-            for key, value in trainer.eval_dataset.items():
-                trainer.eval_dataset[key] = value.map(_train_on_responses_only, batched = True)
-        else:
-            trainer.eval_dataset = trainer.eval_dataset.map(_train_on_responses_only, batched = True)
+    if has_tokenized:
+        if hasattr(trainer, "train_dataset") and trainer.train_dataset is not None:
+            trainer.train_dataset = trainer.train_dataset.map(_train_on_responses_only, batched = True)
         pass
-    pass
+        
+        if hasattr(trainer, "eval_dataset")  and trainer.eval_dataset  is not None:
+            # Eval datasets could be a dict!
+            if type(trainer.eval_dataset) is dict:
+                for key, value in trainer.eval_dataset.items():
+                    trainer.eval_dataset[key] = value.map(_train_on_responses_only, batched = True)
+            else:
+                trainer.eval_dataset = trainer.eval_dataset.map(_train_on_responses_only, batched = True)
+            pass
+        pass
 
-    # Check if all labels randomnly got masked to nothing - maybe wrong chat template?
-    from .training_utils import fix_zero_training_loss
-    fix_zero_training_loss(None, tokenizer, trainer.train_dataset)
+        # Check if all labels randomnly got masked to nothing - maybe wrong chat template?
+        from .training_utils import fix_zero_training_loss
+        fix_zero_training_loss(None, tokenizer, trainer.train_dataset)
+    else:
+        class data_collator:
+            old_collator = trainer.data_collator
+            def __call__(self, examples):
+                batch = self.old_collator(examples)
+                batch["labels"] = _train_on_responses_only(batch)
+                return batch
+        trainer.data_collator = data_collator()
+
     return trainer
 pass
 
--- a/unsloth_zoo/dataset_utils.py
+++ b/unsloth_zoo/dataset_utils.py
@@ -182,7 +182,10 @@ def train_on_responses_only(
     """
     # All Unsloth Zoo code licensed under LGPLv3
     tokenizer = trainer.processing_class if hasattr(trainer, "processing_class") else trainer.tokenizer
+    # if input haven't been tokenized by the trainer
     has_tokenized = not trainer.args.dataset_kwargs["skip_prepare_dataset"]
+    # for vlms, get the text tokenizer
+    tokenizer = tokenizer.tokenizer if hasattr(tokenizer, "tokenizer") else tokenizer
     
     if  not hasattr(tokenizer, "_unsloth_input_part") or \
         not hasattr(tokenizer, "_unsloth_output_part"):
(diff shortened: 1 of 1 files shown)


patel-zeel:
Hi @coding-famer! This PR has most of the changes and will be completed with the following:
- Since the vision tokenizer has image_processor as well, `_find_common_token_ids` takes `tokenizer.tokenizer` instead of `tokenizer`.
- `batch["labels"] = _train_on_responses_only(batch)` -> `batch["labels"] = _train_on_responses_only(batch)["labels"]` not to create nested dictionary with key "labels"

[This colab](https://colab.research.google.com/drive/18PkC1dYETTdbGzBj6GwUW5cv3T3NmGYf?usp=sharing) will run end-to-end once the above changes are done.

patel-zeel:
@coding-famer Could you please help me understand why we would pass image=None for this case? In normal cases, the following line would need a check for VLM and need to access the text tokenizer. 

https://github.com/unslothai/unsloth-zoo/blob/0389d45e2548c07df72e59491102dd1f001de293/unsloth_zoo/dataset_utils.py#L184

There might be bugs in the way I am using this feature, so I'd appreciate it if you could point to the required changes in [my colab](https://colab.research.google.com/drive/18PkC1dYETTdbGzBj6GwUW5cv3T3NmGYf?usp=sharing) as well.

coding-famer:
> @coding-famer Could you please help me understand why we would pass image=None for this case? In normal cases, the following line would need a check for VLM and need to access the text tokenizer explicitly with `tokenizer.tokenizer`.
> 
> https://github.com/unslothai/unsloth-zoo/blob/0389d45e2548c07df72e59491102dd1f001de293/unsloth_zoo/dataset_utils.py#L184
> 
> There might be bugs in the way I am using this feature, so I'd appreciate it if you could point to the required changes in [my colab](https://colab.research.google.com/drive/18PkC1dYETTdbGzBj6GwUW5cv3T3NmGYf?usp=sharing) as well.

@patel-zeel Sorry I mean here `processor(ids)` will have same output with `processor.tokenizer(ids)`, so this modification is unnecessary. 
And I think the colab notebook works correctly? Please LMK what error will occur if we don't set `tokenizer = tokenizer.tokenizer`

patel-zeel:
@coding-famer Oh, I got your point. Actually, the first argument to processor is `images`, and thus, my colab is running into an error. We need to make sure that text part goes to text argument but using keyword argument may break for some tokenizers?

coding-famer:
> @coding-famer Oh, I got your point and also found the issue. Actually, the first argument to processor is `images`, and thus, my colab is running into an error. We need to make sure that text part goes to text argument, but, using keyword argument may break for some tokenizers?

Oh yes you are right. I incorrectly set `tokenizer=processor.tokenizer` when I initialized the trainer(though this seems a solution for this issue but let's make it handle general cases). So what's the best way you think to solve it? Set tokenizer here for vlm?
https://github.com/unslothai/unsloth-zoo/blob/0389d45e2548c07df72e59491102dd1f001de293/unsloth_zoo/dataset_utils.py#L184

patel-zeel:
@coding-famer IMO, we can continue with your logic of using `has_tokenized` to detect VLM and create an if-else condition to use `processor(text=...)` explicitly when a VLM is detected. We can also add what we have discovered so far in comments to save some hours for future devs!

coding-famer:
@patel-zeel After thinking about this, I hope this pr can handle the general case: where inputs haven't been tokenized when init the trainer, instead of just handling vlm. So I think my `has_tokenized` flag is good.

patel-zeel:
Yes, I agree, @coding-famer. After solving the merge conflicts with @danielhanchen's new edits, this PR will be ready for review.

patel-zeel:
@coding-famer A small edit I thought about: Changing the `tokenizer` in place might create an issue in case a full tokenizer is needed later. How about something like the following to make it more explicit and readable?

```py
# Get the text tokenizer from the combined tokenizer
if not has_tokenized:
    text_tokenizer = tokenizer.tokenizer
else:
    text_tokenizer = tokenizer
pass

# Get most common tokens since tokenizers can tokenize stuff differently!
Q_must, Q_left, Q_right = _find_common_token_ids(instruction_part, text_tokenizer)
A_must, A_left, A_right = _find_common_token_ids(response_part,    text_tokenizer)
```

patel-zeel:
@coding-famer A few more thoughts:
1. I'm not sure what the following code does, but shall we keep it outside of the `if else` condition to apply it for VLMs, too? 

```py
# Check if all labels randomnly got masked to nothing - maybe wrong chat template?
from .training_utils import fix_zero_training_loss
fix_zero_training_loss(None, tokenizer, trainer.train_dataset)
```

2. I was thinking about what would happen if someone calls `train_on_response_only` multiple times (say executing the same cell of Jupyter notebook multiple times). Will it wrap one more time? I was thinking of something like this to avoid it:

```py
class CustomDataCollator:
    def __init__(self, collator, modifier_fn):
        self.collator = collator
        self.modifier_fn = modifier_fn

    def __call__(self, examples):
        batch = self.collator(examples)
        batch["labels"] = self.modifier_fn(batch)["labels"]
        return batch

from unsloth_zoo.vision_utils import UnslothVisionDataCollator
import warnings

if not hasattr(trainer, "data_collator"):
    warnings.warn(
        "Vision model detected, but data_collator is not set. Setting it to UnslothVisionDataCollator."
    )
    trainer.data_collator = UnslothVisionDataCollator(model=trainer.model, processor=tokenizer)
if not isinstance(trainer.data_collator, CustomDataCollator):
    trainer.data_collator = CustomDataCollator(trainer.data_collator, _train_on_responses_only)

pass
```

coding-famer:
@patel-zeel Good points! For 1 I'm not sure, too. Let me have a look at the logic of that function. I'll add 2.

unsloth_zoo/dataset_utils.py
```diff
@@ -285,35 +292,42 @@ def _train_on_responses_only(examples):
...
+            if type(trainer.eval_dataset) is dict:
+                for key, value in trainer.eval_dataset.items():
+                    trainer.eval_dataset[key] = value.map(_train_on_responses_only, batched = True, num_proc = num_proc)
+            else:
+                trainer.eval_dataset = trainer.eval_dataset.map(_train_on_responses_only, batched = True, num_proc = num_proc)
+            pass
         pass
-    pass
 
-    # Edit data collator as well if not DataCollatorForSeq2Seq
```
patel-zeel, line 365: I think this was part of the incoming changes, shall we restore it?

coding-famer: Yes, I will combine it with your previous suggestions.

patel-zeel: @coding-famer A point that created a bug for me: The `CustomDataCollator` class should be defined outside of the `train_on_response_only`; otherwise, `isinstance(trainer.data_collator, CustomDataCollator)` will be `False`.

unsloth_zoo/dataset_utils.py
```diff
@@ -285,35 +292,57 @@ def _train_on_responses_only(examples):
...
-    pass
-
-    # Edit data collator as well if not DataCollatorForSeq2Seq
+    # Edit data collator if not DataCollatorForSeq2Seq or UnslothVisionDataCollator, then apply _train_on_responses_only
     from transformers import DataCollatorForSeq2Seq
-    if hasattr(trainer, "data_collator") and \
-        not isinstance(trainer.data_collator, DataCollatorForSeq2Seq):
-        trainer.data_collator = DataCollatorForSeq2Seq(tokenizer = tokenizer)
+    from unsloth_zoo.vision_utils import UnslothVisionDataCollator
+    if hasattr(trainer, "data_collator"):
```
coding-famer: Do we need to ensure the trainer has a `data_collator`?

patel-zeel: If people follow Unsloth Notebooks, it's unlikely to be missed, but I thought it's not hurting existing functionality. Also, it attaches the collator with a warning. In case an advanced user does not want this behavior, they can modify it after applying `train_on_response_only`. 

This was mainly inspired by the recent `DataCollatorForSeq2Seq` change a few lines above.

unsloth_zoo/dataset_utils.py
```diff
@@ -285,35 +292,57 @@ def _train_on_responses_only(examples):
...
+            # Eval datasets could be a dict!
+            if type(trainer.eval_dataset) is dict:
+                for key, value in trainer.eval_dataset.items():
+                    trainer.eval_dataset[key] = value.map(_train_on_responses_only, batched = True, num_proc = num_proc)
+            else:
+                trainer.eval_dataset = trainer.eval_dataset.map(_train_on_responses_only, batched = True, num_proc = num_proc)
+            pass
+        pass
+    else:
+        class CustomDataCollator:
```
patel-zeel: @coding-famer Shall we move this class outside the function to ensure its working?

unsloth_zoo/dataset_utils.py
coding-famer: How about leaving this for @danielhanchen to review? 
I mean I didn't add your
```
if not hasattr(trainer, "data_collator"):
    warnings.warn(
        "Vision model detected, but data_collator is not set. Setting it to UnslothVisionDataCollator."
    )
    trainer.data_collator = UnslothVisionDataCollator(model=trainer.model, processor=tokenizer)
```
part. And the recent `DataCollatorForSeq2Seq` change doesn't ensure the `trainer` has `data_collator`, too.
It's good to add this but the old logic doesn't do this (or maybe it is better to do it outside the `train_on_response_only` cus this is not quite related).

unsloth_zoo/dataset_utils.py
coding-famer: It will only be used inside the `else` I think? 
But one more thing here, what if user pass their own custom datacollator that is also named `CustomDataCollator`?

unsloth_zoo/dataset_utils.py
patel-zeel: Yes, I agree. It's not directly related and not having it is not hurting anything.

unsloth_zoo/dataset_utils.py
patel-zeel: > It will only be used inside the else I think?

Right, but somehow, keeping it inside the function wasn't working when I tested it.

> But one more thing here, what if user pass their own custom datacollator that is also named CustomDataCollator

Technically, it'll still work because we are not string matching, and thus, it'll try to match with our custom class only and will fail. But, `CustomDataCollator` does seem like a catchy and generic name. How about something like `UnslothResponseOnlyCollator`?

unsloth_zoo/dataset_utils.py
```diff
@@ -285,35 +292,57 @@ def _train_on_responses_only(examples):
...
-
-    # Edit data collator as well if not DataCollatorForSeq2Seq
+    # Edit data collator if not DataCollatorForSeq2Seq or UnslothVisionDataCollator, then apply _train_on_responses_only
     from transformers import DataCollatorForSeq2Seq
-    if hasattr(trainer, "data_collator") and \
-        not isinstance(trainer.data_collator, DataCollatorForSeq2Seq):
-        trainer.data_collator = DataCollatorForSeq2Seq(tokenizer = tokenizer)
+    from unsloth_zoo.vision_utils import UnslothVisionDataCollator
+    if hasattr(trainer, "data_collator"):
+        if not isinstance(trainer.data_collator, DataCollatorForSeq2Seq) and not hasattr(trainer.data_collator, "tokenizer"):
```
patel-zeel: @coding-famer Oh, the `DataCollatorForSeq2Seq` related lines would go inside the `if has_tokenized` because they're for language models.

coding-famer: I think this is for both llm and vlm? Or we don't need to do this for vlm?
For `has_tokenized` and `hasattr(tokenizer, "tokenizer")`, I'm trying to separate the cases that input hasn't been tokenized and it is for vlm.

mikeknapp: The problem with this code - I think - is it that it will prefer `DataCollatorForSeq2Seq` over `UnslothVisionDataCollator`. I had to comment it out. 

Also, the current code doesn't seem to support a dataset where there is a mixture of text only examples and text + image examples - or am I missing something?

unsloth_zoo/dataset_utils.py
coding-famer: updated. It wasn't working because I made a mistake. Can you please test on your side again to see if it works now? If it can work, it's better to put it inside the `else` I think.

unsloth_zoo/dataset_utils.py
coding-famer: Got it. Moved this into `if has_tokenized`. Yes I believe there is another issue related to this. The difficulty for that is the processor must have an image input.

unsloth_zoo/dataset_utils.py
patel-zeel: @coding-famer I tested again and understood the problem. Actually, everytime the `train_on_response_only` function reruns, class `CustomDataCollator` is redefined and thus the object which is instantiated in the previous run will not be an instance of the class defined in the current run, thus we need to move `CustomDataCollator` outside the `train_on_response_only` to make the class definition constant with respect to it.

coding-famer: Do you think 
```python
if hasattr(trainer.data_collator, "collator"):
...
```
a good alternative solution for this?

patel-zeel: @coding-famer Yes, that should work I think.

patel-zeel:
@coding-famer This looks mostly good with one caveat: If someone changes `instruction_part` and `response_part` and reapplies the function, it'll silently execute and will have no effect because we will simply check if `trainer.data_collator` already has a `collator` and will return the same object. One possible solution is:

```py
if hasattr(trainer, "data_collator"):
    # If `UnslothResponseOnlyCollator` is found, extract internal collator to avoid double wrapping
    if hasattr(trainer.data_collator, "collator"):
        trainer.data_collator = trainer.data_collator.collator
    trainer.data_collator = UnslothResponseOnlyCollator(trainer.data_collator, _train_on_responses_only)
```

patel-zeel:
Amazing @coding-famer!

Hi @danielhanchen, this PR is ready for review. [This colab](https://colab.research.google.com/drive/18PkC1dYETTdbGzBj6GwUW5cv3T3NmGYf?usp=sharing) is for testing this feature.

mikeknapp:
@coding-famer @patel-zeel I believe we can get text_only and image examples to work correctly if we do something like this: 

```python
class UnslothResponseOnlyCollator:
  def __init__(self, collator, modifier_fn):
      self.collator = collator
      self.modifier_fn = modifier_fn
  def __call__(self, examples):
      image_examples = []
      text_only_examples = []
      for example in examples:
          msgs = example.get("messages", [])
          if any(msg.get("type") == "image" for msg in msgs):
              image_examples.append(example)
          else:
              text_only_examples.append(example)
      batch = {}
      if len(image_examples) > 0:
          image_batch = self.collator(image_examples)
          image_batch["labels"] = self.modifier_fn(image_batch)["labels"]
          batch = image_batch
      if len(text_only_examples) > 0:
          text_batch = self.collator(text_only_examples)
          text_batch["labels"] = self.modifier_fn(text_batch)["labels"]
          batch = {**batch, **text_batch}
      return batch
```

This probably just needs some massaging to deal with the case where people have a 2 column dataset, with the second column being an image.

coding-famer:
> @coding-famer @patel-zeel I believe we can get text_only and image examples to work correctly if we do something like this:
> 
> ```python
> class UnslothResponseOnlyCollator:
>   def __init__(self, collator, modifier_fn):
>       self.collator = collator
>       self.modifier_fn = modifier_fn
>   def __call__(self, examples):
>       image_examples = []
>       text_only_examples = []
>       for example in examples:
>           msgs = example.get("messages", [])
>           if any(msg.get("type") == "image" for msg in msgs):
>               image_examples.append(example)
>           else:
>               text_only_examples.append(example)
>       batch = {}
>       if len(image_examples) > 0:
>           image_batch = self.collator(image_examples)
>           image_batch["labels"] = self.modifier_fn(image_batch)["labels"]
>           batch = image_batch
>       if len(text_only_examples) > 0:
>           text_batch = self.collator(text_only_examples)
>           text_batch["labels"] = self.modifier_fn(text_batch)["labels"]
>           batch = {**batch, **text_batch}
>       return batch
> ```
> 
> This probably just needs some massaging to deal with the case where people have a 2 column dataset, with the second column being an image.

For me, I think the mixture of text only and text image case should be done in a new datacollator, and this issue is trying to handle the train_on_response_only case(imaging someone wants to train with mixture of text and text-image but doesn't want to train on response only). You can open a new PR and I'm willing to contribute to it, too.

danielhanchen:
Oh my wait - apologies I totally missed this

danielhanchen:
I actually I think accidentally in parallel did this during the release for Geamm 3: https://github.com/unslothai/unsloth-zoo/blob/main/unsloth_zoo/vision_utils.py#L259

danielhanchen:
I can accept the PR and provide the bounty for parts which are not in the current code base - the extra checks and names and etc

danielhanchen:
main now covers this in unsloth_zoo/vision_utils.py, where UnslothVisionDataCollator applies the response-only masking at collate time, plus the text tokenizer unwrap in dataset_utils.py, so I do not think this is needed any more. Sorry for the very slow review, and thanks to you and patel-zeel for working it through so carefully.

danielhanchen:
main now tokenizes VLM datasets before masking in unsloth_zoo/dataset_utils.py and routes vision masking through UnslothVisionDataCollator with train_on_responses_only set, so this branch no longer has a job. Sorry for the slow follow up, and thanks for working it through on the thread.

PR #75, opened 2025-03-14
Add support for multiple instruction patterns in train_on_responses_only

marcelodiaz558:
This PR enhances the `train_on_responses_only` utility function by allowing the `instruction_part` argument to accept a list of instruction patterns rather than a single string.

### Changes:
- Updated the function signature to accept either a string or a list of strings for the `instruction_part` parameter.
- Implemented logic to compute token patterns separately for each provided instruction pattern.
- Updated the internal matching logic within the helper function to correctly identify and process each instruction pattern.

### Benefits:
- **Flexibility**: Now supports datasets containing various instruction types within the same training run (e.g., tool calls, multiple chat roles).
- **Enhanced use cases:** Particularly beneficial when handling tool calls, system messages, or multiple user roles within the same training dataset.

### Usage Examples:
- For ChatML-style chat templates:
```python
instruction_part = ["\n<|im_start|>user\n", "\n<|im_start|>system\n"]
response_part = "\n<|im_start|>assistant\n"

trainer = train_on_responses_only(
    trainer,
    instruction_part = instruction_part,
    response_part = response_part,
)
```

- For advanced chat templates, like Llama-3.1/3.2/3.3 with tool usage:
```python
instruction_part = [
    "<|start_header_id|>user<|end_header_id|>\n\n",
    "<|start_header_id|>system<|end_header_id|>\n\n",
    "<|start_header_id|>ipython<|end_header_id|>\n\n"
]
response_part = "<|start_header_id|>assistant<|end_header_id|>\n\n"

trainer = train_on_responses_only(
    trainer,
    instruction_part = instruction_part,
    response_part = response_part,
)
```

### DIFF
--- a/unsloth_zoo/dataset_utils.py
+++ b/unsloth_zoo/dataset_utils.py
@@ -376,7 +376,7 @@ def train_on_responses_only(
        not isinstance(trainer.data_collator, DataCollatorForSeq2Seq):
         trainer.data_collator = DataCollatorForSeq2Seq(tokenizer=tokenizer)
 
-    from unsloth_zoo.training_utils import fix_zero_training_loss
+    from .training_utils import fix_zero_training_loss
     fix_zero_training_loss(None, tokenizer, trainer.train_dataset)
     return trainer
 pass
--- a/unsloth_zoo/dataset_utils.py
+++ b/unsloth_zoo/dataset_utils.py
@@ -192,42 +192,55 @@ def train_on_responses_only(
     """
     Trains only on responses and not on the instruction by masking out
     the labels with -100 for the instruction part.
+    Now supports a list of instruction patterns.
     """
     # All Unsloth Zoo code licensed under LGPLv3
     if tokenizer is None and trainer is not None:
         tokenizer = trainer.processing_class if hasattr(trainer, "processing_class") else trainer.tokenizer
-    # Get non vision tokenizer
+    # Get non-vision tokenizer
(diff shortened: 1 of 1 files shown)


TomieAi:
hmm yeah that make sense

danielhanchen:
The direction is right, and multiple instruction markers for system and tool turns is something we want, but this also restyles the whole function and changes the `"labels" in examples` branch, and `train_on_responses_only` in dataset_utils.py has been rewritten since. Could you rebase and narrow it to just accepting a list for `instruction_part`?

danielhanchen:
The multiple instruction markers idea is useful, but this branch also rewrites the surrounding loop and changes the labels handling in _train_on_responses_only so an existing labels column arriving as a list of lists is replaced by None instead of being copied over, which main now handles in unsloth_zoo/dataset_utils.py. Could you rebase and narrow it to just the list-of-instruction-patterns support so the rest of the function stays as it is?

danielhanchen:
The multi-role masking idea is useful, but this rewrites the whole matching loop and also changes the `"labels" in examples` branch so list-valued labels no longer go through the old-label path, and the surrounding function has moved on since March. Could you rebase and narrow it to just looping over a list of instruction patterns, leaving the single-string path byte-identical?

danielhanchen:
Main still takes a single instruction_part string in train_on_responses_only, so this is still useful, but the branch also reformats unrelated lines and changes how existing labels are read. Could you rebase onto the current function and narrow it to just accepting a list of instruction patterns?

PR #80, opened 2025-03-16
Save compile cache to only home directory

Datta0:
This is to make sure unsloth_compile_cache isn't being duplicated across multiple folders.


Datta0:
Interesting. But if patches don't conflict with other patches, maintaining a global cache is equivalent to maintaining a local cache (even from a debugging perspective)
So I sort of find local cache unnecessary... It creates lot of duplicate folders and hard to keep track of where is what...

PR #84, opened 2025-03-18
fixed a typo in L119, removing unnecessary len()

SpaceHunterInf:
Removing the unnecessary len() assertion that prevents model.load_lora().


danielhanchen:
Thanks!

PR #89, opened 2025-03-21
VLM Data Collator - Make text & image mixing work efficiently

mmathew23:
Currently finetuning VLM with mixed text and text/image rows will not work. The processing class complains if you pass non images, and even if you pass that the vision models requires full tenors for every batch. 

This PR adds a custom data sampler to solve for this. Some alternatives were forcing single batch training, sub batching the mixed batches, and passing dummy images.

Dummy images would force extra flops and masking is requires to make sure the dummy image doesn't influence the model. Not efficient and hard to maintain for new models with new masking strategies.

Single batch woudl work, but that feels to strict. And sub batching would require patching the VLM to loop over image vs non image sub batches. Could be viable but still not as efficient, and harder to extend to multi gpu cases.

Instead I opted for a data sampler to control the order of batches. It will yield batches with all images+text or only text. The model will entirely bypass the vision model so it's most efficient. [This colab notebook](https://colab.research.google.com/drive/1vPE8KY84E2SHaX_LM2DrmJahZB4e5pyj?usp=sharing) was tested for llama, qwen, and pixtral. 

We also note that setting truncation to True was causing training errors (reproducible with the unsloth vlm notebook) and that is no longer set. In order to pass a sampler I follow how Trainer's inner training loop get's patched, and patch training sampler to check for a now config argument, group_mixed_image_text.

### DIFF
--- a/unsloth_zoo/compiler.py
+++ b/unsloth_zoo/compiler.py
@@ -1000,7 +1000,7 @@ def apply_fused_lm_head(forward):
 
         cross_entropy_replacement = cross_entropy_replacement\
             .replace(
-                "$KWARGS$", 
+                "$KWARGS$",
                 "locals().get('loss_kwargs', {}) or locals().get('kwargs', {})"
             )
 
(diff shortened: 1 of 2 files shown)


mmathew23:
Also depends on unsloth [PR #2133](https://github.com/unslothai/unsloth/pull/2133)

rolandtannous:
People can still create their custom collator. I would moreover add that when you have a special case like this, you preprocess in the Dataset Class (People are not even creating dataset classes..)
But if people come to a consensus that going ahead with a collator to support mixed datasets is the right way, I suggest that a new MixedDataCollator class is created instead of trying to patch the VisionDataCollator itself to work with mixed datasets. That would be a slightly more elegant approach and a better separation of concerns approach.

mmathew23:
That's reasonable. 

I was working under the assumption that we should be fixing the current collator to support mixed datasets. But yea it would be good to get an opinion on which way to go.

PR #95, opened 2025-03-22
saving utils check both model.base_model and base_model.model

toranb:
Without this safe check, Phi 4 mini instruct blew up when I attempted to save it

```text
  File "/home/toran/code/sloth/sftune.py", line 104, in <module>
    model.save_pretrained_merged("model", tokenizer, save_method = "merged_16bit")
    ....
  File "/home/toran/code/sloth/.venv/lib/python3.11/site-packages/unsloth_zoo/saving_utils.py", line 318, in create_lora_statistics
    if return_state_dict: assert_same_keys(model, state_dict)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/toran/code/sloth/.venv/lib/python3.11/site-packages/unsloth_zoo/saving_utils.py", line 199, in assert_same_keys
    inner_model = model.base_model.model if hasattr(model, "base_model") else model
                  ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/toran/code/sloth/.venv/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1931, in __getattr__
    raise AttributeError(
AttributeError: 'Phi3Model' object has no attribute 'model'
```


danielhanchen:
Thanks for catching this and for the fix. The bug is real and still live: `assert_same_keys` does the unguarded `model.base_model.model` on current main, which is the exact crash site from your traceback (reached from `create_lora_statistics` on every merged_16bit save). One of the two spots this PR touched (`create_lora_statistics`) has since been refactored to resolve the base via a `find_lora_base_model` helper, so this PR no longer applies cleanly there, but the `assert_same_keys` site was never guarded.

I've opened #877, which applies the guard at that site using the shared `find_lora_base_model` helper (so the key comparison also stays consistent with how the state_dict is built), plus a regression test for the Phi-3 / Phi-4-mini layout. Closing this in favor of #877, which credits your report. Thanks again for digging in.

Issue #90, opened 2025-03-21
Mistral 3.1 bug


maziyarpanahi:
Hi,

When trying to load Mistral Small 3.1 new multimodal model (either bnb 4bit by unsloth or the original) it throws this error:

```python
from unsloth import FastModel
import torch

fourbit_models = [
    # 4bit dynamic quants for superior accuracy and low memory use
    "unsloth/gemma-3-1b-it-unsloth-bnb-4bit",
    "unsloth/gemma-3-4b-it-unsloth-bnb-4bit",
    "unsloth/gemma-3-12b-it-unsloth-bnb-4bit",
    "unsloth/gemma-3-27b-it-unsloth-bnb-4bit",

    "unsloth/Mistral-Small-3.1-24B-Instruct-2503-unsloth-bnb-4bit",
] # More models at https://huggingface.co/unsloth

model, tokenizer = FastModel.from_pretrained(
    model_name = "mistralai/Mistral-Small-3.1-24B-Instruct-2503",
#    model_name =  "unsloth/Mistral-Small-3.1-24B-Instruct-2503-unsloth-bnb-4bit", (this one too)
    max_seq_length = 2048, # Choose any for long context!
    load_in_4bit = False,  # 4 bit quantization to reduce memory
    load_in_8bit = False, # [NEW!] A bit more accurate, uses 2x memory
    full_finetuning = False, # [NEW!] We have full finetuning now!
    # token = "hf_...", # use one if using gated models
)
```

Here is the error:
```
AttributeError: module 'transformers.models.mistral3.modeling_mistral3' has no attribute 'logger'
```

Telsbat:
You can use this temporary workaround:

```python
# Before loading the model
import transformers.models.mistral3.modeling_mistral3 as modeling_mistral3
import logging

# Add the missing logger attribute
if not hasattr(modeling_mistral3, 'logger'):
    modeling_mistral3.logger = logging.getLogger('transformers.models.mistral3.modeling_mistral3')
```
But then, I have no idea why it sometimes WORKS and load model correctly, and sometimes DON'T WORK AT ALL, it just throws
```console
[...]

PR #97, opened 2025-03-24
Fix gradient checkpointing warning filter implementation

rolandtannous:
### Description

This PR fixes a bug in the `unsloth_compile_transformers` method in `compiler.py` that causes an AttributeError when trying to suppress gradient checkpointing warnings. The current implementation incorrectly assumes that the model file has a `logger` attribute, but logger instances are typically module-level variables, not attributes.

### Changes

- Replaced the problematic `exec()` call that was looking for `modeling_file.logger`

- Implemented a proper logging filter that gets the correct logger instance via the transformers logging system

- Ensures the filter works across different model architectures (Gemma, Mistral, etc.)

### Bug Details

The current code attempts to suppress warnings with:

```python

exec("modeling_file.logger.addFilter(HideLoggingMessage('Setting `use_cache=False`'))", globals(), locals())

```

This fails with 
```
AttributeError: module 'transformers.models.mistral3.modeling_mistral3' has no attribute 'logger'` because the logger is a module-level variable, not a model attribute.
```

### Solution

The fix gets the appropriate logger directly from the transformers logging module and applies a filter to target only the specific gradient checkpointing warning message.

### Testing

Verified the solution works with:

- Gemma 3 models

- Mistral 3 models

### Related Issues

Fixes:
- unsloth-zoo issue [#90](https://github.com/unslothai/unsloth-zoo/issues/90)
- unsloth issue [#2146](https://github.com/unslothai/unsloth/issues/2146)

### DIFF
--- a/unsloth_zoo/compiler.py
+++ b/unsloth_zoo/compiler.py
@@ -1000,7 +1000,7 @@ def apply_fused_lm_head(forward):
 
         cross_entropy_replacement = cross_entropy_replacement\
             .replace(
-                "$KWARGS$", 
+                "$KWARGS$",
                 "locals().get('loss_kwargs', {}) or locals().get('kwargs', {})"
             )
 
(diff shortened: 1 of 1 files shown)


rolandtannous:
This is a bug , resulting in a runtime error and actually preventing most users from even loading models using FastLanguageModel.from_pretrained. Example:

```
    model_name="meta-llama/Llama-3.3-70B-Instruct",
    max_seq_length=max_seq_length,
    dtype=None,
    load_in_4bit=False,
    token=hf_key,
)
```

results in the runtime error:
```
AttributeError: module 'transformers.models.mistral3.modeling_mistral3' has no attribute 'logger'
```

and prevents most people from progressing forward.

The underlying code in compiler.py throwing the runtime error isn't even trivial. It is  just trying to suppress a category of logs by filtering it out to un-clutter the console.

PR #98, opened 2025-03-24
Fixes to support IterableDataset

marcandrelarochelle:
Description

This PR fixes a bug when using a dataset of type IterableDataset in the `dataset_utils.py`, method `sft_prepare_dataset`.

Changes

Instead of using an index to access the `test_text`, we iterate to the first element using `iter`

Bug Details

IterableDataset do not have `__getitem__` implemented and when using an index to access the first element `dataset[0]`, it would throw an exception.

Solution

Replace `dataset[0]` with `next(iter(dataset))` to access the test_text

### DIFF
--- a/unsloth_zoo/dataset_utils.py
+++ b/unsloth_zoo/dataset_utils.py
@@ -572,6 +572,8 @@ def sft_prepare_dataset(
 
         if not isinstance(dataset, IterableDataset):
             map_kwargs["num_proc"] = getattr(args, "dataset_num_proc", 2)
+        else:
+            map_kwargs["batch_size"] = dataset.ex_iterable.batch_size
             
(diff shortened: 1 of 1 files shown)


unsloth_zoo/dataset_utils.py
```diff
@@ -570,7 +570,9 @@ def _tokenize(example):
             )
         pass
 
-        map_kwargs["num_proc"] = getattr(args, "dataset_num_proc", 2)
+        if not isinstance(dataset, IterableDataset):
+            map_kwargs["num_proc"] = getattr(args, "dataset_num_proc", 2)
```
marcandrelarochelle, line 583: Reason for this is the `IterableDataset` doesn't take any `num_proc` arguments

danielhanchen:
Appreciate it and thanks a lot!

PR #99, opened 2025-03-25
Fix bugs in generate_batches

5k5000:
Original `outputs= []` will result in duplication of returned outputs.


danielhanchen:
Nice catch - dumb of me not to notice!

5k5000:
Thanks, I really really love Unsloth.

PR #104, opened 2025-03-27
`num_proc=os.cpu_count()` when counting unused tokens

IsaacBreen:
This just sets num_proc=os.cpu_count() when counting unused tokens. Tested on a 64-core machine. Runs approximately 64 times faster.

This should be the default for all those HF map functions imho.

### DIFF
--- a/unsloth_zoo/tokenizer_utils.py
+++ b/unsloth_zoo/tokenizer_utils.py
@@ -417,7 +417,7 @@ def fix_untrained_tokens(model, tokenizer, train_dataset, IGNORED_TOKENIZER_NAME
         counter = np.fromiter(itertools.chain.from_iterable(input_ids), dtype = np.int32)
         np.add.at(final_counts, counter, 1)
     pass
-    train_dataset.map(mapping, batched = True, desc = "Counting untrained tokens", num_proc=os.cpu_count())
+    train_dataset.map(mapping, batched = True, desc = "Counting untrained tokens", num_proc = os.cpu_count())
 
     # Get sum of all items
     sum_embedding = torch.sum(embedding_matrix, dtype = torch.float32, axis = 0)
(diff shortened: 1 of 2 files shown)


IsaacBreen:
https://github.com/unslothai/unsloth/issues/1125

unsloth_zoo/tokenizer_utils.py
```diff
@@ -416,7 +417,7 @@ def mapping(examples):
         counter = np.fromiter(itertools.chain.from_iterable(input_ids), dtype = np.int32)
         np.add.at(final_counts, counter, 1)
     pass
-    train_dataset.map(mapping, batched = True, desc = "Counting untrained tokens")
+    train_dataset.map(mapping, batched = True, desc = "Counting untrained tokens", num_proc = os.cpu_count())
```
marcandrelarochelle, line 420: This will break `IterableDataset` supports as they do not support the `num_proc` argument.

mmathew23:
@IsaacBreen I'm unsure about why we need the change to loss_utils. Do you have examples where `labels` and `attention_mask` are keys that are explicitly set to None?

Would you be able to update similar to the comment I made here, https://github.com/unslothai/unsloth-zoo/pull/126#discussion_r2285980803? That way we can support both IterableDataset and fast unused token counting.

And apologies for the many notifications! Thanks again for contributing

danielhanchen:
The counting callback in unsloth_zoo/tokenizer_utils.py accumulates into a closed-over numpy array, so with num_proc the increments happen in worker processes and the parent array stays empty, and IterableDataset has no num_proc at all. Could you rework it to return counts from the map and gate num_proc on the dataset type as suggested on #126, and drop the loss_utils change unless you have a case where labels is explicitly None?

PR #105, opened 2025-03-27
Fix: SmolVLM indentation error in compiled module

rolandtannous:
## Issue
Resolves [unsloth issue #2179](https://github.com/unslothai/unsloth/issues/2179)
SmolVLM models cause an indentation error in generated compiled modules when used with Unsloth.

## Problem Description
When using SmolVLM models (specifically SmolVLM2 Video Instruct) with Unsloth, the compilation process generates invalid Python code with indentation errors. The error manifests as:

```
File "/home/hermeschen/Repo/taiwan-license-plate-recognition/unsloth_compiled_cache/unsloth_compiled_module_smolvlm.py", line 481
  = loss_fct(shift_logits, shift_labels)
IndentationError: unexpected indent
```

The generated file shows several lines with incorrect indentation and missing variable names:

```python
# Flatten the tokens
loss_fct = CrossEntropyLoss()
shift_logits = shift_logits.view(-1, self.config.text_config.vocab_size)
t_labels = shift_labels.view(-1)
t_labels = shift_labels.to(shift_logits.device)
= loss_fct(shift_logits, shift_labels)  # Note the missing variable name and improper indent
```
## Reproduction Steps
The issue can be reproduced with the latest versions of Unsloth and Unsloth Zoo using SmolVLM2 Video Instruct model. The error occurs consistently during the compilation process, before model inference.

## Root Cause
We discovered that this issue stems from how the loss calculation is implemented in the SmolVLM forward method, specifically the lack of explicit device management during loss calculation.

The compilation error occurs when torch.compile tries to optimize the forward pass. The absence of explicit device placement for tensors passed to the loss function creates complexity that the compiler cannot resolve correctly, resulting in malformed Python code.

## Solution
This PR patches the `SmolVLMForConditionalGeneration.forward` method to include an explicit device conversion when passing tensors to the loss function:

```python
loss = loss_fct(
    shift_logits.view(-1, shift_logits.size(-1)),
    shift_labels.view(-1).to(shift_logits.device)  # Added explicit device conversion
)
```

This ensures that labels are explicitly placed on the same device as logits before loss calculation, resolving both the compilation error and preventing potential runtime device mismatch issues.

## Why This Fix Works
By explicitly managing device placement in the forward method, we prevent torch.compile from generating invalid code during optimization. This approach:

1. Eliminates ambiguity in tensor device placement that confused the compiler
2. Ensures consistent behavior regardless of compilation status
3. Follows best practices for torch.compile compatibility

## Implementation Details
- Added a new patch function `patch_SmolVLMForConditionalGeneration_forward` in `temporary_patches.py`
- The patch is applied using the same pattern as other model-specific patches in Unsloth
- Added safety checks to ensure it only applies to the correct model class
- The patch is automatically applied when Unsloth is imported

## Testing
The patch has been tested with SmolVLM2 Video Instruct model:
- The compilation now succeeds without indentation errors
- The model functions correctly with compilation enabled

### DIFF
--- a/unsloth_zoo/temporary_patches.py
+++ b/unsloth_zoo/temporary_patches.py
@@ -44,8 +44,15 @@ def patch_Gemma3Processor():
     except:
         return
     from transformers.models.gemma3.processing_gemma3 import (
-        BatchFeature, Gemma3ProcessorKwargs, ImageInput, PreTokenizedInput,
-        TextInput, Unpack, make_nested_list_of_images, to_py_obj)
+        BatchFeature,
+        Gemma3ProcessorKwargs,
+        ImageInput,
+        PreTokenizedInput,
+        TextInput,
+        Unpack,
+        make_nested_list_of_images,
+        to_py_obj,
+    )
 
     def __call__(
         self,
@@ -185,8 +192,12 @@ def patch_Gemma3ForConditionalGeneration():
     except:
         return
     from transformers.models.gemma3.modeling_gemma3 import (
-        Cache, Gemma3CausalLMOutputWithPast, HybridCache,
-        is_torchdynamo_compiling, logger)
+        Cache,
+        Gemma3CausalLMOutputWithPast,
+        HybridCache,
+        is_torchdynamo_compiling,
+        logger,
+    )
 
     def forward(
         self,
@@ -395,8 +406,7 @@ def patch_Gemma3ForConditionalGeneration_causal_mask():
         import transformers.models.gemma3.modeling_gemma3
     except:
         return
-    from transformers.models.gemma3.modeling_gemma3 import (HybridCache,
-                                                            StaticCache)
+    from transformers.models.gemma3.modeling_gemma3 import HybridCache, StaticCache
 
     def _update_causal_mask(
         self,
@@ -609,8 +619,14 @@ def patch_Gemma3Attention():
     except:
         return
     from transformers.models.gemma3.modeling_gemma3 import (
-        ALL_ATTENTION_FUNCTIONS, Cache, FlashAttentionKwargs, Unpack,
-        apply_rotary_pos_emb, eager_attention_forward, logger)
+        ALL_ATTENTION_FUNCTIONS,
+        Cache,
+        FlashAttentionKwargs,
+        Unpack,
+        apply_rotary_pos_emb,
+        eager_attention_forward,
+        logger,
+    )
 
     scaled_dot_product_attention = torch.nn.functional.scaled_dot_product_attention
 
@@ -724,10 +740,14 @@ def patch_SmolVLMForConditionalGeneration_forward():
     from typing import List, Optional, Tuple, Union
 
     from transformers.models.smolvlm.modeling_smolvlm import (
-        CrossEntropyLoss, SmolVLMCausalLMOutputWithPast,
-        SmolVLMForConditionalGeneration)
+        CrossEntropyLoss,
+        SmolVLMCausalLMOutputWithPast,
+        SmolVLMForConditionalGeneration,
+    )
 
-    # Check if already patched
+    # Check if the fix is already present (either from Transformers library or previous Unsloth patch)
+    # We look for the specific device handling code that fixes torch.compile indentation errors,
+    # rather than an Unsloth marker, to handle cases where the fix comes from upstream
     current_forward_source = inspect.getsource(
         transformers.models.smolvlm.modeling_smolvlm.SmolVLMForConditionalGeneration.forward
     )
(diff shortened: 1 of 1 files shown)


unsloth_zoo/temporary_patches.py
```diff
@@ -557,17 +694,165 @@ def forward(
...
+
+    from transformers.models.smolvlm.modeling_smolvlm import (
+        CrossEntropyLoss, SmolVLMCausalLMOutputWithPast,
+        SmolVLMForConditionalGeneration)
+
+    # Check if already patched
+    current_forward_source = inspect.getsource(
+        transformers.models.smolvlm.modeling_smolvlm.SmolVLMForConditionalGeneration.forward
+    )
+    if "shift_labels.view(-1).to(shift_logits.device)" in current_forward_source:
```
Datta0, line 595: This is a very weird/specific way to check. 
Maybe we can add a comment in the forward funciton like "unsloth patched forward function" and check for the existence of that comment?
Or `inspect.getfile` should give you the location of the function. We can check if contains `unsloth`
Or there might be another better way. Can you please check?

rolandtannous: @Datta0 this section of the code you're commenting on, is from the smolvlm forward method as coded in the transformers library modeling file for smolvlm. The only part I changed/added is the explicit type conversion. My PR is pretty self explanatory in terms of details.
We should also avoid going into the habit of trying to over-optimize code from other libraries in cases like these, or else you'll end up rewriting 90% of the model files from the transformers library.

rolandtannous: Hello @Datta0 
Thanks for the feedback!
In this specific case, checking for the actual fix implementation (`shift_labels.view(-1).to(shift_logits.device)`) is intentional.

The reason is that this patch addresses a specific compilation error by adding explicit device management. If the Transformers library team implements this same fix upstream, or I push a PR to their repo as i was planning to do, they wouldn't include an Unsloth-specific marker comment. By checking for the actual implementation detail instead of a comment marker, we ensure that:

1. We don't re-apply the patch if the Transformers library fixes it themselves
2. We don't miss applying the patch if it's not yet present

I suggest adding more explanatory inline comments to make the purpose of this check clearer in the code instead. This way, it'll be more maintainable while still correctly detecting if the fix is already present regardless of its source. I added a follow-up commit in that direction.

Datta0: Oh if transformers is probably gonna get this fixed, then yeah I think this check does make some sense...
In that case, should the check be based on a transformers version (assuming that it gets merged in say 4.52), 
`if transformers_version < 4.52: apply_patch` ?

rolandtannous: @Datta0 This approach makes more sense to me than version checking for several reasons:

1. This isn't a bug from the Transformers library perspective - it only causes issues with Unsloth's torch.compile optimizations.

2. According to Transformers' contributing guidelines, any PR would need to be labeled as an "improvement" rather than a fix, requiring discussion first.

3. Since we can't know when (or if) they would merge such a change, a version-based check would create a maintenance burden requiring updates to this file until they do incorporate it at their end.

4. The current check is simpler and more reliable - it directly verifies whether the actual solution is present, regardless of which version added it.

5. I've followed the existing pattern for patch detection used throughout the codebase. This approach maintains consistency with the other patches in the same file, which all use inspect checks in a somewhat similar fashion.

Given that this PR addresses a functional issue that's affecting users, I believe we should prioritize moving it forward rather than delaying for stylistic preferences. I've added the explanatory comments to make the purpose of the check clearer.

Datta0: Cool, this makes sense to me :)
Thanks for the changes

Datta0 (approved):
LGTM

unsloth_zoo/temporary_patches.py
jarrelscy: This is amazing work but I don't quite understand - how does the explicit device management added here somehow fix the indentation? I've got a similar problem in Unsloth where the _move_model_to_vllm gets indented wrongly after compilation...

rolandtannous: @jarrrelscy I traced it back to torch.compile (used by unsloth) and graph breaks. In this case not having the labels and features on the same device throws torch compile off, and as a consequence impacts the "compiled" file and results in the indentation error. It's one of the oddest errors I have seen for sure.

kmamine:
was this merged ?

rolandtannous:
@kmamine maybe @shimmyshimmer can answer the question

rolandtannous:
@kmamine @hermeschen1116 This is now merged to main. make sure to update your unsloth-zoo installation

kmamine:
Thanks

PR #106, opened 2025-03-28
fix double bos

tamewild:
It was mistakenly added here https://github.com/unslothai/unsloth-zoo/commit/4a66f8b08952fc148f5c74cd15aec52cb0113e2d

image

### DIFF
--- a/unsloth_zoo/dataset_utils.py
+++ b/unsloth_zoo/dataset_utils.py
@@ -547,7 +547,7 @@ def sft_prepare_dataset(
                 )
             test_text = test_text[0]
         else:
-            test_text = next(iter(dataset))[dataset_text_field][0]
+            test_text = next(iter(dataset))[dataset_text_field]
 
         # Get chat template
         chat_template = getattr(processing_class, 'chat_template', '')


Datta0:
Hey @tamewild can you please add samples of before and after the change?

danielhanchen:
main now does exactly this in unsloth_zoo/dataset_utils.py, where test_text is taken from the field without indexing the first character, so I do not think this is still needed. Sorry for the slow review, and thanks for catching it and putting the fix together.

tamewild:
> main now does exactly this in unsloth_zoo/dataset_utils.py, where test_text is taken from the field without indexing the first character, so I do not think this is still needed. Sorry for the slow review, and thanks for catching it and putting the fix together.

Thanks for letting me know! Just saw the notification now, will close issue. I know this is self-promotey and I'm not sure if you already saw it but I released a blog post recently that involved training small base models (3B, 4B, 9B) for 6.5 to 40 minutes (on a single GPU) and they approached their official post-trained baselines. It also involves a custom loss function that was derived from KTO and I think it is potentially very simple to implement with unsloth. I understand that you might be interested in stuff like this, and I would love any feedback (especially negative feedback) on the math or whatever. 

I realize you're incredibly busy so I'm not expecting a response or anything, but I just thought I should just try. I really apologize for doing this. You can check it out here: https://github.com/tamewild/pcss (blog post with rendered math: https://huggingface.co/blog/tamewild/pcss). There's also jupyter notebooks (for reproduction) which you can skim if you want. I use unsloth's mirror for Qwen 3 4B Base and I use unsloth's corrected chat template for Qwen 3.5.

(For raw markdown of the blog post, if you want to use an agent to review it first: https://raw.githubusercontent.com/tamewild/pcss/main/BLOG_POST.md)

PR #108, opened 2025-04-03
Patch Gemma3Processor to handle integer num_crops

rupaut98:
Addresses the issue reported in [#2265](https://github.com/unslothai/unsloth/issues/2265) regarding the documentation claim that Gemma 3 vision models work out-of-the-box with existing vision fine-tuning notebooks.

### Problem

Following the documentation at `https://docs.unsloth.ai/basics/vision-fine-tuning`, attempting to fine-tune a Gemma 3 vision model (e.g., `unsloth/gemma-3-4b-it-bnb-4bit`) using the provided notebook examples results in a `TypeError: 'int' object is not iterable` during the `trainer.train()` step.

This occurs because the current Unsloth patch for `transformers.models.gemma3.processing_gemma3.Gemma3Processor.__call__` (located in `unsloth_zoo/temporary_patches.py`) assumes the `num_crops` variable for each item in a batch is always iterable. However, when the image processor doesn't generate extra crops (a common case), `num_crops` for an item can be a single integer (e.g., `0`). The code then attempts `zip(num_crops, image_indexes)`, which fails when `num_crops` is an integer.

### Solution

This PR modifies the `patch_Gemma3Processor` function in `unsloth_zoo/temporary_patches.py` to correctly handle the `num_crops` variable.

Specifically, within the batch processing loop, before the inner `zip` call:

*   It checks if `num_crops_for_item` (the crop value for the current batch item) is an `int`.
*   If it is an `int`, it converts it into a list matching the length of `image_indexes` (e.g., `[int_value, 0, 0,...]`) to ensure it's iterable.
*   It also includes robust handling for cases where `num_crops_for_item` might be some other unexpected non-iterable type, defaulting to a list of zeros.

This ensures that the subsequent `zip(iterable_num_crops_for_item, image_indexes)` call always receives valid iterables, resolving the `TypeError`. Some minor robustness improvements (using logger, safer attribute access, etc.) from the development process are also included.

### Original Code Snippet (Causing Error)

```python
# [...]
            for batch_idx, (prompt, images, num_crops) in enumerate(zip(text, batched_images, batch_num_crops)):
                # [...]
                # ERROR occurs here if num_crops is int:
                for num, idx in reversed(list(zip(num_crops, image_indexes))):
                    # [...]
```
Patched Code Snippet (With Fix)
```python
# [...]
            for batch_idx, (prompt, images_for_item, num_crops_for_item) in enumerate(zip(text, batched_images, batch_num_crops)):
                image_indexes = [m.start() for m in re.finditer(self.boi_token, prompt)]

                if len(images_for_item) != len(image_indexes):
                    raise ValueError(
                        f"Prompt contained {len(image_indexes)} image tokens but received {len(images_for_item)} images."
                    )
                
                iterable_num_crops = num_crops_for_item
                
                if isinstance(num_crops_for_item, int):
                        if len(image_indexes) > 0:
                            iterable_num_crops = [num_crops_for_item] + [0] * (len(image_indexes) - 1)
                        else:
                            iterable_num_crops = []

                # Insert additional image tokens for Pan-and-Scan crops
                for num, idx in reversed(list(zip(iterable_num_crops, image_indexes))):
                    if isinstance(num, int) and num > 0:
                        formatted_image_text = (
                            f"Here is the original image {self.boi_token} and here are some crops to help you see better "
                            + " ".join([self.boi_token] * num)
                        )
                        prompt = prompt[:idx] + formatted_image_text + prompt[idx + len(self.boi_token) :]
                        text_with_crops[batch_idx] = prompt
```

Use the Colab notebook provided: [Gemma3-4b-vision](https://colab.research.google.com/drive/16J6TkNWjf6UnbV88yPCLWcnf9t5KP0zm?usp=sharing)

Expected: Training should now proceed without the TypeError: 'int' object is not iterable.

Related Documentation

This fix makes the documentation claim at https://docs.unsloth.ai/basics/vision-fine-tuning correct. Consider updating the documentation to explicitly confirm Gemma 3 support or link to a verified Gemma 3 notebook example after this merge.

Edit: This only works for GPU that supports bf16.

### DIFF
--- a/unsloth_zoo/temporary_patches.py
+++ b/unsloth_zoo/temporary_patches.py
@@ -56,7 +56,6 @@ def patch_Gemma3Processor():
     try:
         import transformers.models.gemma3.processing_gemma3
     except:
-        logger.info("Gemma3 processor not found. Skipping patch.") # Use logger
         return
     from transformers.models.gemma3.processing_gemma3 import (
         ImageInput,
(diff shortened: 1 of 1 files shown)


pySilver:
Thanks! I've just faced this issue

gnizamaaa:
Thanks, I was racking my brains here trying to understand why this error was happening
Saved me just in time to get results for a presentation

unsloth_zoo/temporary_patches.py
```diff
@@ -34,6 +37,21 @@
 global TEMPORARY_PATCHES
 TEMPORARY_PATCHES = []
 
+# Unsloth Zoo - Utilities for Unsloth
+# Copyright 2023-present Daniel Han-Chen, Michael Han-Chen & the Unsloth team. All rights reserved.
+# ... (License header remains the same) ...
+
```
Datta0: I don't think we need these here
Also move the imports up top

rupaut98: done, thank you for the review :)

pySilver:
@rupaut98 can you please take a look if this PR isn't the root of the problem described here? https://github.com/unslothai/unsloth/issues/2297

rupaut98:
> @rupaut98 can you please take a look if this PR isn't the root of the problem described here? [unslothai/unsloth#2297](https://github.com/unslothai/unsloth/issues/2297)

Hey @pySilver thanks for sharing the dataset. I would like to see your code (how you load the dataset) so i can reproduce that error. However, I trained with your dataset and I don't see any errors during training and inference as well. Here is the link to the notebook https://colab.research.google.com/drive/1HMQmG-p1cf9mw76BwP5Y0Ay0NnGOhdQh?usp=sharing. Hope thats helpful.

pySilver:
@rupaut98 Ive shared the training code in last message in the linked ticket here https://github.com/unslothai/unsloth/issues/2297#issuecomment-2781512405

Our setups is mostly identical. The only difference is that I'm doing 1 epoch instead of running few steps. My training always fail at 1285 step and by logged batch I cannot see anything specific why it happens.

aniJani:
Thank you! I was facing this. I hope the reviewers merge it quick.

rolandtannous:
@MoAnasSiddiqui I've looked at this, your issue is a completely different issue that affects Gemma3. the issue you're facing is because UNSLOTH_FORCE_FLOAT32 is being set to 1 for Gemma3. So it has nothing to do with this particular PR. Do you mind opening a separate github issue please

rolandtannous:
@shimmyshimmer LGTM


I tested it further for Gemma3 inference code on vision dataset from colab notebooks. 
Here is the test code:

```
from datasets import load_dataset

dataset = load_dataset("unsloth/Radiology_mini", split="train")

FastVisionModel.for_inference(model) # Enable for inference!

image = dataset[0]["image"]
image = image.convert('RGB') # Gemma3 requires images to be in RGB
instruction = "You are an expert radiographer. Describe accurately what you see in this image."

messages = [
    {"role": "user", "content": [
        {"type": "image"},
        {"type": "text", "text": instruction}
    ]}
]
input_text = tokenizer.apply_chat_template(messages, add_generation_prompt = True)
inputs = tokenizer(
    image,
    input_text,
    add_special_tokens = False,
    return_tensors = "pt",
).to("cuda")

from transformers import TextStreamer
text_streamer = TextStreamer(tokenizer, skip_prompt = True)
_ = model.generate(**inputs, streamer = text_streamer, max_new_tokens = 128,
                   use_cache = True, temperature = 1.0, top_p = 0.95, top_k=64)
```
fails without this fix. Works perfectly with this fix!


We need this pushed. Current issues in progress depend on this getting pushed.
Thank you

shimmyshimmer:
Thank you @rupaut98 and @rolandtannous merged the PR

PR #110, opened 2025-04-05
Fix wrong index causing double BOS

Oseltamivir:
In Gemma-4B colab,

`tokenizer.decode(trainer.train_dataset[100]["input_ids"])` gives:
`<bos><bos><start_of_turn>user\nWhat...`

Modifying the code by adding `print("\n\nTest text:", test_text)` prints: `Test text: <` , showing that indexing with [0] is unnecessary as it indexes the first letter of the test text.

After modifying, `print("\n\nTest text:", test_text)` now gives: `Test text: <bos><start_of_turn>user\nExplain what boolean...`

Does not apply to `formatting_func` as  the `formatting_func` should return a list of processed strings.

### DIFF
--- a/unsloth_zoo/dataset_utils.py
+++ b/unsloth_zoo/dataset_utils.py
@@ -547,7 +547,7 @@ def sft_prepare_dataset(
                 )
             test_text = test_text[0]
         else:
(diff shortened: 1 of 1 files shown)


Oseltamivir:
Fixed

Issue #109, opened 2025-04-05
UnicodeDecodeError in unsloth GRPO Trainer when locale preferred encoding changes


Hmzbo:
Here I follow the bug submission format from the Unsloth library repository.

**Describe the bug**
When the locale preferred encoding changes from its default value "utf-8" to "ascii" (more precisely "ANSI_X3.4-1968") the GRPOTrainer from unsloth produces a UnicodeDecodeError that stems from the get_trl_metrics function in unsloth_zoo.

1. **Environment Setup:**
   - OS: Linux-6.6.56+-x86_64-with-glibc2.35
   - Python Version: 3.10.12
   - Frameworks/Libraries: unsloth==2025.3.18, trl==0.16.0, triton==3.1.0, unsloth_zoo==2025.3.16, peft==0.14.0 (complete [requirements.txt](https://github.com/user-attachments/files/19617076/requirements.txt) )
   - Kaggle (link for notebook below)

2. **Dataset Details:**
   - Dataset Name: Private dataset

3. **Model Details:**
   - Model ID: unsloth/gemma-3-4b-it-unsloth-bnb-4bit
   - Model Configuration: 4bit quantization with lora (r=alpha=64)

4. **Training Configuration:**
   - Trainer Args: 
```
GRPOConfig(
    learning_rate = 5e-6,
    adam_beta1 = 0.9,
    adam_beta2 = 0.99,
    weight_decay = 0.1,
    warmup_ratio = 0.1,
    lr_scheduler_type = "cosine",
    optim = "adamw_torch_fused",
    logging_steps = 1,
    bf16 = is_bfloat16_supported(),
    fp16 = not is_bfloat16_supported(),
    per_device_train_batch_size = 2,
    gradient_accumulation_steps = 1, # Increase to 4 for smoother training
    num_generations = 2, # Decrease if out of memory
    max_prompt_length = max_prompt_length,
    max_completion_length = max_seq_length - max_prompt_length,
    # num_train_epochs = 1, # Set to 1 for a full training run
    max_steps = 200,
    save_steps = 200,
    max_grad_norm = 0.1,
    #report_to = "wandb", # Can use Weights & Biases
    output_dir = "outputs",
)
```

5. **Reproduction Steps:**
   - Kaggle notebook [link](https://www.kaggle.com/code/hamzaboulahia/reproduce-encoding-error)

8. **Additional notes:**
[...]

PR #112, opened 2025-04-09
Indicating encoding explicitly in logging_utils.py

Hmzbo:
As I detailed in the issue number #109 
without indicating explicitly the encoding when reading the Trainer files from the trl library, a UnicodeDecodeError could arise if the the locale preferred encoding is not set to utf-8. 
I have tested the GRPO Trainer after with this minor change and it worked as intended.


mmathew23:
Hi @Hmzbo I believe this was fixed in a larger utf-8 encodinf refactor, https://github.com/unslothai/unsloth-zoo/commit/3d173e6887600effd83cb6211af97681f32c9cfa. If you could check that the issue if resolved on your end, that would be great! Thank you!

Issue #114, opened 2025-04-12
Failed of renaming .int. to [int]. inside training_utils.py:prepare_model_for_training


fish4terrisa-MSDSM:
When I'm training olmoe using unsloth, I got this error:
```
🦥 Unsloth: Will patch your computer to enable 2x faster free finetuning.
Unsloth: Failed to patch Gemma3ForConditionalGeneration.
🦥 Unsloth Zoo will now patch everything to make training faster!
==((====))==  Unsloth 2025.3.19: Fast Olmoe patching. Transformers: 4.51.1.
   \\   /|    NVIDIA GeForce RTX 4060 Laptop GPU. Num GPUs = 1. Max memory: 7.633 GB. Platform: Linux.
O^O/ \_/ \    Torch: 2.5.1+cu124. CUDA: 8.9. CUDA Toolkit: 12.4. Triton: 3.1.0
\        /    Bfloat16 = TRUE. FA [Xformers = 0.0.29. FA2 = False]
 "-____-"     Free license: http://github.com/unslothai/unsloth
Unsloth: Fast downloading is enabled - ignore downloading bars which are red colored!
Loading checkpoint shards: 100%|██████████████████| 3/3 [00:05<00:00,  1.80s/it]
Unsloth: Making `model.base_model.model.model` require gradients
Traceback (most recent call last):
  File "/home/fish4terrisa/.conda/envs/unsloth/lib/python3.11/site-packages/unsloth_zoo/training_utils.py", line 167, in prepare_model_for_training
    exec(f"{name}.to({str(dtype)})")
  File "<string>", line 1
    model.model.model.layers[0].mlp.experts.1.gate_proj.lora_A.default.to(torch.float32)
                                           ^^
SyntaxError: invalid syntax

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/fish4terrisa/unsloth/train_olmoe.py", line 32, in <module>
    model = FastLanguageModel.get_peft_model(
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/fish4terrisa/.conda/envs/unsloth/lib/python3.11/site-packages/unsloth/models/llama.py", line 2035, in get_peft_model
    return FastBaseModel.get_peft_model(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #115, opened 2025-04-12
Fix bug in training_utils.py

fish4terrisa-MSDSM:
A solution for issue #114


danielhanchen:
Apologies on the delay - great work!

PR #116, opened 2025-04-17
Fix LoRA hot swapping

pbarker:
LoRA hot swapping is broken as documented in https://github.com/unslothai/unsloth/issues/2322

This is due to an overly permissive regex which when loading multiple adapters which will apply the first adapter correctly but when applying the second adapter it will target layers in the first adapter

The main change is a tighter regex in `get_peft_regex`

This also formats some of the code to be more up to spec with ruff linting.

I tested this with qwen-vl-2.5 3b and everything appears to work

### DIFF
--- a/unsloth_zoo/peft_utils.py
+++ b/unsloth_zoo/peft_utils.py
@@ -24,135 +24,173 @@ __all__ = [
 ]
 
 import inspect
-import torch
 import os
-from typing import Any, Callable, Dict, List, Optional, Set, Tuple, Type, TypeVar, Union
-from collections import OrderedDict
 import re
+from collections import OrderedDict
+from typing import Any, Callable, Dict, List, Optional, Set, Tuple, Type, TypeVar, Union
+
+import torch
 
 # Skip some modules sensitive to quantization
 SKIP_QUANTIZATION_MODULES = [
     "lm_head",
-    "multi_modal_projector", # Llama 3.2 Vision, Pixtral, Llava
-    "merger",                # Qwen2 VL
-    "modality_projection",   # Idefics, SmolVLM
-    "router",                # MoE Router
-    "gate",                  # MoE Router
-    'mamba',
+    "multi_modal_projector",  # Llama 3.2 Vision, Pixtral, Llava
+    "merger",  # Qwen2 VL
+    "modality_projection",  # Idefics, SmolVLM
+    "router",  # MoE Router
+    "gate",  # MoE Router
+    "mamba",
 ]
+
 
 def get_peft_regex(
     model,
-    finetune_vision_layers     : bool = True,
-    finetune_language_layers   : bool = True,
-    finetune_attention_modules : bool = True,
-    finetune_mlp_modules       : bool = True,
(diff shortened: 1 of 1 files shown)


danielhanchen:
Thanks for the PR - sorry on the delay! @rolandtannous or @mmathew23 could you guys check if this logic works as expected - appreciate it :)

rolandtannous:
@pbarker @danielhanchen 

# Test results
Just finished testing this and this fix failed for me.
The reason why I think it fails, is that load_adapter method, is a peft library method and not an unsloth method. In fact,  
```
get_peft_regex()
``` 
is not called when applying the 
```
load_adapter()
```
method

## **when is get_peft_regex() actually called?** 
get_peft_regex() is called when applying the get_peft_model for FastVisionModel or FastLanguageModel

In fact, If you look at the Error trace strack in the testing notebook i've attached, you'll see that the call to load_adapter calls the peft library immediately and there is no intermediary call to get_peft_regex()

Here is the E2E testing notebook I used based on the code shared by @pbarker [here](https://github.com/unslothai/unsloth/issues/2322#issuecomment-2800022323) :
[test_notebook](https://colab.research.google.com/drive/1ECkeCnhQJxPDKqYNQ-I1-rhsJM5WfBvN?usp=sharing)

# Existing peft issue
This seems to be an issue related to the peft library and has been discussed a few times in the peft repo github issues section. Examples:
https://github.com/huggingface/peft/issues/957
https://github.com/huggingface/peft/issues/2388



# Workaround

There is a workaround for swapping lora adapters using the ```hotswap_adapter``` method. [test_notebook](https://huggingface.co/docs/peft/en/package_reference/hotswap).

The code would look something like this
```
model_id = "unsloth/Qwen2.5-VL-7B-Instruct"

base_model, model_processor = FastVisionModel.from_pretrained(
    model_id,
    load_in_4bit=False,
)
FastVisionModel.for_inference(base_model)

model = FastVisionModel.get_peft_model(
    base_model,
    finetune_vision_layers=True,  # False if not finetuning vision layers
    finetune_language_layers=True,  # False if not finetuning language layers
    finetune_attention_modules=True,  # False if not finetuning attention layers
    finetune_mlp_modules=True,  # False if not finetuning MLP layers
    r=8,  # The larger, the higher the accuracy, but might overfit
    lora_alpha=16,  # Recommended alpha == r at least
    lora_dropout=0.1,
    bias="none",
    random_state=3407,
    use_rslora=False,  # We support rank stabilized LoRA
    loftq_config=None,  # And LoftQ
    use_fast=True,
    # target_modules = "all-linear", # Optional now! Can specify a list if needed
)

hotswap_adapter(model, "./outputs/checkpoint-15", adapter_name="default")
hotswap_adapter(model, "./outputs/checkpoint-30", adapter_name="default")
from peft.utils.hotswap import hotswap_adapter
```

Here is an End to End notebook that illustrates the use of hotswap_adapter :
[workaround_notebook](https://colab.research.google.com/drive/1dfzX2diJtuZwHzr3Tb4G654Emj9F1mdc?usp=sharing)

PR #124, opened 2025-04-28
Fix for updated llama.cpp syntax.

reedmayhew18:
Fixed syntax issue resolving: NotImplementedError: Unsloth: llama.cpp GGUF conversion does not yet support converting model types of `Gemma3ForCausalLM`.

### DIFF
--- a/unsloth_zoo/llama_cpp.py
+++ b/unsloth_zoo/llama_cpp.py
@@ -356,7 +356,7 @@ def _download_convert_hf_to_gguf(
         )
 
     # Get all supported models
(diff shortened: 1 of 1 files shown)


rolandtannous:
This is not sufficient.
llama.cpp not only changed the class name from Model to BaseModel but also introduced separate TExt and Vision Models. This does not accommodate those other changes. PLease refer to the codebase:

https://github.com/ggml-org/llama.cpp/blob/master/convert_hf_to_gguf.py

and update this PR

reedmayhew18:
Please see updated pull request: https://github.com/unslothai/unsloth-zoo/pull/125

PR #125, opened 2025-04-28
Updated llamacpp regrex.

reedmayhew18:
Updated llama.cpp regrex to handle all models, including Gemma3ForCausalLM.

Resolving the error:

NotImplementedError: Unsloth: llama.cpp GGUF conversion does not yet support converting model types of `Gemma3ForCausalLM`.

This pull request is **just** for resolving the BaseModel issue, since users (like myself) are unable to use unsloth-zoo properly with Gemma 3 finetuning. If others wish to expand on this moving forward for other class fixes, that's fine, but at minimum, this will get some users back up and running without any negative consequences.

### DIFF
--- a/unsloth_zoo/llama_cpp.py
+++ b/unsloth_zoo/llama_cpp.py
@@ -356,7 +356,7 @@ def _download_convert_hf_to_gguf(
         )
 
     # Get all supported models
-    supported_types = re.findall(rb"@ModelBase\.register\(([^)]{1,})\)", converter_latest)
+    supported_types = re.findall(rb"@(?:Model|ModelBase)\.register\(\s*([^)]+?)\s*\)",converter_latest)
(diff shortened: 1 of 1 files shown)


reedmayhew18:
My apologies, my coding knowledge is limited, so I'm not sure how much I'll be able to assist regarding vision capabilities.

Is this the correct commit where the changes were implemented causing the current issue?

https://github.com/ggml-org/llama.cpp/commit/2016f07bd106c73699ecbaace80f55db5ed95dac

rolandtannous:
That's ok no worries.
I'll push a PR shortly and tag you so you can see.
Also tag me in the discord if you want to know more the why behind some of the changes i'll make to the file.
will be up in a few

shimmyshimmer:
Thank you @reedmayhew18 we accepted it! :D

PR #126, opened 2025-04-30
Fix 'standardize_data_formats' when using iterable datasets

marcandrelarochelle:
Adds support to Iterable datasets for the 'standardize_data_formats'

`num_proc` is not supported by `IterableDataset` and cannot be accessed via index

PS: Sorry had to recreated my fork, did a mistake on my side


danielhanchen:
Much apologies on the delay @marcandrelarochelle ! Thanks for the PR!

unsloth_zoo/dataset_utils.py
```diff
@@ -405,10 +405,10 @@ def standardize_data_formats(
     if "conversations" not in column_names:
         return dataset
 
-    convos = dataset[:10]["conversations"]
+    examples = itertools.islice(dataset, 10)
```
danielhanchen, line 410: Is this specifically for iterable datasets? Ie `itertools.islice` works for iterable datasets whilst `dataset[:10]` does not?

unsloth_zoo/dataset_utils.py
```diff
@@ -405,10 +405,10 @@ def standardize_data_formats(
...
     uniques = collections.defaultdict(list)
-    for convo in convos:
-        for message in convo:
+    for example in examples:
+        for message in example["conversations"]:
```
danielhanchen, line 413: The main reason why I did "conversations" outside is to make it somewhat faster, but I guess since its 10 examples, no big deal

unsloth_zoo/dataset_utils.py
```diff
@@ -464,15 +464,23 @@ def _standardize_dataset(examples):
...
-        _standardize_dataset,
-        batched = True,
-        desc = "Unsloth: Standardizing formats",
-        num_proc = num_proc,
-    )
+    if isinstance(dataset, IterableDataset):
+        return dataset.map(
+            _standardize_dataset,
+            batched = True,
+            batch_size = dataset._ex_iterable.batch_size,
```
danielhanchen: Interesting on `._ex_iterable` - I actually am not super interested with this

danielhanchen (commented):
@Erland366 Could you take a final review and confirm if iterable datasets work fine? Appreciate it

unsloth_zoo/dataset_utils.py
marcandrelarochelle: Exactly, you can't use an index or in this case `[:10]` on iterators

unsloth_zoo/dataset_utils.py
marcandrelarochelle: The reason I did this is to pass the batch_size, if you are using a iterable dataset, you might have a resource restrained system and by default it would use a batch size of 1000, but this way it will keep the same batch size it had before the `_standardize_datasetformatting`.

mmathew23: Could we instead keep one dataset.map as the logic doesn't seem to be based on dataset type? And instead create a new default argument to standardize_dataset_format called batch_size that can be passed to map? that way users can change functionality if they are on a resource constrained system.

marcandrelarochelle: @mmathew23 even if I add a default parameter to `standardize_dataset_format` for the batch size, we'll still need 2 dataset.map due to the `num_proc` not supported by iterable dataset, is that still ok with you?

mmathew23: Good point @marcandrelarochelle! You can add a kwarg dict and add num_processes if it not iterable. Then unpack it. 

so something like
```
ds_map_kwargs = {
    'batched': True,
    'desc': "Unsloth: Standardizing formats",
    'batch_size': batch_size,
}
if not isinstance(dataset, IterableDataset):
    ds_map_kwargs['num_proc'] = num_proc
dataset.map(standardize_dataset, **ds_map_kwargs)
```
and batch_size would be an argument with a default value.

marcandrelarochelle: @mmathew23 Done!

PR #131, opened 2025-05-06
update the null check for logit_softcapping to is not None

mmathew23:
There are current issues for full finetuning gemma3. [#2101](https://github.com/unslothai/unsloth/issues/2101). logit_softcapping gets passed as None. The compiler checks for != () instead of is not None like transformers. This PR swaps the check for empty tuple to is not None. Gemma3 notebook below showing that it works.

https://colab.research.google.com/drive/1dKVqnEWOBBcjpiPk3GVScQG-Iir5uV5H?usp=sharing

### DIFF
--- a/unsloth_zoo/compiler.py
+++ b/unsloth_zoo/compiler.py
@@ -803,7 +803,7 @@ else:
         logits = logits * (\\2)
     if (\\3) != ():
         logits = logits / (\\3)
-    if (\\4) is not None:
+    if (\\4) is not None or (\\4) != ():
         logits = logits / (\\4)
         logits = torch.tanh(logits)
         logits = logits * (\\4)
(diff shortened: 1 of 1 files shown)


rolandtannous:
LGTM!

mmathew23:
We get syntax warnings now if the regex parses logit_softcapping and returns () instead of using the full logit_softcapping name. We can use `is` with a tuple literal, ie  `() is not None` results in a SyntaxWarning. I think instead we can do () not in (None, ()). 

From:
```
logit_softcapping = (\\4) if (\\4) is not None or (\\4) != () else 0,
```

To:
```
logit_softcapping    = (\\4) if (\\4) not in (None, ()) else 0,,
```

I updated the notebook with this fix to verify it works. https://colab.research.google.com/drive/1dKVqnEWOBBcjpiPk3GVScQG-Iir5uV5H?usp=sharing

Then I ran a notebook which produced the syntax warnings, and we don't see the warnings on my branch. https://colab.research.google.com/drive/1wylRdL5ri3Df6LCRE-ee3THr8h-OIfbO?usp=sharing

PR #139, opened 2025-05-15
Convert n_items type from torch.Tensor to int

MoonRainy21:
Change the type of n_items from a tensor to an integer.
On `compiler.py`, type of `n_items` is shown as `int`, but the variable passed into the model was typed `torch.Tensor`.
It may not be issue in most of cases, but it may cause issues related to device. (such as `n_items` on `cuda:0` and operate with `loss` on `cuda:7`)

### DIFF
--- a/unsloth_zoo/compiler.py
+++ b/unsloth_zoo/compiler.py
@@ -623,7 +623,7 @@ else:
         logit_scale_divide : float = 0,
         logit_softcapping : float = 0,
         vocab_size : int = 0,
-        n_items : int = 0,
+        n_items : int | torch.Tensor = 0,
     ):
         device = output_logits.device
         if logit_scale_multiply != 0:
@@ -658,6 +658,8 @@ else:
             )
         pass
         if n_items != 0:
+            if torch.is_tensor(n_items):
+                n_items = n_items.to(loss.device)
             loss = loss / n_items
         else:
             loss = loss / (shift_labels != -100).sum()
@@ -743,7 +745,7 @@ elif self.loss_function.__name__.endswith("ForCausalLMLoss") and labels is not N
         logit_scale_divide : float = 0,
         logit_softcapping : float = 0,
         vocab_size : int = 0,
-        n_items : int = 0,
+        n_items : int | torch.Tensor = 0,
     ):
         device = output_logits.device
         if logit_scale_multiply != 0:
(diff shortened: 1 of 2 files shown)


danielhanchen:
Thanks for the PR! Would the previous line `if device is not None and torch.is_tensor(num_items_in_batch):
                num_items_in_batch = num_items_in_batch.to(device)` solve that issue?

MoonRainy21:
I’ve tested multi-GPU training using my fork, and everything seems to be working well. @danielhanchen, do you have any additional comments or suggestions?

Erland366:
I want to discuss abit about this

I tested the behavior in vanilla HuggingFace and it also got the same issue : 

Here's my testing on Kaggle notebook on 2 T4 -> https://www.kaggle.com/code/erlandpg/test-multigpu-bitsandbytes

I tested to move the `num_items_in_batch` to the loss device and it works. but GPU utilization is around 20%. I need more testing whether this is a good number of utilization? Also whether we want to do this at all or not since HF itself did not support it.

cc: @danielhanchen

MoonRainy21:
It seems there's code for the device of `num_items_in_batch` on `loss_utils.py:304`. I wasn't able to understand when `_unsloth_get_batch_samples` is called, but it seems it would become unnecessary.

MoonRainy21:
@Erland366 For utilization, GPU utilization of the running GPU was pretty high for me (~80%) when I have tried with higher batch size but only one of the GPUs were running. We might need pipeline parallelism support for better utilization.

cc: @danielhanchen

MoonRainy21:
Have tested latest commit with 8 GPUs training Qwen3 235B A22B, but not really sure about some code which seems to be string and used for making unsloth compiled cache

danielhanchen:
main now does this in unsloth_zoo/loss_utils.py, where num_items_in_batch is moved to the loss device before the division, and the compiled loss block in compiler.py this also patched no longer exists, so I do not think this is needed any more. Sorry for the slow review, and thanks for working through it with us.

danielhanchen:
main now moves a tensor num_items_in_batch onto the loss device in unsloth_zoo/loss_utils.py, and the compiler loss template this patched no longer exists, so I do not think this is needed any more. Sorry for the slow review, and thanks for digging into the multi GPU case.

MoonRainy21:
Thanks for the kind comment. Confirmed that main have resolved the purpose of this PR.

Thank you for your all effort putting in Unsloth and open source community!

PR #153, opened 2025-05-28
Add _get_vocab_size utility function for config handling

AshAnand34:
Introduced a new utility function, _get_vocab_size, to safely retrieve the vocab_size from both old and new config structures. Updated relevant files to utilize this function, ensuring consistent vocab_size access across the codebase.


danielhanchen:
Oh this is nice! Thanks!

PR #157, opened 2025-05-31
Fix/fix save merge logic new transformers architecture

rolandtannous:
## Solves

Save and merge logic breaks with the new transformers (version==52.2.3) naming convention structure for VLMs.

## Problem

HuggingFace recently updated transformers with a new layer naming convention and structure for VLMs. Model loading with .from_pretrained automatically converts to the new naming convention if necessary. 

As a result lora adapters/peft models also use the new naming convention. Our save and merge logic loads the base model using the safetensors library which does not apply any conversions to the new naming convention. As a result, matching between lora layers and the base model layers fails and no merging takes place. A runtime exception is thrown warning the user that 0 saved modules were saved while there are X number of lora modules

## Error Reproduction code

Using unsloth with transformers 4.51.3, run
```
model.save_pretrained_merged(...)
```
or 
```
model.push_to_hub_merged(...)
```

## Solution

Introduced logic to:
- Check format of base model (new vs old layer names)
- check format of lora adapter (new vs old layer names)
- check for model conversion map
- Build naming reverse conversion logic where base model uses old convention and lora adapter uses new convention.

Updated: 
- Added _convert_lora_keys_to_safe_tensors_format method
- Added _get_checkpoint_conversion_mapping
- Added find_lora_base_model method
- Added detect_keys_format method
- Modified _merge_and_overwrite_lora 
- Introduced minor modification to create_lora_statistics and merge_and_overwrite_lora

## Testing

Tested extensively using logic from: https://github.com/unslothai/unsloth/blob/440419a67df913d40fa12c397219288c0dc80f76/tests/saving/save_merge_comparative_tests/test_unsloth_finetuning_save_merge_notebook.ipynb

Checked that merge and saving takes place without errors
Checked that merged model behaves as expected when loaded

Additional tests will be uploaded to unslothai/unsloth test folder with the PR for text model saving and merging

### DIFF
--- a/unsloth_zoo/saving_utils.py
+++ b/unsloth_zoo/saving_utils.py
@@ -259,8 +259,8 @@ def create_lora_statistics(model, merge_into_original = False, return_state_dict
 
     remove_keys = set()
     keep_keys   = set()
-
-    inner_model = model.base_model.model if hasattr(model, "base_model") else model
+    # Use the new function to find the right model level
+    inner_model = find_lora_base_model(model)
     for name, module in inner_model.named_modules():
         if name == "": continue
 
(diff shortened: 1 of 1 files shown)


rolandtannous:
superseded by #158

PR #159, opened 2025-06-04
N items qwen

mmathew23:
currently qwen 2.5 VL notebook fails on inference based on some recent changes to compiler.py. This is PR is a draft at a solution that aim's to be safe and backward compatible with old transformers versions.

Sometimes loss_kwargs and kwargs aren't ever matched and end up with an empty tuple, and then we fail to get `num_items_in_batch`. So now we make KWARGS a named group, place a sentinel in the replacement if needed, check for a match, and if not default to check loss_kwargs and kwargs.

Added bonus that ConditionalGeneration get's compiled for fast fuse linear.

Tested on Qwen 2.5VL notebook
https://colab.research.google.com/drive/1ANPMeUhUIdHaCy8qNecOdk42iSs_1KV2?usp=sharing

### DIFF
--- a/unsloth_zoo/compiler.py
+++ b/unsloth_zoo/compiler.py
@@ -964,7 +964,7 @@ def apply_fused_lm_head(forward):
                      r"self\.config\.vocab_size|"\
                      r"self\.config\.text_config\.vocab_size"\
                      ")")\
-            .replace("$KWARGS$",       r"(?:, \*\*(loss_kwargs|kwargs))?")\
+            .replace("$KWARGS$",       r"(?:,\s*\*\*(?P<KWARGS>(?:loss_kwargs|kwargs)))?")\
             .replace("$LOGITSUPCAST$", r"(?:logits = logits\.float\(\))?")\
             .replace("$LABELSDEVICE$", r"(?:labels = labels\.to\([^\)]{1,}\))?")\
             .replace("$LOGITSCALINGMULTIPLY$",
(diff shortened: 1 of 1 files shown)


unsloth_zoo/compiler.py
```diff
@@ -1061,12 +1060,35 @@ def apply_fused_lm_head(forward):
...
+                    kwarg_text = match.group("KWARGS")
+                    out = out.replace("$KWARGS$", kwarg_text or DEFAULT_KWARGS)
+                    return out
+            else:
+                # Handle case where KWARGS is not in the regex
+                safe_replacement = replacement.replace("$KWARGS$", DEFAULT_KWARGS)
+
+                def _kw_safe_sub(match):
+                    return match.expand(safe_replacement)
+
```
rolandtannous: There might be a simpler approach though. 
Leave the .replace statement on line 967 as is, meaning
```
.replace("$KWARGS$", r"(?:, \*\*(loss_kwargs|kwargs))?")\
```
do not make the additions between lines 1065 and 1087, and inside cross_entropy_replacement_2 , use the same pattern used in cross_entropy_replacement_1, meaning replace the following:
```
n_items = (\\9).get("num_items_in_batch", None) or (\\9).get("n_items", None)
```
with the pattern from cross_entropy_replacement_1:
```
all_locals = locals()
n_items = None
for __kwargs in all_locals.values():
    if type(__kwargs) is dict:
        n_items = __kwargs.get("num_items_in_batch", None) or __kwargs.get("n_items", None)
        break
```
This works because it iterates through all local variables to find any dict that might contain the num_items_in_batch parameter, making it work regardless of whether it's named kwargs, loss_kwargs, or even passed as a different parameter name entirely.

Tested successfully with qwen2-vl vision notebook.

what do you think?

PR #161, opened 2025-06-10
Update patching_utils.py

pluesclues:
This makes sure when loaded sequence classification score is not quantized, as transformers also should not quantize score when 4bit quantization is done.


Datta0 (approved):
LGTM

PR #164, opened 2025-06-10
trainer.eval_dataset relaced with value

ved1beta:
- Fix isinstance check for IterableDataset in dict eval datasets 

this should resolve  [#1019 - Unsolth No Validation Loss logged (possibly related to train_on_responses_only?)](https://github.com/unslothai/unsloth/issues/1019)


danielhanchen:
Ok I think this is fine! Thanks!

PR #167, opened 2025-06-15
Update compiler.py - resolution of the encoding related error on windows

FS-G:
On Linux and macOS, this defaults to UTF-8.
On many Windows systems, this defaults to cp1252.
So the current change will fix the encoding related error on windows machine especially the windows server


danielhanchen:
Oh thank you!

PR #176, opened 2025-06-24
Use pinned memory for async CUDA-to-CPU transfer

shubhamduhan107:
Ensures CUDA-to-CPU tensor transfers use pinned memory when non_blocking=True, preventing invalid argument errors and enabling safe async copying during LoRA weight merging.

### DIFF
--- a/unsloth_zoo/saving_utils.py
+++ b/unsloth_zoo/saving_utils.py
@@ -378,7 +378,9 @@ def _merge_and_overwrite_lora(save_directory, filename, lora_weights, output_dty
                     torch.save(W.to(output_dtype), temp_file, pickle_module = pickle, pickle_protocol = pickle.HIGHEST_PROTOCOL)
                     W = torch.load(temp_file, map_location = "cpu", mmap = True, weights_only = False)
                 else:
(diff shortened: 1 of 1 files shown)


danielhanchen:
@shubhamduhan107 Thanks - could you elaborate on invalid argument errors?

shubhamduhan60:
> 
@danielhanchen 

While calling model.save_pretrained_merged("model", tokenizer, save_method="merged_16bit") as part of saving the LoRA PEFT fine-tuned model and tokenizer, we occasionally encounter the following runtime error:

W = [W.to](http://w.to/)(device = "cpu", dtype = output_dtype, non_blocking = True)
RuntimeError: CUDA error: invalid argument
CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.

- The issue occurs when non_blocking=True is used in .to() calls during model weight transfer from GPU to CPU, which assumes the source tensor is in pinned (page-locked) memory

danielhanchen:
Ok thanks - I'll merge this since it doesn't affect functionality other than saving - thanks!

PR #183, opened 2025-06-29
Update compiler.py [FIX issue#2820]

bg-l2norm:
Problem (issue #2820 in the main repo)"
Unsloth previously used a plain string-replace to add `enable_gqa=` to every call of `scaled_dot_product_attention`.
If a model still passed the attention-mask **positionally** (e.g., Qwen-2.5-VL), the new kw-arg landed before that positional argument → Python raises "SyntaxError: positional argument follows keyword argument"

Fix:
This patch rewrites every SDPA call to:
- parse the argument list, respecting nested parentheses.
- if the mask is still positional, rewrites it as `attn_mask=…`.
- append `enable_gqa=self.num_key_value_groups != 1` **after** all existing arguments.
- leaves already-patched calls untouched.

No assumptions about the exact argument order is taken, so the code should be robust to future HF helpers that may reorder parameters

### DIFF
--- a/unsloth_zoo/compiler.py
+++ b/unsloth_zoo/compiler.py
@@ -228,47 +228,73 @@ def no_update_causal_mask(*args, **kwargs): return None
 # Patch SDPA
 def replace_with_grouped_query_attention(module, source):
     # All Unsloth Zoo code licensed under LGPLv3
-    if "enable_gqa" not in torch.nn.functional.scaled_dot_product_attention.__doc__: return source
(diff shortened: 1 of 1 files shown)


Datta0:
Hey @bg-l2norm the cause of the issue was fixed in https://github.com/unslothai/unsloth-zoo/pull/182
Can you please check if this fix works properly for you guys?

rolandtannous:
@SagarRavuri-Awone that's because colab notebook installs from pypi via pip install which is still not updated. please leave your feedback here ```https://github.com/unslothai/unsloth/issues/2820```

PR #190, opened 2025-07-03
Merge and Save - Windows safetensors mmap open file error fix

rolandtannous:
## Solves

https://github.com/unslothai/unsloth/issues/2157
https://github.com/unslothai/unsloth/issues/2813
https://discord.com/channels/1179035537009545276/1390396373660340314/1390427576698142870
https://discord.com/channels/1179035537009545276/1179777624986357780/1369207454638014494
https://discord.com/channels/1179035537009545276/1179777624986357780/1350976781787598918
https://discord.com/channels/1179035537009545276/1179777624986357780/1350949700320235540

## Problem
Users on Windows machines intermittently encounter a SafetensorError when using save_pretrained_merged():
```
SafetensorError: Error while serializing: IoError(Os { code: 1224, kind: Uncategorized, message: "The requested operation cannot be performed on a file with a user-mapped section open." })
```
Root Cause: Windows enforces strict file locking when files are memory-mapped. The existing code attempts to overwrite the same safetensors file that is still memory-mapped for reading, which Windows blocks as a security measure. This is not an issue in Linux.

- safe_open() creates memory-mapped tensors for efficiency
- save_file() tries to write to the same file while memory mapping is active
- Windows Error 1224 = "operation cannot be performed on a file with a user-mapped section open"

## Solution 

Implemented atomic file replacement using a temporary file approach that works cross-platform:
- Write merged tensors to a temporary .safetensors file in the same directory, then atomically replace the original file using os.replace()
- Added explicit garbage collection and brief delay on Windows to ensure file handles are released before attempting file replacement
- Fallback to copy-and-replace if atomic rename fails, with retry logic for Windows permission issues

## Tests

Tested on Windows 10 and 3090 GPU machine using unsloth/mistral-7b-instruct-v0.3-bnb-4bit as a base model.
Retested on H100 GPU with linux 
Special thanks to @marcandrelarochelle for helping me by testing on his windows machine

### DIFF
--- a/unsloth_zoo/saving_utils.py
+++ b/unsloth_zoo/saving_utils.py
@@ -414,7 +414,7 @@ def _merge_and_overwrite_lora(save_directory, filename, lora_weights, output_dty
 
     save_file(tensors, temp_filename_safetensors, metadata={"format": "pt"})  # Save to the temporary safetensors file
 
-    # Replace the original file with the temporary file
+    # Replace the temporary file with the original file
     try:
         os.replace(temp_filename_safetensors, filename_original)  # Attempt atomic rename
     except OSError as e:


unsloth_zoo/saving_utils.py
```diff
@@ -347,21 +347,35 @@ def create_lora_statistics(model, merge_into_original = False, return_state_dict
 pass
 
 
+import os
```
Datta0: There seem to be some duplicate imports (like os for eg)
Also should we move these to the top of the file?

unsloth_zoo/saving_utils.py
```diff
@@ -373,26 +387,70 @@ def _merge_and_overwrite_lora(save_directory, filename, lora_weights, output_dty
...
                 count += 1
-                W = _merge_lora(W, lora_stats, key)
+                W = _merge_lora(W, lora_stats, key)  # Assume _merge_lora is defined elsewhere
                 if psutil.virtual_memory().available <= limit:
-                    temp_file = tempfile.NamedTemporaryFile(suffix = ".pt")
-                    torch.save(W.to(output_dtype), temp_file, pickle_module = pickle, pickle_protocol = pickle.HIGHEST_PROTOCOL)
-                    W = torch.load(temp_file, map_location = "cpu", mmap = True, weights_only = False)
+                    temp_file = tempfile.NamedTemporaryFile(suffix=".pt", delete=False)  # Create temporary file
+                    temp_filename = temp_file.name  # Get temporary file name
+                    torch.save(W.to(output_dtype), temp_filename, pickle_module=pickle, pickle_protocol=pickle.HIGHEST_PROTOCOL)
```
Datta0, line 388: So we're essentially saving to a location than to a file object here?

unsloth_zoo/saving_utils.py
```diff
@@ -373,26 +387,70 @@ def _merge_and_overwrite_lora(save_directory, filename, lora_weights, output_dty
...
             else:
                 if lora_key in converted_lora_weights:
                     lora_stats_info = converted_lora_weights[lora_key]
+
             tensors[key] = W
-        pass
-    pass
 
-    save_file(tensors, filename, metadata = {"format": "pt"})
+    # CRITICAL: Force cleanup to release file handles on Windows
```
Datta0, line 406: So file.close() isn't instantaneous/atomic in Windows?

unsloth_zoo/saving_utils.py
```diff
@@ -373,26 +387,70 @@ def _merge_and_overwrite_lora(save_directory, filename, lora_weights, output_dty
...
+        gc.collect()
+        time.sleep(0.1)  # Give Windows a moment to release file handles
+
+    # Create a temporary file in the same directory for atomic rename
+    with tempfile.NamedTemporaryFile(suffix=".safetensors", dir=save_directory, delete=False) as tmpfile:
+        temp_filename_safetensors = tmpfile.name
+
+    save_file(tensors, temp_filename_safetensors, metadata={"format": "pt"})  # Save to the temporary safetensors file
+
+    # Replace the original file with the temporary file
```
Datta0: NIT: (comment) we replace the temp file with original file

unsloth_zoo/saving_utils.py
```diff
@@ -373,26 +387,70 @@ def _merge_and_overwrite_lora(save_directory, filename, lora_weights, output_dty
...
+    try:
+        os.replace(temp_filename_safetensors, filename_original)  # Attempt atomic rename
+    except OSError as e:
+        # If rename fails (e.g., due to permissions), fall back to copy and remove temporary file
+        print(f"Error renaming temporary file: {e}. Attempting copy and replace.")
+        import shutil
```
Datta0, line 423: NIT: Restructure imports

unsloth_zoo/saving_utils.py
rolandtannous: We tried that approach and it failed on testing. it seems that the file handle closes immediately, but since the tensors in the tensors dictionary still contain references to memory-mapped data from the original file, and Windows Blocks writing to a file with active memory mapping

unsloth_zoo/saving_utils.py
rolandtannous: yes will fix duplicates

unsloth_zoo/saving_utils.py
rolandtannous: Yes  We never simultaneously read and write the same file path, temp file has its own independent file handle so there are no file handle conflicts. os.replace (later int he code) is an atomic file operation. It's handled by OS kernel i think.

unsloth_zoo/saving_utils.py
```diff
@@ -373,26 +381,70 @@ def _merge_and_overwrite_lora(save_directory, filename, lora_weights, output_dty
...
+        gc.collect()
+        time.sleep(0.1)  # Give Windows a moment to release file handles
+
+    # Create a temporary file in the same directory for atomic rename
+    with tempfile.NamedTemporaryFile(suffix=".safetensors", dir=save_directory, delete=False) as tmpfile:
+        temp_filename_safetensors = tmpfile.name
+
+    save_file(tensors, temp_filename_safetensors, metadata={"format": "pt"})  # Save to the temporary safetensors file
+
+    # Replace the original file with the temporary file
```
Datta0: This comment needs to be adjusted
It should be 
"replace temporary file with original file" ig?

Datta0 (approved):
LGTM

PR #191, opened 2025-07-07
Add `typing.Any` import to compiled files

fizzAI:
Looks like `peft==0.16.0` started type hinting functions and we hardcode our own imports, so update that to pull in the type hints it wants now If `peft<0.16.0` or something non-PEFT is being patched then this'll just be a useless import that doesn't do anything but doesn't waste any time

### DIFF
--- a/unsloth_zoo/compiler.py
+++ b/unsloth_zoo/compiler.py
@@ -342,6 +342,7 @@ def create_new_function(
     imports += "import torch\n"
     imports += "import torch.nn as nn\n"
     imports += "from torch.nn import functional as F\n"
+    imports += "from typing import Any\n"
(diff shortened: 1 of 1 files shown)


unsloth_zoo/compiler.py
```diff
@@ -342,6 +342,7 @@ def create_new_function(
     imports += "import torch\n"
     imports += "import torch.nn as nn\n"
     imports += "from torch.nn import functional as F\n"
+    imports += "from typing import Any\n"
```
danielhanchen:
```suggestion
    imports += "from typing import Any, List, Optional, Tuple, Union, Dict, Set, Callable\n"
```

danielhanchen:
Thank you! I added more imports just in case!

PR #193, opened 2025-07-08
fix vram estimation when loading a 16bit model with load_in_4bit=True

yizhe-skyfall:


danielhanchen:
main now takes the explicit load_in_4bit = use_bitsandbytes argument into approximate_vllm_memory_usage in unsloth_zoo/vllm_utils.py, so I do not think this is needed any more. Sorry for the slow review, and thanks for putting it together.

danielhanchen:
main now takes load_in_4bit directly in approximate_vllm_memory_usage and passes use_bitsandbytes into it from load_vllm in unsloth_zoo/vllm_utils.py, so I do not think this is needed any more. Sorry for the slow review, and thanks for putting it together.

PR #197, opened 2025-07-10
Qwen25vl finfo fix

mmathew23:
Qwen 2.5 VL, Qwen 2 VL, and GLM 4 VL all have the potential to call torch.finfo on an attention_mask of type torch.bool. finfo raises an error when this happens (generally when cache_implementation is static) but we shouldn't need to force a float dtype for an attention_mask. Since there are three models that can apply to, I do create a find replace patch that runs inside compiler.py. The attention mask itself is created inside the GenerationMixin class using the new mask api.

the patch is admittedly brittle, but I do expect HF to fix at some point, and only 3 models potentially impacted.  

Here's a notebook showing qwen 2.5 VL working on inference and training. Before it used to fail with the finfo error on inference.

qwen 25 VL: https://colab.research.google.com/drive/10TM6vO5h-RV3U_NRiDhdp_vlfzLMEs_R?usp=sharing
llama to make sure it still works: https://colab.research.google.com/drive/1HEF4Mqh91vj80TYinSO2qcRrSnY8iqFp?usp=sharing

Edit: 
Before patch
image

After patch
image

### DIFF
--- a/unsloth_zoo/compiler.py
+++ b/unsloth_zoo/compiler.py
@@ -35,6 +35,7 @@ import time
 import logging
 import tempfile
 import sys
+import textwrap
 from .utils import (
     Version,
     is_main_process,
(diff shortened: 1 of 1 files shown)


unsloth_zoo/compiler.py
```diff
@@ -1229,6 +1230,47 @@ def patch_gradient_checkpointing(module, source):
...
+"""
+
+def patch_finfo_attention_mask_dtype_mismatch(module, source):
+    try:
+        old_block = textwrap.dedent(DTYPE_MISMATCH_FIND).strip()
+        new_block = textwrap.dedent(DTYPE_MISMATCH_REPLACE).strip()
+        if not old_block or not new_block:
+            return source
+
+        first_line = old_block.split('\n')[0]
```
Datta0, line 1253: NIT: I think we should check for the presence of second line as that is what contains finfo
I know the whole string might be together, but if we're anyway checking for single line, lets just check for the exact thing?

mmathew23: sounds good. updated so now the first line we check is the line that contains finfo.

PR #198, opened 2025-07-10
Patch decoder layers to have field for device

Datta0:
To patch PP, we need have info of which layer is on what device.

Needed for: unslothai/unsloth#2919

### DIFF
--- a/unsloth_zoo/patching_utils.py
+++ b/unsloth_zoo/patching_utils.py
@@ -220,7 +220,7 @@ def verify_and_set_device(module,):
     if len(set_of_devices) > 1:
         raise ValueError(f"Unsloth: All parameters of {module} should be on the same device")
     device = set_of_devices.pop()
-    module._per_layer_device = device
+    module._per_layer_device_index = device.index
 pass
 
 def patch_model_and_tokenizer(


unsloth_zoo/patching_utils.py
```diff
@@ -192,6 +192,36 @@ def patch_torch_compile(debug = False, O3 = False, ignore_errors = True):
...
+    pass
+    return x, found_layers
+pass
+
+
+def verify_and_set_device(module,):
+    """
+    Verify that all parameters of a module are on the same device.
+    """
+    set_of_devices = set([x.device for x in module.parameters()])
```
danielhanchen: can this be a generator set() and not set([])

unsloth_zoo/patching_utils.py
```diff
@@ -192,6 +192,36 @@ def patch_torch_compile(debug = False, O3 = False, ignore_errors = True):
     pass
 pass
 
+def get_model(model):
+    found_layers = False
+    x = model
+    while True:
```
danielhanchen, line 198: can we instead find `torch.nn.ModuleList` or something with layers as the name maybe via `model.named_modules()` maybe

PR #221, opened 2025-07-30
Fix: update from disable_adapter to disable_adapters 

wa008:
Solve unlsoth [issue-3064](https://github.com/unslothai/unsloth/issues/3064)

relate link: 
- [`disable_adapters` function](https://github.com/huggingface/transformers/blob/887054c714f594b1593f17e1860776dc97b862b1/src/transformers/integrations/peft.py#L365)
- https://github.com/unslothai/unsloth/pull/2736
- https://github.com/unslothai/unsloth/issues/2688

### DIFF
--- a/unsloth_zoo/rl_replacements.py
+++ b/unsloth_zoo/rl_replacements.py
@@ -367,12 +367,21 @@ def grpo_accumulated_loss(
     lm_head = trainer.model.get_output_embeddings().weight
 
     with torch.amp.autocast(device_type = trainer.model.device.type, dtype = trainer._autocast_dtype):
-        with torch.inference_mode(), trainer.accelerator.unwrap_model(trainer.model, keep_fp32_wrapper = False).disable_adapter():
-            ref_hidden_states = trainer.model(
-                input_ids = input_ids,
-                attention_mask = attention_mask,
-                logits_to_keep = logits_to_keep + 1,
-            ).logits
+        with torch.inference_mode():
+            unwarp_model = trainer.accelerator.unwrap_model(trainer.model, keep_fp32_wrapper = False)
+            if unwarp_model._hf_peft_config_loaded:
+                with unwarp_model.disable_adapters():
+                    ref_hidden_states = trainer.model(
+                        input_ids = input_ids,
+                        attention_mask = attention_mask,
+                        logits_to_keep = logits_to_keep + 1,
+                    ).logits
+            else:
+                ref_hidden_states = trainer.model(
+                    input_ids = input_ids,
+                    attention_mask = attention_mask,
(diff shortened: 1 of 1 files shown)


mmathew23:
Hi we'd like to be able to reproduce the issue on our side in order to evaluate the PR. Will also comment on the issue you opened, https://github.com/unslothai/unsloth/issues/3067, for further discussion.

unsloth_zoo/rl_replacements.py
```diff
@@ -367,12 +367,21 @@ def grpo_accumulated_loss(
...
 
     with torch.amp.autocast(device_type = trainer.model.device.type, dtype = trainer._autocast_dtype):
-        with torch.inference_mode(), trainer.accelerator.unwrap_model(trainer.model, keep_fp32_wrapper = False).disable_adapter():
-            ref_hidden_states = trainer.model(
-                input_ids = input_ids,
-                attention_mask = attention_mask,
-                logits_to_keep = logits_to_keep + 1,
-            ).logits
+        with torch.inference_mode():
+            unwarp_model = trainer.accelerator.unwrap_model(trainer.model, keep_fp32_wrapper = False)
```
mmathew23, line 371: instead of an if else, could we create a conditional context? Something like
```
from contextlib import nullcontext

unwrap_model = trainer.accelerator.unwrap_model(trainer.model, keep_fp32_wrapper=False)

try:
    disable_adapter_ctx = base.disable_adapter()
except AttributeError:
    disable_adapter_ctx = nullcontext()
```

unsloth_zoo/rl_replacements.py
```diff
@@ -367,12 +367,21 @@ def grpo_accumulated_loss(
...
+            if unwarp_model._hf_peft_config_loaded:
```
mmathew23, line 372: just a small nit: `unwarp_model` should be `unwrap_model`

unsloth_zoo/rl_replacements.py
wa008: Thanks for your comments!

1. I'm sorry, what is `base`? 
2. I think `try except` should be the last method to process this issue, I'm not sure if it's a good time to use it right now.

wa008:
> Hi @wa008, I made some code comments. But one issue is that it should not be calling `disable_adapters` with an `s`. This is not a context manager and just a function that returns a bool.
> 
> If you could update based on the comments I made we should be good to go.

Thanks for your reply and suggestions.

I’m sorry, I didn’t fully understand what you meant.

If you’d like, you can go ahead and submit a new PR for this issue—I’m completely fine with that.

mmathew23:
@wa008 the issue should now be fixed in the latest unsloth and unsloth zoo.

PR #222, opened 2025-07-30
Fix: Remove problematic vision data checks.

Dandandooo:
The `UnslothVisionDataCollator` should follow the conversational format and listen to the `type` field.

For example, in the following entry, it should understand that the type of the entry is `text`, but the old code would look for an image.
```py
{"type": "text", "text": "Hello", "image": None}
```
This scenario occurs when building datasets with HuggingFace, as it fills in None values to maintain column consistency.

It should be upon the user to make sure the `type` field is correct.


danielhanchen:
Thanks!

PR #224, opened 2025-08-02
Apply the correct image factor in smart_resize and adjust the final resizing step to ensure output dimensions are multiples of the image factor.

yorozcogonzalez:
**Issues**
- vision_utils currently uses a fixed IMAGE_FACTOR = 28, which works for most models but appears to be incompatible with Pixtral, where patch_size = 16.

- The final resizing step in the __call__ function (starting at line 442) does not consistently enforce that image dimensions are multiples of IMAGE_FACTOR.

**Proposal**
- Dynamically extract patch_size and spatial_merge_size from the model configuration, and compute the effective IMAGE_FACTOR as factor = patch_size * spatial_merge_size. This ensures compatibility across models by aligning image dimensions with model-specific requirements.

- Modify the final resizing logic to ensure the longer image side does not exceed the maximum allowed size and that both dimensions remain multiples of the computed IMAGE_FACTOR.

### DIFF
--- a/unsloth_zoo/vision_utils.py
+++ b/unsloth_zoo/vision_utils.py
@@ -194,6 +194,7 @@ pass
 
 def process_vision_info(
     conversations: Union[List[Dict], List[List[Dict]]],
+    factor: int = IMAGE_FACTOR,
 ) -> Tuple[Union[List[Image.Image], None], Union[List[Union[torch.Tensor, List[Image.Image]]], None]]:
     vision_infos = extract_vision_info(conversations)
     ## Read images or videos
@@ -201,7 +202,7 @@ def process_vision_info(
     video_inputs = []
     for vision_info in vision_infos:
         if "image" in vision_info or "image_url" in vision_info:
-            image_inputs.append(fetch_image(vision_info))
+            image_inputs.append(fetch_image(vision_info, factor))
         elif "video" in vision_info:
             video_inputs.append(fetch_video(vision_info))
         else:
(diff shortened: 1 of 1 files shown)


yorozcogonzalez:
Hi @mmathew23, thank you for your comment.

Yes, I have a standalone version of vision_utils.py in the following Google Colab notebook, in case yo want to examine the issue in more detail:

https://colab.research.google.com/drive/1Vm42JA9zXXAylKCsBIHve1YbcNGbF28U#scrollTo=Shw56XnoJjKH

There are two main points to consider:

1. **Patch size differences.**
 Most multimodal models currently available use an effective patch size of 28×28 pixels, with the exception of Pixtral, which uses 16×16 pixels. For this reason, using a general factor of 28 is not ideal for Pixtral. **I therefore suggest retrieving the image factor directly from the model parameters whenever possible**.

2. **Resizing inconsistencies** 
Even when fine-tuning models other than Pixtral, the current image resizing process fails under certain conditions. The smart_resize function correctly adjusts both image dimensions to the closest multiple of 28, while keeping them within defined maximum and minimum limits. However, the resulting dimensions can still exceed the model’s maximum allowed size. To address this, you have a second resizing step performed in line 435. At the moment, this only applies if the image “width” exceeds the model’s maximum size, which means the final “height” may still end up larger than allowed and not necessarily multiple of 28. **I therefore propose a fix to this second resizing step, ensuring that both image dimensions are always less than or equal to the maximum allowed size and multiples of the selected factor while keeping aspect ratio.**

I see your point about letting the image processor handle the final processing, and that might be solution. I’m just though this pre-stage was needed because I am not entirely sure how the image processor manages images that fall outside the specified parameter.

I’ve already tested the PR by fine-tuning **Mistral-Small-3.1–24B-Instruct-2503**, and that inspired me to put together this tutorial, which I’d appreciate any feedback you might have :)

https://medium.com/@yoelvis.orozco_42583/how-to-fine-tune-mistral-small-3-1-24b-instruct-2503-2a71e0a591f2

PR #225, opened 2025-08-06
enable cce with intel xpu

leizhenyuan:
As we tested on intel xpu, the ut for cce is all passed.


leizhenyuan:
@mmathew23  Shall we merge this pr?

PR #234, opened 2025-08-12
Update logging_utils.py

songminkyu:
The following error occurred while fine-tuning the Korean model, so code modification is required.

codec can't decode byte 0xf0 in position 30502: illegal multibyte sequence

unsloth_zoo/logging_utils.py
```
@functools.cache
def get_trl_metrics():
    # Gets metrics so we can output them in notebooks

    import trl.trainer
    trainers = dir(trl.trainer)
    trainers = [x for x in trainers if x.endswith("_trainer")]
    filepath = inspect.getfile(trl.trainer)
    filepath = os.path.split(filepath)[0]

    all_metrics = dict()
    for trainer in trainers:
        filename = os.path.join(filepath, f"{trainer}.py")
        if not os.path.exists(filename): continue
        with open(filename, "r", encoding='utf-8') as file: file = file.read()    # I think it needs to be changed to utf-8.
```

### DIFF
--- a/unsloth_zoo/__init__.py
+++ b/unsloth_zoo/__init__.py
@@ -14,7 +14,7 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-__version__ = "2025.8.4"
+__version__ = "2025.8.5"
 
 import os
 # Hugging Face Hub faster downloads


Datta0:
Hey @songminkyu , thanks for the contribution
Can you help us understand which trainer file was causing the error?

songminkyu:
To check the error, go to the url blow example/gpt-oss-fine-tune-kor.py
https://github.com/songminkyu/fine-tunning

songminkyu:
> Hey @songminkyu , thanks for the contribution Can you help us understand which trainer file was causing the error?

The base model used gpt-oss, and the learning dataset used maywell/korean_textbook from huggingface.
To check the error, go to the url blow example/gpt-oss-fine-tune-kor.py
https://github.com/songminkyu/fine-tunning

songminkyu:
fix Thank you for the correction.

songminkyu:
> The following error occurred while fine-tuning the Korean model, so code modification is required.
> 
> codec can't decode byte 0xf0 in position 30502: illegal multibyte sequence
> 
> unsloth_zoo/logging_utils.py
> 
> ```
> @functools.cache
> def get_trl_metrics():
>     # Gets metrics so we can output them in notebooks
> 
>     import trl.trainer
>     trainers = dir(trl.trainer)
>     trainers = [x for x in trainers if x.endswith("_trainer")]
>     filepath = inspect.getfile(trl.trainer)
>     filepath = os.path.split(filepath)[0]
> 
>     all_metrics = dict()
>     for trainer in trainers:
>         filename = os.path.join(filepath, f"{trainer}.py")
>         if not os.path.exists(filename): continue
>         with open(filename, "r", encoding='utf-8') as file: file = file.read()    # I think it needs to be changed to utf-8.
> ```

PR #237, opened 2025-08-13
fix: Add LRUCacheWorkerLoRAManager patching opt-out

bradhilton:
In ART we've found patching the V0 LRUCacheWorkerLoRAManager breaks engine LoRA adapter loading. Sensitive to the fact that this patch was probably added for a good reason, we propose adding an opt-out environment variable. If you think we could remove the patching altogether let us know. Thanks in advance! 🙏


danielhanchen:
Can also allow the env flag as well!

bradhilton:
Thank you @danielhanchen!

PR #252, opened 2025-08-22
Fix standardize_data_formats when using iterable datasets

mmathew23:
Standardzie_data_formats will now work with iterabledatasets.

test notebooks:
basic: https://colab.research.google.com/drive/1NcutJAkV_KBcL_Z-c9oOUyoA2ZnAse3n?usp=sharing
llama notebook that uses the function: https://colab.research.google.com/drive/1m9dd75T9Bu1JGltQxaulne1EV6dej1zv?usp=sharing


mmathew23:
this replaces PR #126 , and note that we use itertools.islice now since [:10] does not work on iterabledatasets.

gpt-oss test for good measure: https://colab.research.google.com/drive/1F17KON0gT9JYTSWm5bdP65bxbDOboqS9?usp=sharing

PR #258, opened 2025-08-29
Remove old version constraint in dependency list

timkpaine:
xref: https://github.com/conda-forge/unsloth_zoo-feedstock/pull/29


danielhanchen:
Ok this works - I just disallowed 4.70 because it was buggy

PR #260, opened 2025-09-02
Update logging_utils.py, put entropy in removed metrics

pluesclues:
So here is the cause of the issue: https://github.com/huggingface/trl/commit/da167d88b2d557b972cb608663344969bb909e56. SFT trainer added entropy as a new recorded metric recently and this pr just removes it from being recorded on jupyternotebooks.


pluesclues:
Here is the notebook used for testing with this PR: 

https://colab.research.google.com/drive/1gC9ehLzgQxPb_rPSedk67KClFNXnoAKL?usp=sharing

PR #281, opened 2025-09-15
fix Gemma3Attention.forward patch fail on transformers versions > 4.55.4

rolandtannous:
## Problem
Similar issue to https://github.com/unslothai/unsloth-zoo/pull/271 for GptOssAttention.Forward
Gemma3Attention.forward takes `past_key_values` as function argument for transformers versions > 4.55.4, while it accepts `past_key_value` for prior transformers versions (4.55.0 -> 4.55.4).

Current Python Patch functionality syntax leads to always patching version meant for previous transformers versions (<4.55.4)

and results in patch failing for transformers > 4.55.4 with log:
```
Unsloth: Skipped Gemma3Attention.forward
Reason: Parameter 'past_key_values' signature changed
```
## Solution

Integrated conditional patching to fix failed patch while keeping backwards compatibiility with
```
if transformers_version <= Version("4.55.4"):
        patch_function(transformers.models.gemma3.modeling_gemma3.Gemma3Attention, "forward", forward_backward_compat)
    else:
        patch_function(transformers.models.gemma3.modeling_gemma3.Gemma3Attention, "forward", forward)
```

Fix applied to both `def patch_Gemma3Attention()` and `def patch_Gemma3Attention_generic()`

### DIFF
--- a/unsloth_zoo/temporary_patches/gemma.py
+++ b/unsloth_zoo/temporary_patches/gemma.py
@@ -18,6 +18,9 @@ from typing import Any, List, Optional, Tuple, Union, Dict, Set, Callable
 import torch
 import torch.nn as nn
 import os
+from importlib.metadata import version as importlib_version
+from ..utils import Version
+transformers_version = Version(importlib_version("transformers"))
 from .common import TEMPORARY_PATCHES, torch_compile
 from .utils import (
     patch_function,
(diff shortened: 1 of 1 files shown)


danielhanchen:
main now handles both the past_key_value and past_key_values signatures in unsloth_zoo/temporary_patches/gemma.py via patch_function_past_key_values, so I do not think this is needed any more. Sorry for the slow review, and thanks for putting it together.

danielhanchen:
main now handles both the past_key_value and past_key_values signatures generically in unsloth_zoo/temporary_patches/gemma.py via patch_function_past_key_values, so I do not think this is needed any more. Sorry for the slow review, and thanks for putting it together.

PR #289, opened 2025-09-17
Mistral 3 Fast Inference

Datta0:


unsloth_zoo/empty_model.py
```diff
@@ -61,11 +61,10 @@ def compare_attributes(original_model, new_model):
         new_attrs = {attr for attr in dir(module) if not attr.startswith('_')}
         buffer_names = {name for name,_ in original_module.named_buffers(recurse=False)}
 
-        assert type(module) == type(original_module), f"Type mismatch for {name}: {type(module)} != {type(original_module)}"
 
         # Find missing attributes (in original but not in new)
         missing_in_new = orig_attrs - new_attrs
-        missing_in_new = missing_in_new - {'hf_device_map'}
+        missing_in_new = missing_in_new - {'hf_device_map', 'source_cls'}
```
danielhanchen, line 67: if this is a set, wrap in set ie set({...})

unsloth_zoo/empty_model.py
```diff
@@ -326,20 +321,23 @@ def _init_weights(self, module):
...
     if is_vision_model:
-        return create_empty_vision_model(config, dtype)
+        new_model, original_meta_model, num_layers = create_empty_vision_model(config, dtype)
     else:
-        return create_empty_causal_lm(config, dtype)
+        new_model, original_meta_model, num_layers = create_empty_causal_lm(config, dtype)
+
+    # Get layer names from config
+    layer_templates = get_model_layer_config(return_non_layered=False)
+    layer_names = sum(layer_templates.values(), [])
```
danielhanchen, line 337: Wait sum?

danielhanchen (changes requested):
Nice work!

unsloth_zoo/empty_model.py
Datta0: Oh basically this is like join but I wanted a single deflated list. So sum everything with empty list and you get a single list of everything :)

Datta0: its like a=[1,2,3]; b = [4,5,6] and doing `a+b` which is [1,2,3,4,5,6] or something

unsloth_zoo/empty_model.py
Datta0: Oh this is already a set right? I don't think we need to do a wrap

PR #302, opened 2025-09-23
[Part 1] Complete llama.cpp Integration Overhaul with Enhanced Build System and Multi-Modal Support

rolandtannous:
## PROBLEM

Requires https://github.com/unslothai/unsloth/pull/3356

 llama.cpp integration in unsloth-zoo suffered from critical failures:
-  **Monolithic conversion function**: Single `convert_to_gguf()` function attempted to handle base conversion and quantization simultaneously, making it impossible to generate multiple quantization formats efficiently or adapt to llama.cpp's new requirement that quantization must occur from high-precision formats only
- **Architecture detection breakage**: Regex-based extraction from llama.cpp converter script broke when llama.cpp reworked its class structures
 - **Module conflicts**: System-wide GGUF installations conflicted with llama.cpp's custom GGUF implementation containing model-specific extensions (Mistral classes, etc.)
 - **Build system failures**: Hardcoded sudo operations caused automatic permission denials, requiring manual installation for most users
- **No multimodal support**: Single conversion path failed on Vision-Language Models requiring dual-file output after llama.cpp restructuring
- **Missing dependency detection**: Build failures . Also missing`mistral_common` as dependency package broke llama.cpp

### Separated Quantization Pipeline

**New two-function architecture:**
- `convert_to_gguf()`: Focuses solely on high-precision base conversion (f32/f16/bf16), handling model loading, architecture detection, and initial GGUF generation
- `quantize_gguf()`: Dedicated quantization function using llama-quantize binary, optimized for parallel processing with thread allocation based on CPU count

This separation enables the two-stage architecture where one expensive base conversion supports multiple quantization targets.

### Dynamic Architecture Detection via Class Introspection

Previous regex-based architecture extraction broke when llama.cpp restructured its converter. New implementation uses proper class introspection:

- Downloads and loads converter script as Python module
- Extracts architectures from `ModelBase._model_classes[ModelType.TEXT]` and `ModelBase._model_classes[ModelType.MMPROJ]`
- Maintains separate text and vision architecture sets for proper model classification
- Adapts automatically to llama.cpp structural changes without code modifications

### Intelligent Permission and Dependency Management

Replaced hardcoded sudo operations with platform-aware permission handling. The new `check_build_requirements()` function validates essential tools (gcc, cmake, curl, git) and libraries (libcurl4-openssl-dev) using multiple detection methods (which, pkg-config, curl-config). 

Cloud environments (Colab/Kaggle) receive automatic sudo elevation with notification, while local environments require explicit user consent. Missing packages are identified upfront with specific installation guidance. Build commands are run as regular user as they do not require sudo privileges.

### GGUF Isolation with Context Management

Introduced `use_local_gguf()` context manager to resolve conflicts between system GGUF and llama.cpp's extended GGUF implementation:

```python
@contextlib.contextmanager
def use_local_gguf():
    # Temporarily prioritize llama.cpp/gguf-py
    # Remove conflicting system modules
    # Complete restoration after conversion
```

This ensures llama.cpp's custom GGUF extensions (Mistral-specific classes, additional model support) are used during conversion, preventing import errors and missing functionality.

### Multi-Modal Conversion Architecture

Enhanced `convert_to_gguf()` now handles Vision-Language Models through dual conversion paths:

- **VLM detection**: Checks architecture against vision_archs set
- **Dual output**: Generates both model.gguf and model-mmproj.gguf files
- **Fallback logic**: Vision models without MMPROJ support convert as text-only with user notification
- **Intelligent naming**: Maintains consistent file naming for inference tools expecting specific patterns

Models with text-only llama.cpp support despite having vision capabilities now convert successfully instead of failing.

### Enhanced Build System with Fallback Strategy

Use  dual-strategy build system:

- **Primary path**: Traditional make-based compilation
- **Automatic fallback**: CMAKE compilation when make fails (required for newer llama.cpp versions)
- **Binary validation**: Verifies executable existence and functionality

### MISC

- Added support for sentence-piece tokenizers (required by llama.cpp for some model conversions) by explicitly downloading `tokenizer.model` during merging process in `saving_utils` if applicable.
- Dropped Metadata regex parsing success tests for python native methods as they are less sensitive to banner or logging changes.

## TESTING
testing branches: Testing branches: https://github.com/unslothai/rolandtannous/unsloth-zoo@fix/llamacpp-compatibility-gguf-conversion and https://github.com/unslothai/rolandtannous/unsloth@fix/llamacpp-compatibility-gguf-conversion

### Build System Validation
- Tested missing dependency scenarios: system correctly identifies and guides installation of build-essential, cmake, curl
- Verified make→cmake fallback on systems requiring modern build process
- Confirmed permission handling: Colab auto-proceeds, local systems request consent

### Architecture Detection
- Validated against llama.cpp converter restructuring
- Tested proper classification of text models (LlamaForCausalLM) vs vision models (LlavaForConditionalGeneration)

### Multi-Modal Conversion
- Tested VLM models generating dual outputs (model.gguf + mmproj.gguf)
- Verified fallback behavior for vision models with text-only support
- Validated file naming conventions for multimodal models

### Environment Isolation
- Confirmed resolution of import conflicts between system and local GGUF versions
- Tested access to llama.cpp-specific GGUF extensions (Mistral classes)
- Validated complete environment restoration after conversion

### Models Tested
Comprehensive testing across model architectures: GPT-OSS, Llama 3.1/3.2, Pixtral (VLM), Gemma variants, Qwen series, Mistral, Phi models. Specific validation of GPT-OSS-20B on Colab T4. [Notebook](https://colab.research.google.com/drive/1KB_HsINAcQTdccHGajGMU5g8u-0LJrlO?usp=sharing)

### End-to-End Validation
- Local and Colab environments
- Save and Hub push operations  
- Post-conversion inference: llama-cli for text, llama-mtmd-cli for multimodal
- Ollama model creation and inference using generated Modelfiles

## Solves
https://github.com/unslothai/unsloth/issues/3348
https://github.com/unslothai/unsloth/issues/3297
https://github.com/unslothai/unsloth/issues/3090
https://github.com/unslothai/unsloth/issues/3229
https://github.com/unslothai/unsloth/issues/3215
https://github.com/unslothai/unsloth/issues/3202
https://github.com/unslothai/unsloth/issues/3194
https://github.com/unslothai/unsloth/issues/3133
https://github.com/unslothai/unsloth/issues/3124
https://github.com/unslothai/unsloth/issues/3040
https://github.com/unslothai/unsloth/issues/2984
https://github.com/unslothai/unsloth/issues/2950
https://github.com/unslothai/unsloth/issues/2860
https://github.com/unslothai/unsloth/issues/2667
https://github.com/unslothai/unsloth/issues/2580
https://github.com/unslothai/unsloth/issues/2526
https://github.com/unslothai/unsloth/issues/2478
https://github.com/unslothai/unsloth/issues/2399
https://github.com/unslothai/unsloth/issues/2370
https://github.com/unslothai/unsloth/issues/2365
https://github.com/unslothai/unsloth/issues/2360
https://github.com/unslothai/unsloth/issues/2326
https://github.com/unslothai/unsloth/issues/2321
https://github.com/unslothai/unsloth/issues/2290
https://github.com/unslothai/unsloth/issues/2209
https://github.com/unslothai/unsloth/issues/2193
https://github.com/unslothai/unsloth/issues/2115
https://github.com/unslothai/unsloth/issues/2058
https://github.com/unslothai/unsloth/issues/2007
https://github.com/unslothai/unsloth/issues/1917
https://github.com/unslothai/unsloth/issues/1905
https://github.com/unslothai/unsloth/issues/1903
https://github.com/unslothai/unsloth/issues/1846
https://github.com/unslothai/unsloth/issues/1781
https://github.com/unslothai/unsloth/issues/1729
https://github.com/unslothai/unsloth/issues/1721
https://github.com/unslothai/unsloth/issues/1645
https://github.com/unslothai/unsloth/issues/1610
https://github.com/unslothai/unsloth/issues/1546
https://github.com/unslothai/unsloth/issues/1504
https://github.com/unslothai/unsloth/issues/965
https://github.com/unslothai/unsloth/issues/835
https://github.com/unslothai/unsloth/issues/748
https://github.com/unslothai/unsloth/issues/785
https://github.com/unslothai/unsloth/issues/2098
https://github.com/unslothai/unsloth/issues/3050

### DIFF
--- a/unsloth_zoo/llama_cpp.py
+++ b/unsloth_zoo/llama_cpp.py
@@ -104,6 +104,8 @@ def use_local_gguf():
     original_modules = set(sys.modules.keys())
     gguf_py_path = os.path.join("llama.cpp", "gguf-py")
 
+    original_gguf_modules = {}
+
     try:
         # Add gguf-py to sys.path if it exists
         if os.path.exists(gguf_py_path):
@@ -113,6 +115,7 @@ def use_local_gguf():
             # Remove system gguf modules to force reimport
             gguf_modules = [key for key in sys.modules.keys() if key.startswith('gguf')]
             for module in gguf_modules:
+                original_gguf_modules[module] = sys.modules[module]  # Store original
                 del sys.modules[module]
                 logger.debug(f"Removed system module {module}")
 
@@ -128,6 +131,11 @@ def use_local_gguf():
         for module in gguf_modules_to_remove:
             del sys.modules[module]
             logger.debug(f"Cleaned up module {module}")
+
+        # Restore original gguf modules
+        for module_name, module_obj in original_gguf_modules.items():
+            sys.modules[module_name] = module_obj
+            logger.debug(f"Restored original module {module_name}")
 
         logger.debug("Restored original Python environment")
 pass


danielhanchen (changes requested):
Nice work!

rolandtannous:
> @mmathew23 @Datta0 Can you guys also review this - appreciate it :)

note that you'll need to install mistral_common manually to make it work, since entries to pyproject.toml don't get installed as dependencies if you're not installing via pypi

unsloth_zoo/llama_cpp.py
```diff
@@ -73,32 +88,80 @@
...
+        sys.path[:] = original_sys_path
+
+        # Remove any new gguf modules that were imported
+        new_modules = set(sys.modules.keys()) - original_modules
+        gguf_modules_to_remove = [m for m in new_modules if m.startswith('gguf')]
+        for module in gguf_modules_to_remove:
+            del sys.modules[module]
+            logger.debug(f"Cleaned up module {module}")
+
+        logger.debug("Restored original Python environment")
```
mmathew23, line 140: Is the intention to restore the originally deleted modules? It looks like sys.path is restored, new modules are deleted, but I don't think the original modules del in the try are restored.

unsloth_zoo/llama_cpp.py
```diff
@@ -73,32 +88,80 @@
...
-            "Please install llama.cpp manually via https://docs.unsloth.ai/basics/troubleshooting-and-faqs#how-do-i-manually-save-to-gguf"
-        )
-    # x = f"{'sudo ' if sudo else ''}apt-get install {package} -y"
-    x = f"{''}apt-get install {package} -y"
+    # Choose package manager based on system type
+    if system_type == "rpm":
+        pkg_manager = "yum" if os.path.exists('/usr/bin/yum') else "dnf"
+        install_cmd = f"{'sudo ' if sudo else ''}{pkg_manager} install {package} -y"
+    else:  # Default to debian/apt-get
+        install_cmd = f"{'sudo ' if sudo else ''}apt-get install {package} -y"
```
mmathew23, line 150: Just to confirm should we allow sudo or not? I remember there were user complaints at one point.

rolandtannous: yes i remember the user complaints as well . but the difference now , is that it asks the user for permission first (Daniel added this and I left the `user permission` mechanism in) but there is no way for a user on linux, even if you are part of the sudoer group to install the system level packages required by `llama.cpp` without prefixing your command with sudo. I think asking for user permission is fair enough . Users can always choose NO, and unsloth will exit the installation. Users can then install the packages manually. But there is no actual risk here, because using sudo , you are not changing the user's assigned permissions : They are either part of the sudoer group and can use sudo or they aren't.

rolandtannous: it's on line 145 in the new file . I can't seem to get a permalink , so i'll copy paste the code block
```
acceptance = input(f"Missing system packages. We need to execute `{install_cmd}` - do you accept? Press ENTER. Type NO if not.")
        if "no" in str(acceptance).lower():
            raise RuntimeError(
                f"Unsloth: Execution of `{install_cmd}` was cancelled!\n"\
                "Please install llama.cpp manually via https://docs.unsloth.ai/basics/troubleshooting-and-faqs#how-do-i-manually-save-to-gguf"
            )
    with subprocess.Popen(install_cmd, shell = True, stdout = subprocess.PIPE, stderr = subprocess.STDOUT) as sp:
        for line in sp.stdout:
            line = line.decode("utf-8", errors = "replace").rstrip()

            if "Permission denied" in line or "not open lock file" in line or "are you root?" in line or "fatal" in line:
                sp.terminate()
                raise RuntimeError(f"[FAIL] Unsloth: Permission denied when installing package {package}\n"\
                                   "This operation requires elevated sudo/root permissions. Please manually install missing packages and retry again"
```

This was previously introduced by Daniel and I left it in place

unsloth_zoo/llama_cpp.py
rolandtannous: Great catch! You're right - there's a bug in the context manager. The original gguf modules that get deleted in the try block are never restored in the finally block, only the new ones are cleaned up.

I'll update shortly.

Thanks!

rolandtannous: fixed in this commit
https://github.com/unslothai/unsloth-zoo/pull/302/commits/d8ff7a936a77401115b77ed900f198b744e094f4

PR #305, opened 2025-09-26
Update compiler.py: resolve AttributeError when loading Qwen3 models with LoRA adapters

MengAiDev:
## PR Description
Fixes [#3376](https://github.com/unslothai/unsloth/issues/3376)

This PR resolves an `AttributeError` that occurs when loading Qwen3 models with LoRA adapters using `FastLanguageModel.from_pretrained()`. The error was caused by unsafe attribute access during model compilation.

## Problem
When compiling certain model types (particularly Qwen3), the code attempted to access attributes that don't exist in the transformers module, specifically:
```python
AttributeError: module 'transformers.models.bit.modeling_bit' has no attribute 'Linear'
```

This prevented users from directly loading models with LoRA adapters, forcing them to use a two-stage loading process.

### DIFF
--- a/unsloth_zoo/compiler.py
+++ b/unsloth_zoo/compiler.py
@@ -2235,6 +2235,7 @@ def unsloth_compile_transformers(
                 if x in source_code:
                     can_remove = False
                     break
+            pass
             if can_remove: remove_causal_masks.append(module)
         pass
     pass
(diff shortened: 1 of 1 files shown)


danielhanchen:
main now guards this in unsloth_zoo/compiler.py, where the class lookup in the causal mask loop is wrapped in a try/except and skipped when the attribute is missing, so I do not think this is needed any more. Sorry for the slow review, and thanks for putting it together.

danielhanchen:
main now guards this same lookup in unsloth_zoo/compiler.py, wrapping the eval of the module attribute in a try/except before the causal-mask check, so I do not think this is needed any more. Sorry for the slow review, and thanks for putting it together.

PR #330, opened 2025-10-20
 fix gpt oss memory calculation for intel device

leizhenyuan:
fix gpt oss memory calculation for intel device


danielhanchen:
Thanks!

PR #334, opened 2025-10-21
return local model in determine_base_model_source with any quantization

noah1510:
This returns the found local model regardless of quantization.
The change also means that local base models and hf base models have more similar behaviour.
This should prevent some confution regarding the error messages if local 4bit models are returned.

I would suggest removing the special case for mxfp4 models, since the caller has to check the returned quant type in any case, but I am not too sure if that is actually desired.

### DIFF
--- a/unsloth_zoo/saving_utils.py
+++ b/unsloth_zoo/saving_utils.py
@@ -2068,22 +2068,27 @@ def determine_base_model_source(model_name, token=None):
     if local_path:
         local_is_quantized, local_quant_type = check_model_quantization_status(local_path)
 
-    # Priority 1: Local unquantized OR Local mxfp4
-    if local_path and (not local_is_quantized or local_quant_type == "mxfp4"):
-        if not local_is_quantized:
-            return (local_path, True, "local_unquantized", False, None)
-        else:  # local_quant_type == "mxfp4"
-            return (local_path, True, "local_mxfp4", True, "mxfp4")
+    # Priority 1: Local unquantized
+    if local_path and not local_is_quantized:
+        return (local_path, True, "local_unquantized", False, None)
 
-    # Priority 2: HF unquantized
+    # Priority 2: Local mxfp4
+    if local_path and local_is_quantized and local_quant_type == "mxfp4":  # local_quant_type == "mxfp4"
+        return (local_path, True, "local_mxfp4", True, "mxfp4")
+
+    # Priority 3: HF unquantized
     if hf_exists and not hf_is_quantized:
         return (model_name, False, "HF_unquantized", False, None)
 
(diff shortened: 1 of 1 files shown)


unsloth_zoo/saving_utils.py
```diff
@@ -2068,22 +2068,27 @@ def determine_base_model_source(model_name, token=None):
...
-            return (local_path, True, "local_mxfp4", True, "mxfp4")
-
-    # Priority 2: HF unquantized
+    # Priority 1: Local unquantized
+    if local_path and not local_is_quantized:
+        return (local_path, True, "local_unquantized", False, None)
+
+    # Priority 2: Local mxfp4
+    if local_path and local_is_quantized and local_quant_type == "mxfp4":  # local_quant_type == "mxfp4"
+        return (local_path, True, "local_mxfp4", True, "mxfp4")
```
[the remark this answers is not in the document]
rolandtannous: these two blocks are actually functionally equivalent.

unsloth_zoo/saving_utils.py
```diff
@@ -2068,22 +2068,27 @@ def determine_base_model_source(model_name, token=None):
...
+    # Priority 4: HF quantized (covers both "both quantized" and "just HF quantized")
     if hf_exists and hf_is_quantized:
         return (model_name, False, f"HF_{hf_quant_type}", True, hf_quant_type)
 
-    # Priority 4: Nothing suitable found
+    # Priority 5: Local other quantization
+    if local_path and local_is_quantized:
+        return (local_path, True, f"local_{local_quant_type}", True, local_quant_type)
+
+    # Priority 6: Nothing suitable found
```
rolandtannous, line 2091: we've intentionally restricted using a 4bit bnb base model when the merge method is 16bit , because the upcasting from 4bit-> 32 bit to merge will result in a high precision loss, so we don't want the case for Priotrity 5 when we are merging in 16bits.
in fact, we explicitely prohibit this and issue a warning. a 4 bit model should be mostly used for inference , and a 16bit base model should be available if we wish to finetune.

noah1510: While I understand the reason behind restricting merging of 4bit models, this is not the place where it should happen.
The code calling this has to check the quantization anyway, since a remote model is always returned if found, even if that is 4bit.
Filtering here is just reduntant and makes it more difficult to find out why the merging isn't working, since you get a model not found error instead of a model not compatible error.

rolandtannous: I agree with a better error message if that's the intended purpose of this change, cause  the priority logic seems almost the same (end point is no merging happens) , and i wasn't sure whether there was a functional bug motivating the change, hence had to ask. LGTM if that's the case.

PR #348, opened 2025-11-07
packed seq lengths token count correction

djsaunde:
Should be merged alongside [#3525](https://github.com/unslothai/unsloth/pull/3525).

### DIFF
--- a/unsloth_zoo/loss_utils.py
+++ b/unsloth_zoo/loss_utils.py
@@ -351,6 +351,8 @@ def _unsloth_get_batch_samples(self, epoch_iterator, num_batches, device = None,
                 count = token_count.sum()
                 seq_lengths = x.get("packed_seq_lengths")
                 if seq_lengths is not None:
+                    # When packing N sequences, there are N-1 internal boundaries
+                    # that shouldn't be counted as valid training positions
                     count -= torch.count_nonzero(seq_lengths > 0).item() - 1
                 token_counts.append(count)
             pass


unsloth_zoo/loss_utils.py
```diff
@@ -348,7 +348,11 @@ def _unsloth_get_batch_samples(self, epoch_iterator, num_batches, device = None,
                     token_type_ids = x["token_type_ids"]
                     mark_static (token_type_ids, 0)
                     mark_dynamic(token_type_ids, 1)
-                token_counts.append(token_count.sum())
+                count = token_count.sum()
+                seq_lengths = x.get("packed_seq_lengths")
+                if seq_lengths is not None:
+                    count -= torch.count_nonzero(seq_lengths > 0).item() - 1
```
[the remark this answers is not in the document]
djsaunde: If seq_lengths exist, why would none of them be non-zero? Silly

unsloth_zoo/dataset_utils.py
```diff
@@ -364,8 +364,12 @@ def _train_on_responses_only(examples):
...
     from transformers import DataCollatorForSeq2Seq
-    if hasattr(trainer, "data_collator") and \
-        not isinstance(trainer.data_collator, DataCollatorForSeq2Seq):
+    packing_enabled = getattr(trainer.args, "packing", False)
+    if (
+        hasattr(trainer, "data_collator")
+        and not isinstance(trainer.data_collator, DataCollatorForSeq2Seq)
+        and not packing_enabled
+    ):
         trainer.data_collator = DataCollatorForSeq2Seq(tokenizer = tokenizer)
```
djsaunde, line 373: oh @danielhanchen I wanted to ask why we do this, and if there's anything that might be breaking about my change? FWIW I haven't encountered any issues with this change when running all the notebooks in the unsloth main repo README.

danielhanchen:
@codex review

unsloth_zoo/loss_utils.py
danielhanchen, line 356:
```suggestion
                    # When packing N sequences, there are N-1 internal boundaries
                    # that shouldn't be counted as valid training positions
                    count -= torch.count_nonzero(seq_lengths > 0).item() - 1
```

unsloth_zoo/dataset_utils.py
danielhanchen: It's because we must use `DataCollatorForSeq2Seq` for training on completions only since `DataCollatorForSeq2Seq` allows us to have `input_ids` and `labels` separately. The other `DataCollatorForLanguageModel` doesn't allow `labels`.

djsaunde: hmm I think we should actually use `DataCollatorForLanguageModel` and pass `completion_only_loss = True` there. I don't think we should use a custom (non-default) collator here. that breaks packing

djsaunde: hmm I think we should actually use DataCollatorForLanguageModel and pass completion_only_loss = True there. I don't think we should use a custom (non-default) collator here. that breaks packing

PR #351, opened 2025-11-11
Enable FP8 + RL training

andrewor14:
This commit works around a bug in PyTorch that doesn't let us use inference mode with tensor subclasses. TorchAO's fp8 quantization leverages tensor subclasses so we are affected by this error. For more details, see https://github.com/pytorch/pytorch/issues/164872. The workaround is to use `torch.no_grad` in this case instead of `torch.inference_mode`.


danielhanchen:
@andrewor14 Very nice! Good work!

PR #357, opened 2025-11-21
Chunk Across Batch and Context length for logprob calculations for grpo 

pluesclues:
Relies on: https://github.com/unslothai/unsloth/pull/3628

### DIFF
--- a/unsloth_zoo/rl_replacements.py
+++ b/unsloth_zoo/rl_replacements.py
@@ -73,10 +73,7 @@ def chunked_hidden_states_selective_log_softmax(
     logit_softcapping: float = 0.0,
     temperature: float = 1.0,
 ) -> torch.Tensor:
-    """
-    More memory efficient by chunking on (bsz+qlen) dimension.
-    Exactly equivalent to the non-chunked version.
-    """
+    # All Unsloth Zoo code licensed under AGPL3 
     flat_hidden_states = hidden_states.reshape(-1, hidden_states.shape[-1]) 
     flat_index = index.reshape(-1)                                    
 


unsloth_zoo/rl_replacements.py
```diff
@@ -581,78 +575,158 @@ def grpo_accumulated_loss(
...
+
+    input_ids_chunks = torch.chunk(input_ids, chunks=B, dim=0)
+    attention_mask_chunks = torch.chunk(attention_mask, chunks=B, dim=0)
+    completion_ids_chunks = torch.chunk(completion_input_ids, chunks=B, dim=0)
+
+    def chunk_optional(tensor, chunks):
+        if tensor is None:
+            return [None] * chunks
+        return torch.chunk(tensor, chunks=chunks, dim=0)
+
```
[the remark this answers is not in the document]
pluesclues: I removed one of these chunk_optionals.

unsloth_zoo/rl_replacements.py
```diff
@@ -496,7 +482,7 @@ def accumulate_chunk(
             #     mark_dynamic(old_hidden_states_j)
             # mark_dynamic(input_ids_j)
             # mark_dynamic(mask_j)
-            
+            #breakpoint()
```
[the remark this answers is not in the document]
pluesclues: I removed this breakpoint just now.

unsloth_zoo/rl_replacements.py
```diff
@@ -190,13 +238,11 @@ def grpo_update_SamplingParams(SamplingParams, generation_kwargs, vllm_sampling_
...
-
-
 # Custom compiled GRPO loss - creates 3 Triton kernels
 def grpo_compute_loss(
-    ref_logits,
-    new_logits,
-    old_logits,
+    ref,
+    new,
+    old,
```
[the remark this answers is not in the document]
pluesclues: Changed naming covnentions for these just now.

unsloth_zoo/rl_replacements.py
```diff
@@ -559,6 +545,14 @@ def grpo_accumulated_loss(
...
+    sampling_per_token_logps = kwargs.get("sampling_per_token_logps", None) if getattr(trainer, "vllm_importance_sampling_correction", False) else None
+    temperature = kwargs.get("temperature", 1.0)
+    logit_scale_multiply = kwargs.get("logit_scale_multiply", 0.0)
+    logit_scale_divide   = kwargs.get("logit_scale_divide", 0.0)
+    logit_softcapping    = kwargs.get("logit_softcapping", 0.0)
+
+    prev_max_left_pad    = kwargs.get("max_left_pad", None)
+
     #delete this from kwargs so less issues 
     sampling_per_token_logps = kwargs.pop("sampling_per_token_logps", None)
```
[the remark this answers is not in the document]
pluesclues: I just changed the pop kwargs to be equal to `_` since this would be a bug, I do not fully recall why that was there. will investigate further into that.

unsloth_zoo/rl_replacements.py
```diff
@@ -126,6 +174,7 @@ def left_pack_padding(tensor: torch.Tensor, pad_id: int) -> torch.Tensor:
 pass
 RL_REPLACEMENTS["left_pack_padding"] = left_pack_padding
 
+import torch
```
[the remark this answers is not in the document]
pluesclues: I removed this for code cleanliness

unsloth_zoo/rl_replacements.py
```diff
@@ -223,53 +263,15 @@ def grpo_compute_loss(
...
-        pass
         if use_vllm and sampling_per_token_logps is not None:
             #must filter out extra prompt tokens in begining after making input_ids left padded
             importance_sampling_ratio = torch.exp((old * mask) - sampling_per_token_logps)
             importance_sampling_ratio = torch.clamp(
                 importance_sampling_ratio, max=vllm_importance_sampling_cap
             )
-        pass
     pass
+    # pass
```
danielhanchen: Remove

unsloth_zoo/rl_replacements.py
```diff
@@ -559,8 +542,16 @@ def grpo_accumulated_loss(
...
     image_sizes = kwargs.get('image_sizes',None)
+    sampling_per_token_logps = kwargs.get("sampling_per_token_logps", None) if getattr(trainer, "vllm_importance_sampling_correction", False) else None
+    temperature = kwargs.get("temperature", 1.0)
+    logit_scale_multiply = kwargs.get("logit_scale_multiply", 0.0)
+    logit_scale_divide   = kwargs.get("logit_scale_divide", 0.0)
+    logit_softcapping    = kwargs.get("logit_softcapping", 0.0)
+
+    prev_max_left_pad    = kwargs.get("max_left_pad", None)
+
     #delete this from kwargs so less issues
```
danielhanchen: Spacing and capitalize and mention we enable by default

unsloth_zoo/rl_replacements.py
```diff
@@ -581,78 +572,153 @@ def grpo_accumulated_loss(
 
         max_left_pad = max(left_pad_tokens_per_prompt).item()
 
+        if max_left_pad < max(prev_max_left_pad).item() and (logits_to_keep +max_left_pad) != old_logps.shape[1]:
```
danielhanchen: Can you use torch ops

unsloth_zoo/rl_replacements.py
```diff
@@ -581,78 +572,153 @@ def grpo_accumulated_loss(
...
+                    logits_to_keep = logits_to_keep + 1, 
+                ).logits
+                
+                new_hidden_states_chunk = new_hidden_states_chunk[:, :-1, :]
+
+            logprobs_chunk = chunked_hidden_states_selective_log_softmax(
+                new_hidden_states_chunk,
+                lm_head, 
+                completion_ids, 
+                chunks = 8,
```
danielhanchen: Does increasing this reduce VRAM more but makes it slower?

unsloth_zoo/rl_replacements.py
```diff
@@ -581,78 +572,153 @@ def grpo_accumulated_loss(
 
         max_left_pad = max(left_pad_tokens_per_prompt).item()
 
+        if max_left_pad < max(prev_max_left_pad).item() and (logits_to_keep +max_left_pad) != old_logps.shape[1]:
+            max_left_pad =  max(prev_max_left_pad).item()
```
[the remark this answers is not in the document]
pluesclues: Given that we are working with the LLM path, we will always have this variable, this should won't ever be None.

unsloth_zoo/rl_replacements.py
pluesclues: Done

unsloth_zoo/rl_replacements.py
pluesclues: Yes.

unsloth_zoo/rl_replacements.py
pluesclues: I understand the spacing, but what needs to be capitalized?

unsloth_zoo/rl_replacements.py
pluesclues: I changed `max` to use `torch.max` instead I am not sure if this is what you meant though.

unsloth_zoo/rl_replacements.py
```diff
@@ -579,80 +568,155 @@ def grpo_accumulated_loss(
...
-        lm_head,
-        completion_input_ids,
-        completion_mask,
-        advantages,
-        trainer.beta,
-        trainer.accelerator.scaler,
-        n_chunks,
-        kwargs # pass kwargs as a dict
-    )
+        for (
```
Datta0: This looks very similar [to this](https://github.com/unslothai/unsloth/pull/3628/files#diff-6c2de9d986f1bfea9d390a1d7980faf9b706afae4a38b962131d164f8fe79fdeR635-R672)
Why do we need it at both places?

unsloth_zoo/rl_replacements.py
```diff
@@ -579,80 +568,155 @@ def grpo_accumulated_loss(
...
+                logit_scale_divide = logit_scale_divide,
+                logit_softcapping = logit_softcapping, 
+                temperature = temperature
+            )
+            
+            all_logprobs_list.append(logprobs_chunk)
+
+        new_logprobs = torch.cat(all_logprobs_list, dim=0)
+
+        loss, completion_length, mean_kl, delta, flat_is_ratio = UnslothEfficientGRPO.apply(
```
Datta0, line 901: Shouldn't this be within the for loop for chunked loss calculation? At this point once we materialise all the logits, with seemingly no changes to `UnslothEfficientGRPO`, aren't we just doing what we always were?

unsloth_zoo/rl_replacements.py
pluesclues: So we have to itterate through it again because of the strange way TRL deals with gradient accumulation here vs in `_generate_score_completions`. In that function, recall that TRL materializes the ENTIRE effective batch so if we have `train_per_device_batch_size=4` and `gradient_accumulation=2` the incoming `old_logprob` tensor for example is `[train_per_device_batch_size*gradient_accumulation, context_length]` large. On the other hand, in this function when we receive `[train_per_device_batch_size, context_length]` in this function.

unsloth_zoo/rl_replacements.py
danielhanchen: Yes torch.max works

unsloth_zoo/rl_replacements.py
danielhanchen:
```
# Delete
```

unsloth_zoo/rl_replacements.py
danielhanchen: Ok hm interesting

unsloth_zoo/rl_replacements.py
pluesclues: Done.

unsloth_zoo/rl_replacements.py
pluesclues: The difference here is that previously, all of the hidden states would be materialized before getting sent into the compiled autograd function. Now what we can do is process the hidden_states batch wise to which instead of `[bsz, cl, hidden_states_dim]` we materialize `[1, cl, hidden_states_dim]` and the most VRAM ever materialized in this loop compared to before will be `[1, cl//num_chunks, vocab_dim]` rather than  `[1, cl, vocab_dim]` previously because of this https://github.com/pluesclues/unsloth-zoo/blob/9bb5a9628a6b93c212ddeaedfe3e76517361b37d/unsloth_zoo/rl_replacements.py#L67-L109.

unsloth_zoo/rl_replacements.py
```diff
@@ -583,83 +613,272 @@ def grpo_accumulated_loss(
...
+        """
+        Manual Gradient Checkpointing/CPU Offloading for Log Softmax.
+        """
+        @staticmethod
+        def forward(ctx, hidden_states, lm_head, index, chunks, 
+                    logit_scale_multiply, logit_scale_divide, 
+                    logit_softcapping, temperature):
+            
+            ctx.saved_hidden_states = to_device(hidden_states, "cpu", non_blocking=True)
+            ctx.device = hidden_states.device
```
Datta0, line 776: Are you sure this works? Cuz once you do `to_device` which I see is inplace, `hidden_states.device` would be `cpu` right?

unsloth_zoo/rl_replacements.py
```diff
@@ -583,83 +613,272 @@ def grpo_accumulated_loss(
...
+            return output
+
+        @staticmethod
+        def backward(ctx, grad_output):
+            # 1. Restore hidden_states from CPU to GPU
+            hidden_states = to_device(ctx.saved_hidden_states, ctx.device)
+            hidden_states = hidden_states.to(ctx.dtype)
+            hidden_states.requires_grad_(True)
+            
+            lm_head = ctx.lm_head.detach()
```
Datta0: Can you please explain the two lines
Why detach and then set requires_grad?

pluesclues: I believe we first detach in order to get rid of any hooks on the hidden states and lm head, and then aftewrards we set it to require grad as we are about to recompute the forward pass and get the gradients with `torch.autograd.backward`.

unsloth_zoo/rl_replacements.py
pluesclues: I am positive this works, as we use it for gradient checkpointing to offload the internal model activtions, which this is basd off of. https://github.com/unslothai/unsloth-zoo/blob/main/unsloth_zoo/gradient_checkpointing.py#L146-L176

danielhanchen:
Pushed a fix for two issues found during testing:

**Issue 1: AttributeError when old_logps is None**
- Location: Line 662
- Error: `AttributeError: 'NoneType' object has no attribute 'shape'`
- Fix: Added `old_logps is not None and` guard before accessing `old_logps.shape[1]`

**Issue 2: Shape mismatch in non-vLLM path**
- Location: After completion_mask creation
- Error: `RuntimeError: The size of tensor a (X) must match the size of tensor b (Y)`
- Cause: Precomputed `ref_logps` and `old_logps` were created with a different `max_left_pad` than the current batch
- Fix: Added alignment using `align_logprobs_with_mask()` for both `ref_logps` and `old_logps`, matching how `sampling_per_token_logps` is already handled

**Test Results (Llama-3.2-1B)**

| Path | Main Branch | PR (Fixed) | Speedup |
|------|-------------|------------|---------|
| non-vLLM | 132.0s | 116.6s | ~12% faster |
| vLLM | 99.2s | 83.8s | ~16% faster |

Both paths now work correctly with similar gradient norms.

danielhanchen:
Updated the fix based on your feedback. You were right that `align_logprobs_with_mask` is not the correct approach here.

**New approach:**

Instead of computing `max_left_pad` from the current batch and comparing with precomputed logprobs, we now derive it directly from the precomputed logprobs shape:

```python
if old_logps is not None:
    max_left_pad = old_logps.shape[1] - logits_to_keep
elif ref_logps is not None:
    max_left_pad = ref_logps.shape[1] - logits_to_keep
else:
    max_left_pad = torch.max(left_pad_tokens_per_prompt).item()
```

**Why this works:**

- Precomputed logprobs have shape `[batch, logits_to_keep + prev_max_left_pad]`
- Extracting `shape[1] - logits_to_keep` gives us `prev_max_left_pad`
- New logprobs are then computed with the same `max_left_pad`, ensuring shapes match

**Benefits:**

- No `.item()` call when precomputed logprobs exist (avoids GPU sync)
- Handles both `old_logps` and `ref_logps` being None
- Simpler logic, no complex conditionals

**Test results (Llama-3.2-1B):**

| Path | Time | Status |
|------|------|--------|
| non-vLLM | 116.4s | OK |
| vLLM | ~84s | OK |

PR #363, opened 2025-11-26
Remove include_buffers from init_empty_weights

pluesclues:
This PR is able to fix some issues during the loading of `fast_inference` models, gemma 3 for example was not able to get the correct `embed_scale` in its `language_model.embed_tokens` module.

### DIFF
--- a/unsloth_zoo/empty_model.py
+++ b/unsloth_zoo/empty_model.py
@@ -320,6 +320,12 @@ def create_empty_vision_model(config, dtype = torch.float16):
     try:
         # Use accelerate's init_empty_weights, not transformers.modeling_utils
         from accelerate import init_empty_weights
+        # [NOTE] init_empty_weights(include_buffers = True) is wrong
+        # include_buffers=False is required because buffers (non-trainable tensors like
+        # embed_scale, position_ids) must be initialized with actual values, not on meta
+        # device. Models like Gemma 3 use embed_scale as a buffer in their embedding layer.
+        # With include_buffers=True, buffers become empty meta tensors with no data,
+        # causing attribute access failures during inference.
         with init_empty_weights():
             original_meta_model = model_cls(config)
     except Exception as e:


unsloth_zoo/empty_model.py
```diff
@@ -243,7 +243,7 @@ def create_empty_causal_lm(config, dtype = torch.float16):
     kwargs = {"torch_dtype" if HAS_TORCH_DTYPE else "dtype" : dtype_from_config(config)}
     original_meta_model = None
     error = None
-    with init_empty_weights(include_buffers = True):
+    with init_empty_weights():
```
danielhanchen:
```suggestion
    # [NOTE] init_empty_weights(include_buffers = True) is wrong
    # include_buffers=False is required because buffers (non-trainable tensors like
    # embed_scale, position_ids) must be initialized with actual values, not on meta
    # device. Models like Gemma 3 use embed_scale as a buffer in their embedding layer.
    # With include_buffers=True, buffers become empty meta tensors with no data,
    # causing attribute access failures during inference.
    with init_empty_weights(include_buffers = False):
```

unsloth_zoo/empty_model.py
```diff
@@ -314,7 +314,7 @@ def _init_weights(self, module):
     try:
         # Use accelerate's init_empty_weights, not transformers.modeling_utils
         from accelerate import init_empty_weights
-        with init_empty_weights(include_buffers = True):
+        with init_empty_weights():
```
danielhanchen, line 329:
```suggestion
        # [NOTE] init_empty_weights(include_buffers = True) is wrong
        # include_buffers=False is required because buffers (non-trainable tensors like
        # embed_scale, position_ids) must be initialized with actual values, not on meta
        # device. Models like Gemma 3 use embed_scale as a buffer in their embedding layer.
        # With include_buffers=True, buffers become empty meta tensors with no data,
        # causing attribute access failures during inference.
        with init_empty_weights():
```

PR #369, opened 2025-12-02
FIX: Ensure modules_to_save are merged

oKatanaaa:
This PR fixes https://github.com/unslothai/unsloth/issues/3444. Current merging code does not properly save `modules_to_save` what makes `lm_head`/`embed_token` left untrained after merging.

Ensured that `modules_to_save` (if used during finetuning) are properly propagated into the merged model.

Tested with `.save_pretrained_merged(..., save_method="merged_16bit"), model Qwen3-8B-Base (training in 4bit, 2 new tokens in the vocab).

Also a small fix with stripping `transformers` version from the model config after merge because it led to an error when loading the merged model.

### DIFF
--- a/unsloth_zoo/saving_utils.py
+++ b/unsloth_zoo/saving_utils.py
@@ -1447,8 +1447,9 @@ def merge_and_overwrite_lora(
     has_embed = any(key.endswith("embed_tokens") for key in lora_weights)
     has_head  = any(key.endswith("lm_head") for key in lora_weights)
     if has_embed and has_head:
-        has_embed_tensor = any("embed_tokens" in key for key in safetensor_keys_seen)
-        has_head_tensor  = any("lm_head"      in key for key in safetensor_keys_seen)
+        # Only count actual weight tensors; lm_head.bias alone should not mask tied-embedding cases.
+        has_embed_tensor = any(key.endswith("embed_tokens.weight") for key in safetensor_keys_seen)
+        has_head_tensor  = any(key.endswith("lm_head.weight")      for key in safetensor_keys_seen)
         if has_embed_tensor ^ has_head_tensor:  # exactly one side present on disk
             effective_loras -= 1
 


DmitryDiTy:
@danielhanchen Do you have any known deadlines for implementing this?

unsloth_zoo/saving_utils.py
```diff
@@ -1364,9 +1439,22 @@ def upload_items(filename = None):
...
 
     # Step 7: Check for errors
-    if len(lora_weights) != n_saved_modules:
+    effective_loras = len(lora_weights)
+    # For tied embeddings, PEFT can register both embed_tokens and lm_head as modules_to_save even
+    # though only one tensor exists on disk. If we see both logical modules but only one backing
+    # tensor key across shards, treat them as a single merged target to avoid an off-by-one on the
+    # sanity check while keeping the check meaningful for non-tied models.
+    has_embed = any(key.endswith("embed_tokens") for key in lora_weights)
+    has_head  = any(key.endswith("lm_head") for key in lora_weights)
```
Datta0, line 1448: NIT: rename this to `has_lm_head`

unsloth_zoo/saving_utils.py
```diff
@@ -171,6 +171,27 @@ def _merge_lora(W, lora_stats, name):
 pass
 
 
+def _get_modules_to_save_weight(module):
+    modules_to_save = getattr(module, "modules_to_save", None)
+    if modules_to_save is None:
+        return None
+
+    # Prefer the active/default adapter if present, else first entry
+    for key in ("default",):
```
Datta0, line 180: Why `for` when there's only one value?

unsloth_zoo/saving_utils.py
```diff
@@ -171,6 +171,27 @@ def _merge_lora(W, lora_stats, name):
...
+        try:
+            candidate = modules_to_save[key]
```
Datta0, line 182: If you expect that `modules_to_save[key]` would return error, we prefer using `getattr` for objects or `dict.get()` for dicts

unsloth_zoo/saving_utils.py
```diff
@@ -1167,6 +1235,7 @@ def upload_items(filename = None):
     if save_method == "merged_16bit":
         config.save_pretrained(save_directory)
         _remove_quantization_config(config_path = Path(save_directory) / "config.json")
+        _remove_transformers_version(config_path = Path(save_directory) / "config.json")
```
Datta0, line 1238: Any rational behind removing this? I know the field is probably not necessary but...

DmitryDiTy:
Lets merge it pls!

PR #380, opened 2025-12-11
FP8: Load model on-the-fly in vLLM

andrewor14:
Supports https://github.com/unslothai/unsloth/pull/3717

### DIFF
--- a/unsloth_zoo/vllm_utils.py
+++ b/unsloth_zoo/vllm_utils.py
@@ -845,7 +845,13 @@ def vllm_dynamic_quant_supported(
 pass
 
 
-def get_vllm_state_dict(llm, return_state_dict = False, config = None, is_vision_model = False):
+def get_vllm_state_dict(
+    llm,
+    return_state_dict = False,
+    config = None,
+    is_vision_model = False,
+    load_in_fp8 = False,
+):
     # If the vllm state dict was quantized using torchao, we will run into
     # the following error when calling ops like aten.t() in inference mode.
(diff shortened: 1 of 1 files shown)


unsloth_zoo/vllm_utils.py
```diff
@@ -822,7 +828,7 @@ def get_vllm_state_dict(llm, return_state_dict = False, config = None, is_vision
     #
     # For now, we work around this issue by using torch.no_grad in this case.
     # See https://github.com/pytorch/pytorch/issues/164872 for more details
-    if get_quant_type(config) == "torchao":
+    if get_quant_type(config) == "torchao" or load_in_fp8:
```
Datta0, line 863: Is this strictly necessary? I don't remember this being needed for offline quant FP8.

andrewor14: Yeah this was actually necessary for offline FP8 quant as well. It's just that now we don't require the config to explicitly say "torchao FP8" (it can be any bf16 checkpoint), and instead we handle the FP8 quantization dynamically through vllm, but we still need `no_grad` cause it's still using tensor subclasses

danielhanchen:
Rebased on latest main. The zoo changes look good -- `_get_torchao_fp8_config`, `fp8_mode` param in `load_vllm`, `load_in_fp8` param in `get_vllm_state_dict`, and the `torch.no_grad()` fix are all correct.

The mapper bypass fix was applied on the unsloth side (PR #3717).

PR #383, opened 2025-12-13
update compiler for XLMRobertaModel

electroglyph:
This fixes an issue when trying to compile XLMRobertaModel.

the XLMRobertaModel models have a "forward" entry in their dir(), but they are not classes, so issubclass fails.

this just adds a check to make sure issubclass is only used on classes.


unsloth_zoo/compiler.py
```diff
@@ -2663,7 +2663,7 @@ def replaced_tqdm(*args, **kwargs):
                 continue
 
             module_class = eval(f"modeling_file.{module}")
```
danielhanchen:
```suggestion
            module_class = getattr(modeling_file, module)
```

PR #397, opened 2025-12-29
fix: psutil import at function level and GGUF sharding support

yurekami:
## Summary

This PR addresses two issues:

### 1. Fix psutil import for compiled code (related to unsloth#3800)

**Problem:** When `SFTTrainer` code is compiled/cached via `inspect.getsource()`, the `import psutil` statement inside conditional blocks (`if dataset_num_proc is None:`) was being lost, causing `NameError: name 'psutil' is not defined`.

**Solution:** Moved `import psutil` to function level in both:
- `train_on_responses_only()` 
- `sft_prepare_dataset()`

This ensures the import is captured during code compilation.

### 2. Add support for sharded GGUF files (fixes unsloth#3773)

**Problem:** When converting large models (>50GB) to GGUF, `llama.cpp` automatically shards the output (e.g., `model.BF16-00001-of-00002.gguf`), but Unsloth only checked for the single file name and threw `RuntimeError`.

**Solution:** 
- Added detection for sharded output files using glob pattern matching
- Properly handles sharded file lists in `all_output_files`
- Added proper size reporting that sums all shard sizes

## Test plan

- [ ] Tested psutil fix by running SFTTrainer with `TrainingArguments`
- [ ] Tested GGUF sharding with a large model (>50GB)

### DIFF
--- a/unsloth_zoo/dataset_utils.py
+++ b/unsloth_zoo/dataset_utils.py
@@ -195,6 +195,7 @@ def train_on_responses_only(
     the labels with -100 for the instruction part.
     """
     # All Unsloth Zoo code licensed under LGPLv3
+    import psutil  # Import at function level to ensure it's captured during compilation
     if tokenizer is None and trainer is not None:
         tokenizer = trainer.processing_class if hasattr(trainer, "processing_class") else trainer.tokenizer
     # Get non vision tokenizer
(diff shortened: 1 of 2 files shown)


danielhanchen:
main now handles sharded GGUF output in unsloth_zoo/llama_cpp.py via #446, and the psutil imports already sit inside train_on_responses_only and sft_prepare_dataset, so I do not think this is needed any more. Sorry for the slow review, and thanks for putting it together.

danielhanchen:
main now handles sharded GGUF output in unsloth_zoo/llama_cpp.py, including shard validation and summed size reporting, via #446, so I do not think this is needed any more. Sorry for the slow review, and thanks for putting it together.

yurekami:
Confirmed on main: sharded GGUF output is handled in `llama_cpp.py` (#446), and the psutil imports already sit inside `train_on_responses_only` and `sft_prepare_dataset`. Closing.

PR #398, opened 2025-12-29
fix: pass system_type to do_we_need_sudo and fix peft import

### LINKED ISSUE #311: 'peft' is not defined in patching_utils.py

```
    import bitsandbytes
    else:
        
        namespace = {}
        
            try:
                
                module = eval(x, namespace)
                layers = dir(module)
                
                for fx in layers:
                    try:
                        layer = getattr(module, fx)
                    except:
                        continue
                        continue
                        continue
                    
            except ImportError:
                if "peft" in x:
                continue
    return
pass
```

yurekami:
## Summary

This PR fixes two issues:

### 1. Fix #390: Pass system_type argument to do_we_need_sudo()

**Problem:** `do_we_need_sudo()` requires `system_type` parameter to use the correct package manager (apt-get vs yum/dnf), but was called without arguments at line 391 in `llama_cpp.py`.

**Solution:** Pass `system_type=system_type` to the function call, using the value already available from `check_build_requirements()`.

### 2. Fix #311: Fix peft import in patch_compiling_bitsandbytes()

**Problem:** The `exec()` and `eval()` calls with `globals()`/`locals()` don't properly persist imported modules across subsequent `eval()` calls. This caused `NameError: 'peft' is not defined` when trying to disable torch.compile on peft modules.

**Solution:**
- Use a namespace dict to properly store imported modules
- Use `getattr()` instead of `eval()` for cleaner attribute access
- Add `try/except ImportError` to gracefully handle cases where peft is not installed

## Test plan

- [ ] Tested on Fedora/RHEL system with yum/dnf
- [ ] Tested peft import with bitsandbytes < 0.46.0

### DIFF
--- a/unsloth_zoo/dataset_utils.py
+++ b/unsloth_zoo/dataset_utils.py
@@ -195,6 +195,7 @@ def train_on_responses_only(
     the labels with -100 for the instruction part.
     """
     # All Unsloth Zoo code licensed under LGPLv3
+    import psutil  # Import at function level to ensure it's captured during compilation
     if tokenizer is None and trainer is not None:
         tokenizer = trainer.processing_class if hasattr(trainer, "processing_class") else trainer.tokenizer
     # Get non vision tokenizer
@@ -329,7 +330,6 @@ def train_on_responses_only(
         return _train_on_responses_only
 
     if num_proc is None or type(num_proc) is not int:
-        import psutil
         num_proc = min(max(psutil.cpu_count()+4, 2), 64)
         # Check memory left so we can reduce multiprocessing to converse memory
         memory_gb_left = psutil.virtual_memory().available / (1024**3)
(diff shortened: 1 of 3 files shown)


danielhanchen:
Thanks for these. Status after today's merges:

- The patching_utils peft import fix is superseded by #710, which just landed on main.
- The `do_we_need_sudo` fix already landed via #506.

That leaves two unique changes here: the function level psutil imports in dataset_utils and the sharded GGUF output detection in convert_to_gguf. Please rebase onto main and drop the superseded files so the PR only contains those two. The sharded output detection looks useful. For the psutil change, please add a short repro of the compilation failure it fixes so we can verify it.

danielhanchen:
I checked out this branch and merged latest main to trim it down as planned, and the resulting diff against main is empty. Everything here has since landed through other PRs, in more complete form:

- `do_we_need_sudo(system_type)` via #506
- the patching_utils peft import fix via #710 and #743
- sharded GGUF output detection via the convert path overhaul on main, which also validates shard completeness and mismatched totals
- the num_proc / psutil logic was reworked on main with a multiprocessing start method guard and finer memory based capping, with psutil imported where it is used

So there is nothing left for this PR to add. Closing as superseded. Thanks for the fixes, every one of them pointed at a real bug.

PR #400, opened 2025-12-29
fix: use thinking content in encode_conversations_with_harmony

### LINKED ISSUE #246: encode_conversation_with_harmony try to refer "content" when "thinking" in message.
https://github.com/unslothai/unsloth-zoo/blob/755a74ab43d6cdd05675730e25491860d7a286e1/unsloth_zoo/temporary_patches/gpt_oss.py#L630-L637

https://github.com/unslothai/unsloth-zoo/blob/755a74ab43d6cdd05675730e25491860d7a286e1/unsloth_zoo/temporary_patches/gpt_oss.py#L688-L691

First of all, thank you for providing this package.

I have a quick question: 
As I mentioned in title, function 'encode_conversation_with_harmony" try to refer "content" when "thinking" in message.
Is that line intended?

I just want to confirm whether this line is there by design or if it might be a mistake.

yurekami:
## Summary
- Fixes `encode_conversations_with_harmony` to properly use `message["thinking"]` instead of `message["content"]` when the "thinking" key is present
- Also includes the final content with "final" channel when both thinking and content are present
- Previously, the thinking/analysis content was completely lost

## Related Issue
Fixes #246

## Test plan
- [ ] Verify messages with `thinking` key now properly encode the thinking content with "analysis" channel
- [ ] Verify messages with both `thinking` and `content` include both pieces of content
- [ ] Verify messages without `thinking` still work as expected

### DIFF
--- a/unsloth_zoo/temporary_patches/gpt_oss.py
+++ b/unsloth_zoo/temporary_patches/gpt_oss.py
@@ -1345,8 +1345,16 @@ def encode_conversations_with_harmony(
             )
         elif message["role"] == "assistant":
             if "thinking" in message:
-                x = Message.from_role_and_content(Role.ASSISTANT, message["content"])
+                # Add the thinking content with "analysis" channel
+                x = Message.from_role_and_content(Role.ASSISTANT, message["thinking"])
                 x = x.with_channel("analysis")
+                convos.append(x)
+                # Also add the final content if present
+                if "content" in message and message["content"]:
+                    x = Message.from_role_and_content(Role.ASSISTANT, message["content"])
+                    x = x.with_channel("final")
+                    convos.append(x)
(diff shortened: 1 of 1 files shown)


danielhanchen:
main now handles this in unsloth_zoo/temporary_patches/gpt_oss.py, emitting the thinking text on the analysis channel and the content on final, so I do not think this is needed any more. Sorry for the slow review, and thanks for putting it together.

danielhanchen:
main now handles this in unsloth_zoo/temporary_patches/gpt_oss.py, emitting the thinking field on the analysis channel and the final content separately, so I do not think this is needed any more. Sorry for the slow review, and thanks for putting it together.

yurekami:
Confirmed on main: `encode_conversations_with_harmony` now emits `thinking` on the analysis channel and the content on final. Closing.

PR #401, opened 2025-12-29
Add Alpine, openSUSE and Gentoo to the llama.cpp package-manager paths

### LINKED ISSUE #251: RuntimeError: [FAIL] Unsloth: apt-get does not exist? Is this NOT a Linux / Mac based computer?
Hello! I tried fine-tune a model and convert it to gguf format, but got this error:
`Traceback (most recent call last):
  File "/mnt/floppy/finetune/unsloth/./Qwen3_(4B)-Instruct.py", line 218, in <module>
    model.save_pretrained_gguf("model", tokenizer, quantization_method = "f16")
  File "/mnt/floppy/finetune/unsloth/.unsloth/lib/python3.12/site-packages/torch/utils/_contextlib.py", line 116, in decorate_context
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/floppy/finetune/unsloth/.unsloth/lib/python3.12/site-packages/unsloth/save.py", line 2287, in save_to_gguf_generic
    install_llama_cpp(just_clone_repo = True)
  File "/mnt/floppy/finetune/unsloth/.unsloth/lib/python3.12/site-packages/unsloth_zoo/llama_cpp.py", line 284, in install_llama_cpp
    sudo = do_we_need_sudo()
           ^^^^^^^^^^^^^^^^^
  File "/mnt/floppy/finetune/unsloth/.unsloth/lib/python3.12/site-packages/unsloth_zoo/llama_cpp.py", line 118, in do_we_need_sudo
    raise RuntimeError("[FAIL] Unsloth: apt-get does not exist? Is this NOT a Linux / Mac based computer?")
RuntimeError: [FAIL] Unsloth: apt-get does not exist? Is this NOT a Linux / Mac based computer?`

I'm running Gentoo Linux and obviously on't have apt-get. I 'fixed' it by creating empty file called apt-get and then conversion to gguf went successfully. Previously I didn't face this issue. And I'm not sure my 'solution' will work after future updates of this package. Please, think about users of other distros. Seriously, Ubuntu is not only one!

yurekami:
Fixes #251.

`check_linux_type` only knew Debian, RPM and Arch, so on Alpine, openSUSE and Gentoo the llama.cpp build path fell through to `apt-get`. This adds the three:

- detection: `/etc/alpine-release`, `/etc/gentoo-release`, and `suse` or `sles` in the `ID` / `ID_LIKE` of `/etc/os-release` (openSUSE and SLES no longer ship `/etc/SuSE-release`)
- install and refresh commands for `apk`, `zypper` and `emerge`. Gentoo skips the refresh probe in `do_we_need_sudo`, because `emerge --sync` takes minutes and would trip the 180 s no-internet check, and answers the sudo question from the effective uid instead
- package names for build tools, libgomp, libssl and the libcurl headers on each, folded into one `DISTRO_PACKAGES` map next to the existing RPM and Arch entries

Per review, the "package does not exist" check no longer matches any line containing "not found". `PACKAGE_NOT_FOUND` lists the phrase each manager prints for that case, so ordinary installer output cannot abort a working install.

Rebased onto current main; the earlier version of this branch predates the Arch and Windows support that landed in the meantime.

**Tests.** `tests/test_llama_cpp_distro_support.py` covers detection from release files and from os-release, the per-distro maps, and that each manager's not-found phrase is matched while lines such as `Package 'libgomp1' is not found in cache` are not. 27 pass locally on Python 3.12. I have not run the install path on the three distros themselves; the commands are the standard non-interactive forms for each manager.

### DIFF
--- a/tests/test_llama_cpp_distro_support.py
+++ b/tests/test_llama_cpp_distro_support.py
@@ -1,0 +1,132 @@
+"""Distro detection and package-manager mapping in llama_cpp.py.
+
+Covers the Alpine / openSUSE / Gentoo additions: `check_linux_type` picks them
+up from their release files or /etc/os-release, the per-distro package maps
+have an entry for each, and the package-not-found detection matches each
+manager's own phrase without matching an ordinary "not found" line.
+"""
+
+from __future__ import annotations
+
+import importlib.util
+import os
+import sys
+from pathlib import Path
+
+import pytest
+
+
+def _load_llama_cpp_module():
+    repo_root = Path(__file__).resolve().parents[1]
+    module_path = repo_root / "unsloth_zoo" / "llama_cpp.py"
+    spec = importlib.util.spec_from_file_location("llama_cpp_under_test_distro", module_path)
+    module = importlib.util.module_from_spec(spec)
+    sys.modules[spec.name] = module
+    spec.loader.exec_module(module)
+    return module
+
+
+@pytest.fixture(scope="module")
+def llama_cpp():
+    return _load_llama_cpp_module()
+
+
+def _only_these_files_exist(monkeypatch, present):
+    real_exists = os.path.exists
+
+    def fake_exists(path):
+        if str(path).startswith("/etc/"):
+            return str(path) in present
+        return real_exists(path)
+
+    monkeypatch.setattr(os.path, "exists", fake_exists)
+
+
+@pytest.mark.parametrize(
+    "release_file, expected",
+    [
+        ("/etc/debian_version", "debian"),
+        ("/etc/fedora-release", "rpm"),
+        ("/etc/arch-release", "arch"),
+        ("/etc/alpine-release", "alpine"),
+        ("/etc/gentoo-release", "gentoo"),
+    ],
+)
+def test_check_linux_type_from_release_file(llama_cpp, monkeypatch, release_file, expected):
+    monkeypatch.setattr(llama_cpp.platform, "system", lambda: "Linux")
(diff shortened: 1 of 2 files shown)


danielhanchen:
Confirmed the gap for Alpine, openSUSE and Gentoo in check_linux_type in unsloth_zoo/llama_cpp.py, though main has since added Arch, Windows and the libgomp/libssl package maps, so this needs a rebase before it applies. Could you also narrow the new "not found" substring check, since it can match ordinary installer output and abort a working install?

yurekami:
Rebased onto main and narrowed as asked. The branch now adds only Alpine, openSUSE and Gentoo on top of the Arch and Windows support that landed in the meantime, and the not-found check matches each manager's own phrase (`PACKAGE_NOT_FOUND`) instead of any line containing "not found". Description updated with what was and was not tested.

danielhanchen:
Confirmed check_linux_type in unsloth_zoo/llama_cpp.py still stops at Arch on main, so this is the path that breaks the GGUF build on Gentoo in #251, and the rebase plus the narrowed PACKAGE_NOT_FOUND phrases both look right. Will get this reviewed.

danielhanchen:
Confirmed check_linux_type in unsloth_zoo/llama_cpp.py still stops at debian, rpm and arch on main, and the rebase plus the per-manager PACKAGE_NOT_FOUND phrases address both points from last time, so this also covers #251. Will get this reviewed.

PR #405, opened 2025-12-29
fix: multiple bug fixes (#390, #311, #316)

### LINKED ISSUE #311: 'peft' is not defined in patching_utils.py

### LINKED ISSUE #316: UnslothVisionDataCollator cannot handle a tuple as patch size

yurekami:
## Summary
- **Fix #390**: Pass `system_type` to `do_we_need_sudo()` for proper package manager detection on non-Debian systems (e.g., Fedora using yum/dnf)
- **Fix #311**: Use namespace dictionary for `exec/eval` in `patch_compiling_bitsandbytes` to properly access imported modules, add `ImportError` handling for optional peft dependency
- **Fix #316**: Handle tuple patch sizes (e.g., `(14, 14)` for InternVL3 models) in `UnslothVisionDataCollator` by extracting the first element

## Changes
1. `llama_cpp.py`: Pass `system_type` argument to `do_we_need_sudo()` call
2. `patching_utils.py`: Refactor import handling with proper namespace and error handling
3. `vision_utils.py`: Normalize tuple patch sizes to single integer

## Test plan
- [ ] Verify `do_we_need_sudo()` works correctly with different Linux distros
- [ ] Verify bitsandbytes/peft patching works without errors when peft is not installed
- [ ] Verify InternVL3-2B and other models with tuple patch sizes work correctly

Fixes #390, #311, #316

### DIFF
--- a/unsloth_zoo/llama_cpp.py
+++ b/unsloth_zoo/llama_cpp.py
@@ -388,7 +388,7 @@ def install_llama_cpp(
 
     print_outputs = []
     missing_packages, system_type = check_build_requirements()
-    sudo = do_we_need_sudo()
+    sudo = do_we_need_sudo(system_type)
     kwargs = {"sudo" : sudo, "print_output" : print_output, "print_outputs" : print_outputs, "system_type": system_type}
 
     if not missing_packages:
--- a/unsloth_zoo/patching_utils.py
+++ b/unsloth_zoo/patching_utils.py
@@ -44,15 +44,22 @@ def patch_compiling_bitsandbytes():
         # Disable dynamo on Linear4bit, Linear8bit and other future modules
         if os.environ.get("UNSLOTH_ENABLE_LOGGING", "0") == "1":
             print("Unsloth: Bitsandbytes < 0.46.0 does not support torch.compile - disabling.")
+        # Use a namespace dictionary to properly store imported modules
+        namespace = {}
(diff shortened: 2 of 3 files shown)


danielhanchen:
Thanks for the batch of fixes. Status after today's merges:

- The patching_utils #311 fix is superseded by #710, which just landed on main.
- The `do_we_need_sudo(system_type)` fix already landed via #506.
- #316 is closed, so the vision_utils tuple patch size change is the only remaining part.

This PR now conflicts with main on patching_utils. If the InternVL3 tuple patch size handling is still needed, please rebase onto main and trim this down to just the vision_utils change, or open a fresh focused PR.

danielhanchen:
I checked out this branch and merged latest main to trim it down as planned, and the resulting diff against main is empty. All three changes have since landed through other PRs:

- `do_we_need_sudo(system_type)` via #506
- the patching_utils exec/eval fix via #710 and #743
- tuple patch size handling in vision_utils via the fix that closed #316 (main now also covers empty tuples, non square sizes and None)

So there is nothing left for this PR to add. Closing as superseded. Thanks for the report and the fixes, they all pointed at real bugs.

PR #410, opened 2025-12-29
fix: use importlib for dynamic module imports in patch_compiling_bitsandbytes

### LINKED ISSUE #311: 'peft' is not defined in patching_utils.py

yurekami:
## Summary
- Fixed `'peft' is not defined` error in `patch_compiling_bitsandbytes()` function
- Replaced problematic `exec()`/`eval()` pattern with proper `importlib.import_module()`

## Problem
The previous code used `exec(f"import {x}", globals(), locals())` followed by `eval(x)` to dynamically import modules. This pattern doesn't reliably persist imports in the execution context, causing `NameError: name 'peft' is not defined` errors.

## Changes
- Added `import importlib` at top of file  
- Use `importlib.import_module(x)` for dynamic imports
- Use `getattr(module, fx)` instead of `eval(f"{x}.{fx}")`
- Added try/except to gracefully skip optional dependencies (peft) if not installed

Fixes #311

### DIFF
--- a/unsloth_zoo/patching_utils.py
+++ b/unsloth_zoo/patching_utils.py
@@ -18,6 +18,7 @@ import torch
 import os
 import re
 import ast
+import importlib
 
 __all__ = [
     "patch_compiling_bitsandbytes",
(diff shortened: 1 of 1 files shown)


danielhanchen:
main now does this in unsloth_zoo/patching_utils.py via #710 and #743, using importlib.import_module with the same getattr access, so I do not think this is needed any more. Sorry for the slow review, and thanks for putting it together.

danielhanchen:
main now does this in unsloth_zoo/patching_utils.py, where the import is done via importlib.import_module and the forward patch uses getattr, so I do not think this is needed any more. Sorry for the slow review, and thanks for putting it together.

yurekami:
Confirmed on main: `patch_compiling_bitsandbytes` now goes through `importlib.import_module` and `getattr` (#710, #743). Closing.

PR #413, opened 2025-12-29
fix: video fps accumulation bug in _collate_prompt_completion

### LINKED ISSUE #319: Vision Utils Breaks on Video Only Samples
If we try to use `UnslothVisionDataCollator` on a dataset of conversations that **do not** have images and **only has** samples that have videos, as multimodal inputs the script fails in two places

1. **Empty Images List** If no images are provided as a part of the message, `images` defaults to an empty list (not None). Downstream processors (like Qwen-VL-2.5's ImageProcessor) does not know how to handle an empty list for a sample's images and throws and error.

https://github.com/unslothai/unsloth-zoo/blob/ea85a264ef188a4ea1b20edab9864f90907e59a7/unsloth_zoo/vision_utils.py#L775-L807

2. **`_cast_pixel_values_dtype_inplace` expects `pixel_values`** If there are no images, there is `no pixel_values` value and the `_cast_pixel_values_dtype_inplace` function errors out.

https://github.com/unslothai/unsloth-zoo/blob/ea85a264ef188a4ea1b20edab9864f90907e59a7/unsloth_zoo/vision_utils.py#L950-L965

yurekami:
## Summary
- Fixed broken video fps handling in prompt/completion collation path
- Enables video-only samples to work correctly

## Problem
In `_collate_prompt_completion`, there were multiple bugs with fps accumulation:

1. **No batch-level accumulator**: `video_kwargs` was not initialized, so fps values couldn't be accumulated across examples
2. **Self-extending bug**: Line `vids_kwarg['fps'].extend(vids_kwarg['fps'])` was extending the list with itself, doubling values on each iteration
3. **Wrong variable used**: After the loop, `vids_kwarg` (last example's kwargs) was used instead of accumulated values

This caused video-only batches to fail because fps list length didn't match video count.

## Changes
- `unsloth_zoo/vision_utils.py`:
  - Initialize `video_kwargs = {'fps': []}` at batch level
  - Fix accumulation: `video_kwargs['fps'].extend(vids_kwarg['fps'])` 
  - Use `video_kwargs` instead of `vids_kwarg` after loop

Fixes #319

### DIFF
--- a/unsloth_zoo/vision_utils.py
+++ b/unsloth_zoo/vision_utils.py
@@ -1123,6 +1123,7 @@ class UnslothVisionDataCollator:
 
     def _collate_prompt_completion(self, examples):
         prompt_texts, completion_texts, images, videos = [], [], [], []
+        video_kwargs = {'fps': []}
 
         for ex in examples:
             p, c = ex["prompt"], ex["completion"]
@@ -1164,9 +1165,8 @@ class UnslothVisionDataCollator:
 
             if vids and len(vids) > 0:  # Works for list, tuple or tensor
                 videos.append(vids)
-                if vids_kwarg is None:
-                    vids_kwarg = {"fps": []}
-                vids_kwarg['fps'].extend(vids_kwarg['fps'])
+                if vids_kwarg is not None and 'fps' in vids_kwarg:
+                    video_kwargs['fps'].extend(vids_kwarg['fps'])
 
         prompt_kwargs = dict(
             padding=True,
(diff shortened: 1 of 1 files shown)


danielhanchen:
main now does this in unsloth_zoo/vision_utils.py, where _collate_prompt_completion accumulates fps into a batch level video_kwargs, via #475, so I do not think this is needed any more. Sorry for the slow review, and thanks for putting it together.

danielhanchen:
main now does this in unsloth_zoo/vision_utils.py, where the fps list is accumulated per example and collapsed before the processor call, via #475, so I do not think this is needed any more. Sorry for the slow review, and thanks for putting it together.

yurekami:
Confirmed on main: `_collate_prompt_completion` now accumulates fps into a batch-level `video_kwargs` and collapses it before the processor call (#475), which is what this PR did. Closing.

PR #427, opened 2026-01-02
fix: add revision to vLLM loader

majiayu000:
## Summary
- Add revision parameter to `load_vllm` and pass through to vLLM engine args

## Test Plan
- Not run (no supported GPU in local env)


Datta0 (approved):
LGTM!

PR #474, opened 2026-02-08
Make bitsandbytes optional on ROCm

danielhanchen:
This makes bitsandbytes an optional dependency so ROCm environments without bitsandbytes can still run Unsloth-Zoo workflows (16bit / bf16) end to end.

Changes:
- patching_utils: guard patch_compiling_bitsandbytes when bitsandbytes is not installed
- patching_utils: treat Linear4bit types as optional and avoid raising when bnb is missing
- saving_utils: guard bitsandbytes import and isinstance checks
- peft_utils: skip LoRA layer modules that fail to import (eg peft.tuners.lora.bnb) instead of crashing

This is NVIDIA-safe: when bitsandbytes is installed, behavior is unchanged.

### DIFF
--- a/unsloth_zoo/patching_utils.py
+++ b/unsloth_zoo/patching_utils.py
@@ -36,7 +36,12 @@ def patch_compiling_bitsandbytes():
     # All Unsloth Zoo code licensed under LGPLv3
     os.environ["UNSLOTH_PATCHED"] = "1"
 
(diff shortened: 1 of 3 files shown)


danielhanchen:
AMD Support has been added into Unsloth with Bitsandbytes support.

PR #492, opened 2026-02-10
Fix Gemma3 Vision + Gemma3N audio inference on transformers 5.x

danielhanchen:
## Summary
- Fix Gemma3 Vision inference crash (`index_copy_` dtype mismatch + flex_attention dtype validation) when using `dtype=torch.float16` on transformers 5.x
- Fix Gemma3N Conversational audio inference crash (`AttributeError: 'Tensor' object has no attribute 'audio_mel_mask'`) on transformers 5.x (upstream transformers bug)

### 1. `compiler.py` -- RoPE mixed dtype fix
When `FORCE_FLOAT32=1` (Gemma3 float16 path), the model loads in bfloat16 but inference uses float16 autocast. RoPE returns bfloat16 cos/sin (from hidden_states dtype), attention projections produce float16 q/k/v (under autocast), and `apply_rotary_pos_emb` promotes `float16 * bfloat16 = float32`. This breaks flex_attention (requires matching q/k/v dtypes) and causes cache dtype mismatches.

Adds `fix_apply_rotary_pos_emb_mixed_dtype()` source transformation that inserts `cos/sin` dtype casting to match `q` dtype inside `apply_rotary_pos_emb` during compilation. Applied in both `create_standalone_class()` and `create_new_function()`.

### 2. `temporary_patches/misc.py` -- Static cache dtype mismatch
Adds `patch_static_cache_dtype_mismatch()` that patches `StaticLayer` and `StaticSlidingWindowLayer` to handle key/value dtype mismatches during cache initialization and updates. Uses lower precision dtype for cache allocation and casts incoming states to match cache dtype. Belt-and-suspenders defense for the FORCE_FLOAT32 path.

### 3. `temporary_patches/gemma3n.py` -- Audio mel mask fix
Upstream transformers bug in `modeling_gemma3n.py:2114-2115`: `audio_features = audio_features.pooler_output` overwrites the dataclass with a raw Tensor before `.audio_mel_mask` is extracted. Patches `get_audio_features` to attach `audio_mel_mask` onto the `pooler_output` tensor so it survives the variable reassignment.

## Test plan
- [x] Gemma3 4B Vision inference with `dtype=torch.float16` on transformers 5.1.0 -- PASS
- [x] Gemma3 4B Vision multi-inference (cache reuse) -- PASS
- [x] Gemma3 4B bfloat16 inference regression -- PASS
- [x] Gemma3 4B training + post-training inference on transformers 5.1.0 -- PASS (loss=1.486)
- [x] Gemma3N audio_mel_mask patch verified (patch registered, upstream bug detected, fix applied)

### DIFF
--- a/unsloth_zoo/temporary_patches/misc.py
+++ b/unsloth_zoo/temporary_patches/misc.py
@@ -1269,3 +1269,72 @@ def patch_vllm_safe_apply_chat_template():
         pass
 pass
 TEMPORARY_PATCHES.append(patch_vllm_safe_apply_chat_template)
+
+
+def patch_static_cache_dtype_mismatch():
+    """Fix StaticLayer/StaticSlidingWindowLayer index_copy_ dtype mismatch.
+
+    When using float16 autocast on a bfloat16 model (FORCE_FLOAT32 path for Gemma3),
+    RoPE produces float32 keys (float16 * bfloat16) while values stay float16.
+    The cache is allocated with key_states.dtype (float32), causing index_copy_ to
+    fail when storing float16 values. Fix: eagerly initialize with consistent dtype
+    in update wrappers, then cast all incoming states before delegating.
+    """
+    if os.environ.get("UNSLOTH_FORCE_FLOAT32", "0") == "0":
+        return
+    try:
+        from transformers.cache_utils import StaticLayer, StaticSlidingWindowLayer
+    except Exception:
+        return
+
+    if hasattr(StaticLayer.update, "_unsloth_patched"):
+        return
+
+    def _resolve_dtype(key_states, value_states):
+        """Pick lower-precision dtype when key/value dtypes differ."""
+        if key_states.dtype == value_states.dtype:
+            return key_states.dtype
+        return (
+            value_states.dtype
+            if value_states.element_size() <= key_states.element_size()
+            else key_states.dtype
+        )
+
+    # Patch StaticLayer.update
+    _orig_update = StaticLayer.update
+    def patched_update(self, key_states, value_states, cache_kwargs=None):
+        if not self.is_initialized:
+            # Eagerly initialize with consistent dtype so the original update
+            # sees is_initialized=True and uses our casted states
+            if key_states.dtype != value_states.dtype:
+                target = _resolve_dtype(key_states, value_states)
+                self.lazy_initialization(key_states.to(target), value_states.to(target))
+            else:
+                self.lazy_initialization(key_states, value_states)
+        # Cast incoming states to match the cache dtype
+        if key_states.dtype != self.keys.dtype:
+            key_states = key_states.to(self.keys.dtype)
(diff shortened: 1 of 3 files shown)


unsloth_zoo/temporary_patches/misc.py
```diff
@@ -1269,3 +1269,72 @@ def _patched_safe_apply(model_config, tokenizer, conversation, *,
         pass
 pass
 TEMPORARY_PATCHES.append(patch_vllm_safe_apply_chat_template)
+
+
+def patch_static_cache_dtype_mismatch():
+    """Fix StaticLayer/StaticSlidingWindowLayer index_copy_ dtype mismatch.
+
+    When using float16 autocast on a bfloat16 model (FORCE_FLOAT32 path for Gemma3),
```
Datta0, line 1277: This was effecting gpt_oss 
That is why I had to [do this conversion](https://github.com/unslothai/unsloth-zoo/blob/b6dbba1ad5aa41f64f78f69435e06339b2a3467a/unsloth_zoo/temporary_patches/gpt_oss.py#L1965-L1967)
Once this lands, can we explore removing that ?

danielhanchen:
Closing as obsolete. Verified against current main (unsloth-zoo 2026.9.3, unsloth 2026.9.4) on transformers 5.17.0, torch 2.13.0, B200.

All three parts of this PR no longer apply:

**Gemma3N audio (`audio_mel_mask`)** - fixed upstream. transformers 5.17 no longer overwrites the dataclass before reading the mask:

```python
audio_outputs = self.get_audio_features(...)
audio_features = audio_outputs.pooler_output
audio_mask = audio_outputs.audio_mel_mask
```

gemma-3n-E2B-it conversational audio inference runs clean on unpatched main. The `get_audio_features` replacement here is now identical to upstream apart from the setattr, so it would install a patch that does nothing.

**Gemma3 vision float16 dtype (RoPE + static cache)** - no longer reproduces. gemma-3-4b-it with `dtype=torch.float16` under `UNSLOTH_FORCE_FLOAT32=1`, two generation passes for cache reuse, passes on main with no `index_copy_` mismatch and no flex_attention dtype error. Main reaches the same result via `fix_rotary_embedding_dtype` + `fix_attention_dtype_consistency` in compiler.py and the boundary dtype work in #1017.

Separately, `patch_static_cache_dtype_mismatch` reads `UNSLOTH_FORCE_FLOAT32` at import time, before model loading sets it, so it would be a no-op anyway, and its `(self, key_states, value_states, cache_kwargs=None)` wrapper has drifted from upstream's current `(self, key_states, value_states, *args, **kwargs)`.

A genuinely separate bug turned up while testing this, tracked on its own: Gemma3 vision inference with torch.compile enabled hits a dynamo data-dependent branching error in `masking_utils.flex_attention_mask`, identically for float16 and bfloat16, so it is not a dtype problem and is not addressed here.

PR #496, opened 2026-02-11
Handle missing CSM depth decoder loss during loss aggregation

danielhanchen:
## Summary
- fixes CSM temporary patch loss aggregation when depth decoder loss is absent
- avoids `Tensor + NoneType` during training loss computation

## Root Cause
In the patched `CsmForConditionalGeneration.forward`, `loss` was always computed as:
- `backbone_loss + depth_decoder_loss`

When `depth_decoder_loss` is `None`, this raises:
- `TypeError: unsupported operand type(s) for +: 'Tensor' and 'NoneType'`

## Changes
- `unsloth_zoo/temporary_patches/misc.py`
  - replace unconditional sum with guarded aggregation:
    - if only one loss exists, use that loss
    - if both exist, sum both

## Validation
- `python -m compileall unsloth_zoo/temporary_patches/misc.py`
- static assertion confirms guarded aggregation block is present in patched source

## Notes
- behavior is unchanged when both losses are present
- this only changes the edge case where one loss is missing


danielhanchen:
Pushed follow-up fixes to this PR branch for transformers 5.x CSM compatibility.

New commits:
- a2283ac csm: improve transformers 5.x forward patch compatibility
- 622dda9 csm: restore misc.py file mode

What changed:
- Added a dedicated CsmDepthDecoderModel forward patch to avoid the leaf-view in-place autograd failure.
- Updated CSM forward/depth forward fallback patching so 5.x signatures are patched correctly.
- Replaced 5.x fallback wrappers with full patched implementations so the loss aggregation guard and kwargs handling apply on 5.x.
- Preserved output_hidden_states/output_attentions in fallback kwargs handling to avoid generation hidden state regressions.
- Added defensive depth-decoder input id clamping in patched paths.

Validation notes (local, no CSM short-circuit):
- The previous training-time failures are fixed on transformers 5.1.0 (in-place autograd + Tensor+None loss).
- CSM training now runs and completes.
- There is still a remaining CUDA device-side assert in depth-decoder generation (index out of bounds in IndexKernel), which appears to be a deeper generation-path issue.

danielhanchen:
Additional updates pushed after further validation:

- ea8eb1b csm: preserve generation hidden states in 5.x fallbacks
- dd26d59 csm: restore misc.py file mode after update

Current local status on transformers 5.1.0 (no CSM short-circuit):
- CSM training path now completes.
- Generation proceeds deeper than before, but there is still a remaining CUDA device-side assert in the depth-decoder generation path (index out of bounds in CUDA IndexKernel).

Smoke harness run with standard CSM shortcut remains passing after these updates.

PR #499, opened 2026-02-14
Fix MoE target_parameters module_count alignment (#3405, #3701)

GoldenGrapeGentleman:
## Summary

Resolves the `TODO: handle MoE target_parameters to align counts` at line 410 of `saving_utils.py`.

### Problem

When using PEFT `target_parameters` for MoE expert LoRA (e.g., gpt-oss), the `create_lora_statistics` function reports a misleading diagnostic warning:

```
[Unsloth merge debug] LoRA count mismatch: modules=120, lora_A=144, lora_B=144, scaling=144
```

This happens because MoE `ParamWrapper` entries have `lora_A`/`lora_B`/`scaling` counted, but lack a `.base_layer` module, so `module_count` is never incremented for them.

### Fix

After the LoRA collection loop, count MoE expert entries that have `lora_A`/`lora_B` but no `.module`, aligning `module_count` with the other counts. Only applies to `.mlp.experts` entries to avoid affecting non-MoE models.

### Changes

- `saving_utils.py`: +12 lines / -3 lines (removes TODO comment)

### Verified on 8× AMD Instinct MI355X (gfx950), ROCm 7.1

- gpt-oss-20b BF16 + MoE expert LoRA: 46.2M trainable params (vs 2.0M attention-only)
- `save_pretrained_merged`: ✅ success, **no count mismatch warning**
- Before fix: `modules=120, lora_A=144` warning printed
- After fix: counts aligned, clean merge

Related: unslothai/unsloth#3405, unslothai/unsloth#3701

cc @danielhanchen @Datta0

### DIFF
--- a/unsloth_zoo/saving_utils.py
+++ b/unsloth_zoo/saving_utils.py
@@ -390,15 +390,15 @@ def create_lora_statistics(model, merge_into_original = False, return_state_dict
             remove_keys.add(name)
         pass
     pass
-    # MoE target_parameters (ParamWrapper) entries have lora_A/B/scaling but
-    # may lack a corresponding .base_layer module, leaving module_count short.
-    # Count these so the diagnostic check below stays accurate (#3405, #3701).
-    for key, stats in lora_weights.items():
+    # PEFT target_parameters (ParamWrapper for nn.Parameter, not nn.Linear)
+    # have lora_A/B/scaling but no .base_layer, leaving module_count short.
+    # nn.Linear targets always get .base_layer set, so module=None reliably
+    # identifies nn.Parameter targets. Count them to align (#3405, #3701).
+    for _key, _stats in lora_weights.items():
         if (
-            stats.lora_A is not None
-            and stats.lora_B is not None
-            and stats.module is None
-            and ".mlp.experts" in key
+            _stats.lora_A is not None
+            and _stats.lora_B is not None
+            and _stats.module is None
         ):
             module_count += 1
 
--- a/unsloth_zoo/saving_utils.py
+++ b/unsloth_zoo/saving_utils.py
@@ -390,6 +390,18 @@ def create_lora_statistics(model, merge_into_original = False, return_state_dict
             remove_keys.add(name)
         pass
     pass
+    # MoE target_parameters (ParamWrapper) entries have lora_A/B/scaling but
+    # may lack a corresponding .base_layer module, leaving module_count short.
+    # Count these so the diagnostic check below stays accurate (#3405, #3701).
+    for key, stats in lora_weights.items():
+        if (
+            stats.lora_A is not None
+            and stats.lora_B is not None
+            and stats.module is None
+            and ".mlp.experts" in key
+        ):
+            module_count += 1
+
     if not (module_count == lora_A_count == lora_B_count == scaling_count):
         print(
             f"[Unsloth merge debug] LoRA count mismatch: modules={module_count}, "
@@ -405,9 +417,6 @@ def create_lora_statistics(model, merge_into_original = False, return_state_dict
                 print(f"  key={k} param={param_name} A={a_shape} B={b_shape}")
         except Exception:
             pass
-        # Allow merge to continue; downstream checks will still fail loudly if tensors are missing
-        # but this avoids silent assertion without context.
-        # TODO: handle MoE target_parameters to align counts.
 
     # Also return state_dict if needed
     if return_state_dict:
--- a/unsloth_zoo/saving_utils.py
+++ b/unsloth_zoo/saving_utils.py
@@ -2585,6 +2585,46 @@ def detect_keys_format(keys_to_check, forward_mapping):
     return "new" # Default, assuming most models/keys will be in the "new" (current HF) format.
 pass
 
+def _infer_prefix_and_remap(lora_weights, safetensor_keys):
+    """Infer a missing key prefix by matching LoRA keys against safetensor keys.
+
+    Some composite models (e.g. Qwen3.5) store safetensors with an extra
+    prefix like ``model.language_model.`` that differs from the runtime key
+    namespace ``model.``.  When no explicit ``_checkpoint_conversion_mapping``
+    exists, this helper detects the discrepancy and remaps LoRA keys so that
+    the merge loop can match them.
+
+    Returns a remapped ``defaultdict`` on success, or ``None`` if no prefix
+    could be inferred (caller should fall back to returning keys unchanged).
+    """
+    if not safetensor_keys:
+        return None
+
+    inferred_prefix = None
+    for lora_key in lora_weights:
+        if not isinstance(lora_key, str):
+            continue
+        suffix = lora_key + ".weight"
+        for sf_key in safetensor_keys:
+            if sf_key.endswith(suffix):
+                candidate = sf_key[: -len(suffix)]
+                if candidate:  # non-empty extra prefix
+                    inferred_prefix = candidate
+                break
+        if inferred_prefix is not None:
+            break
+
+    if inferred_prefix is None:
+        return None
+
+    remapped = defaultdict(lora_weights.default_factory)
+    for k, v in lora_weights.items():
+        new_key = inferred_prefix + k if isinstance(k, str) else k
+        remapped[new_key] = v
+    return remapped
+pass
+
+
 def _convert_lora_keys_to_safetensor_format(
     lora_weights,        # Global dict of LoraStats objects
     safetensor_keys,     # List of keys from the CURRENT shard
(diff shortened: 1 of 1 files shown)


Datta0:
Happy new year @GoldenGrapeGentleman. Wishing you a great time :)

GoldenGrapeGentleman:
Hi @Datta0, updated per your feedback!

**Code change**: Replaced `.mlp.experts` string check with `stats.module is None`. Since `nn.Linear` targets always get `.base_layer` set (`module != None`), this condition precisely identifies `nn.Parameter` (ParamWrapper) targets without path matching. `nn.ModuleList(nn.Linear)` models in transformers v4 are unaffected.

**Test results** (8× MI355X, ROCm 7.1):

| Model | Trainable | Train | Merge | Mismatch |
|-------|-----------|-------|-------|----------|
| `unsloth/gpt-oss-20b-BF16` | 46.2M | ✅ | ✅ | None ✅ |
| `unsloth/Qwen3-30B-A3B-Instruct-2507` | 211.0M | ✅ | ✅ | None ✅ |
| `imdatta0/tiny_qwen3_moe_2.8B_0.7B` | 29.0M | ✅ | ✅ | None ✅ |
| `unsloth/GLM-4.7-Flash` | 106.0M | ✅ | ✅ | None ✅ |
| `unsloth/Qwen3-VL-30B-A3B-Instruct` | 161.0M | ✅ | N/A | None ✅ |

Notes:
- `tiny_qwen3_moe`: 4-bit has numerical instability on ROCm (inf in LoRA weights after training), works fine with bf16 loading
- `GLM-4.7-Flash`: `torch._grouped_mm` not supported on ROCm, works with `UNSLOTH_MOE_BACKEND=native_torch`
- `Qwen3-VL`: get_peft_model + save_pretrained verified, full SFT needs VL dataset format

GoldenGrapeGentleman:
Hey @Datta0! I'm back from the holidays and excited to dive back into this 🎉 Thanks so much for testing with Qwen3.5 and providing the dummy model — that was super helpful!

## What happened with Qwen3.5

Great catch. The `n_saved_modules = 0` failure was NOT a count alignment issue — it was a **key prefix mismatch**:

- Safetensor keys: `model.language_model.layers.X.mlp.experts.N.gate_proj.weight`
- LoRA keys (in memory): `model.layers.X.mlp.experts`

`Qwen3_5MoeForCausalLM` has no `_checkpoint_conversion_mapping`, so `_convert_lora_keys_to_safetensor_format` returned LoRA keys unchanged, and the merge loop couldn't match anything.

## Fix: prefix inference fallback

Added a fallback in `_convert_lora_keys_to_safetensor_format`: when no explicit mapping exists, infer the prefix by matching a LoRA key suffix against safetensor keys. For Qwen3.5, `shared_expert.gate_proj.weight` provides the match → prefix `model.language_` is detected → all LoRA keys remapped correctly.

## Verified on 8× MI355X (gfx950)

**`imdatta0/small_qwen3_5_20b` (MoE-only LoRA):**
- Trainable: 99.7M
- Training: ✅ (5 steps, lr=1e-3)
- Merge: ✅ (was `n_saved_modules=0`, now succeeds)
- Weight diff after merge: `max=4.88e-4` ✅ LoRA changes preserved

**Previous models (on transformers 5.1.0):**

| Model | Train | Merge | Mismatch |
|-------|-------|-------|----------|
| `unsloth/gpt-oss-20b-BF16` | ✅ | ✅ | None ✅ |
| `unsloth/Qwen3-30B-A3B-Instruct-2507` | ✅ | ✅ | None ✅ |
| `imdatta0/tiny_qwen3_moe_2.8B_0.7B` | ✅ | ✅ | None ✅ |
| `unsloth/GLM-4.7-Flash` | ✅ | ✅ | None ✅ |
| `unsloth/Qwen3-VL-30B-A3B-Instruct` | ✅ | ✅ | None ✅ |

Note: Qwen3.5 requires transformers 5.3.0.dev0 which breaks gpt-oss patches (`gate_up_projs` → `gate_up_proj` API change), so cross-testing in a single environment isn't possible. Each model family was tested on its compatible transformers version.

Let me know if you'd like me to test anything else!

unsloth_zoo/saving_utils.py
```diff
@@ -2596,6 +2596,33 @@ def _convert_lora_keys_to_safetensor_format(
...
+                for sf_key in safetensor_keys:
+                    if sf_key.endswith(suffix):
+                        candidate = sf_key[: -len(suffix)]
+                        if candidate:  # non-empty extra prefix
+                            inferred_prefix = candidate
+                        break
+                if inferred_prefix is not None:
+                    break
+
+            if inferred_prefix is not None:
```
Datta0: NIT: What would be the right course of action if we can't find something suitable?
For eg, say `gate_up_proj` changes to `gate_up_projs` should we error here? Or do we want to handle it elsewhere

unsloth_zoo/saving_utils.py
```diff
@@ -2596,6 +2596,33 @@ def _convert_lora_keys_to_safetensor_format(
...
 
     if not forward_mapping:
+        # No explicit mapping. Some composite models (e.g. Qwen3.5) store
+        # safetensors with an extra prefix like "model.language_model." that
+        # differs from the runtime key namespace "model.". Try to detect
+        # this by matching a LoRA key against the safetensor keys and, if a
+        # consistent prefix is found, remap all LoRA keys accordingly.
+        if safetensor_keys:
+            inferred_prefix = None
+            for lora_key in lora_weights:
```
Datta0: ig we can move this out as a function, something like `sanitize_module_name` perhaps?

unsloth_zoo/saving_utils.py
```diff
@@ -390,6 +390,18 @@ def create_lora_statistics(model, merge_into_original = False, return_state_dict
...
+    # have lora_A/B/scaling but no .base_layer, leaving module_count short.
+    # nn.Linear targets always get .base_layer set, so module=None reliably
+    # identifies nn.Parameter targets. Count them to align (#3405, #3701).
+    for _key, _stats in lora_weights.items():
+        if (
+            _stats.lora_A is not None
+            and _stats.lora_B is not None
+            and _stats.module is None
+        ):
+            module_count += 1
```
Datta0, line 403: So this pretty much fixes only the count part. I think my previous changes (in the #450 perhaps) would automatically handle the right tensor and file placement things I presume.

Datta0 (commented):
Hey @GoldenGrapeGentleman thanks for the changes and updating to fix Qwen3.5
Also can you point me to the change you're referring to on transformers (gate_up_proj to gate_up_projs)?

unsloth_zoo/saving_utils.py
GoldenGrapeGentleman: When no suitable prefix can be inferred (e.g. `gate_up_proj` → `gate_up_projs` API change), `_infer_prefix_and_remap` returns `None`, and the caller falls back to returning keys unchanged — same behavior as before this PR.

I think silently falling back is the right call here rather than erroring, because:
1. The downstream merge loop already handles mismatches explicitly (it counts `n_saved_modules` and raises a clear error if 0 modules matched)
2. A hard error here would break models that genuinely have no prefix discrepancy

So the worst case is: prefix inference fails → keys pass through unchanged → merge loop catches the real mismatch and reports it. No silent data loss.

unsloth_zoo/saving_utils.py
GoldenGrapeGentleman: Done! ✅ Extracted it into `_infer_prefix_and_remap()` as a standalone function with a docstring. See commit 643f3d1.

I went with `_infer_prefix_and_remap` instead of `sanitize_module_name` since the function does two things: infer the missing prefix AND remap all keys. Happy to rename if you prefer a different name though!

unsloth_zoo/saving_utils.py
GoldenGrapeGentleman: Exactly right — this fix only addresses the `module_count` alignment so the mismatch warning no longer fires for `nn.Parameter` targets. The actual tensor placement and file writing is handled by your work in #450.

The two fixes are complementary: #450 handles the merge mechanics, this PR ensures the diagnostic counts are correct so users don't see a misleading warning during an otherwise successful merge.

Datta0 (approved):

danielhanchen:
Thanks for this, and apologies it sat open for a while. Both changes here have since landed on main through later work, so I'm closing this as superseded rather than merged.

- The MoE `target_parameters` count alignment (the `TODO: handle MoE target_parameters to align counts` at the old line 410) is now handled in `create_lora_statistics` in `saving_utils.py`, using the same `_stats.module is None` check to count the `nn.Parameter` (ParamWrapper) targets that lack a `.base_layer`, and referencing the same issues (#3405, #3701). The misleading `[Unsloth merge debug] LoRA count mismatch` diagnostic no longer fires for gpt-oss-style MoE target_parameters.
- The composite-model key-prefix inference (`_infer_prefix_and_remap`) is also in main now, extended beyond the Qwen3.5 extra-prefix case to also cover Mistral 3 style `model.language_model.` to `language_model.model.` reorderings.

So the behavior you fixed is in place. Appreciate you digging into that TODO and filing this.

PR #504, opened 2026-02-18
fix: handle LFM2/Mamba hybrid layers in _get_vllm_state_dict for fast_inference

devchilll:
## Summary
Fixes `UnboundLocalError` in `_get_vllm_state_dict` when using `fast_inference=True` with LFM2/Mamba hybrid models, e.g., `LiquidAI/LFM2.5-1.2B-Thinking`

## Problem
When loading LFM2 models with `fast_inference=True`, `_get_vllm_state_dict` crashes with:
```
UnboundLocalError: cannot access local variable 'prefix'
```
This happens because the layer iteration loop only handles `self_attn` and `cross_attn` layer types. LFM2 is a hybrid architecture with **short convolution layers** that have neither attribute, so `prefix` is never assigned. Additionally, LFM2 uses different naming conventions: `out_proj` instead of `o_proj`, `feed_forward.w1/w2/w3` instead of `mlp.gate_proj/up_proj/down_proj`, `embedding_norm` instead of `norm`

## Fix
- Added `short_conv` branch to extract conv `in_proj`, `out_proj`, and `conv` weights from non-attention layers
- Moved `o_proj`, `out_proj` extraction inside each attention branch so conv layers don't crash..
- Added conditional MLP extraction: `mlp` path (standard) vs `feed_forward` path w LFM2 w1/w2/w3
- Handle `embedding_norm` (LFM2) alongside `norm` (standard) for final model norm
- Added LFM2 layer templates to `get_model_layer_config` for HF model reconstruction

## Changes
- `vllm_utils.py`: `_get_vllm_state_dict`
- `empty_model.py`: `get_model_layer_config` && `set_additional_modules`

Related: unslothai/unsloth#4073

CC @danielhanchen @Datta0


unsloth_zoo/empty_model.py
```diff
@@ -539,6 +546,17 @@ def get_model_layer_config(return_non_layered=True):
...
             "model.layers.{kk}.mlp.gate_up_proj", # for extracting from vLLM (phi3 architecture)
             "model.layers.{kk}.mlp.down_proj",
+
+            # LFM2 hybrid model layers (attention + short convolution)
+            "model.layers.{kk}.self_attn.out_proj",  # LFM2 attention uses out_proj instead of o_proj
+            "model.layers.{kk}.self_attn.q_layernorm",
+            "model.layers.{kk}.self_attn.k_layernorm",
+            "model.layers.{kk}.conv.in_proj",
+            "model.layers.{kk}.conv.out_proj",
+            "model.layers.{kk}.conv.conv",
```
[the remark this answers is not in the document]
devchilll: Good catch. updated. removed `conv.conv`

unsloth_zoo/empty_model.py
```diff
@@ -539,6 +546,17 @@ def get_model_layer_config(return_non_layered=True):
             "model.layers.{kk}.mlp.up_proj",
             "model.layers.{kk}.mlp.gate_up_proj", # for extracting from vLLM (phi3 architecture)
             "model.layers.{kk}.mlp.down_proj",
+
+            # LFM2 hybrid model layers (attention + short convolution)
+            "model.layers.{kk}.self_attn.out_proj",  # LFM2 attention uses out_proj instead of o_proj
+            "model.layers.{kk}.self_attn.q_layernorm",
+            "model.layers.{kk}.self_attn.k_layernorm",
```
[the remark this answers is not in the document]
devchilll: good catch. done.

unsloth_zoo/vllm_utils.py
```diff
@@ -1160,9 +1219,15 @@ def _is_fused_module(name: str) -> bool:
...
-    quant_state_dict[norm_prefix] = state_dict[norm_prefix]
+    # LFM2 uses embedding_norm instead of norm
+    if hasattr(vllm_text_model, "norm"):
+        norm_prefix = f"{vllm_text_model_prefix}.norm.weight"
+        state_dict[norm_prefix] = vllm_text_model.norm.weight.data
+        quant_state_dict[norm_prefix] = state_dict[norm_prefix]
+    elif hasattr(vllm_text_model, "embedding_norm"):
+        norm_prefix = f"{vllm_text_model_prefix}.embedding_norm.weight"
+        state_dict[norm_prefix] = vllm_text_model.embedding_norm.weight.data
+        quant_state_dict[norm_prefix] = state_dict[norm_prefix]
```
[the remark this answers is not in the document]
devchilll: low key valid but in this case there are only 2 conditions... doesn't make much diff imho but could also update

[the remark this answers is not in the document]
devchilll: atually I think keep as is - would be more readable so gonna reject this gemini code comment

unsloth_zoo/vllm_utils.py
```diff
@@ -1118,22 +1127,72 @@ def _is_fused_module(name: str) -> bool:
...
+            # HF expects nn.Conv1d weight with same shape (hidden, 1, kernel) for depthwise conv
+            conv_module = short_conv.conv
+            conv_weight = getattr(conv_module, "base_layer", conv_module).weight
+            conv_weight.requires_grad_(False)
+            state_dict[f"{conv_prefix}.conv.weight"] = conv_weight
+            quant_state_dict[f"{conv_prefix}.conv.weight"] = conv_weight
+            # Handle conv bias if present
+            conv_bias = getattr(conv_module, "bias", None)
+            if conv_bias is None:
+                conv_bias = getattr(getattr(conv_module, "base_layer", conv_module), "bias", None)
```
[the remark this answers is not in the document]
devchilll: done

unsloth_zoo/vllm_utils.py
```diff
@@ -1118,22 +1127,72 @@ def _is_fused_module(name: str) -> bool:
             get_state_dict(f"{prefix}.q_proj", 0, state_dict, q_proj)
             get_state_dict(f"{prefix}.k_proj", 1, state_dict, kv_proj)
             get_state_dict(f"{prefix}.v_proj", 2, state_dict, kv_proj)
+            get_state_dict(f"{prefix}.o_proj", 0, state_dict, o_proj)
+
+        elif hasattr(layer, "short_conv"):
```
Datta0: We have similar processing for conv layers [in here](https://github.com/unslothai/unsloth-zoo/blob/e120cac148bbc557a0040d100f16b49d3f77bf8d/unsloth_zoo/empty_model.py#L852) Can we put this there?
This function is specifically for LLMs and we'd like to keep it clean that way

unsloth_zoo/vllm_utils.py
```diff
@@ -1107,6 +1106,16 @@ def _is_fused_module(name: str) -> bool:
                 get_state_dict(f"{prefix}.q_proj", 0, state_dict, qkv_proj)
                 get_state_dict(f"{prefix}.k_proj", 1, state_dict, qkv_proj)
                 get_state_dict(f"{prefix}.v_proj", 2, state_dict, qkv_proj)
+
+            # Extract o_proj or out_proj depending on model architecture
+            # LFM2 uses out_proj, most other models use o_proj
+            if hasattr(layer.self_attn, "o_proj"):
```
Datta0: All this branching looks way too much for a single model. Can you please think of a simpler way that preferably doesn't disturb a lot of existing code?

Datta0 (commented):
Hey @devchilll thanks for your contributions. A few comments 

Also we'd like you to run [this function](https://github.com/unslothai/unsloth-zoo/blob/e120cac148bbc557a0040d100f16b49d3f77bf8d/unsloth_zoo/vllm_utils.py#L2901) on both the new model (LFM in this case) and older models like llama, qwen (and VL), gemma (text and vision) to ensure that we are not breaking anything?

unsloth_zoo/vllm_utils.py
devchilll: Thanks for the feedback! -- the branching is getting heavy. I'll move the `short_conv` extraction into a helper function `_extract_short_conv_layer()` to keep the main loop clean. however, the branching itself is necessary because LFM2 layers are architecturally distinct cause its attention vs convolution, but this will make `_get_vllm_state_dict` much more readable. wdyt?

unsloth_zoo/vllm_utils.py
devchilll: super valid! like mentioned in the reply above, I think moving conv to some helper function would make code look leaner. Else the added support for this new model category will indeed add more if else conditions.  Will verify with test() as suggested to make sure the foundamental things don't break.

devchilll: Turns out that my macbook pro supports torch up to 2.2.2 cause its on intel chips... I gotta find a cloud instance or something that supports torch 2.4.0 or above to install unsloth-zoo locally

Datta0: You can try on google colab.

gaztrabisme: @devchilll Hey there, I just got back from holidays. Thanks for opening this PR! Btw I got GPU, do you need testing anything?

Datta0:
Hey @devchilll , we'd ideally appreciate if you can find a cleaner or simpler way to handle this instead of doing `if else` everywhere that too only for one model

unsloth_zoo/vllm_utils.py
devchilll: The other way around, is to add a if else condition at the beginning and like separate the path for LFM2. So this makes the LFM2 a special case, and then leaves everything else (every other path) untouched. This `if lfm2?` check will happen before the `for kk in range(len(vllm_text_model.layers)):` for loop for other models.

devchilll: @gaztrabisme any ideas?

devchilll (commented):
## Refactor
Moved all LFM2-specific logic out of the main `_get_vllm_state_dict` loop into a dedicated `_extract_lfm2_layer` function. The main loop now uses an outer if `model_type == "lfm2" / else` dispatch so: 

- LFM2 gets its own clean loop calling `_extract_lfm2_layer`
- The standard model loop body is unchanged from main (just re-indented)
- Zero `elif hasattr(...)` branches for LFM2 polluting the standard path

Easy to add other hybrid architectures (e.g. Mamba, RWKV) the same way later

Also extracted `_extract_short_conv_layer` as a helper for the conv layer weight extraction.

## Bug Fixes 

- Conv `in_proj `weight shape mismatch
-  LoRA registration fails on conv modules
-  `lm_head.weight` not initialized warning - No code change needed. 

CC @Datta0 @gaztrabisme

unsloth_zoo/vllm_utils.py
devchilll: @gaztrabisme  yes! did 1) refactor 2) bug fixes 3) sync to head. Plz help test and also feel free to edit files. thanks

unsloth_zoo/empty_model.py
```diff
@@ -442,10 +442,37 @@ def set_embedding(module, embed_tokens_key, pad_token_id, requires_grad=False):
...
+        norm = torch.nn.Parameter(norm, requires_grad = False)
+        language_model.norm.weight = norm
+    elif embedding_norm_key in quant_state_dict:
+        norm = quant_state_dict[embedding_norm_key]
+        norm = torch.nn.Parameter(norm, requires_grad = False)
+        language_model.embedding_norm.weight = norm
+
+    # LFM2 conv.conv weights — must be assigned directly to preserve nn.Conv1d module type
+    # (standard_layers uses nn.Linear for all entries, which would break Conv1d)
+    text_config_tmp = getattr(config, "text_config", config)
```
Datta0: I'd recommend having a function like `_extract_lfm2_layer` to deal with these differences

unsloth_zoo/vllm_utils.py
```diff
@@ -1088,65 +1171,72 @@ def _is_fused_module(name: str) -> bool:
...
-        if hasattr(layer, "self_attn"):
-            prefix = f"{vllm_text_model_prefix}.layers.{kk}.self_attn"
-            qkv_proj = layer.self_attn.qkv_proj
-            o_proj = layer.self_attn.o_proj
-
-            use_fused_qkv = _is_fused_module("qkv_proj")
-            if use_fused_qkv:
-                # For some model types like phi3 vllm will expect fused qkv (e.g. Phi3, Phi3.5-mini-instruct, Phi4-mini-instruct)
+    if model_type == "lfm2":
+        for kk in range(len(vllm_text_model.layers)):
```
Datta0: I'd say 
```
for kk ...:
  if lfm: extract_lfm; return
  rest of the code with no `else`
```
is perhaps better?

unsloth_zoo/vllm_utils.py
```diff
@@ -1868,6 +1964,12 @@ def load_vllm(
             f"This may cause an out-of-memory crash with standby mode. Consider lowering gpu_memory_utilization."
         )
 
+    # LFM2 conv modules are not BaseLayerWithLoRA — disable LoRA to avoid
+    # AssertionError during LoRA registration in vLLM's create_lora_manager.
+    if getattr(config, "model_type", "") == "lfm2":
```
Datta0, line 1969: Wait on, if we disable LoRA whats the point of this anymore?

unsloth_zoo/empty_model.py
gaztrabisme: On it, will push a commit shortly.

unsloth_zoo/vllm_utils.py
gaztrabisme: Makes sense, will switch to early-return style.

unsloth_zoo/vllm_utils.py
gaztrabisme: Sorry, I let Claude Code run amok on that fix. Looked into it properly — vLLM's `Lfm2ForCausalLM` doesn't define `supported_lora_modules` at all, so LoRA isn't supported for LFM2 in vLLM yet. `enable_lora = False` just avoids the crash. We could open a PR on vLLM to add LoRA support for LFM2 — @devchilll want to work on that together?

gaztrabisme: Update: turns out vLLM already merged LoRA support for LFM2 (vllm-project/vllm#34921, merged Feb 20). It's not in any release yet (our 0.11.2 doesn't have it), but once Unsloth upgrades vLLM we can drop the `enable_lora = False` guard. So no action needed from us — just a version bump down the line.

unsloth_zoo/vllm_utils.py
devchilll: @Datta0 No. You can't do early return right there because there's common code after the `for kk` loop. the latest commit of the `for kk: if lfm, extract lfm layer, continue;` does make sense in turns of extracting layers correctly.

danielhanchen:
Confirmed the gap is still there on main, where the layer loop in vllm_utils.py only handles self_attn and cross_attn and empty_model.py indexes model.norm.weight unconditionally, so LFM2 conv layers break extraction. Could you rebase onto current main (that loop now has _unwrap_tensor and the shared-KV handling) and fold the conv reconstruction into the existing conv path in empty_model.py rather than a separate LFM2 setter? Note #531 covers the same ground, so it is worth agreeing which branch carries this.

danielhanchen:
Confirmed main still has no LFM2 handling in unsloth_zoo/vllm_utils.py or empty_model.py, so the conv/feed_forward/embedding_norm extraction here does fill a real gap, and #565 looks like the same prefix crash on NemotronH which this does not cover. Note #531 is open covering the same ground, so could you check it against yours and keep the LFM2 branch as narrow as possible so the standard loop stays untouched?

danielhanchen:
Confirmed the loop in unsloth_zoo/vllm_utils.py still only assigns `prefix` for self_attn and cross_attn layers, so the conv layers in LFM2 do break there, and the GPU run posted here is useful evidence. There is an overlapping branch in #531 doing the same job, so let us settle which one we take before another round of changes here.

PR #510, opened 2026-02-23
fix: skip non-attention layers in _get_vllm_state_dict (fixes unslothai/unsloth#4073)

stakeswky:
## Problem

`_get_vllm_state_dict` crashes with `UnboundLocalError: cannot access local variable 'prefix'` when processing LFM2/Mamba models (e.g. `LiquidAI/LFM2.5-1.2B-Thinking`).

The layer iteration loop only sets `prefix` inside `if hasattr(layer, "self_attn")` and `elif hasattr(layer, "cross_attn")` branches. Mamba/SSM layers use `mixer` instead, so neither branch executes and `prefix` is never assigned before the subsequent `get_state_dict(f"{prefix}.o_proj", ...)` call.

## Fix

Add an `else` branch that logs and skips layers without `self_attn` or `cross_attn`. This is the minimal-invasive fix — it prevents the crash and allows models with mixed layer types (attention + Mamba) to still extract the attention layers correctly.

## Future work

A more complete fix would add an `elif hasattr(layer, "mixer")` branch that properly extracts Mamba layer state dicts (A, B, C, D matrices, dt projections, etc.) for full round-trip fidelity. This skip-based approach is a safe stopgap.

Fixes unslothai/unsloth#4073

### DIFF
--- a/unsloth_zoo/vllm_utils.py
+++ b/unsloth_zoo/vllm_utils.py
@@ -1121,7 +1121,7 @@ def _get_vllm_state_dict(llm, return_state_dict = False, config = None, is_visio
         else:
             # Skip layers that don't have self_attn or cross_attn (e.g. Mamba/SSM layers
             # like LFM2's mixer layers). Full Mamba state dict extraction can be added later.
-            logger.info(f"Unsloth: Skipping layer {kk} — no self_attn or cross_attn found.")
+            logger.info(f"Unsloth: Skipping layer {kk} - no self_attn or cross_attn found.")
             continue
 
         get_state_dict(f"{prefix}.o_proj", 0, state_dict, o_proj)
--- a/unsloth_zoo/vllm_utils.py
+++ b/unsloth_zoo/vllm_utils.py
@@ -1118,6 +1118,11 @@ def _get_vllm_state_dict(llm, return_state_dict = False, config = None, is_visio
             get_state_dict(f"{prefix}.q_proj", 0, state_dict, q_proj)
             get_state_dict(f"{prefix}.k_proj", 1, state_dict, kv_proj)
             get_state_dict(f"{prefix}.v_proj", 2, state_dict, kv_proj)
+        else:
+            # Skip layers that don't have self_attn or cross_attn (e.g. Mamba/SSM layers
+            # like LFM2's mixer layers). Full Mamba state dict extraction can be added later.
+            logger.info(f"Unsloth: Skipping layer {kk} — no self_attn or cross_attn found.")
+            continue
 
         get_state_dict(f"{prefix}.o_proj", 0, state_dict, o_proj)
 


danielhanchen (approved):
Tested on 1x NVIDIA B200 (CUDA 12.8, torch 2.9.1, vLLM 0.15.1) with Qwen3_4B_GRPO notebook -- PASS (training + inference OK).

Code review:
- The fix is correct. `continue` properly skips the `o_proj`, MLP, and layernorm extraction that follow the attention prefix assignment, preventing the `UnboundLocalError` from issue #4073.
- Minimal and safe approach. Mamba/SSM layer state dict extraction can be added separately.

One minor nit: the log message uses an em dash (U+2014) in "Skipping layer {kk} --- no self_attn". Consider using a plain ASCII dash for consistency with the rest of the codebase, though this is cosmetic only.

LGTM.

danielhanchen (commented):
Tested on 1x NVIDIA B200 (CUDA 12.8, torch 2.9.1, vLLM 0.15.1).

**Verification:**
- Regression test: `unsloth/Llama-3.2-1B-Instruct` with `fast_inference=True` -- PASS. `_get_vllm_state_dict` processes all attention layers correctly, inference produces valid output.
- LFM2 test: `LiquidAI/LFM2.5-1.2B-Thinking` with `fast_inference=True` -- vLLM's BitsAndBytes loader fails before reaching `_get_vllm_state_dict` (vLLM doesn't fully support Lfm2ForCausalLM quantization yet). Confirmed the `UnboundLocalError` on `prefix` is NOT hit.
- Unit tests: verified `continue` correctly skips `o_proj` (line 1127), `mlp.gate_up_proj` (line 1129), `mlp.down_proj` (line 1140), and layernorm extraction (lines 1143-1154) for layers without `self_attn`/`cross_attn`. Tested all-Mamba, mixed-ordering, and Mamba-first-layer edge cases.

**Code review:** Fix is correct and minimal. Two small nits:

1. Unclosed parenthesis in comment (line 1122): `(e.g. Mamba/SSM layers` -- missing closing `)`.
2. Consider `logger.warning` instead of `logger.info` -- silently dropping an entire layer's weights from the state dict is significant for debugging. A warning makes it more visible when users check extraction completeness. This aligns with the codex bot's inline comment about avoiding silent weight dropping.

Neither nit blocks merge.

danielhanchen (commented):
**Update: End-to-end verification completed.**

Successfully reproduced the bug and confirmed the fix works.

**Bug reproduction (unpatched main, 16-bit LFM2):**
```
  File "vllm_utils.py", line 1122, in _get_vllm_state_dict
    get_state_dict(f"{prefix}.o_proj", 0, state_dict, o_proj)
                      ^^^^^^
UnboundLocalError: cannot access local variable 'prefix' where it is not associated with a value
```

**With PR fix applied (16-bit LFM2):**
```
[WARNING] Unsloth: Skipping layer 0 - no self_attn or cross_attn found.
[WARNING] Unsloth: Skipping layer 1 - no self_attn or cross_attn found.
```
The `else: continue` branch fires correctly for the non-attention layers. No more `UnboundLocalError`.

Note: LFM2 then hits a separate, unrelated issue on its attention layers (`Lfm2Attention` uses `out_proj` instead of `o_proj` -- line 1098). That is out of scope for this PR and would need a separate fix for full LFM2 support.

Earlier 4-bit test did not reach `_get_vllm_state_dict` because vLLM's BnB loader fails for Lfm2ForCausalLM. Loading in 16-bit successfully reaches the function and exercises the fix.

danielhanchen:
main now pulls o_proj inside each attention branch in unsloth_zoo/vllm_utils.py and skips layers without an mlp, so the unbound prefix path this patched is gone. Sorry for the slow review, and thanks for tracking it down.

danielhanchen:
main now assigns and uses `prefix` entirely inside each attention branch in unsloth_zoo/vllm_utils.py (the o_proj call moved to line 1218), so layers without self_attn or cross_attn already fall through without the UnboundLocalError and I do not think this is needed any more. Sorry for the slow review, and thanks for putting it together.

PR #512, opened 2026-02-24
Enable ROCm GPU acceleration for llama.cpp GGUF export

GoldenGrapeGentleman:
# Enable ROCm GPU acceleration for llama.cpp GGUF export

## Summary

Adds automatic ROCm/HIP detection in `install_llama_cpp()` so AMD GPU users get hardware-accelerated GGUF inference.

## Problem

On AMD ROCm systems, llama.cpp was compiled without GPU support, resulting in CPU-only inference that is 5–9x slower than GPU-accelerated.

## Solution

After `gpu_support = "ON"` is confirmed, detect the active GPU backend:
- **ROCm systems**: reads `torch.version.hip`, queries `gcnArchName` for the target arch, sets `-DGGML_HIP=ON` with clang/clang++ compilers and `CMAKE_HIP_ARCHITECTURES`
- **CUDA systems**: falls through to existing `-DGGML_CUDA=ON` flags (no behavior change)
- **`gpu_support=False` (default)**: entire block is skipped (no behavior change for existing users)

## Changes

**unsloth_zoo/llama_cpp.py** (`+33/-2`):
- ROCm detection block inserted after `gpu_support == "ON"` confirmation
- Hardcoded `-DGGML_CUDA={gpu_support}` replaced with dynamic `gpu_cmake_flags` in both Linux and Windows cmake paths
- Fixed guard bug from review: `if gpu_support == "ON":` (was `if gpu_support:` which treats `"OFF"` as truthy)

## Testing

Tested on AMD Radeon PRO W7900 (gfx1100, ROCm 7.1):
- GPU inference: **312.4 t/s** (5.9x faster than CPU-only)
- ROCm libraries correctly linked via HIP flags
- CUDA systems: no regression (guard prevents ROCm block from running)

## Notes

- Rebased cleanly onto current upstream main (1 commit ahead, 0 behind)
- Resolves conflict with upstream Windows cmake refactor
- Companion PR unslothai/unsloth#4103 (save.py `gpu_support=True`) was closed by @danielhanchen pending investigation of compile time. This PR keeps the detection logic ready for when that decision is revisited.

### DIFF
--- a/unsloth_zoo/llama_cpp.py
+++ b/unsloth_zoo/llama_cpp.py
@@ -673,6 +673,37 @@ def install_llama_cpp(
     if gpu_support == "ON":
         print("Unsloth: Building llama.cpp with GPU support")
 
+    # Detect GPU backend for CMake flags (ROCm vs CUDA)
+    gpu_cmake_flags = f"-DGGML_CUDA={gpu_support}"
+    if gpu_support == "ON":
+        try:
+            import torch
+            if hasattr(torch.version, 'hip') and torch.version.hip is not None and torch.cuda.is_available():
+                # ROCm detected
+                try:
+                    gpu_arch = torch.cuda.get_device_properties(0).gcnArchName.split(":")[0]
+                except Exception:
+                    gpu_arch = ""
+                rocm_path = os.environ.get('ROCM_PATH', '/opt/rocm')
+                if gpu_arch:
+                    gpu_cmake_flags = (
+                        f"-DGGML_HIP=ON "
+                        f"-DCMAKE_C_COMPILER={rocm_path}/llvm/bin/clang "
+                        f"-DCMAKE_CXX_COMPILER={rocm_path}/llvm/bin/clang++ "
+                        f"-DCMAKE_HIP_ARCHITECTURES={gpu_arch}"
+                    )
+                else:
+                    gpu_cmake_flags = (
+                        f"-DGGML_HIP=ON "
+                        f"-DCMAKE_C_COMPILER={rocm_path}/llvm/bin/clang "
+                        f"-DCMAKE_CXX_COMPILER={rocm_path}/llvm/bin/clang++"
+                    )
+                print(f"Unsloth: Detected ROCm GPU{' (' + gpu_arch + ')' if gpu_arch else ''} -- building with HIP support")
+            else:
+                gpu_cmake_flags = f"-DGGML_CUDA={gpu_support}"
+        except Exception:
+            gpu_cmake_flags = f"-DGGML_CUDA={gpu_support}"
+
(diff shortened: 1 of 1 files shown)


danielhanchen (changes requested):
Code review -- found a bug in the gpu_support guard logic.

**Bug**: `gpu_support` is converted from bool to string (`"ON"/"OFF"`) at line 427 *before* the new detection block. The check `if gpu_support:` on line ~430 evaluates `"OFF"` as truthy (non-empty string), so the detection block always runs regardless of the original `gpu_support` parameter.

When `gpu_support=False` is passed (e.g. from `unsloth/save.py` lines 1212-1218 where `install_llama_cpp` is called with `gpu_support=False` and the comment says "GGUF conversion doesn't need CUDA"):
- Original behavior: `-DGGML_CUDA=OFF` is set
- With this PR: detection block runs, `torch.cuda.is_available()` returns True on GPU machines, sets `-DGGML_CUDA=ON` instead of `-DGGML_CUDA=OFF`
- This means GGUF builds attempt CUDA compilation even when explicitly disabled

**Fix**: Change `if gpu_support:` to `if gpu_support == "ON":` to match the already-converted string type.

The ROCm detection logic itself looks correct and well-structured. Just needs the guard fix.

GoldenGrapeGentleman:
Hi @danielhanchen, thanks for the review!

Fixed the `gpu_support` guard bug — changed `if gpu_support:` to `if gpu_support == "ON":` so it correctly respects `gpu_support=False` (which becomes `"OFF"` string earlier in the function). Previously `"OFF"` was truthy and the GPU detection block always ran, potentially overriding an explicit OFF request.

Please take another look when you get a chance. 🙏

danielhanchen:
Confirmed llama_cpp.py still passes -DGGML_CUDA in both cmake paths, so an explicit gpu_support=True on ROCm gets a CUDA-only build, and the gpu_support == "ON" guard fix is in. Will get this reviewed.

danielhanchen:
Confirmed the source-build path in unsloth_zoo/llama_cpp.py still passes only -DGGML_CUDA at the Linux and Windows cmake calls, so this is still needed, but main now has a _detect_gpu_target() helper that already returns ("rocm", gfx) plus _rocm_gfx_family. Could you rebase onto current main and reuse that helper instead of the inline torch.version.hip block, so there is one detector?

danielhanchen:
Confirmed llama_cpp.py still passes a hardcoded -DGGML_CUDA on both the Windows and Linux configure paths, so the HIP detection is still needed; one thing before I review, on Windows the flags go into a cmake argv list, so the multi flag ROCm string would be passed as a single argument, could you split it into separate list entries there?

PR #520, opened 2026-02-26
Fix vLLM vision GRPO compatibility for issue #4081

danielhanchen:
## Summary
This fixes two vLLM compatibility issues that broke GRPO vision notebook runs referenced from unsloth issue `unslothai/unsloth#4081`.

1. Add `WorkerLoRAManager.supports_tower_connector_lora()` fallback in `unsloth_zoo/vllm_lora_worker_manager.py`.
2. Patch multimodal `_seq2text` decode handling in `unsloth_zoo/vllm_utils.py` to normalize nested token payloads (`dict`/`list`/`tensor`) before decode.

## Why
GRPO vision runs were failing with:
- missing `supports_tower_connector_lora` on the worker manager path
- multimodal decode path receiving nested/non-flat token payloads

Both failure paths are vLLM integration edges that should be handled in the Zoo patch layer.

## Validation
Reproduced and reran issue notebooks with `max_steps=5` where needed:
- `qwen2_5_7b_vl_grpo_postfix4`: PASS (`train_runtime=197.8613s`)
- `qwen3_vl_8b_vision_grpo_postfix1`: PASS (`train_runtime=105.9599s`)
- `openenv_gpt_oss_20b_postfix2`: PASS (`train_runtime=817.5183s`)

Logs and rerun summary:
- `4_prs_check/logs/issue4081/postfix_reruns_summary.txt`
- `4_prs_check/logs/issue4081/*.log`

## Notes
This PR is intentionally scoped to the vLLM compatibility layer and keeps behavior unchanged when payloads are already in standard token-id form.


unsloth_zoo/vllm_lora_worker_manager.py
```diff
@@ -271,6 +271,24 @@ def list_adapters(self) -> Set[int]:
         else:
             return set(self._adapter_manager.list_adapters())
 
+    def supports_tower_connector_lora(self) -> bool:
```
Datta0, line 274: I see where this comes from. Is this a necessity or a good to have?
Cuz from the looks of it, this is primarily aimed at mm_proj or those kind of modules LoRA and is kinda experimental [acc to their docs](https://github.com/vllm-project/vllm/commit/ce1eafd1a56bd5a4eb1ee743933a5e96e63d2116#diff-f5a83dbb60cced22f546e6ff6390b5625cfe71879bcfd041545a2cf5c007b4fdR280)

unsloth_zoo/vllm_utils.py
```diff
@@ -372,6 +372,79 @@ def patch_vllm_lora_load_tensors():
...
+    def patch_vllm_multimodal_seq2text():
+        # vLLM multimodal prompt updates can pass token payloads as nested
+        # dict/list/tensor objects. Normalize these payloads before decode.
+        try:
+            import vllm.multimodal.processing.processor as mm_processor
+            original_seq2text = mm_processor._seq2text
+            if getattr(original_seq2text, "__unsloth_patched_seq2text__", False):
+                return
+
+            def _extract_token_ids(payload, depth = 0):
```
Datta0: Looks like we're heavily over complicating this. We should ideally check for 3-4 known things
list(int), list(list(int)), tensor. All the depth thing seems complicated

danielhanchen (commented):
## Review Summary

The two patches are necessary -- `supports_tower_connector_lora` prevents `AttributeError` in vLLM v1 vision code, and the `_seq2text` normalizer handles non-standard payload types that upstream `_seq2text` (which only accepts `str | list[int]`) cannot process.

Validated `_extract_token_ids` with unit tests covering: `list[int]`, `torch.Tensor`, `np.ndarray`, `dict` extraction, nested list flattening, string passthrough, single int, tuple, empty list, `np.integer` scalars. All pass.

### Issues

**1. `supports_mm` default should be `False` not `True`** (vllm_lora_worker_manager.py:281)
`getattr(adapter_manager, "supports_mm", True)` assumes multimodal support when the attribute is absent. Upstream vLLM sets `supports_mm` as an explicit instance attribute -- if it's missing, the adapter manager likely doesn't support multimodal. Default should be `False` to be conservative.

**2. `_extract_token_ids` over-engineered** (vllm_utils.py:384-431)
Agree with @Datta0's comment. The depth-8 recursion and 8 preferred dict keys are complex for what should handle 3-4 known types: `torch.Tensor`, `np.ndarray`, `list[list[int]]`, and `dict` with `token_ids`/`input_ids`. A simpler version would be more maintainable and less likely to mask bugs by silently normalizing unexpected inputs.

**3. Inconsistent return type from `_extract_token_ids`**
The function returns `list[int]`, `None`, or `str`. The string path causes: `_extract_token_ids(["a", "b"])` returns `None` because recursive calls return strings which fail `isinstance(found, list)`. Not a crash (wrapper falls back to original `seq`), but it means multi-element string lists silently bypass normalization.

**4. Bare `except:` on line 444** -- should be `except Exception:` to avoid catching `KeyboardInterrupt`/`SystemExit`. Consistent with codebase convention but easy to fix.

**5. Issue #4081 reference** -- Issue #4081 is about a tensor shape mismatch (828 vs 824) in `compute_loss`. The errors this PR fixes (missing `supports_tower_connector_lora` + nested payloads in `_seq2text`) are different. The PR description should clarify the relationship.

unsloth_zoo/vllm_utils.py
```diff
@@ -372,6 +372,76 @@ def patch_vllm_lora_load_tensors():
...
+                if isinstance(payload, (int, np.integer)):
+                    return [int(payload)]
+                if isinstance(payload, (tuple, list)):
+                    if len(payload) == 0:
+                        return payload
+                    # Already flat list[int]
+                    if all(isinstance(x, (int, np.integer)) for x in payload):
+                        return [int(x) for x in payload]
+                    # One level of nesting: list[list[int]] -> list[int]
+                    if all(isinstance(x, (list, tuple)) for x in payload):
```
Datta0, line 412: Is there any good reason to do that? is it not like merging tokens of diff unrelated sequences?

PR #519, opened 2026-02-25
Fix missing ParameterModule export in GPT-OSS compiler path

danielhanchen:
## Summary
Fixes GPT-OSS compiled-cache failures where generated modules can reference `ParameterModule` but the symbol is not available from `transformers.models.gpt_oss.modeling_gpt_oss`.

## Root cause
`ParameterModule` is defined in `unsloth_zoo.temporary_patches.gpt_oss`, but compiler-generated code expects symbols to be resolvable from the GPT-OSS modeling namespace.

## Changes
- Added `patch_gpt_oss_compiler_exports()` in `unsloth_zoo/temporary_patches/gpt_oss.py`.
- Exported compiler-required symbols into `transformers.models.gpt_oss.modeling_gpt_oss`:
  - `ParameterModule`
  - `swiglu_torch_forward`
  - `dtype_from_config`
  - `transformers_version`
  - `Version`
- Registered the hook via `TEMPORARY_PATCHES.append(patch_gpt_oss_compiler_exports)`.
- Normalized GPT-OSS guard matching with `UNSLOTH_MODEL_NAME.replace("-", "_")` usage across key 4-bit/load guards.
- Cleaned compiler export patch flow in exception path and removed redundant `pass`.
- Updated compiler decorator stripping to handle:
  - `@use_kernel_forward_from_hub(...)`
  - `@auto_docstring`
  This removes noisy unknown-decorator warnings in GPT-OSS compile paths.

## Validation
Run dirs:
- `logs/pr519_verify_20260226_031808/test_gpt_oss_compiler_export.log`
- `logs/pr519_verify_20260226_031808/test_gpt_oss_compiler_export_after_autodoc_reinstall.log`

Results:
- GPT-OSS compiler export smoke test: PASS
- `@use_kernel_forward_from_hub` unknown decorator warning: removed
- `@auto_docstring` unknown decorator warning: removed

## Notes
This change is scoped to symbol export and compile-time compatibility. It does not modify model math.

PR #521, opened 2026-02-26
Guard GPT-OSS allocator warmup on low-memory 4-bit loads

danielhanchen:
## Summary
- Guard `transformers.modeling_utils.caching_allocator_warmup` in Unsloth patch flow.
- Auto-skip warmup on low effective accelerator memory (`< 24 GiB`) across model loads.
- Keep explicit override:
  - `UNSLOTH_ALLOCATOR_WARMUP=on|off|auto`
- Rename warmup guard identifiers to generic names.
- Normalize GPT-OSS model-name guards to handle hyphenated names (`gpt-oss` -> `gpt_oss`) consistently.
- Remove legacy `UNSLOTH_GPT_OSS_ALLOCATOR_WARMUP` alias support.
- Use the active accelerator index (not hardcoded device `0`) for memory and per-process-fraction checks.
- Strip `@use_kernel_forward_from_hub(...)` and `@auto_docstring` decorators in class rewrite path to remove unknown-decorator warning noise.

## Why
- `caching_allocator_warmup` can allocate a large single chunk before weights load. On low-memory setups this can OOM before model load completes.
- GPT-OSS patch activation checks were using raw `UNSLOTH_MODEL_NAME` substring checks in multiple places. Hyphenated names can skip patch paths that should apply for 4-bit GPT-OSS loads.
- Multi-GPU/non-default-device runs can mis-detect memory when checks are hardcoded to device `0`.

## Validation
Run dir:
- `logs/memory_investigation_20260226_000809/warmup_24gb_global_20260226_014122`

On `Unsloth 2026.2.1` / `transformers 4.56.2`:
- `cap16_auto_unsloth2026_generic_names.log`: PASS
  - `GPU_RESERVED_GB_AFTER_LOAD 11.85`
  - `GPU_PEAK_RESERVED_GB_AFTER_LOAD 12.932`
- `cap16_force_on_unsloth2026_generic_names.log`: FAIL (expected control)
  - `OutOfMemoryError` with `Tried to allocate 18.93 GiB`
- `test_llama_compile_smoke.log`: PASS

Hyphen-guard/load checks:
- `logs/hyphen_guard_verify_20260226_020925/repro_uninitialized_after_local_patch.log`: PASS (`LOAD_OK`)

Version matrix (`transformers==4.56.2`):
- Run dir: `logs/version_matrix_gpt_oss_20260226_020432`
- `old_2025_12` (`unsloth==2025.12.10`, `unsloth-zoo==2025.12.8`):
  - `load_ok=true`, `train_ok=true`, `exception_type=null`
  - `losses=[1.165579, 4.171052, 3.063763]`
  - `grad_norms=[2.650939, inf, 1.8630026331731638e+17]`
- `new_2026_2_1` (`unsloth==2026.2.1`, `unsloth-zoo==2026.2.1`):
  - `load_ok=true`, `train_ok=true`, `exception_type=null`
  - `losses=[1.156882, 4.137518, 3.050902]`
  - `grad_norms=[2.590222, 5.900850786987662e+17, 9147114459174.867]`

Sanity check for active-device memory helpers:
- imported patched module from installed wheel path and verified:
  - `_get_active_accelerator_index` present
  - `_get_accelerator_total_memory_bytes` / `_get_effective_accelerator_memory_bytes` return values

## Notes
- High-memory accelerators keep warmup by default (`auto`).
- In this environment, the uninitialized-weights load exception did not reproduce on either `2025.12` or `2026.2.1`; both loaded successfully.

### DIFF
--- a/unsloth_zoo/compiler.py
+++ b/unsloth_zoo/compiler.py
@@ -1154,6 +1154,7 @@ def create_standalone_class(
                     if (
                         "use_experts_implementation" in stripped
                         or "use_kernel_forward_from_hub" in stripped
+                        or "use_kernelized_func" in stripped
                         or stripped.startswith("@auto_docstring")
                     ):
                         decorator_name = stripped.split("(")[0].lstrip("@")
@@ -1274,6 +1275,7 @@ def create_standalone_class(
 
     # Remove @auto_docstring
     source = re.sub(r"@auto_docstring[\s]{0,}(\([^\)]{0,}\))?", "", source)
+    source = re.sub(r"@use_kernelized_func[\s]{0,}(\([^\)]{0,}\))?", "", source)
     source = re.sub(r"@check_model_inputs[\s]{0,}(\([^\)]{0,}\))?", "", source)
     # source = source.replace("@auto_docstring", "")
 
--- a/unsloth_zoo/compiler.py
+++ b/unsloth_zoo/compiler.py
@@ -1151,8 +1151,12 @@ def create_standalone_class(
             for line in lines:
                 stripped = line.strip()
                 if stripped.startswith("@"):
-                    if "use_experts_implementation" in stripped:
-                        logger.info(f'Unsloth: stripped use_experts_implementation decorator from {module}')
+                    if (
+                        "use_experts_implementation" in stripped
+                        or "use_kernel_forward_from_hub" in stripped
+                    ):
+                        decorator_name = stripped.split("(")[0].lstrip("@")
+                        logger.info(f"Unsloth: stripped {decorator_name} decorator from {module}")
                         continue # Strip it
                     else:
                         logger.warning(f"Unsloth: Warning: Unknown decorator {stripped} found for {module}.")
--- a/unsloth_zoo/compiler.py
+++ b/unsloth_zoo/compiler.py
@@ -1154,6 +1154,7 @@ def create_standalone_class(
                     if (
                         "use_experts_implementation" in stripped
                         or "use_kernel_forward_from_hub" in stripped
+                        or stripped.startswith("@auto_docstring")
                     ):
                         decorator_name = stripped.split("(")[0].lstrip("@")
                         logger.info(f"Unsloth: stripped {decorator_name} decorator from {module}")
--- a/unsloth_zoo/temporary_patches/gpt_oss.py
+++ b/unsloth_zoo/temporary_patches/gpt_oss.py
@@ -1394,15 +1394,19 @@ def _should_use_gpt_oss_bnb4bit() -> bool:
     Default: True when load_in_4bit is active.
     Set UNSLOTH_GPT_OSS_BNB4BIT_DISABLE=1 to force BF16 path.
     """
-    if "gpt_oss" not in os.environ.get("UNSLOTH_MODEL_NAME", ""):
+    if "gpt_oss" not in _normalized_unsloth_model_name():
         return False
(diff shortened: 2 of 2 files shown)


danielhanchen:
Additional follow-up on #521 before final matrix runs:

- Added compiler decorator stripping for `@use_kernel_forward_from_hub(...)` and `@auto_docstring` in class rewrite path.
- This removes the previously observed unknown-decorator warning noise for GPT-OSS/Llama compile paths.
- Kept unknown-decorator warnings for truly unrecognized decorators.

Commit included on this branch: `eaf1f4a` (with prior `684f684`).

unsloth_zoo/temporary_patches/gpt_oss.py
```diff
@@ -1094,13 +1094,105 @@ def patch_gpt_oss_bnb4bit_auto():
...
+                return int(torch.xpu.current_device())
+            return 0
+        if hasattr(torch, "cuda") and hasattr(torch.cuda, "current_device"):
+            return int(torch.cuda.current_device())
+    except Exception:
+        pass
+    return 0
+
+
+def _get_accelerator_total_memory_bytes():
```
Datta0, line 1113: NIT: We should put these in utils.py (away from temporary_patches) as these have potential to be reused.
Also there was a PR about DeviceContext iirc which might better handle this
Here it is https://github.com/unslothai/unsloth/pull/3875

unsloth_zoo/temporary_patches/gpt_oss.py
```diff
@@ -1094,13 +1094,105 @@ def patch_gpt_oss_bnb4bit_auto():
...
+
+
+def _get_effective_accelerator_memory_bytes():
+    total_memory = _get_accelerator_total_memory_bytes()
+    if total_memory is None:
+        return None
+    if DEVICE_TYPE != "xpu" and hasattr(torch.cuda, "get_per_process_memory_fraction"):
+        try:
+            device_index = _get_active_accelerator_index()
+            fraction = float(torch.cuda.get_per_process_memory_fraction(device_index))
```
Datta0, line 1130: Interesting. Are we restricting this somewhere else or is this to respect some hypothetical user setting such a limit?

unsloth_zoo/temporary_patches/gpt_oss.py
```diff
@@ -1316,15 +1408,19 @@ def _should_use_gpt_oss_bnb4bit() -> bool:
     Default: True when load_in_4bit is active.
     Set UNSLOTH_GPT_OSS_BNB4BIT_DISABLE=1 to force BF16 path.
     """
-    if "gpt_oss" not in os.environ.get("UNSLOTH_MODEL_NAME", ""):
+    if "gpt_oss" not in _normalized_unsloth_model_name():
```
Datta0, line 1411: I see this being a repeat of #519  (some of it). We should try to consolidate to avoid merge conflicts

unsloth_zoo/compiler.py
```diff
@@ -1151,8 +1151,14 @@ def create_standalone_class(
             for line in lines:
                 stripped = line.strip()
                 if stripped.startswith("@"):
-                    if "use_experts_implementation" in stripped:
-                        logger.info(f'Unsloth: stripped use_experts_implementation decorator from {module}')
+                    if (
+                        "use_experts_implementation" in stripped
+                        or "use_kernel_forward_from_hub" in stripped
```
Datta0, line 1156: Actually nice. Never noticed this. Seems like a very recent change.
But this function is to strip the decorator. Why not have the decorator and use the kernel?

PR #529, opened 2026-03-03
Moe kernels refactor

Datta0:
Tried to preserve as much authorship as possible with the help from codex. 

Ref: https://github.com/unslothai/unsloth/pull/4145/changes

### DIFF
--- a/unsloth_zoo/kernels/__init__.py
+++ b/unsloth_zoo/kernels/__init__.py
@@ -1,0 +1,1 @@
+# Unsloth Zoo kernel package namespace.
--- a/unsloth_zoo/kernels/moe/LICENSE
+++ b/unsloth_zoo/kernels/moe/LICENSE
@@ -1,0 +1,661 @@
+                    GNU AFFERO GENERAL PUBLIC LICENSE
+                       Version 3, 19 November 2007
+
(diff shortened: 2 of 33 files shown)


danielhanchen:
The direction is right since unsloth_zoo/temporary_patches/moe_utils.py currently reaches into unsloth.kernels.moe, but this also drags in the benchmark and test trees, whose top-level imports like "from grouped_gemm.kernels.tuning import ..." will not resolve inside the package. Could you narrow it to the kernel and interface modules with package-relative imports?

danielhanchen:
The direction of owning the MoE kernels in zoo is right, but this lands a second full copy of unsloth/kernels/moe including benchmarks and tests while the unsloth tree keeps its own, and the branch has drifted from the current moe_utils.py. Could you rebase and split out the kernel move from the test and benchmark additions, and say how the unsloth side gets retired?

danielhanchen:
I checked that this copies the grouped GEMM MoE kernels into unsloth_zoo/kernels/moe and repoints the imports in temporary_patches/moe_utils.py, while unsloth/kernels/moe still exists on main. Could you rebase and confirm the landing order against unsloth#4145 so the two copies do not diverge, and say whether the benchmark and test trees need to come along in the same change?

PR #534, opened 2026-03-06
Double-buffer GPU activations for overlapping H2D copy with backward compute

ruixiang63:
This PR implements a double-buffering optimization for activation reloading during the backward pass of gradient checkpointing with `use_gradient_checkpointing="unsloth"`. This means any model fine-tuned with Unsloth's gradient checkpointing will benefit from this PR.

Problem: With the current single-buffer implementation, the H2D memory copy for reloading offloaded activations during backward cannot overlap with GPU computation. The copy stream must wait for the compute stream to finish using the shared GPU buffer before writing the next activation into it, creating GPU idle bubbles.

Solution: Allocate a second GPU buffer (`GPU_BUFFERS_B`) and alternate between the two buffers based on layer index (CPU_INDEX % 2) with cuda event to prevent race condition. While the compute stream operates on buffer A, the copy stream can simultaneously load the next activation into buffer B, and vice versa — fully overlapping H2D transfers with backward computation. 

e.g. 
```text
Layer 23 backward:
  1. H2D copy into buffer A (EXTRA_STREAM)
  2. MAIN_STREAM.wait_stream(EXTRA_STREAM)  // wait for copy
  3. Compute on buffer A                   
Layer 22 backward:
  1. H2D copy into buffer B (EXTRA_STREAM)  // overlaps with step 3 above
  2. MAIN_STREAM.wait_stream(EXTRA_STREAM)
  3. Compute on buffer B                    
Layer 21 backward:
  1. H2D copy into buffer A (EXTRA_STREAM)  // Layer 23's compute already done
  2. MAIN_STREAM.wait_stream(EXTRA_STREAM)
  3. Compute on buffer A
```

Here are the nsys profiling traces:
* Without this PR
image

* With this PR
image

### Performance: 
This method doesn’t take much memory as it only takes a buffer size in VRAM, e.g. in gpt-oss-20B model fine-tuning, one buffer takes around 47.18MB with batch size 8. 
This PR also implemented a smart double buffering enabling strategy, i.e. if there are enough VRAM we will enable double-buffering otherwise disable. 
In my experiments, this achieves approximately 10% speedup per backward pass during GPT-OSS-20B fine-tuning, though results may vary across different models, hyperparameters (e.g. seq_length, batch size etc.) and hardware configurations.

### DIFF
--- a/unsloth_zoo/gradient_checkpointing.py
+++ b/unsloth_zoo/gradient_checkpointing.py
@@ -360,16 +360,16 @@ def initialize_unsloth_gradient_checkpointing(dtype = None):
     # Allocate buffers to how many GPUs
     n_gpus = torch.cuda.device_count() if DEVICE_TYPE in ("cuda", "hip") else torch.xpu.device_count()
     try:
-        GPU_BUFFERS = tuple([torch.empty(2*256*2048, dtype = dtype, device = f"{DEVICE_TYPE_TORCH}:{i}") for i in range(n_gpus)])
+        GPU_BUFFERS = tuple([torch.empty(INITIAL_GPU_BUFFER_SIZE, dtype = dtype, device = f"{DEVICE_TYPE_TORCH}:{i}") for i in range(n_gpus)])
         # Double buffering: try to allocate buffer B
         try:
-            GPU_BUFFERS_B = tuple([torch.empty(2*256*2048, dtype = dtype, device = f"{DEVICE_TYPE_TORCH}:{i}") for i in range(n_gpus)])
+            GPU_BUFFERS_B = tuple([torch.empty(INITIAL_GPU_BUFFER_SIZE, dtype = dtype, device = f"{DEVICE_TYPE_TORCH}:{i}") for i in range(n_gpus)])
             USE_DOUBLE_BUFFER = True
             # Per-buffer events to prevent race conditions in double buffering.
             # Each event tracks when compute on that buffer finishes
             BUFFER_EVENTS_A = tuple([torch.cuda.Event() for _ in range(n_gpus)])
             BUFFER_EVENTS_B = tuple([torch.cuda.Event() for _ in range(n_gpus)])
-        except:
+        except Exception as e:
             GPU_BUFFERS_B = None
             USE_DOUBLE_BUFFER = False
             BUFFER_EVENTS_A = None
@@ -494,7 +494,7 @@ class UnslothCheckpointFunction(torch.autograd.Function):
                                 try:
                                     GPU_BUFFER_B.resize_(new_size)
                                     print("Unsloth: Double buffering enabled (parallel H2D + compute) for backward pass.")
-                                except:
+                                except Exception as e:
                                     # OOM - disable double buffering
                                     USE_DOUBLE_BUFFER = False
                         x = x[:new_size].view(shape)
--- a/unsloth_zoo/gradient_checkpointing.py
+++ b/unsloth_zoo/gradient_checkpointing.py
@@ -312,6 +312,7 @@ global FIRST_PASS
 global CURRENT_GC_INDEX
 global BUFFER_EVENTS_A
 global BUFFER_EVENTS_B
+global NEXT_BUFFER_SLOT
 
 if DEVICE_TYPE in ("cuda", "hip"):
     torch_gpu_stream = torch.cuda.stream
@@ -338,6 +339,7 @@ def initialize_unsloth_gradient_checkpointing(dtype = None):
     global CURRENT_GC_INDEX
     global BUFFER_EVENTS_A
     global BUFFER_EVENTS_B
+    global NEXT_BUFFER_SLOT
     CPU_BUFFERS = []
     CPU_INDEX = 0
 
@@ -359,6 +361,7 @@ def initialize_unsloth_gradient_checkpointing(dtype = None):
 
     # Allocate buffers to how many GPUs
     n_gpus = torch.cuda.device_count() if DEVICE_TYPE in ("cuda", "hip") else torch.xpu.device_count()
+    NEXT_BUFFER_SLOT = [0] * n_gpus
     try:
         GPU_BUFFERS = tuple([torch.empty(INITIAL_GPU_BUFFER_SIZE, dtype = dtype, device = f"{DEVICE_TYPE_TORCH}:{i}") for i in range(n_gpus)])
         # Double buffering: try to allocate buffer B
@@ -367,9 +370,17 @@ def initialize_unsloth_gradient_checkpointing(dtype = None):
             USE_DOUBLE_BUFFER = True
             # Per-buffer events to prevent race conditions in double buffering.
             # Each event tracks when compute on that buffer finishes
-            BUFFER_EVENTS_A = tuple([torch.cuda.Event() for _ in range(n_gpus)])
-            BUFFER_EVENTS_B = tuple([torch.cuda.Event() for _ in range(n_gpus)])
-        except Exception as e:
+            if DEVICE_TYPE in ("cuda", "hip"):
+                event_ctor = torch.cuda.Event
+            elif DEVICE_TYPE == "xpu":
+                event_ctor = torch.xpu.Event
+            else:
+                raise RuntimeError(f"Double buffering unsupported on {DEVICE_TYPE}")
+            BUFFER_EVENTS_A = tuple([event_ctor() for _ in range(n_gpus)])
+            BUFFER_EVENTS_B = tuple([event_ctor() for _ in range(n_gpus)])
+        except RuntimeError as e:
+            if "out of memory" not in str(e).lower():
+                raise
             GPU_BUFFERS_B = None
             USE_DOUBLE_BUFFER = False
             BUFFER_EVENTS_A = None
@@ -493,8 +504,10 @@ class UnslothCheckpointFunction(torch.autograd.Function):
                             if new_size > GPU_BUFFER_B.numel():
                                 try:
                                     GPU_BUFFER_B.resize_(new_size)
-                                    print("Unsloth: Double buffering enabled (parallel H2D + compute) for backward pass.")
-                                except Exception as e:
+                                    # print("Unsloth: Double buffering enabled (parallel H2D + compute) for backward pass.")
+                                except RuntimeError as e:
+                                    if "out of memory" not in str(e).lower():
+                                        raise
                                     # OOM - disable double buffering
                                     USE_DOUBLE_BUFFER = False
                         x = x[:new_size].view(shape)
@@ -504,7 +517,10 @@ class UnslothCheckpointFunction(torch.autograd.Function):
                         with torch_gpu_stream(EXTRA_STREAM):
                             x.copy_(arg, non_blocking = True)
 
-                        ctx._saved_metadata = (new_size, shape, CPU_INDEX, device_index, MAIN_STREAM, EXTRA_STREAM,)
+                        global NEXT_BUFFER_SLOT
+                        buffer_slot = NEXT_BUFFER_SLOT[device_index]
+                        NEXT_BUFFER_SLOT[device_index] ^= 1
+                        ctx._saved_metadata = (new_size, shape, CPU_INDEX, device_index, MAIN_STREAM, EXTRA_STREAM, buffer_slot,)
                         CPU_INDEX += 1
                         tensor_inputs.append(None)
 
@@ -513,7 +529,7 @@ class UnslothCheckpointFunction(torch.autograd.Function):
                             print("Unsloth: Will smartly offload gradients to save VRAM!")
                             USE_UNSLOTH_GC = False
                     else:
-                        ctx._saved_metadata = (None, None, None, None, None, None,)
+                        ctx._saved_metadata = (None, None, None, None, None, None, None,)
                         tensor_inputs.append(arg)
                     pass
                 else:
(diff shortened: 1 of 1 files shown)


unsloth_zoo/gradient_checkpointing.py
```diff
@@ -523,11 +546,25 @@ def backward(ctx, *args):
...
+                buffer = GPU_BUFFERS[device_index][:new_size].view(shape)
+
             x = CPU_BUFFERS[CPU_INDEX][:new_size].view(shape)
 
             # See https://pytorch.org/docs/stable/notes/cuda.html#cuda-streams
-            EXTRA_STREAM.wait_stream(MAIN_STREAM)
+            if USE_DOUBLE_BUFFER:
+                # Double buffer mode: No need to wait for MAIN_STREAM because
+                # we're using a different buffer than the one being computed on
+                pass
```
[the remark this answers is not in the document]
ruixiang63: Resolved with cuda event.

unsloth_zoo/gradient_checkpointing.py
```diff
@@ -523,11 +546,25 @@ def backward(ctx, *args):
...
+            else:
+                # Single buffer mode: Must wait for MAIN_STREAM to finish
+                EXTRA_STREAM.wait_stream(MAIN_STREAM)
```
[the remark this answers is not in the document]
ruixiang63: Resolved with cuda event.

mmathew23:
yes thanks for flagging. I guess we can keep it for safety :)

unsloth_zoo/gradient_checkpointing.py
```diff
@@ -352,7 +360,20 @@ def initialize_unsloth_gradient_checkpointing(dtype = None):
...
-        GPU_BUFFERS = tuple([torch.empty(2*256*2048, dtype = dtype, device = f"{DEVICE_TYPE_TORCH}:{i}") for i in range(n_gpus)])
+        GPU_BUFFERS = tuple([torch.empty(INITIAL_GPU_BUFFER_SIZE, dtype = dtype, device = f"{DEVICE_TYPE_TORCH}:{i}") for i in range(n_gpus)])
+        # Double buffering: try to allocate buffer B
+        try:
+            GPU_BUFFERS_B = tuple([torch.empty(INITIAL_GPU_BUFFER_SIZE, dtype = dtype, device = f"{DEVICE_TYPE_TORCH}:{i}") for i in range(n_gpus)])
+            USE_DOUBLE_BUFFER = True
+            # Per-buffer events to prevent race conditions in double buffering.
+            # Each event tracks when compute on that buffer finishes
+            BUFFER_EVENTS_A = tuple([torch.cuda.Event() for _ in range(n_gpus)])
+            BUFFER_EVENTS_B = tuple([torch.cuda.Event() for _ in range(n_gpus)])
```
danielhanchen: **[6/8 reviewers] Hard-coded `torch.cuda.Event()` on backend-generic path.**

This file supports `DEVICE_TYPE == "xpu"` for streams and buffers, but the event allocation is CUDA-only. On XPU, `torch.cuda.Event()` will raise, be silently caught by the broad `except`, and double buffering will be silently disabled for all XPU users without any diagnostic.

Suggested approach -- use a backend-aware event constructor:

```python
if DEVICE_TYPE in ("cuda", "hip"):
    event_ctor = torch.cuda.Event
elif DEVICE_TYPE == "xpu":
    event_ctor = torch.xpu.Event
else:
    event_ctor = None

if event_ctor is None:
    raise RuntimeError(f"Double buffering unsupported on {DEVICE_TYPE}")
BUFFER_EVENTS_A = tuple([event_ctor() for _ in range(n_gpus)])
BUFFER_EVENTS_B = tuple([event_ctor() for _ in range(n_gpus)])
```

unsloth_zoo/gradient_checkpointing.py
```diff
@@ -523,11 +563,29 @@ def backward(ctx, *args):
         new_size, shape, CPU_INDEX, device_index, MAIN_STREAM, EXTRA_STREAM = ctx._saved_metadata
         if CPU_INDEX is not None:
             global GPU_BUFFER
-            buffer = GPU_BUFFERS[device_index][:new_size].view(shape)
+            global USE_DOUBLE_BUFFER
+            global GPU_BUFFERS_B
+            global BUFFER_EVENTS_A
```
danielhanchen, line 621: **[6/8 reviewers] Buffer selection is keyed off global `CPU_INDEX` parity, not per-device state.**

On multi-GPU / `device_map` workloads, layers are interleaved across devices. One device may always see even `CPU_INDEX` values (always buffer A), another always odd (always buffer B), collapsing back to single-buffer behavior per device and losing the intended overlap.

Suggested fix: introduce a per-device `NEXT_BUFFER_SLOT` list. In forward, capture `buffer_slot = NEXT_BUFFER_SLOT[device_index]` and toggle it (`^= 1`). Save `buffer_slot` in `ctx._saved_metadata`. In backward, use the saved `buffer_slot` to select the buffer and the matching event, instead of `CPU_INDEX % 2`.

```python
# initialize
NEXT_BUFFER_SLOT = [0] * n_gpus

# forward: right before saving metadata
buffer_slot = NEXT_BUFFER_SLOT[device_index]
NEXT_BUFFER_SLOT[device_index] ^= 1
ctx._saved_metadata = (..., buffer_slot)

# backward
..., buffer_slot = ctx._saved_metadata
if USE_DOUBLE_BUFFER and buffer_slot == 1:
    buffer = GPU_BUFFERS_B[device_index][:new_size].view(shape)
else:
    buffer = GPU_BUFFERS[device_index][:new_size].view(shape)
```

unsloth_zoo/gradient_checkpointing.py
```diff
@@ -464,6 +489,21 @@ def forward(ctx, run_function, preserve_rng_state, *args):
                         shape = arg.shape
                         if new_size > x.numel(): x.resize_(new_size)
                         if new_size > GPU_BUFFER.numel(): GPU_BUFFER.resize_(new_size)
+                        # resize buffer B as needed if double buffering is enabled, disable if OOM
+                        if USE_DOUBLE_BUFFER:
+                            GPU_BUFFER_B = GPU_BUFFERS_B[device_index]
+                            if new_size > GPU_BUFFER_B.numel():
+                                try:
+                                    GPU_BUFFER_B.resize_(new_size)
+                                except Exception as e:
```
danielhanchen: **[4/8 reviewers] Broad `except Exception` swallows non-OOM errors.**

A programming error (wrong type, negative size, sticky CUDA error from prior illegal memory access) would be silently caught here, disabling double-buffering without any diagnostic. Should catch only OOM:

```python
except RuntimeError as e:
    if "out of memory" not in str(e).lower():
        raise
    # OOM - disable double buffering and free buffer B
    USE_DOUBLE_BUFFER = False
    ...
```

unsloth_zoo/gradient_checkpointing.py
```diff
@@ -352,7 +360,20 @@ def initialize_unsloth_gradient_checkpointing(dtype = None):
...
+        # Double buffering: try to allocate buffer B
+        try:
+            GPU_BUFFERS_B = tuple([torch.empty(INITIAL_GPU_BUFFER_SIZE, dtype = dtype, device = f"{DEVICE_TYPE_TORCH}:{i}") for i in range(n_gpus)])
+            USE_DOUBLE_BUFFER = True
+            # Per-buffer events to prevent race conditions in double buffering.
+            # Each event tracks when compute on that buffer finishes
+            BUFFER_EVENTS_A = tuple([torch.cuda.Event() for _ in range(n_gpus)])
+            BUFFER_EVENTS_B = tuple([torch.cuda.Event() for _ in range(n_gpus)])
+        except Exception as e:
+            GPU_BUFFERS_B = None
```
danielhanchen, line 370: **[4/8 reviewers] Same broad `except Exception` issue in initialization.**

If `torch.cuda.Event()` fails for a non-OOM reason (e.g., CUDA not available on an XPU system), the real error is swallowed. Consider narrowing to `except RuntimeError`.

unsloth_zoo/gradient_checkpointing.py
```diff
@@ -890,13 +971,24 @@ def reset_unsloth_gradient_checkpointing_buffers():
...
+    pass
+
     # Reset state variables for fresh training run
     CPU_INDEX = 0
     BACKWARD_PASS = True
     LAST_GC_INDEX = 0
     FIRST_PASS = True
     CURRENT_GC_INDEX = 0
     USE_UNSLOTH_GC = True  # Re-enable the "Will smartly offload" message
+    # Re-enable double buffering if buffer B still exists (may have been
```
danielhanchen: **[2/8 reviewers] After OOM, double buffering is permanently disabled with no recovery path.**

When OOM triggers during forward, `GPU_BUFFERS_B = None`. This reset function correctly skips re-enabling (`GPU_BUFFERS_B is not None` is False). But there is no mechanism to re-allocate buffer B for subsequent training runs -- even after the memory pressure that caused the OOM is gone.

Consider re-attempting allocation in the `else` branch:

```python
if GPU_BUFFERS_B is not None:
    USE_DOUBLE_BUFFER = True
else:
    # Try to re-allocate buffer B for fresh training runs
    try:
        GPU_BUFFERS_B = tuple([torch.empty(INITIAL_GPU_BUFFER_SIZE, ...) for i in range(n_gpus)])
        BUFFER_EVENTS_A = tuple([event_ctor() for _ in range(n_gpus)])
        BUFFER_EVENTS_B = tuple([event_ctor() for _ in range(n_gpus)])
        USE_DOUBLE_BUFFER = True
    except Exception:
        pass  # Stay in single-buffer mode
```

danielhanchen (changes requested):
Thank you for the PR! The goal of this PR is to add double-buffering to Unsloth's gradient checkpointing activation reloading, alternating between two GPU buffers (A and B) during the backward pass to overlap H2D copies with gradient computation. As a summary, this PR allocates a second GPU buffer (B), per-buffer CUDA events, and uses CPU_INDEX parity to select which buffer to copy into during backward, while adding OOM fallback, cleanup, and reset logic for the new state.

**Benchmark results** (Llama-3.2-1B-Instruct, 4-bit LoRA, 61 steps, B200 GPU):

| Method | Tokens/s | Peak Memory | Train Loss |
|--------|----------|-------------|------------|
| Baseline (main) | 22,410 | 1.528 GB | 1.3790 |
| PR (double-buf ON) | 24,527 | 1.532 GB | 1.3790 |
| PR (double-buf OFF) | 24,436 | 1.530 GB | 1.3790 |

Double-buffer ON vs OFF: +0.4% (within noise). Losses and grad-norms are numerically equivalent across all modes (max delta 0.0035). The double-buffering mechanism is numerically correct and the core idea is sound. The OOM fallback and cleanup paths are safe.

| 6/8 | High | Buffer slot selection uses global `CPU_INDEX` parity, not per-device state -- breaks multi-GPU/device_map |
| 6/8 | High | `torch.cuda.Event()` hard-coded; breaks/silently disables on XPU devices |
| 4/8 | Medium | Broad `except Exception` on buffer-B resize swallows non-OOM errors |
| 2/8 | Low | OOM permanently disables double buffering with no recovery path |

Concrete suggestions for each finding below.

Datta0:
I'm just thinking of an extreme case workload, say on a 16GB machine, our old setup takes ~15 ish GB. We do proactively allocate the second buffer during forward pass, but during backward pass (which consumes more memory than forward I presume), the process OOMs. 
I know this is very rare especially given the small small overhead we're adding. But ig we're much safer if we just restrict this only to cases where after original model load, we have a certain amount of memory free (relative or absolute)?

One other thing I couldn't benchmark is, how adding a second buffer effects forward/backward performance in cases where memory pressure is already high?

ruixiang63:
> I'm just thinking of an extreme case workload, say on a 16GB machine, our old setup takes ~15 ish GB. We do proactively allocate the second buffer during forward pass, but during backward pass (which consumes more memory than forward I presume), the process OOMs.
I know this is very rare especially given the small small overhead we're adding. But ig we're much safer if we just restrict this only to cases where after original model load, we have a certain amount of memory free (relative or absolute)?

Make sense. I've added a few safety layers to handle this:

* Double buffering starts disabled. After the first training step completes (thanks to `FIRST_PASS` parameter), we check `torch.cuda.mem_get_info()` (this reflects post-backward peak headroom on GPU for the first training step), and only enable double buffering if *CUDA free memory > 512MB* (can be changed) for the following training steps. 

* OOM guards on buffer resize: If buffer A or B resize OOMs during forward, buffer B is freed and we fall back to single buffer.

* Environment variable: For extreme cases we can set `UNSLOTH_DISABLE_DOUBLE_BUFFER=1` to skip buffer B allocation entirely, which is more reliable than any heuristic when memory is that tight.


> One other thing I couldn't benchmark is, how adding a second buffer effects forward/backward performance in cases where memory pressure is already high?

Yeah, double buffering is now only enabled when CUDA free memory > 512MB after the first training step. Under high memory pressure it stays disabled, so there's zero performance impact. Single buffer is always the default, double buffering is only activated as a bonus when there's enough VRAM headroom.

Feel free to share any suggestions or concerns!

Datta0 (approved):
LGTM !

ruixiang63:
@danielhanchen @Datta0 Since this PR has already been approved, when can we expect it to be merged?

mmathew23:
@ruixiang63 I think we were asked to hold off on merging from the nvidia side. Unsure if that position has changed.

PR #548, opened 2026-03-16
[MoE] FP8 support for MoE, specifically GLM 4.7 flash

Datta0:
`grouped_mm` is not a supported operation on FP8, so we try to make use of `_scaled_grouped_mm`, but that is only supported on newer GPU architectures. If the operation is not supported, we de-quantize the tensors on the fly and pass it to the same `grouped_mm` function.
There seems to be some issue where transformers are not very keen on loading the weight scales for the FPI tensors, so we needed to patch it up. Currently, this is very specific to GLM; still not sure about Qwen3 MoE or other MOEs.

Note that this is only tested on B200 so far, and the losses seem to pretty closely match up with the BF16 training counterpart.  

To Test: 
1. Use [Qwen3 MoE notebook](https://colab.research.google.com/github/unslothai/notebooks/blob/main/nb/Qwen3_MoE.ipynb) but replace the model name with `Qwen/Qwen3-Coder-30B-A3B-Instruct-FP8`. The memory usage should be <40GB
2. Use [GLM 4.7 flash](https://colab.research.google.com/github/unslothai/notebooks/blob/main/nb/GLM_Flash_A100(80GB).ipynb) notebook but replace the model name with `unsloth/GLM-4.7-Flash-FP8-Dynamic`. Same as above 
3. Sanity: Ensure the BF16 variants (unchanged notebooks) work too
4. Ensure that [Qwen3 8B dense fp8 grpo](https://colab.research.google.com/github/unslothai/notebooks/blob/main/nb/Qwen3_8B_FP8_GRPO.ipynb#scrollTo=DkIvEkIIkEyB) works

### DIFF
--- a/unsloth_zoo/patching_utils.py
+++ b/unsloth_zoo/patching_utils.py
@@ -30,6 +30,7 @@ __all__ = [
 from .compiler import UNSLOTH_COMPILE_LOCATION
 from .utils import _get_dtype, Version
 from .hf_utils import dtype_from_config, set_dtype_in_config, HAS_TORCH_DTYPE
+from .temporary_patches.moe_utils_fp8 import maybe_patch_stacked_moe_expert_fp8_scales
 
 # Also disable compiling on bitsandbytes
 def patch_compiling_bitsandbytes():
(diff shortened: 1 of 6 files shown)


Datta0:
Closed for #659

PR #551, opened 2026-03-16
Fix FP8 MoE scale patching for compressed-tensors models

danielhanchen:
## Summary

Fixes the FP8 MoE scale patching in `moe_utils_fp8.py` which was silently failing for all compressed-tensors FP8 MoE models (e.g., GLM-4.7-Flash-FP8-Dynamic). Based on the cleanup branch from #548.

The scale patcher (`_maybe_patch_glm4_stacked_moe_fp8_scales`) was not firing due to two bugs:

1. **Enum vs string comparison**: `quantization_config.quant_method` returns a `QuantizationMethod` enum in newer transformers, not the string `"compressed-tensors"`. The comparison `quant_method != "compressed-tensors"` always returned True, causing an early return.

2. **Wrong layer probe**: The probe for scale keys in safetensors checked only the first routed layer. Some layers (e.g., layers 1, 39, 46 in GLM-4.7-Flash) are in the quantization ignore list and have no scales. If the first routed layer happened to be ignored, the probe returned False and the entire patcher bailed out.

Additionally, the patcher only supported single-file `model.safetensors` checkpoints. Sharded checkpoints (`model-00001-of-NNNNN.safetensors`) would crash with `EntryNotFoundError`.

## Changes

- Normalize `quant_method` to string before comparison (handles both enum and string)
- Scan all routed layers to find those with scale keys, skip layers without
- Read the full set of available keys from safetensors once instead of probing per-layer
- Add `_resolve_safetensors_files()` supporting both single-file and sharded layouts via `model.safetensors.index.json`
- Add `_open_safetensors_for_keys()` that opens only the relevant shard(s)

## Test results

Tested on NVIDIA RTX PRO 6000 Blackwell (SM 12.0, 98 GB):

- GLM-4.7-Flash-FP8-Dynamic: 43 FP8 expert layers get scales attached (previously 0)
- Forward pass produces valid output, no NaN/Inf
- 21-step LoRA training completes successfully
- Normal (non-FP8, non-MoE) models are unaffected -- Llama-3.2-1B 4-bit training passes

## Backwards compatibility

- Works with transformers 4.57.6 (string quant_method) and 5.3.0 (enum quant_method)
- Falls back gracefully if safetensors files are not found
- No changes to any other code paths

### DIFF
--- a/unsloth_zoo/patching_utils.py
+++ b/unsloth_zoo/patching_utils.py
@@ -30,6 +30,7 @@ __all__ = [
 from .compiler import UNSLOTH_COMPILE_LOCATION
 from .utils import _get_dtype, Version
 from .hf_utils import dtype_from_config, set_dtype_in_config, HAS_TORCH_DTYPE
+from .temporary_patches.moe_utils_fp8 import maybe_patch_stacked_moe_expert_fp8_scales
 
 # Also disable compiling on bitsandbytes
 def patch_compiling_bitsandbytes():
@@ -395,6 +396,8 @@ def patch_model_and_tokenizer(
     # patch to dict makes sure that any torch.dtype is converted to
     # string when trying to save the config or serialize it
     patch_to_dict()
+
+    maybe_patch_stacked_moe_expert_fp8_scales(model)
 
     # Check all params and patch!
     for name, module in model.named_modules():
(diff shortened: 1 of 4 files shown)


danielhanchen:
Closing as superseded.

FP8 MoE scale handling now lives in the load-time quant handlers plus `saving_utils` on main, so the `maybe_patch_stacked_moe_expert_fp8_scales` hook this PR adds no longer has an integration point.

The compressed-tensors GLM4 stacked-scale patching idea can be folded into a fresh PR against current main if it is still needed.

PR #554, opened 2026-03-17
Fix bugs in FP8 MoE support

danielhanchen:
## Summary

Fixes multiple bugs identified during review of the FP8 MoE PRs (unsloth #4312 + #548). These changes sit on top of the PR branch and address correctness, safety, and compatibility issues.

**Companion PR:** unslothai/unsloth#4348

### Fixes in `moe_utils_fp8.py`

- **B7**: Support sharded safetensors (multi-shard) in GLM4 FP8 scale patching. Pre-computes needed tensor keys, builds tensor-to-shard lookup, opens all needed shards via a multi-shard reader
- **B8**: Add `quant_kind` parameter to `_dequantize_expert_slice` and `_dequantize_full_expert_weights` so `weight_scale_inv` scales get `reciprocal()` applied before dequantization
- **B18**: Guard `fp8_linear` import with `try/except ImportError` and add dequant+F.linear fallback
- **B19**: Flatten 3D `top_k_index`/`top_k_weights` alongside `hidden_states` to prevent shape mismatch in expert mask
- **B23**: Fix `_make_grouped_mm_rhs_column_major` -- was a no-op double transpose (`transpose(-2,-1).contiguous().transpose(-2,-1)`), now correctly uses `weight.mT.contiguous()`
- **B25**: Add `act_fn` fallback to `F.silu` when module attribute is missing
- **B26**: Remove dead `_forward_native_moe_loop_fp8` function (never called)
- Fix `use_separated_lora` to call `_should_use_separated_lora()` instead of hardcoded `True` (respects `UNSLOTH_MOE_LORA_MERGED=1`)

### Fixes in `moe_utils.py`

- **B10**: Add 3D input reshape at start/end of `forward_native_moe_loop`
- **B12**: Prefer generic backend over unconditional FP8 in `get_forward_moe_backend`; use local `forward_moe_backend` (which auto-detects FP8) as final fallback
- Remove no-op `try/except RuntimeError as e: raise e`
- Remove duplicate comment

## Test plan

- [ ] Syntax check all modified files
- [ ] Verify imports work correctly
- [ ] Run FP8 MoE model (GLM4) to verify scale patching works with sharded checkpoints
- [ ] Run standard non-FP8 MoE model to verify no regression from B12 backend priority change
- [ ] Verify `UNSLOTH_MOE_LORA_MERGED=1` is respected in scaled_grouped_mm path

### DIFF
--- a/unsloth_zoo/patching_utils.py
+++ b/unsloth_zoo/patching_utils.py
@@ -30,6 +30,7 @@ __all__ = [
 from .compiler import UNSLOTH_COMPILE_LOCATION
 from .utils import _get_dtype, Version
 from .hf_utils import dtype_from_config, set_dtype_in_config, HAS_TORCH_DTYPE
+from .temporary_patches.moe_utils_fp8 import maybe_patch_stacked_moe_expert_fp8_scales
 
 # Also disable compiling on bitsandbytes
 def patch_compiling_bitsandbytes():
@@ -395,6 +396,8 @@ def patch_model_and_tokenizer(
     # patch to dict makes sure that any torch.dtype is converted to
     # string when trying to save the config or serialize it
     patch_to_dict()
+
+    maybe_patch_stacked_moe_expert_fp8_scales(model)
 
     # Check all params and patch!
     for name, module in model.named_modules():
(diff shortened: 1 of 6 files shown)


danielhanchen:
Closing as superseded.

The core FP8 MoE forward from this PR has since landed and evolved on main (`temporary_patches/moe_utils_fp8.py`), so the new-file add here no longer applies cleanly.

Two pieces here are still genuinely novel: the `qwen3_moe` router dtype cast and the GLM4 sharded-scale patching. If we want those, I will bring them in as a small targeted PR against current main, and drop the `misc.py` `create_block_mask` change, which regresses on torch builds whose `create_block_mask` does not accept `_compile`.

PR #562, opened 2026-03-24
Add GraniteMoeHybridForCausalLM compiler support

Maxusmusti:
## Summary

Adds compiler support for `GraniteMoeHybridForCausalLM` (IBM Granite 4 h-family models) and fixes the `cross_entropy_find_2` regex to handle multi-line `loss_function` call formatting.

### Changes

1. **Add `GraniteMoeHybridForCausalLM` to the forwards list** — the compiler now processes its `forward` method for fused cross-entropy and `UNSLOTH_RETURN_HIDDEN_STATES` support.

2. **Fix multi-line `loss_function` regex** — the `cross_entropy_find_2` template used literal `, ` (comma-space) between arguments, which only matched single-line formatting (e.g. Llama). GraniteMoeHybrid (and potentially other models) format the call across multiple lines:
   ```python
   # Single-line (Llama) - matched before
   loss = self.loss_function(logits=logits, labels=labels, vocab_size=self.config.vocab_size, **kwargs)

   # Multi-line (GraniteMoeHybrid) - now also matched
   loss = self.loss_function(
       logits,
       labels,
       vocab_size=self.config.vocab_size,
       **kwargs,
   )
   ```

3. **Fix `$KWARGS$` regex** — same issue: `, **kwargs` had a literal space that didn't match `\n    **kwargs`.

4. **Trailing comma tolerance** — added `,?` before the closing paren to handle the trailing comma in multi-line formatting.

### Testing

Verified that `apply_fused_lm_head` successfully patches all three model types:
- `LlamaForCausalLM` (single-line format) ✅
- `Qwen3ForCausalLM` (single-line format) ✅
- `GraniteMoeHybridForCausalLM` (multi-line format) ✅

Also ran end-to-end RL training (ART/GRPO with LoRA) on `granite-4.0-h-tiny` to confirm the patched forward works correctly with `UNSLOTH_RETURN_HIDDEN_STATES=1`.

### Related

Companion PR for loader.py approach: https://github.com/unslothai/unsloth/pull/4373

## Test plan
- [x] `apply_fused_lm_head` matches Llama, Qwen3, GraniteMoeHybrid
- [x] End-to-end training with GraniteMoeHybrid completes without shape mismatch errors

### DIFF
--- a/unsloth_zoo/compiler.py
+++ b/unsloth_zoo/compiler.py
@@ -1556,7 +1556,7 @@ $LOGITSCALINGMULTIPLY$
 $LOGITSCALINGDIVISION$
 $LOGITSOFTCAPPING$
 loss = None
-if labels is not None:$SPACES$loss = self.loss_function($NEWLINES$$LOGITS$, $LABELS$, $VOCABSIZE$$KWARGS$$NEWLINES$)
+if labels is not None:$SPACES$loss = self.loss_function($NEWLINES$$LOGITS$,$NEWLINES$$LABELS$,$NEWLINES$$VOCABSIZE$$KWARGS$,?$NEWLINES$)
 """
 
 cross_entropy_replacement_2 = """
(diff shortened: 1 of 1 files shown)


danielhanchen:
Confirmed the cross_entropy_find_2 template in unsloth_zoo/compiler.py still assumes a single-line self.loss_function call, so the multi-line Granite form never matches, and the widened pattern still covers the existing comma-space form. Will get this reviewed.

danielhanchen:
Confirmed the template at compiler.py:2631 still only matches single-line loss_function calls, so the multi-line fix is still needed; main has since widened the kwargs capture to any **name, so could you rebase and keep that broader group before I review?

danielhanchen:
Confirmed the single-line only patterns are still there in unsloth_zoo/compiler.py (cross_entropy_find_2 and the KWARGS replacement), so the multi-line loss_function call is not matched today. The looser regex now applies to every compiled causal LM, so could you confirm the match set is unchanged for Llama, Qwen3 and Gemma, and rebase onto main since the template moved?

PR #579, opened 2026-04-07
Add last_response_only parameter to train_on_responses_only

maximedb:
# Add `last_response_only` parameter to `train_on_responses_only`

> **Note: This change was generated by an AI assistant (Claude). The repo owner should review it carefully before merging.**

---

## What & Why

`train_on_responses_only` currently unmasks the labels for **every** assistant turn in a multi-turn conversation. This is useful for general SFT, but there are common fine-tuning scenarios — e.g. training a model to follow a final instruction given a conversation history — where you only want to supervise the **last** assistant response and leave all earlier ones masked.

## Change

A single new parameter `last_response_only=False` is added to `train_on_responses_only`. The default is `False`, so existing behaviour is completely unchanged.

```python
train_on_responses_only(
    trainer,
    instruction_part="<|im_start|>user\n",
    response_part="<|im_start|>assistant\n",
    last_response_only=True,  # only the final assistant turn is trained on
)
```

## Implementation

The inner `_train_on_responses_only` was refactored slightly: instead of immediately writing labels when each `(assistant_start, response_end)` span is found, the loop now **collects all spans first**, then applies labels from either all of them or just `spans[-1:]` depending on the flag. This avoids duplicating the parsing logic.

The refactor is a no-op when `last_response_only=False`.

### DIFF
--- a/unsloth_zoo/dataset_utils.py
+++ b/unsloth_zoo/dataset_utils.py
@@ -184,16 +184,20 @@ pass
 
 def train_on_responses_only(
     trainer,
-    instruction_part = None,
-    response_part    = None,
-    force_match      = True,  # Match newlines as well!
-    tokenizer        = None,  # Optional
-    return_function  = False, # Useful for iterating over lists
-    num_proc         = None,
+    instruction_part  = None,
+    response_part     = None,
+    force_match       = True,  # Match newlines as well!
+    tokenizer         = None,  # Optional
+    return_function   = False, # Useful for iterating over lists
+    num_proc          = None,
+    last_response_only = False, # Train only on the last assistant turn
 ):
     """
     Trains only on responses and not on the instruction by masking out
     the labels with -100 for the instruction part.
+
+    If last_response_only=True, only the final assistant turn has its
+    labels unmasked; all earlier assistant turns remain masked at -100.
     """
     # All Unsloth Zoo code licensed under LGPLv3
     if tokenizer is None and trainer is not None:
@@ -249,13 +253,16 @@ def train_on_responses_only(
         for input_ids, old_labels in zip(input_ids_, labels_):
             n = len(input_ids)
             labels = [-100] * n
-            
+
             use_old_labels = False
             if old_labels is not None:
                 use_old_labels = True
                 assert(n == len(old_labels))
             n_minus_1 = n - 1
             j = 0
+
+            # Collect all (assistant_k, user_j) spans for this sample
+            spans = []
             while j < n:
                 # Find <assistant>
                 if (input_ids[j] == A_first) and \
@@ -308,13 +315,7 @@ def train_on_responses_only(
                                 k = n
                             pass
 
-                            if not use_old_labels:
-                                # Now copy input_ids to labels
-                                labels[assistant_k : user_j] = input_ids [assistant_k : user_j]
-                                # print(assistant_j, assistant_k, user_j, user_k)
-                            else:
-                                # Copy over from old labels!
-                                labels[assistant_k : user_j] = old_labels[assistant_k : user_j]
+                            spans.append((assistant_k, user_j))
                             break
                         pass
                         j += 1
@@ -322,6 +323,17 @@ def train_on_responses_only(
                 pass
                 j += 1
             pass
+
+            # Apply labels: only the last assistant turn when last_response_only=True
+            apply_spans = spans[-1:] if last_response_only else spans
+            for assistant_k, user_j in apply_spans:
+                if not use_old_labels:
+                    # Now copy input_ids to labels
+                    labels[assistant_k : user_j] = input_ids [assistant_k : user_j]
+                else:
+                    # Copy over from old labels!
+                    labels[assistant_k : user_j] = old_labels[assistant_k : user_j]
+
             all_labels.append(labels)
         pass
         return { "labels" : torch.tensor(all_labels, dtype = torch.int64) if use_tensors else all_labels }
--- a/unsloth_zoo/dataset_utils.py
+++ b/unsloth_zoo/dataset_utils.py
@@ -197,7 +197,9 @@ def train_on_responses_only(
     the labels with -100 for the instruction part.
 
     If last_response_only=True, only the final assistant turn has its
-    labels unmasked; all earlier assistant turns remain masked at -100.
+    labels unmasked; all earlier assistant turns remain masked at -100
+    (they are never written, so they keep the initialized -100 values
+    and are not copied from old_labels either).
     """
     # All Unsloth Zoo code licensed under LGPLv3
     if tokenizer is None and trainer is not None:
@@ -324,7 +326,9 @@ def train_on_responses_only(
                 j += 1
             pass
 
-            # Apply labels: only the last assistant turn when last_response_only=True
+            # Apply labels: only the last assistant turn when last_response_only=True.
+            # Note: spans[-1:] safely returns [] when spans is empty (no assistant turn
+            # was found), so a sample with no assistant turn stays fully masked at -100.
             apply_spans = spans[-1:] if last_response_only else spans
             for assistant_k, user_j in apply_spans:
                 if not use_old_labels:
--- a/unsloth_zoo/vision_utils.py
+++ b/unsloth_zoo/vision_utils.py
@@ -656,6 +656,7 @@ class UnslothVisionDataCollator:
         response_part    = None,
         force_match      = True, # Match newlines as well!
         num_proc         = None,
+        last_response_only = False, # Train only on the last assistant turn
         completion_only_loss = True,
         pad_to_multiple_of = None,
         resize_dimension = 0, # can be 0, 1, 'max' or 'min' (max resizes based on the max of height width, min the min size, 0 the first dim, etc)
@@ -733,12 +734,13 @@ class UnslothVisionDataCollator:
             assert(isinstance(instruction_part, str) and isinstance(response_part, str))
             self.train_on_responses_only = _train_on_responses_only(
                 None,
-                instruction_part = instruction_part,
-                response_part    = response_part,
-                force_match      = force_match,
-                tokenizer        = processor,
-                return_function  = True,
-                num_proc         = num_proc,
+                instruction_part   = instruction_part,
+                response_part      = response_part,
+                force_match        = force_match,
+                tokenizer          = processor,
+                return_function    = True,
+                num_proc           = num_proc,
+                last_response_only = last_response_only,
             )
         else:
             self.train_on_responses_only = None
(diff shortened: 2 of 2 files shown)


danielhanchen (commented):
Thank you for the PR! The goal of this PR is to enable training only on the final assistant turn in multi-turn conversations, which is a common SFT scenario when you want supervision limited to the last response given a long conversation history. As a summary, this PR adds a new `last_response_only=False` keyword argument to `train_on_responses_only` and refactors the inner masking loop to collect `(assistant_k, user_j)` spans first, then applies labels from either all spans or `spans[-1:]` depending on the flag.

We ran 11 independent reviewers (7 from reviewer.py, 1 from gemini-code-assist, and 3 Sonnet subagents). Consolidated findings:

| 3/11 | Medium | `last_response_only` is not forwarded to `UnslothVisionDataCollator` in `unsloth_zoo/vision_utils.py`, so the VLM response-only masking path cannot use the new behavior |
| 3/11 | Low | `spans[-1:]` on empty `spans` silently returns `[]` (correct, but should be documented) |
| 2/11 | Low | Docstring could be clearer about the `old_labels` interaction when `last_response_only=True` |
| 2/11 | Low | Parameter alignment style is slightly off (`last_response_only` is longer than the other parameter names) |
| 1/11 | Low | The default path now allocates a `spans` list per sample instead of writing labels inline (very minor; dataset.map is one-shot so overhead is not measurable) |
| 1/11 | Low | Adding the new kwarg before `num_proc` would be safer against any callers passing `num_proc` positionally (unlikely) |
| 7+/11 | Info | PR description mentions `tests/test_dataset_utils.py` but no test file is in the actual diff |

Strong consensus across all 11 reviewers: the default path (`last_response_only=False`) is behaviorally a no-op relative to `main` in every scenario tested (single-turn, multi-turn, tensor inputs, `old_labels` present, empty conversations). Multiple reviewers ran differential harnesses comparing pre-PR and post-PR outputs across hundreds of randomized samples and confirmed bit-exact equivalence.

The only finding worth fixing in this PR is the VLM integration (Medium, 3/11). I am going to push a small follow-up commit to this PR branch to:

1. Forward `last_response_only` to `UnslothVisionDataCollator.__init__` in `unsloth_zoo/vision_utils.py`
2. Add an inline comment explaining that `spans[-1:]` is intentionally safe on empty `spans`

See inline comments below for the concrete suggested changes.

unsloth_zoo/dataset_utils.py
```diff
@@ -184,16 +184,20 @@ def _find_common_token_ids(component, tokenizer, force_match = False):
...
-    tokenizer        = None,  # Optional
-    return_function  = False, # Useful for iterating over lists
-    num_proc         = None,
+    instruction_part  = None,
+    response_part     = None,
+    force_match       = True,  # Match newlines as well!
+    tokenizer         = None,  # Optional
+    return_function   = False, # Useful for iterating over lists
+    num_proc          = None,
+    last_response_only = False, # Train only on the last assistant turn
```
danielhanchen, line 193: **[3/11 reviewers]** The new `last_response_only` parameter is not forwarded to `UnslothVisionDataCollator.__init__` in `unsloth_zoo/vision_utils.py`. `UnslothVisionDataCollator` currently calls `_train_on_responses_only(None, ..., return_function=True)` on line 734 without passing `last_response_only`, so VLM fine-tuning cannot access the new behavior. Since the function signature is unchanged (kwarg with default), this is a forward-compatible addition. Suggested forwarding in `vision_utils.py` (separate file, see that file for the actual edit):

```suggestion
    last_response_only = False, # Train only on the last assistant turn
```

No change needed on this line itself — this comment flags the VLM integration to also be added in `vision_utils.py` in the follow-up.

unsloth_zoo/dataset_utils.py
```diff
@@ -308,20 +315,25 @@ def _train_on_responses_only(examples):
...
                             break
                         pass
                         j += 1
                     pass
                 pass
                 j += 1
             pass
+
+            # Apply labels: only the last assistant turn when last_response_only=True
+            apply_spans = spans[-1:] if last_response_only else spans
```
danielhanchen, line 332: **[3/11 reviewers]** `spans[-1:]` on an empty list returns `[]` rather than raising `IndexError`, which is the correct behavior (sample with no assistant turn stays fully masked), but it is non-obvious. Adding a short inline comment clarifies intent:

```suggestion
            # Apply labels: only the last assistant turn when last_response_only=True.
            # Note: spans[-1:] safely returns [] when spans is empty (no assistant turn found),
            # so a sample with no assistant turn stays fully masked at -100.
            apply_spans = spans[-1:] if last_response_only else spans
```

unsloth_zoo/dataset_utils.py
```diff
@@ -184,16 +184,20 @@ def _find_common_token_ids(component, tokenizer, force_match = False):
...
+    return_function   = False, # Useful for iterating over lists
+    num_proc          = None,
+    last_response_only = False, # Train only on the last assistant turn
 ):
     """
     Trains only on responses and not on the instruction by masking out
     the labels with -100 for the instruction part.
+
+    If last_response_only=True, only the final assistant turn has its
+    labels unmasked; all earlier assistant turns remain masked at -100.
```
danielhanchen: **[2/11 reviewers]** Minor docstring nit — the phrase "remain masked at -100" is accurate but could be clearer about *why* they remain masked (they are never written; `labels` is initialized to `[-100] * n` on line 255). Also worth noting the `old_labels` interaction: when `last_response_only=True`, earlier assistant spans are not copied from `old_labels` either. Suggested:

```suggestion
    If last_response_only=True, only the final assistant turn has its
    labels unmasked; all earlier assistant turns remain masked at -100
    (they are never written, so they keep the initialized -100 values
    and are not copied from old_labels either).
```

unsloth_zoo/vision_utils.py
```diff
@@ -656,6 +656,7 @@ def __init__(
         response_part    = None,
         force_match      = True, # Match newlines as well!
         num_proc         = None,
+        last_response_only = False, # Train only on the last assistant turn
```
danielhanchen: **[7/11 reviewers]** Inserting `last_response_only` between `num_proc` and `completion_only_loss` shifts every subsequent parameter by one position. Any external caller passing `completion_only_loss`, `pad_to_multiple_of`, `resize_dimension`, or `snap_to_patch_size` positionally would silently get wrong behavior. Move it to the end of the parameter list instead:

```suggestion
        completion_only_loss = True,
        pad_to_multiple_of = None,
        resize_dimension = 0, # can be 0, 1, 'max' or 'min' (max resizes based on the max of height width, min the min size, 0 the first dim, etc)
        snap_to_patch_size = False,
        last_response_only = False, # Train only on the last assistant turn
```

danielhanchen (commented):
Round 2 review after the follow-up commit (`6f9f66c`) that forwarded `last_response_only` through `UnslothVisionDataCollator`. We ran the same review pipeline again: 7 reviewers from reviewer.py, 1 gemini-code-assist pass, and 3 fresh Sonnet subagents.

Consolidated findings:

| 7/11 | Medium | In `unsloth_zoo/vision_utils.py`, `last_response_only` was inserted between `num_proc` and `completion_only_loss` in `UnslothVisionDataCollator.__init__`. This shifts every subsequent parameter by one position and would silently break any external caller that passed `completion_only_loss`, `pad_to_multiple_of`, `resize_dimension`, or `snap_to_patch_size` positionally. The only in-repo caller (`studio/backend/core/training/trainer.py:3020`) passes just `(model, tokenizer)` positionally, so there is no immediate breakage, but external callers are at risk. |
| 1/11 | Low | In `UnslothVisionDataCollator.__init__`, `last_response_only=True` is silently ignored when `train_on_responses_only=False`. Adding a warning or assertion would improve the UX. Low priority since standard Python kwarg behavior. |
| 1/11 | Low | Docstring paragraph in `train_on_responses_only` could be trimmed — the parenthetical about `old_labels` over-explains an implementation detail. The public contract is simply "earlier turns stay masked." |
| 1/11 | Low | Minor alignment inconsistency between `dataset_utils.py` and `vision_utils.py` (cosmetic). |

Consensus on the rest of the change set:

- Default path (`last_response_only=False`) is a bit-exact no-op vs `main` — confirmed by multiple differential harnesses.
- `last_response_only=True` correctly masks all earlier assistant turns and leaves only the final one unmasked.
- Closure capture of `last_response_only` is correct.
- The `spans[-1:]` slice is safe on empty spans and the new inline comment documents that.
- `UnslothVisionDataCollator` forwarding threads the flag correctly and does not need `__slots__` updates (the value is consumed in `__init__` and not stored on `self`).
- No cross-version (transformers 4.57.6/5.4.0, TRL 0.22.2/0.27.1/1.0.0) compatibility concerns — pure Python label massaging.

I'm pushing a second follow-up commit that moves `last_response_only` to the end of `UnslothVisionDataCollator.__init__`'s parameter list to preserve the positional ABI. The other low-priority nits are not blocking.

danielhanchen (approved):
Round 3 review after the positional-ABI fix (`deac818`) that moved `last_response_only` to the end of `UnslothVisionDataCollator.__init__`'s parameter list. Same pipeline: 7 reviewers from reviewer.py, 1 gemini-code-assist pass, 3 fresh Sonnet subagents.

Consolidated result: **10/11 APPROVE, 1/11 COMMENT (cosmetic nits only).**

| 1/11 | Low (cosmetic) | Docstring paragraph in `train_on_responses_only` parenthetical about `old_labels` is implementation detail that could be trimmed. Non-blocking. |
| 1/11 | Low (cosmetic) | Minor `=` alignment inconsistency between `dataset_utils.py` (col 19) and `vision_utils.py` call site (col 22). Non-blocking. |
| 11/11 | Info | No full-repo test suite exists; no VLM end-to-end run was executed. Pre-existing testing gap, not a PR defect. |

Strong consensus on the current state:

- Positional ABI of `UnslothVisionDataCollator.__init__` is now preserved (the kwarg is the last entry, after all pre-existing params).
- No in-repo caller of `UnslothVisionDataCollator(` uses more than 2 positional args, so there is no risk of breakage.
- `last_response_only` is threaded correctly through to `_train_on_responses_only(..., last_response_only=last_response_only, ...)` at `vision_utils.py:743`.
- `dataset_utils.py` default path (`last_response_only=False`) remains bit-exact with `main` in all differential harnesses across hundreds of synthetic samples.
- `last_response_only=True` correctly masks earlier assistant turns and leaves only the final span unmasked. Verified with single-turn, multi-turn, tensor-backed `input_ids`, and `old_labels`-present cases.
- `spans[-1:]` safely returns `[]` on empty spans (no assistant turn) — documented inline.
- Closure capture of `last_response_only` is correct.
- No `__slots__` update needed (parameter is consumed in `__init__`, not stored on `self`).

No blocking findings remain. The PR is ready to merge.

PR #587, opened 2026-04-09
Fix inf grad_norm on Qwen3.5 at seq_len > 65536 with tighter SDPA guards

danielhanchen:
## Summary
- Fixes #4906: LoRA training of Qwen3.5-4B/9B produces `grad_norm=inf` at `seq_len > 65536` when flash-attn is not installed
- Wraps `sdpa_attention_forward` to detect materialized bool causal masks and replace with `attention_mask=None, is_causal=True`
- Includes tighter guards than #582 to protect sliding-window, bidirectional, and packed-sequence masks
- Adds bool-to-float fallback for non-pure-causal masks to avoid the Cutlass bug while preserving mask semantics

## Root cause
`patch_transformers_masks` wraps `create_causal_mask` with `torch.compile(dynamic=True)`. Under tracing, `find_packed_sequence_indices` takes the `is_tracing` branch, forcing `allow_is_causal_skip=False`. This materializes a dense `[1,1,Q,K]` bool causal mask. PyTorch SDPA's Cutlass backend with bf16 + bool mask at `seq_len > 65536` produces wrong outputs/gradients (int16 sequence-index overflow at 2^16).

## Guards (all O(1), no CUDA ops)
1. `module.is_causal` check -- protects BERT/bidirectional encoders
2. `sliding_window` kwarg check -- protects Gemma2/Mistral/Qwen2/3
3. Dict-mask unwrap with safe fallback when `layer_type` not in dict
4. 4D bool + square + Q==K shape check -- not cross-attention or kv-cache decode
5. Upper-triangle spot-check `m[0,0,0,1]==False` -- distinguishes pure causal from packed-sequence masks
6. Bool-to-float additive bias fallback for non-pure-causal masks -- avoids Cutlass bug while preserving mask semantics

## Test results (NVIDIA B200, torch 2.9.1, transformers 5.5.1, no flash-attn)

**Qwen3.5-4B at seq_len=69632, 4-bit LoRA, 9 steps:**

| Method | Time | Peak Mem | Grad Norms [1st, 5th, 9th] |
|--------|------|----------|---------------------------|
| Without fix | 3649s | 74.27 GB | [inf, nan, nan] |
| With this fix | 2765s | 74.27 GB | [3.319, 3.219, 1.057] |

**Llama-3.2-1B regression at seq_len=2048, 21 steps:**
- Max loss delta vs baseline: 0.003
- Max grad_norm delta vs baseline: 0.06
- No regression

## Test plan
- [x] Reproduce `grad_norm=inf` at seq_len=69632 without fix
- [x] Verify finite grad_norms with fix at seq_len=69632
- [x] Llama short-context regression (max loss delta < 0.01)
- [ ] Sliding-window model test (Gemma2/Mistral)
- [ ] Packed-sequence training test

### DIFF
--- a/unsloth_zoo/temporary_patches/misc.py
+++ b/unsloth_zoo/temporary_patches/misc.py
@@ -511,39 +511,12 @@ TEMPORARY_PATCHES.append(patch_transformers_masks)
 
 
 def patch_sdpa_bool_causal_mask():
-    """Drop materialised bool causal masks before dispatching to SDPA.
+    """Fix unslothai/unsloth#4906: inf grad_norm on Qwen3.5 at seq_len > 65536.
 
-    Fixes unslothai/unsloth#4906 -- inf grad_norm on Qwen3.5 at seq_len > 65536.
-
-    When `patch_transformers_masks` (above) wraps `create_causal_mask` with
-    `torch.compile(dynamic=True)`, HF's `find_packed_sequence_indices` takes
-    the `is_tracing` branch and returns a non-None packed_sequence_mask even
-    for sequential positions. That sets `allow_is_causal_skip=False` inside
-    `create_causal_mask`, so `sdpa_mask` materialises a dense [1, 1, Q, K]
-    bool causal mask instead of returning None.
-
-    On builds that fall through to PyTorch SDPA's memory-efficient (Cutlass)
-    backend -- e.g. head_dim=256 models without flash-attn installed --
-    that backend with bf16 + a dense bool causal mask at seq_len > 65536
-    produces wrong forward outputs AND wrong gradients. The 65536 = 2**16
-    boundary matches a hard Cutlass int16 sequence-index limit.
-
-    Fix: wrap `sdpa_attention_forward` so that when the incoming
-    attention_mask is a plain lower-triangular 4D bool causal mask with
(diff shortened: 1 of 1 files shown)


danielhanchen:
Updated with a second commit to address the packed-sequence mask regression identified by review.

**What changed:** The single-cell spot-check `m[0,0,0,1]==False` was insufficient -- packed masks like `[[1,0,0,0],[1,1,0,0],[0,0,1,0],[0,0,1,1]]` also satisfy it. Added a second O(1) check: `m[0,0,-1,0]==True` (last query can see first key), which correctly rejects packed, padded, and sliding-window masks while accepting pure lower-triangular causal masks.

**Verification:**
- Unit test confirms all 5 mask types are correctly classified:
  - Pure causal: accepted (is_causal=True fast path)
  - Packed sequence: rejected (falls through to float fallback)
  - Right-padded: rejected (falls through to float fallback)
  - Left-padded: rejected (falls through to float fallback)
  - Sliding window: rejected (falls through to float fallback)
- GPU test: Qwen3.5-4B at seq_len=69632, 9 steps -- all grad_norms finite (3.319, 3.320, 3.012, 2.932, 3.209, 3.671, 2.848, 1.612, 1.058)

PR #592, opened 2026-04-14
GRPO: auto-fallback to eager when torch.compile selective_log_softmax OOMs

danielhanchen:
## Summary

The three RL `selective_log_softmax` helpers in `unsloth_zoo/rl_replacements.py` are decorated with `@torch.compile(dynamic=True, fullgraph=True, options=torch_compile_options)`. On some TorchInductor builds this picks oversized workspace pools for dynamic shapes and blows up memory on large-vocab models (observed e.g. on Blackwell sm_120 with torch >= 2.11 and Qwen 3.5's ~248k vocab), causing the compiled `chunked_hidden_states_selective_log_softmax` to OOM at step one of GRPO even when the actual logits chunks are small.

This PR keeps `torch.compile` as the fast path but wraps each of the three helpers in a tiny dispatcher that:
1. Tries the compiled function first.
2. Catches `torch.cuda.OutOfMemoryError` and `RuntimeError: CUDA driver error ... device not ready` (the symptoms reported in unslothai/unsloth#4985).
3. Logs a one-time warning, empties the CUDA cache, and falls through to the plain Python function for the rest of the process.
4. Unrelated `RuntimeError`s still propagate normally.

No configuration required - environments where compile works keep the fast path unchanged; environments where it OOMs now recover automatically.

Refs unslothai/unsloth#4985.

## Diff size

`unsloth_zoo/rl_replacements.py`, +83/-3 lines, one file.

## Validation

| Test | Result |
|---|---|
| Happy path (compile succeeds) | Correct output, `_use_compiled` stays `True` |
| Simulated workspace OOM | Wrapper swallows, flips to eager for the rest of the session, output identical to eager baseline |
| Non-OOM `RuntimeError` | Propagates untouched; `_use_compiled` stays `True` |
| Numerical equivalence compile vs eager | max \|diff\| = 3.8e-6 (bf16 noise) |
| `__name__` / docstring / `eager` attribute | Preserved for debugging |
| End-to-end Qwen 3.5 9B GRPO (5 steps, max_completion=4096) | Completes; peak alloc 80.95 GB (matches what this workload actually needs, not the inductor workspace inflation) |

Before this PR, users on the bad torch/GPU combo were reduced to monkey-patching `torch.compile` globally and swapping the function in every loaded module. This lands the fix in the right place.

## Test plan

- [ ] CI green on `unslothai/unsloth-zoo`
- [ ] Spot-check on an environment where compile works (no regression in training speed or memory)
- [ ] Spot-check on a reported-OOM environment (torch >= 2.11 + sm_120 + large vocab) - should see a one-time warning and training continues

### DIFF
--- a/unsloth_zoo/rl_replacements.py
+++ b/unsloth_zoo/rl_replacements.py
@@ -29,8 +29,88 @@ from .device_type import DEVICE_TYPE, device_synchronize
 from .temporary_patches.common import torch_compile_options
 RL_REPLACEMENTS = dict()
 
+
+# The three *selective_log_softmax functions below were originally decorated
+# with ``@torch.compile(dynamic=True, fullgraph=True, options=torch_compile_options)``.
+# On some TorchInductor builds that pick oversized workspace pools for dynamic
+# shapes (observed e.g. on NVIDIA Blackwell sm_120 with torch >= 2.11 and
+# large vocabularies like Qwen 3.5's ~248k vocab) the compiled kernel can
+# request multi-GB workspace buffers even when the actual logits chunks are
+# small, causing CUDA OOM inside ``chunked_hidden_states_selective_log_softmax``
+# during GRPO training. See https://github.com/unslothai/unsloth/issues/4985.
+#
+# We keep ``torch.compile`` as the fast path but wrap each function with an
+# auto-fallback: on the first CUDA OOM or ``RuntimeError: CUDA driver error``
+# raised by the compiled path, the wrapper transparently re-runs the inputs
+# on the plain (uncompiled) Python function and remembers the decision for
+# the rest of the process. No configuration is required -- environments
+# where compile works stay fast, and environments where it OOMs are
+# automatically recovered without the caller ever seeing the error.
+import functools as _functools
+
+
+def _is_workspace_oom(exc: BaseException) -> bool:
+    """Heuristic: is this the inductor workspace OOM we want to trap?"""
+    if isinstance(exc, torch.cuda.OutOfMemoryError):
+        return True
+    if isinstance(exc, RuntimeError):
+        msg = str(exc).lower()
+        if "out of memory" in msg:
+            return True
+        if "cuda driver error" in msg and "device not ready" in msg:
+            return True
+    return False
+
+
+class _AutoFallbackCompile:
+    """Call ``compiled`` first; on workspace OOM, fall back to ``eager`` for life."""
+
+    def __init__(self, fn):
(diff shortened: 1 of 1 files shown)


danielhanchen:
Closing this PR because I cannot reproduce the workspace OOM that motivated it.

## What I tried

Built a standalone reproducer using the exact `chunked_hidden_states_selective_log_softmax` function body and the same 22 inductor `torch_compile_options` that `unsloth_zoo` uses at production time. No `unsloth` or TRL - just torch + the decorated function. Ran on NVIDIA RTX PRO 6000 Blackwell (sm_120), which is the same arch as OP's RTX 5090.

| torch | seq=4096 peak_alloc | seq=4480 peak_alloc | OOM? |
|---|---|---|---|
| 2.9.1+cu128 | 7.95 GB | 8.53 GB | No |
| 2.11.0+cu128 | 7.95 GB | 8.53 GB | No |

Four stress scenarios on 2.11.0 + sm_120 trying to trigger inductor's worst-case workspace sizing:

| Scenario | Shapes | Worst peak_alloc |
|---|---|---|
| A: seqlen cycling 512→4480 × 3 iters | batch=2 | 8.53 GB |
| B: batch∈{1,2,3,4} × seqlen∈{1024,2048,4096} (multi-symbol dynamic) | b=4, s=4096 | 14.10 GB |
| C: huge warmup (b=8, s=8192) then OP's shape | b=2, s=4096 | 7.95 GB |
| D: vocab=512k (> Qwen 3.5's 248k) | b=2, s=4096 | 16.33 GB |

Memory scales linearly and predictably in batch×seqlen×vocab. At no point does inductor allocate the 44GB workspace pool that unslothai/unsloth#4985 describes.

## What this means

The 44GB `empty_strided_cuda((max(2818572288, 535040*((15 + s47*s87) // 16)), ), ...)` allocation OP observed is not something this function produces in isolation on torch 2.11 + sm_120. It is likely one of:
1. Environment-specific (WSL2, specific torch nightly, specific RTX 5090 SKU behavior)
2. Misattributed - the OOM trace comes from the outer GRPO training loop or `GradientCheckpointingLogSoftmax` autograd.Function wrapper, not this compiled kernel
3. State-dependent (stale inductor cache, vLLM-related state, interaction with gradient checkpointing)

Shipping a defensive wrapper here without a verified reproducer is noise: maintainers cannot confirm it fixes anything, and the warning message points to an issue that may not be the root cause.

If someone can attach a standalone repro that triggers the workspace OOM on this function in isolation, happy to revisit.

PR #598, opened 2026-04-16
Add per-model logit_matmul_upcast for fp16 RL safety

danielhanchen:
## Summary

- Adds a `logit_matmul_upcast: bool = False` parameter threaded through the GRPO logit computation path
- When `True`, the `hidden_states @ lm_head` matmul uses float32 instead of the model's native dtype
- Auto-detects Gemma-3/3n/4 model types in `grpo_accumulated_loss` and enables the upcast automatically
- Default `False` for all other models -- zero overhead

## Changes

`unsloth_zoo/rl_replacements.py`:
- `chunked_hidden_states_selective_log_softmax`: new `logit_matmul_upcast` param, conditional float32 matmul
- `Unsloth_Offloaded_Log_Softmax`: threads the param through forward/backward via `ctx.args`
- `efficient_log_softmax`: passes through to both fast and offloaded paths
- `grpo_accumulated_loss`: auto-detects model type from `trainer.model.config.model_type` and `text_config.model_type`, sets `logit_matmul_upcast=True` for known problematic models. The model set is inlined in the function body so the compiled trainer can see it.

## Motivation

Gemma-4 E2B GRPO training can produce NaN on float16. The hidden states grow to magnitudes of ~2000+ through 35 decoder layers with residual connections. While the lm_head matmul does not currently overflow for Gemma-4 E2B specifically (lm_head weights are small at max 0.56), this provides a safety guard for:
1. Models with larger embedding weights
2. Longer training runs where hidden states may grow further
3. Future Gemma variants

## Extensibility

To add another model type, add its `model_type` string to the `_upcast_models` set inside `grpo_accumulated_loss`. Can also be forced externally via `kwargs["logit_matmul_upcast"] = True`.

## Companion PR

unslothai/unsloth#5055 adds `gemma4` to the `FORCE_FLOAT32` list (the primary fix that auto-switches fp16 to float32 mixed precision).

## Test plan

- [x] fp16 Gemma-4 E2B GRPO 10 steps -- NaN-free
- [x] bf16 Gemma-4 E2B GRPO 10 steps -- unaffected
- [x] Compiled GRPO trainer regenerates correctly with new parameter


danielhanchen:
Closing in favor of a clean PR -- this one accumulated wrong commits and accidentally uploaded compiled cache files.

PR #599, opened 2026-04-16
Add per-model logit_matmul_upcast for Gemma-4 fp16 RL

danielhanchen:
## Summary

Adds a `logit_matmul_upcast: bool` parameter to the GRPO logit computation path. When True, the `hidden_states @ lm_head` matmul uses float32 instead of the model's native dtype, preventing fp16 overflow during RL training.

Only one file changed: `unsloth_zoo/rl_replacements.py`

## What changed

**Module level:**
- `LOGIT_MATMUL_UPCAST_MODELS` frozenset with `gemma4` and `gemma4_text`

**`chunked_hidden_states_selective_log_softmax`:**
- New `logit_matmul_upcast: bool = False` parameter
- `lm_head.float().t()` hoisted outside the chunk loop (cast once, not per chunk)
- When False (default), behavior is identical to before

**`Unsloth_Offloaded_Log_Softmax`:**
- Parameter threaded through forward/backward via `ctx.args`
- Extra `None` in backward return tuple

**`efficient_log_softmax`:**
- Parameter passed through to both fast and offloaded paths

**`grpo_accumulated_loss`:**
- Auto-detection placed after `_autocast_dtype` initialization
- Only fires when `trainer._autocast_dtype == torch.float16` (not bf16/fp32)
- Checks `model.config.model_type` and `text_config.model_type` against the frozenset
- Uses `from unsloth_zoo.rl_replacements import LOGIT_MATMUL_UPCAST_MODELS` so the compiled trainer can resolve it
- Passes `logit_matmul_upcast=logit_matmul_upcast` to both `efficient_log_softmax` call sites

## Test results

Gemma-4 E2B GRPO, ga=4, num_generations=2, max_steps=10:

- fp16 with this PR: 10/10 steps NaN-free
- bf16 control: 10/10 steps NaN-free, upcast not triggered
- fp16 without this PR: NaN at step 5

## Extensibility

Add model types to `LOGIT_MATMUL_UPCAST_MODELS`. Or force manually via `kwargs["logit_matmul_upcast"] = True`.

### DIFF
--- a/unsloth_zoo/rl_replacements.py
+++ b/unsloth_zoo/rl_replacements.py
@@ -29,6 +29,7 @@ from .device_type import DEVICE_TYPE, device_synchronize
 from .temporary_patches.common import torch_compile_options
 RL_REPLACEMENTS = dict()
 
+
 # https://github.com/huggingface/trl/blob/main/trl/trainer/utils.py#L1674
 @torch.compile(dynamic = True, fullgraph = True, options = torch_compile_options,)
 def selective_log_softmax(logits, index):
@@ -76,6 +77,7 @@ def chunked_hidden_states_selective_log_softmax(
     logit_scale_divide: float = 0.0,
     logit_softcapping: float = 0.0,
     temperature: float = 1.0,
+    logit_matmul_upcast: bool = False,
 ) -> torch.Tensor:
     # All Unsloth Zoo code licensed under AGPL3
     flat_hidden_states = hidden_states.reshape(-1, hidden_states.shape[-1])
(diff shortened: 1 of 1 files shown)


unsloth_zoo/rl_replacements.py
```diff
@@ -87,7 +89,10 @@ def chunked_hidden_states_selective_log_softmax(
     all_per_token_logps = []
 
     for chunk_hidden_states, chunk_index in zip(chunked_hidden_states, chunked_index):
-        chunk_logits = chunk_hidden_states.to(lm_head.dtype) @ lm_head.t()
+        if logit_matmul_upcast:
+            chunk_logits = chunk_hidden_states.float() @ lm_head.float().t()
```
Datta0, line 93: Ideally we should upcast before the for loop to avoid unnecessary repetition of upcast.

PR #600, opened 2026-04-17
Add Gemma-4 float16 UNSLOTH_FORCE_FLOAT32 patches for GRPO stability

danielhanchen:
## Problem

`unsloth/gemma-4-E2B-it` and `unsloth/gemma-4-E4B-it` GRPO training NaNs and crashes with a CUDA device-side assert at step 2 when the user passes `dtype=torch.float16`. `torch.bfloat16` is unaffected.

Minimal reproduction, 8 steps on a single B200:

```
python scripts/gemma4_grpo_sudoku_repro.py --model unsloth/gemma-4-E2B-it --dtype float16 --max-steps 8 --grad-accum 4 --num-generations 2
```

Unpatched result (dtype=float16):

```
step 1: loss=0.0, grad_norm=0.1108, reward=1.416
step 2: CUDA error: device-side assert triggered
```

Hooks on the model show the first overflow occurs in `language_model.layers.0.mlp.down_proj`. `post_stats[base_layer]` hits +inf during step 1's forward, poisons the residual stream, and step 2's generation produces NaN logits that trip the categorical sampler.

## Root cause

Under `UNSLOTH_FORCE_FLOAT32=1`, the loader stores weights as bfloat16 but GRPO enters its fp16 autocast context for training. Gemma-4's text-decoder MLP does

```
down_proj(act_fn(gate_proj(x)) * up_proj(x))
```

with no internal upcast. Inside fp16 autocast, gate and up outputs saturate at `fp16_max`, the product becomes +inf on downcast, and `down_proj`'s fp16 matmul accumulator then produces +inf for the residual add. bf16 avoids this because its range covers fp32 magnitudes.

Gemma-3 already ships `UNSLOTH_FORCE_FLOAT32` patches for the same failure mode. The corresponding Gemma-4 patches existed briefly and were removed in 158e9812 (2026-04-06) after initial testing found it stable. GRPO fp16 training exercises code paths the earlier testing did not.

## Fix

Four text-decoder patches in `unsloth_zoo/temporary_patches/gemma4.py`, each gated on `UNSLOTH_FORCE_FLOAT32=1` and each appended to `TEMPORARY_PATCHES`:

| Patch | What it does |
|-------|--------------|
| `patch_Gemma4RMSNorm` | fp32 variance and weight multiply, clamp to 65280 (one bf16 ulp below fp16_max so later fp16 casts cannot round up to inf), then fp16 cast. |
| `patch_Gemma4TextScaledWordEmbedding` | Embed lookup in fp32 multiplied by the fp32 embed_scale so the first RMSNorm gets full precision input. |
| `patch_Gemma4TextMLP` | Disable fp16 autocast so gate_proj and up_proj execute at the bf16 weight dtype; activation and multiply in fp32; clamp to 65280; fp16 cast; `torch.nan_to_num` safety net rescues the rare case where down_proj's fp16 accumulator still overflows for wide intermediate dims. |
| `patch_Gemma4TextAttention` | Same autocast-disabled pattern for Q/K/V projections, fp32 q/k/v_norm, fp32 RoPE, fp32 SDPA, clamp and fp16 cast before o_proj, same `nan_to_num` safety net. Preserves the KV-sharing and `shared_kv_states` contract. |

The paired change in `unslothai/unsloth` adds `"gemma4"` to `FORCE_FLOAT32` so these patches actually engage for fp16 requests.

No Gemma-4 causal mask patch is added because, unlike Gemma-3, Gemma-4 delegates mask construction to `ALL_ATTENTION_FUNCTIONS` and the existing sdpa / eager / flex_attention paths handle mask dtype correctly once the Q/K/V inputs are finite.

## Verification

All runs on a B200 using `scripts/gemma4_grpo_sudoku_repro.py` with the plan's parameters (`max_steps=8`, `grad_accum=4`, `num_generations=2`, `batch_size=1`, seed 3407).

| Model | dtype | GC | Status | step 1 loss | step 8 loss | step 8 grad_norm | step 8 kl |
|-------|-------|-----|--------|-------------|-------------|-------------------|-----------|
| E2B-it | bf16 baseline | unsloth | pass | -2.79e-06 | 4.92e-07 | 0.3908 | 5.74e-05 |
| E2B-it | fp16 unpatched | unsloth | CUDA assert at step 2 | 0.0 | n/a | n/a | n/a |
| E2B-it | fp16 patched | unsloth | pass | 0.0 | 7.15e-07 | 0.2655 | 1.18e-04 |
| E2B-it | fp16 patched | off | pass | 0.0 | 7.15e-07 | 0.2655 | 1.18e-04 |
| E4B-it | fp16 patched | unsloth | pass | 1.64e-07 | 1.42e-08 | 0.0000 | 1.42e-05 |

`hooks.find_overflows()` on the patched runs reports only the embedding input-id pseudo-overflow (hooks see integer token IDs up to the 256k vocab) and the layer-0 down_proj saturation that the `nan_to_num` safety net catches. No parameter contains NaN or inf at the end of any patched run (`post_train_bad_params: []`).

## Notes for reviewers

- The clamp bound is 65280 rather than `fp16_max`=65504 because bf16's next-representable value above 65504 is 65536, which becomes inf on the fp16 cast taken by the PEFT LoRA wrapper on the down_proj base_layer.
- The `nan_to_num` call is a defense-in-depth measure, not the primary fix. It only fires when the fp16 accumulator in the down_proj / o_proj matmul tips past fp16_max even for a bounded input; the upstream clamp in the MLP and attention paths prevents this in the vast majority of steps.
- The patches are no-ops when `UNSLOTH_FORCE_FLOAT32=0` (i.e. any bf16 run) so bf16 users see no behavior change.


danielhanchen:
Reduced to a single patch after a cleaner bisection.

`patch_Gemma4TextMLP` alone is sufficient to fix the step-2 NaN on `unsloth/gemma-4-E2B-it` and `unsloth/gemma-4-E4B-it` GRPO under `dtype=torch.float16`. The RMSNorm, Embedding, and Attention patches from the earlier version are redundant:

- Upstream `Gemma4RMSNorm` already computes internally in fp32 and uses `type_as(hidden_states)`; for trained weight ranges the output is finite in fp16.
- `Gemma4TextScaledWordEmbedding` multiplies by `sqrt(hidden_size)` = ~45 to 60 for E2B/E4B, well within fp16.
- `Gemma4TextAttention` projections did not overflow in any run.

Once the MLP no longer writes +-inf into the residual stream, the rest of the text pipeline stays finite without modification. Removing the three extra patches also drops the autocast-disable gymnastics and keeps the gate / up / down matmuls on fp16 tensor cores (important for Tesla T4 which has no bf16 tensor cores).

The MLP patch is now three operations:

```python
def forward(self, x):
    gate = self.gate_proj(x)
    up = self.up_proj(x)
    product = self.act_fn(gate.float()) * up.float()
    product = torch.clamp(product, min=-65280.0, max=65280.0)
    out = self.down_proj(product.to(x.dtype))
    return torch.nan_to_num(out, nan=0.0, posinf=65280.0, neginf=-65280.0)
```

Verified on B200 (fp16 GRPO, `max_steps=8`, `grad_accum=4`, `num_generations=2`):

| Model | GC | step 1 grad_norm | step 8 grad_norm | step 8 kl |
|-------|-----|------------------|--------------------|-----------|
| unsloth/gemma-4-E2B-it | unsloth | 0.1110 | 0.4077 | 6.03e-05 |
| unsloth/gemma-4-E2B-it | off | 0.1110 | 0.2655 | 1.18e-04 |
| unsloth/gemma-4-E4B-it | unsloth | 0.0987 | 0.0000 | 1.47e-05 |

All runs complete 8 steps with `post_train_bad_params: []`.

`UNSLOTH_ENABLE_FLEX_ATTENTION` note: flex is enabled by default but Gemma-4 text attention uses sdpa when sdpa is available. Setting the env to 0 produces identical E4B trajectories and near-identical E2B (steps 1-7 identical; step 8 diverges by a small amount but stays finite). No action needed.

danielhanchen:
Final state: shipping only the minimal fp16 MLP trick.

The int8 weight-only variant (`torch.ops.aten._weight_int8pack_mm` + autograd.Function + LoRA-aware forward) was prototyped and removed after a 100-step GRPO parity run showed it fails the bf16-closeness goal: GRPO's `log_pi_new - log_pi_old` ratio amplifies the ~7% per-matmul quantization noise because the rollout path runs under `torch.inference_mode` (dense fp16 fallback) while the training forward would use int8. Symmetric int8 on both paths would need a separate int8 reference model, which is out of scope for a surgical NaN fix.

100-step parity table on `unsloth/gemma-4-E2B-it` GRPO (temperature=0.05, min_p=0.5, seed=3407, grad_accum=4, num_generations=2):

| metric | bf16 | fp16+clamp (shipped) | int8 weight-only (dropped) |
|--------|------|----------------------|------------------------------|
| median \|KL\| | 4.02e-05 | **7.99e-06** | 1.06e-02 |
| p95 \|KL\| | 0.411 | 0.620 | 2.788 |
| max \|KL\| | 8529 | **35.77** | 7.46e+06 |
| total \|KL\| | 9388 | **52.16** | 7,471,000 |
| mean reward | 1.2111 | 1.1359 | 1.1611 |
| calm steps (\|KL\|<1e-3) | 63/100 | **66/100** | 42/100 |
| big spikes (\|KL\|>1.0) | 4/100 | **3/100** | 8/100 |
| training time | 3712s | 3722s (+0.3%) | 4256s (+14.7%) |
| post_train_bad_params | 0 | 0 | 0 |

Reward-identical steps (\|delta\|<0.01): bf16 vs fp16+clamp = 75/100.

The shipped patch body is now three operations:

```python
def forward(self, x):
    gate = self.gate_proj(x)
    up = self.up_proj(x)
    product = self.act_fn(gate.float()) * up.float()
    product = torch.clamp(product, min=-65280.0, max=65280.0)
    out = self.down_proj(product.to(x.dtype))
    return torch.nan_to_num(out, nan=0.0, posinf=65280.0, neginf=-65280.0)
```

Dtype contract unchanged from upstream, no attention / RMSNorm / embedding companion patches, gate / up / down matmuls stay on fp16 tensor cores (full T4 throughput at 65 TFLOPS).

danielhanchen:

Adds patch_Gemma4TextMLP which stabilizes Gemma-4 GRPO training under fp16 on Tesla T4 by running gate*up in fp32, clamping to 65280, and rescuing down_proj accumulator overflows with nan_to_num, gated on UNSLOTH_FORCE_FLOAT32 and the gate_proj output dtype so bf16/fp32 paths stay byte-identical to upstream.

**Reason:** Correct fp16 overflow rescue for Gemma-4 MLP; all accepted review findings fixed, no remaining regressions.

danielhanchen:

Adds patch_Gemma4TextMLP to stabilize Gemma-4 fp16 GRPO training (notably Tesla T4) by computing gate*up in fp32 with safe clamping and nan_to_num rescue; review hardened it with signature-mismatch detection, inf/NaN neutralization before act_fn, and an idempotency sentinel.

**Reason:** All real issues fixed; 16/16 tests pass; core fp32-upcast + clamp + nan_to_num logic is correct and minimal.

PR #602, opened 2026-04-19
Guard GRPO text branch against models returning logits

Rorical:
## Problem

In `grpo_accumulated_loss` (unsloth_zoo/rl_replacements.py), the branch handling text-only inputs (`if pixel_values is None:`) passes the model's `.logits` output unconditionally into `efficient_log_softmax`, which assumes hidden states and computes `hidden_states @ lm_head.t()`. When a model's forward actually returns real logits (last dim = vocab) rather than pre-lm_head hidden states, this produces a shape-mismatched matmul and crashes the RL step.

The vision branch (`else:`, lines 986–1001) already dispatches based on a shape check:

```python
if new_hidden_states_chunk.shape[-1] == lm_head.shape[1]:
    logprobs_chunk = efficient_log_softmax(...)
else:
    logprobs_chunk = chunked_selective_log_softmax(new_hidden_states_chunk, completion_ids, temperature)
```

The text branch is missing the same guard.

## Repro

Training `nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16` (hybrid Mamba+attention MoE, text-only) with Unsloth `PatchFastRL("GRPO", ...)` and `trl.GRPOTrainer`. The model's remote modeling code returns genuine logits from `.logits`, and the text branch crashes with a dimension mismatch inside the compiled `chunked_hidden_states_selective_log_softmax`. Other text models that ship with custom modeling code can hit the same path.

Workaround in the wild: monkey-patching the compiled cache file to add the guard.

## Fix

Mirror the existing vision-branch guard into the text branch. When the model returns actual logits, skip the `lm_head` matmul and go directly to `chunked_selective_log_softmax`. No behavioral change for models that follow Unsloth's usual pattern of returning hidden states through `.logits`.

## Diff

Pure addition of the shape-check branch in the text path; no logic changes elsewhere.

### DIFF
--- a/unsloth_zoo/rl_replacements.py
+++ b/unsloth_zoo/rl_replacements.py
@@ -932,6 +932,29 @@ def grpo_accumulated_loss(
                 logit_softcapping, temperature
             )
 
+    def compute_logprobs_chunk(new_hidden_states_chunk, completion_ids, input_ids_chunk):
+        # Hidden states -> lm_head matmul path; raw logits -> skip matmul and
+        # skip scale/softcap (model forward already applied them).
+        chunks = input_ids_chunk.shape[0] * multiplier
+        if new_hidden_states_chunk.shape[-1] == lm_head.shape[1]:
+            return efficient_log_softmax(
+                new_hidden_states_chunk,
+                lm_head,
+                completion_ids,
+                chunks = chunks,
+                logit_scale_multiply = logit_scale_multiply,
+                logit_scale_divide = logit_scale_divide,
+                logit_softcapping = logit_softcapping,
+                temperature = temperature,
+                batch_size = B,
+            )
+        return chunked_selective_log_softmax(
+            new_hidden_states_chunk,
+            completion_ids,
+            temperature = temperature,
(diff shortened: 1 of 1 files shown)


Datta0:
Nice find. Maybe we should consolidate the two into a function and reuse?

Rorical:
Great suggestion! I've merged them into a shared helper function

Datta0 (approved):
LGTM

bi1101:
Hi @Datta0 , may I ask if this PR is going to be merged soon? This is blocking text-only GRPO for Qwen 3.5

lastrei:
> Hi @Datta0 , may I ask if this PR is going to be merged soon? This is blocking text-only GRPO for Qwen 3.5

Yes，This update breaks the text-only GRPO for Qwen 3.5. Once I load the model using FastVisionModel, it still expects images even when my input data contains only text and no images at all, triggering RuntimeError: The size of tensor a (2) must match the size of tensor b (0) at non-singleton dimension 1.

PR #604, opened 2026-04-21
fix(cpu): return "cpu" from get_device_type() instead of raising when no GPU found

LeoBorcherding:
Companion branch: https://github.com/LeoBorcherding/unsloth/tree/fix/cpu-no-torch-import
Companion issue: unslothai/unsloth#5008

## Changes
- `device_type.py`: return `"cpu"` instead of raising `NotImplementedError` when no GPU accelerator found
- `__init__.py`: gate all torch-dependent imports and allocator config behind `_HAS_TORCH`, add CPU stubs for device type functions
- `temporary_patches/gpt_oss.py`: guard CUDA memory query behind device type check to avoid crash on CPU

## Validation
Validated in LeoBorcherding/unsloth@fix/cpu-no-torch-import. Full validation log posted in [unsloth/pull/5119](https://github.com/unslothai/unsloth/pull/5119).

### DIFF
--- a/unsloth_zoo/__init__.py
+++ b/unsloth_zoo/__init__.py
@@ -248,7 +248,7 @@ else:
 IS_HIP_RUNTIME = (DEVICE_TYPE == "hip") or bool(is_hip())
 
 # Torch >= 2.9 uses PYTORCH_ALLOC_CONF and treats legacy per-backend vars as deprecated.
-if _HAS_TORCH and IS_TORCH_2_9_OR_NEWER:
+if IS_TORCH_2_9_OR_NEWER:
     # Preserve explicit legacy allocator settings when user did not directly set PYTORCH_ALLOC_CONF.
     if not _HAS_ORIGINAL_PYTORCH_ALLOC_CONF:
         promoted = _ORIGINAL_PYTORCH_CUDA_ALLOC_CONF
--- a/unsloth_zoo/__init__.py
+++ b/unsloth_zoo/__init__.py
@@ -263,7 +263,7 @@ if IS_TORCH_2_9_OR_NEWER:
     delete_key("PYTORCH_HIP_ALLOC_CONF")
 
 # Specify PYTORCH_CUDA_ALLOC_CONF or PYTORCH_HIP_ALLOC_CONF
-if _HAS_TORCH and IS_HIP_RUNTIME:
+if IS_HIP_RUNTIME:
     if IS_TORCH_2_9_OR_NEWER:
         # PyTorch >= 2.9 uses PYTORCH_ALLOC_CONF. expandable_segments is unsupported on HIP.
         remove_expandable_segments("PYTORCH_ALLOC_CONF")
--- a/unsloth_zoo/__init__.py
+++ b/unsloth_zoo/__init__.py
@@ -102,11 +102,6 @@ _NO_TORCH_MODE = os.environ.get("UNSLOTH_NO_TORCH", "0").strip().lower() in (
 if (not _HAS_TORCH) and (not _NO_TORCH_MODE):
     _NO_TORCH_MODE = True
     os.environ["UNSLOTH_NO_TORCH"] = "1"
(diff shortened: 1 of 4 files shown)


LeoBorcherding:
Marking PR's as draft, need to test again after latest improvements. Will retest both paths tomorrow against the latest commits before merging.

LeoBorcherding:
This PR is ready and working with unsloth#5119. 

Tested end-to-end on Windows CPU-only:
- grouped-gemm probe fallback working ✅
- unsloth-zoo imports successfully without torch ✅
- All features gate correctly for no-torch mode ✅
- Still required Linux Test and validation ❌

Ready for review with [#5119](https://github.com/unslothai/unsloth/pull/5119).

LeoBorcherding:
Linux validation done - 3/3 pass on Ubuntu 24.04 (WSL2, Python 3.12, `torch==2.11.0+cpu`, no GPU). Same results as Windows.

Both PRs are fully validated across platforms. Ready to merge whenever #5119 is reviewed.

rolandtannous:
check comment on Companion PR: https://github.com/unslothai/unsloth/pull/5119

unsloth_zoo/temporary_patches/moe_utils.py
```diff
@@ -232,7 +232,7 @@ def _check_grouped_gemm_available():
         from unsloth.kernels.moe.grouped_gemm.interface import grouped_gemm, supports_tma
         _GROUPED_GEMM_AVAILABLE = True
         _init_triton_allocator()
-    except (ImportError, ModuleNotFoundError):
+    except Exception:
```
Datta0, line 235: What other exception are we trying to catch/avoid here?

unsloth_zoo/__init__.py
```diff
@@ -175,28 +183,28 @@ def clean_expandable_segments_value(value):
     return ",".join(parts) if len(parts) else None
 
 
-if (major_torch < 2):
+if _HAS_TORCH and (major_torch < 2):
```
Datta0, line 186: NIT: Instead of all this, should we just do, 
```
if not _HAS_TORCH:
   return/exit
# Rest of the code
```
?

LeoBorcherding: cant use return at module level

unsloth_zoo/temporary_patches/moe_utils.py
LeoBorcherding: _init_triton_allocator() can raise RuntimeError on CUDA/Triton init failures even when the import succeeds, that's what we're covering with except Exception. Also, ModuleNotFoundError is a subclass of ImportError so the original was already redundant there

LeoBorcherding (commented):
addressed Datta's concerns, and applied some of Gemini's suggestions

danielhanchen:
Confirmed main still raises in device_type.py where no accelerator is found, so the CPU return is a real fix. The direction is right but this also gates the whole allocator and import block in __init__.py and widens the grouped-gemm probe to except Exception. Could you split those out and keep this branch to the get_device_type change?

danielhanchen:
main now handles this in unsloth_zoo/device_type.py via _cpu_fallback and the UNSLOTH_ZOO_DISABLE_GPU_INIT gate in __init__.py, which keeps the no-GPU answer opt-in rather than making "cpu" the unconditional default, so I do not think this branch is needed any more. Sorry for the slow review, and thanks for the cross-platform validation you did on it.

danielhanchen:
main now takes the CPU path in unsloth_zoo/device_type.py via _cpu_fallback (UNSLOTH_ALLOW_CPU / UNSLOTH_ZOO_DISABLE_GPU_INIT) with matching constants and stubs in __init__.py, so I do not think this branch is needed any more, and returning "cpu" unconditionally would also send a host with a broken CUDA install silently onto CPU. Sorry for the slow review, and thanks for putting it together along with the Linux and Windows validation.

PR #603, opened 2026-04-20
[Qwen 3.5][gemma4] Qwen35 and Gemma 4 fast inference (clean-base mirror of #588)

danielhanchen:
## Summary

Clean-base mirror of #588. Contains only the original author's 9 commits from @Datta0 (7fa143f1 through b7052b55, 2026-03-30 to 2026-04-17), with none of the later auto-review fix commits that accumulated on that PR's branch.

## Why

Provides a fresh starting point to re-run the PR-review pipeline against the author's original diff without interference from prior review-loop commits layered on top of #588. Useful for validating review-pipeline changes end-to-end.

## Branch contents

Nine commits from @Datta0:
- 7fa143f [WIP] initial fast_inference support for qwen3.5
- 31c1bc3 [WIP] fixes for rope deltas
- 112f1b5 cleanup
- a31dee8 fix lm_head detection and remove moe
- 5d07504 [WIP] gemma 4 dense fast inference
- a85a4f4 Fix dtype setting
- d2f7983 Merge remote-tracking branch 'origin/main' into qwen35-shared-final
- 43ed24f Fix gemma4 load on vllm 0.19.0
- b7052b5 fix bnb loader for gemam4

## Note

Not a replacement for #588. See #588 for the actively-maintained PR and associated review history.

PR #608, opened 2026-04-22
Make Qwen3 MoE grouped_mm path CUDA-graph / torch.compile friendly

danielhanchen:
## Summary

Drops two patch-time escape hatches on the `grouped_mm` MoE backend so Qwen3 MoE decode can be captured into a CUDA graph end-to-end. Needed by unslothai/unsloth#5123's new `FlexMoEInference` capture path — without these, CUDA graph capture trips on a CPU→CUDA scalar copy inside `torch.bincount(..., minlength=python_int)`.

1. `qwen3_moe.py`: `_make_qwen_moe_experts_forward` resolves the backend once at patch time. On `grouped_mm` it returns a closure that calls `forward_native_grouped_mm` directly, eliminating the per-forward `select_moe_backend()` dispatch. Other backends keep the generic `forward_moe_backend` dispatcher.

2. `qwen3_moe.py`: `_make_qwen_moe_sparse_moe_block_forward` only wraps the result in `@torch.compiler.disable` when the resolved backend is NOT `grouped_mm`. On `grouped_mm` with pre-merged LoRA the block is Dynamo-clean (no `.item()`, no Python loops over experts, no mutation) and should compile as-is. The decorator is still applied unconditionally for `unsloth_triton` and `native_torch` since those backends use the legacy expert-loop path.

3. `moe_utils.py::forward_native_grouped_mm`: replace `torch.bincount(flat_top_k, minlength=num_experts)` with an equivalent `torch.zeros(...).scatter_add_(0, flat_top_k, torch.ones_like(flat_top_k))`. `bincount`'s `minlength` is a Python int; under CUDA graph capture this triggers a host-side scalar materialisation and fails with `Cannot copy between CPU and CUDA tensors during CUDA graph capture`. `scatter_add_` on a pre-zeroed device tensor is identical semantically and fully graph-capturable.

## Benchmarks

Qwen3-30B-A3B-Instruct-2507, bs=8 × 64 new tokens on a single B200, with unslothai/unsloth#5123's FlexMoEInference CUDA graph decode enabled.

| precision | pure HF naive | Flex + CUDA graphs | speedup |
|-----------|--------------:|-------------------:|--------:|
| 4bit      | 66.7 tok/s    | 374.3 tok/s        | 5.6x    |
| bf16      | 101.1 tok/s   | 421.6 tok/s        | 4.2x    |

Greedy parity:
- flex-captured is bit-exact to flex-eager: 32/32 tokens × 3 prompts at 4bit
- flex-captured vs pure transformers.AutoModelForCausalLM: 32/32 on 5 out of 6 (prompt × precision) pairs; the remaining divergence is a tie-break logit boundary on an open-ended continuation, both outputs coherent English

GRPO 5-step smoke: 4bit 548.3s peak 130.9 GB, bf16 451.8s peak 133.4 GB, loss / KL stable, no NaN.

## Test plan

- [x] UNSLOTH_MOE_BACKEND=grouped_mm smoke: Qwen3 MoE generates coherent text at 36.5 tok/s warm (4bit) / 36.4 tok/s (bf16) through FlexMoEInference with CUDA graphs
- [x] UNSLOTH_MOE_BACKEND=native_torch: unchanged behavior (decorator still applied, legacy loop still disabled from compile)
- [x] bincount → scatter_add_ produces identical num_tokens_per_expert counts (verified by decode parity against pre-change eager)
- [x] GRPO 5-step training runs to completion on both 4bit and bf16 with no divergence

### DIFF
--- a/unsloth_zoo/temporary_patches/qwen3_moe.py
+++ b/unsloth_zoo/temporary_patches/qwen3_moe.py
@@ -39,6 +39,8 @@ from .utils import (
 from .moe_utils import (
     patch_param_wrapper_for_moe,
     get_forward_moe_backend,
+    select_moe_backend,
+    forward_native_grouped_mm,
 )
 
 
@@ -69,11 +71,31 @@ def _make_qwen_moe_lora_extractor():
 
 
 def _make_qwen_moe_experts_forward(module_name: Optional[str] = None):
+    # Resolve the MoE backend once at patch time. For grouped_mm we bind
+    # ``forward_native_grouped_mm`` directly so the decode hot path has no
+    # per-token ``select_moe_backend()`` dispatch and presents a stable
+    # callable to ``torch.compile`` / CUDA graph capture. Other backends
+    # keep the generic dispatcher (``forward_moe_backend``) which still
+    # benefits from the ``lru_cache`` on ``select_moe_backend``.
+    backend = select_moe_backend()
+    if backend == "grouped_mm":
+        def forward(self, hidden_states, top_k_index, top_k_weights):
+            return forward_native_grouped_mm(self, hidden_states, top_k_index, top_k_weights)
+        return forward
     return get_forward_moe_backend()
 
 
 def _make_qwen_moe_sparse_moe_block_forward(use_shared_expert: bool, module_name: Optional[str] = None):
-    @torch.compiler.disable
+    # The legacy expert-loop path (``forward_native_moe_loop``) uses
+    # ``torch.where`` + a Python for-loop over experts and is genuinely
+    # uncapturable — keep ``@torch.compiler.disable`` when that backend
+    # is selected. On ``grouped_mm`` with pre-merged LoRA the sparse
+    # block is Dynamo-clean (no ``.item()``, no Python loops, no
+    # mutation), so we omit the decorator and let ``torch.compile``
+    # trace through the whole block for prefill.
+    _backend = select_moe_backend()
+    _compile_safe = _backend == "grouped_mm"
+
     def sparse_moe_block_forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
         use_shared_expert = hasattr(self, "shared_expert") and hasattr(self, "shared_expert_gate")
         if hidden_states.dim() == 3:
@@ -115,6 +137,8 @@ def _make_qwen_moe_sparse_moe_block_forward(use_shared_expert: bool, module_name
 
     if module_name is not None:
         sparse_moe_block_forward.__module__ = module_name
+    if not _compile_safe:
+        sparse_moe_block_forward = torch.compiler.disable(sparse_moe_block_forward)
     return sparse_moe_block_forward
 
 
--- a/unsloth_zoo/empty_model.py
+++ b/unsloth_zoo/empty_model.py
@@ -21,6 +21,9 @@ __all__ = [
     "get_model_layer_config",
     "compare_attributes",
     "copy_attributes",
+    "patch_gemma4_vllm_lora_support",
+    "patch_gemma4_vllm_k_eq_v_support",
+    "patch_qwen3_5_vllm_lora_support",
 ]
 
 import torch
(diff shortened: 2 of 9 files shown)


danielhanchen:
**Companion benchmark: vLLM direct on Qwen3-30B-A3B-Instruct-2507 (B200).**

Ran vLLM 0.19.2 on the same decode workload as unsloth#5123 to give the
grouped_mm + CUDA-graph changes here a proper external reference. Bench
script is in unsloth#5123 as `tests/flex_moe_vllm_bench.py`.

bf16 decode, tok/s (flex uses the cudagraph + torch.compile-walker path
enabled by the patches in this PR):

| bs | Flex walker | vLLM (no LoRA) |
|---:|------------:|---------------:|
|  8 |       905.0 |         1360.0 |
| 16 |      1401.1 |         2471.6 |
| 32 |      2864.0 |         4721.3 |
| 48 |      3911.1 |         6794.1 |

4bit (bnb) decode:

| bs | Flex walker (bnb-4bit ckpt) | vLLM (bnb on-the-fly) |
|---:|----------------------------:|----------------------:|
|  8 |                       752.6 |                 177.0 |

Notes:
- The grouped_mm patches in this PR (`qwen3_moe._make_qwen_moe_experts_forward`
  pre-binds the backend at patch time; `_make_qwen_moe_sparse_moe_block_forward`
  drops `@torch.compiler.disable` on grouped_mm; `moe_utils.forward_native_grouped_mm`
  uses `scatter_add_` instead of `bincount` for cudagraph-safe routing) are
  what lets the flex walker capture and compile end-to-end. Without them
  the decode CUDA graph refused to capture on Qwen3 MoE.
- Flex decisively wins the 4bit-bnb regime (4.25x) — vLLM has no native bnb
  MoE kernel, so each step dequantises expert weights on the fly. This is
  the most common GRPO configuration.
- vLLM is 1.5-1.76x faster on bf16 dense decode. Headroom on our side is
  primarily attention kernels and weight-freeze compile.
- vLLM LoRA (`enable_lora=True`) pays ~40% throughput for punica per-step
  dispatch; flex merges once per rollout via `torch.baddbmm` so the decode
  hot loop stays branch-free. Details in
  https://github.com/unslothai/unsloth/pull/5123#issuecomment-4301605491.

danielhanchen:
**Companion validation: MoE LoRA merge correctness + perf**

Since the grouped_mm + cudagraph changes here are paired with a
stacked-expert LoRA merge
(`unsloth.inference.flex_moe::refresh_moe_lora_merge_from_pristine`),
added an independent parity + perf test:
https://github.com/unslothai/unsloth/pull/5123#issuecomment-4301714346.
Summary: all 8 numerical cases bitwise-match the textbook reference
`W_ref[e] = W_pristine[e] + Σ scaling · B[e] @ A[e]` at Qwen3-30B MoE
shapes; batched `baddbmm` is 5-9.6x faster than a per-expert `addmm`
loop and 3.6-7.1x **cheaper per expert** than the dense addmm path used
by `refresh_lora_merge_from_pristine`.

Confirms the cudagraph-safe merge-then-decode path used by this PR is
correct and performance-sensible.

danielhanchen:
**Companion: capture-safe expert bias expansion (needed for gpt-oss flex cudagraph)**

Added a single fix in `forward_native_grouped_mm` (commit 32a7888):
`gate_up_proj_bias` / `down_proj_bias` were being expanded via
`repeat_interleave(counts_tensor)`, which forces a GPU→CPU sync (output
size depends on `counts.sum().item()`) and breaks CUDA graph capture.

Replaced with `index_select` on the per-token expert id derived from the
existing `sorted_indices = argsort(flat_top_k)` permutation:
`sorted_expert_ids = flat_top_k[sorted_indices]`. Capturable, numerically
identical.

Only affects paths that actually have these biases (gpt-oss today; Qwen3
MoE has neither), so it's a no-op risk for Qwen3.

This is what lets the flex CUDA-graph decode path from this PR work on
gpt-oss. Throughput measured on B200 with `unsloth/gpt-oss-20b-BF16`:

| bs | HF naive | flex walker (with capture) |
|---:|---------:|---------------------------:|
|  8 |    258.9 |                      832.1 |
| 32 |        — |                     1595.3 |
| 48 |        — |                     2146.4 |

3.2x over HF naive at bs=8, coherent greedy output matches HF bitwise
(24/24 tokens).

Full context + unsloth-side PR:
https://github.com/unslothai/unsloth/pull/5123#issuecomment-4301971384

danielhanchen:
Shipped the same \`return_dict\` kwargs-collision fix to this branch (\`d33e0b6\`) that landed on unslothai/unsloth-zoo#603. Affects the Gemma 4 MoE training patch only — the UNSLOTH_RETURN_HIDDEN_STATES branch of the patched \`Gemma4ForConditionalGeneration.forward\` passed \`return_dict=True\` explicitly alongside \`**kwargs\`, which collided the moment TRL's GRPOTrainer supplied its own \`return_dict\`:

    TypeError: got multiple values for keyword argument 'return_dict'

Fix is to pop \`return_dict\` from kwargs before forwarding.

**Re-verified GRPO 5-step smoke** on \`unsloth/gemma-4-26b-a4b-it\` after the fix (flex backend, \`max_steps=5\`, \`num_generations=2\`, \`seed=3407\`):

| Metric | Value |
|---|---|
| Steps | 5/5 complete |
| train_runtime | 1119.2s |
| Peak VRAM | 116.5 GB |
| Loss | -2.9e-14 → 0 (no NaN) |
| KL | -2.9e-11 → 0 (no NaN) |
| Rewards | -7.5 constant |

Loss/KL being ~0 on these first steps is expected — PEFT initializes \`B=0\` so the LoRA is a numerical identity and policy == reference, hence KL=0 and the reward-baseline loss=0.

danielhanchen:
Extended this branch to also carry the Qwen3.5 / Qwen3.6 vLLM fast_inference + LoRA serving path, plus cherry-picked the Gemma-4 vLLM LoRA patches from #603 so both sets are active together for end-to-end validation.

**New commits** on top of the MoE cudagraph/grouped-mm work:

- `35667e2` `load_vllm: honor explicit max_num_seqs on vision models for text-only` (from #603)
- `0790452` `gemma4_moe_lora_convert: expand stacked Unsloth MoE LoRA to per-expert` (from #603)
- `a5d41f7` `convert_vllm_to_huggingface: skip missing norm attrs on shared-KV layers` (from #603)
- `61acd62` `Gemma 4 vLLM fast_inference: get_expert_mapping + tracing + warmup fixes` (from #603)
- `165b854` `qwen3_5 / qwen3_5_moe vLLM fast_inference + LoRA serving support` (new)
- `3582637` `gemma4 + qwen3_5 MoE vLLM LoRA: config proxy + rotary reinit skip` (new)

Happy to rebase the #603 cherry-picks off this branch before merge if you'd rather keep the PRs separated.

### What the two new commits do

**165b854 — `patch_qwen3_5_vllm_lora_support`** mirrors `patch_gemma4_vllm_lora_support`. Qwen3.5-MoE hits the exact same vLLM MoE LoRA path: `vllm.lora.utils.process_packed_modules_mapping` calls `get_expert_mapping` on the model and its first-level children during `LoRAModelManager.__init__`, BEFORE `patched_create_lora_manager` gets a chance. So we install the method on both:
- `Qwen3_5MoeForCausalLM.get_expert_mapping` — builds per-expert params via `FusedMoE.make_expert_params_mapping` with `gate_proj` / `up_proj` / `down_proj` ckpt names.
- `Qwen3_5MoeForConditionalGeneration.get_expert_mapping` — delegates to `self.language_model.get_expert_mapping()`.
- Dense `Qwen3_5ForConditionalGeneration` gets an empty-list stub (no experts).

The `create_lora_manager` wrapper declares `vllm_config` as a named parameter so that `unsloth_zoo.vllm_lora_worker_manager._call_create_lora_manager`'s signature inspection (`"vllm_config" in sig.parameters`) routes the value through. A generic `(model, *args, **kwargs)` wrapper silently drops `vllm_config` into `**kwargs` and then the original `create_lora_manager` (vLLM's, which requires vllm_config positionally) blows up.

Also touches:
- `vllm_utils.py :: _get_vllm_state_dict` — skip attention extraction for `linear_attn` (Gated DeltaNet) layers and MLP extraction for MoE blocks. The vLLM engine retains the real weights; the HF shell is only used for LoRA adapter construction.
- `vllm_utils.py :: approximate_vllm_memory_usage` — fall back to `moe_intermediate_size` on qwen3_5_moe_text.
- `empty_model.py :: set_additional_modules` — dereference `new_model.model.language_model` for Qwen3_5's `Qwen3_5Model` multimodal container (text model is one level deeper than Gemma 4's).

**3582637 — Gemma-4 MoE vLLM LoRA follow-ups** (two small fixes that surfaced when I enabled Gemma-4 MoE LoRA serving end-to-end):
- `temporary_patches/gemma4.py :: _Gemma4KVSharedSafeProxy.__setattr__` / `__delattr__` — forward attribute writes to the wrapped config. vLLM's `VllmConfig.with_hf_config` does `hf_config.get_text_config().tie_word_embeddings = tie_word_embeddings`, and the `__slots__`-only proxy previously rejected the write (`no attribute ... and no __dict__ for setting new attributes`). `_real` is still routed via `object.__setattr__` in `__init__` / deepcopy.
- `vllm_utils.py :: convert_vllm_to_huggingface` — skip `module.rotary_emb = module.rotary_emb.__class__(config=...)` for gemma4/gemma4_text. Gemma4TextRotaryEmbedding stores `rope_parameters` as a nested `{layer_type: {...}}` dict and its `__init__` expects a `layer_type` kwarg; without it it falls into `self.config.rope_parameters["rope_type"]` and raises KeyError. The rotary buffer is already loaded correctly from vLLM weights.

### Validation

End-to-end LoRA train (fast_inference=False, 5 steps, r=16, attn-only q/k/v/o_proj) + serve (fast_inference=True + `LoRARequest`) passes on:
- `Qwen/Qwen3.5-4B` (qwen3_5) — outputs differ from baseline
- `Qwen/Qwen3.6-35B-A3B` (qwen3_5_moe) — LoRARequest accepts adapter cleanly
- `unsloth/gemma-4-26B-A4B-it` (gemma4 MoE) — LoRARequest accepts adapter after `scripts/strip_lora_vision.py` post-processes out the PEFT-auto-matched `vision_tower.*` entries

Throughput on B200, bf16, enforce_eager, greedy 64 new tokens:

| Model | bs=1 | bs=4 | bs=8 |
|---|---|---|---|
| Qwen3.5-4B | 22.8 tok/s | 23.0 | 22.7 |
| Qwen3.6-35B-A3B | 7.9 | 8.8 | 8.7 |
| Gemma-4-26B-A4B | 13.3 | 13.6 | 13.3 |

Flex-Gemma4-MoE path from the earlier commits in this branch was also re-verified (scatter_add_ capture-safe grouped_mm path, CUDA graph capture at bs=1/2/4, output matches HF naive on all 3 prompts × 16 tokens).

Companion PR: `unslothai/unsloth#5123` adds `qwen3_5` / `qwen3_5_moe` to `VLLM_SUPPORTED_VLM` and forces `enforce_eager=True` for the same FX splitter bug as gemma4.

unsloth_zoo/temporary_patches/qwen3_moe.py
```diff
@@ -69,11 +71,31 @@ def _qwen_moe_lora_extractor(wrapper, weight_A, weight_B, scaling, num_experts):
...
-    @torch.compiler.disable
+    # The legacy expert-loop path (``forward_native_moe_loop``) uses
+    # ``torch.where`` + a Python for-loop over experts and is genuinely
+    # uncapturable — keep ``@torch.compiler.disable`` when that backend
+    # is selected. On ``grouped_mm`` with pre-merged LoRA the sparse
+    # block is Dynamo-clean (no ``.item()``, no Python loops, no
+    # mutation), so we omit the decorator and let ``torch.compile``
+    # trace through the whole block for prefill.
+    _backend = select_moe_backend()
+    _compile_safe = _backend == "grouped_mm"
```
Datta0, line 97: Ideally we should do this compile checks in moe_utils.py so that every moe file benefits from it

danielhanchen:
Closing this one.

Follow-up benchmarking showed the CUDA-graph / torch.compile path for the grouped_mm MoE gives only about 1.03-1.04x in practice, and the CUDA-graph capture is not compatible with gradient checkpointing (it raises a checkpoint error on the recompute path), so it is not worth carrying as-is.

The genuine MoE training win, eliminating the per-step copy of the frozen base expert stack inside the grouped_mm path, is handled in unsloth-zoo #838 instead (about 1.39x on the 16-bit expert-LoRA step, bit-exact). The transformers<5 ModuleList grouped path is covered by #837.

This branch has also drifted from main. Happy to revisit a compile/CUDA-graph pass later if it can be made checkpoint-safe.

PR #611, opened 2026-04-24
gpt-oss MXFP4: cross-version loader patch for transformers 4.x + 5.x

danielhanchen:
## Summary

- Restores MXFP4 loading for `unsloth/gpt-oss-20b` on transformers 5.x: the 4.56.x function-based `load_and_swizzle_mxfp4` was replaced by the `WeightConverter`-based `Mxfp4Deserialize`, which silently skips when the checkpoint key names already match the registered parameter names. MXFP4 models end up with raw `gate_up_proj_blocks` / `_scales` on the module and the first forward raises inside the property fallback.
- Restores dequantize-path correctness on transformers 5.x: `Mxfp4Dequantize` returns `gate_up_proj` in `(E, 2I, H)` and `down_proj` in `(E, H, I)`, but stock `GptOssExperts.forward` expects `(E, H, 2I)` / `(E, I, H)`. The transpose was baked into `convert_moe_packed_tensors` in 4.x and was dropped from the 5.x path.
- Adds a Hopper-only gate in `_process_model_before_weight_loading`: if any visible CUDA device is non-Hopper, force `Mxfp4Config.dequantize = True`. Without this, T4 / A100 / B200 users crash inside `triton_kernels.matmul_ogs` with `Only Hopper swizzling is supported`.

## Approach

Monkey-patch `Mxfp4HfQuantizer` via two wrapping hooks, guarded by idempotence flags set on the class:

- `_process_model_after_weight_loading` always calls the original first (preserves `torch.cuda.empty_cache()` and any future upstream logic), then walks the model to either transpose `GptOssExperts` weights into the stock layout, or call `swizzle_mxfp4_convertops` on `Mxfp4GptOssExperts` modules that are still holding raw blocks/scales. Under 4.x the walker is a no-op because `load_and_swizzle_mxfp4` already fired during weight load.
- `_process_model_before_weight_loading` inspects every visible CUDA device and forces `dequantize=True` when any is non-Hopper, and also when CUDA is unavailable.

Per-projection shape checks guard against future transformers releases producing either weight already in the correct orientation. Failure modes (`triton_kernels` missing, `swizzle_mxfp4_convertops` missing, swizzle error) raise with actionable error messages instead of leaving the model silently unrunnable.

## GPU coverage

| Arch | Compute capability | Path |
|---|---|---|
| T4 | sm_75 | dequantize (bf16) |
| A100 | sm_80 | dequantize (bf16) |
| H100 | sm_90 | native MXFP4 (triton_kernels) |
| B200 | sm_100 | dequantize (bf16) |

## Verification

Tested on NVIDIA B200 (sm_100) with both transformers versions. Output matches byte-for-byte across both versions. Greedy decode, 32 new tokens, chat template.

## Test plan

- [x] transformers 4.57.6 greedy parity vs coherent output
- [x] transformers 5.5.4 greedy parity vs coherent output
- [x] B200 (sm_100) routes through dequantize path
- [x] Idempotence: running `patch_gpt_oss()` twice in the same process is a no-op
- [ ] H100 (sm_90) native MXFP4 path sanity (no hardware access; covered by the idempotence no-op on 4.x and the existing `HAS_TRITON_KERNELS` gate)
- [ ] T4 / A100 smoke (no hardware access; Hopper gate covers them by forcing dequantize)


danielhanchen:
Correction pushed: dropped the Hopper-only gate entirely.

triton_kernels actually supports MXFP4 on every target:
- `tensor_details/layout.py:21-28` picks `BlackwellMXValueLayout` on sm_100, `HopperMXValueLayout` on sm_90, and `StridedLayout` (no swizzle) on older archs
- `matmul_ogs_details/_matmul_ogs.py:114` asserts `SWIZZLE_MX_VALUE == "HOPPER_VALUE" or SWIZZLE_MX_VALUE is None`, so the strided path passes on T4/A100

The B200 "Only Hopper swizzling is supported" crash I originally attributed to a Hopper-only kernel was actually the 5.x WeightConverter skip leaving weights unswizzled. Once `swizzle_mxfp4_convertops` runs (via the post-load walker in the first commit), the native path works on every arch.

Kept only the CPU-only fallback. Re-verified on B200 with transformers 5.5.4: native MXFP4 produces the same byte-identical coherent output as the dequantize path.

Updated GPU coverage table:

| Arch | Compute capability | Path |
|---|---|---|
| T4 | sm_75 | native MXFP4, StridedLayout |
| A100 | sm_80 | native MXFP4, StridedLayout |
| H100 | sm_90 | native MXFP4, HopperMXValueLayout |
| B200 | sm_100 | native MXFP4, BlackwellMXValueLayout |
| CPU-only | - | forced dequantize to bf16 |

danielhanchen:
**Possible duplicate of a trusted maintainer's PR.** This PR looks like it solves the same underlying problem as [unslothai/unsloth-zoo#471](https://github.com/unslothai/unsloth-zoo/pull/471) by @Datta0 (trusted maintainer).

> Same file and GPT-OSS MXFP4 loading/dequantization compatibility; broader MoE work but includes the target loader repair area.

Canonical PR summary: This PR fixes GPT-OSS MoE support in Unsloth Zoo by adding grouped_mm-based LoRA inference/training paths, MXFP4 dequantization, GRPO hidden-state returns, and compatibility patches for compiler imports, expert/router patching, and attention forward flows.

The auto-review is still running against this PR — reviewers will factor in the canonical above. If this PR is genuinely different, call out the delta in the review discussion so the maintainer can decide which to merge.

danielhanchen:

PR repairs gpt-oss MXFP4 loading on transformers 5.x by adding post-load layout/swizzle walkers and a CPU-only dequantize gate to Mxfp4HfQuantizer, restoring correct gate_up_proj/down_proj orientation (including the gpt-oss-20b H==I square case) and avoiding crashes on hosts without CUDA/XPU.

**Reason:** All accepted findings fixed across two iterations; remaining iter-3 findings were all rejected (re-raises of prior rejections, hypothetical edge cases, or contradicted by upstream). Tests pass.

danielhanchen:
Confirmed main's gpt_oss.py only patches the 4.x load_and_swizzle_mxfp4 path, so on transformers 5.x gpt-oss MXFP4 still lands with raw blocks and scales and the wrong gate_up_proj orientation. Will get this reviewed.

danielhanchen:
Closing this as superseded. Re-checked against current main and current transformers, and all three things this PR fixes are now handled elsewhere.

**1. The 5.x dequantize transpose is fixed in main by #1114.** That PR patches `transformers.integrations.mxfp4.dequantize_convertops` in `unsloth_zoo/temporary_patches/mxfp4.py`, which is the real choke point on 5.x, rather than repairing the layout after load the way the walker here does.

Measured on `tiny-random/gpt-oss-mxfp4` with transformers 5.17.0 on one GPU. With hidden_size 32 and intermediate_size 64, the correct stock layout is gate_up `(E, 32, 128)` and down `(E, 64, 32)`:

| build | gate_up_proj | down_proj | logits mean |
|---|---|---|---|
| stock transformers | (32, 32, 128) | (32, 64, 32) | 0.035183172672986984 |
| zoo main | (32, 32, 128) | (32, 64, 32) | 0.035183172672986984 |
| zoo main + this PR | (32, 32, 128) | (32, 64, 32) | 0.035183172672986984 |

Byte identical. Applied on top of main this PR is inert: the walker's outer guard looks for the wrong `(2I, H)` layout, main already produces `(H, 2I)`, so it never fires.

**2. The `Mxfp4Deserialize` skip described here does not happen on any released 5.x.** Checked `integrations/mxfp4.py` at v5.0.0, v5.5.0, v5.10.0, v5.16.0 and installed 5.17.0. In every one of them `Mxfp4GptOssExperts` registers `gate_up_proj` and `down_proj` as packed uint8 parameters, while the converter's `source_patterns` are `gate_up_proj_blocks` and `gate_up_proj_scales` against target `gate_up_proj$`. Source and target never coincide, so the converter always runs, and `Mxfp4Deserialize.convert` calls `swizzle_mxfp4_convertops` eagerly and in place. The premise the post-load swizzle walker was built on does not hold.

**3. The CPU-only gate is upstream's job now.** `Mxfp4HfQuantizer._process_model_before_weight_loading` has forced `dequantize = True` on CPU-only hosts since at least v5.5.0, which is zoo's current transformers cap, alongside a non cuda/xpu/cpu accelerator arm. The `_patched_before_load` hook here duplicates it. The Hopper gate was already withdrawn in the second commit on this branch.

What I did not exercise on this host: the native, non-dequantize MXFP4 path needs `kernels` and `triton_kernels`, neither of which is installed, so transformers fell back to bf16 in all three runs above. That path is off by default in zoo (it needs `UNSLOTH_MXFP4_NO_DEQUANTIZE=1`), and point 2 covers it by source reading rather than by a live run.

Merging this would add roughly 190 lines of monkey-patch that provably do nothing on current transformers, so closing.

PR #615, opened 2026-04-28
fix: use backend device type in GGUF merge path

andomeder:
Use `DEVICE_TYPE` / `DEVICE_TYPE_TORCH` in `unsloth_zoo/saving_utils.py` instead of assuming CUDA directly in the GGUF merge/export path.

This updates:
- `_merge_lora` to move tensors with `torch.device(DEVICE_TYPE_TORCH, index)`
- post-merge cache clearing to use the active backend
- Tesla T4 detection to run only on `DEVICE_TYPE == "cuda"`

This was failing on Intel XPU during export with:

```text
Torch not compiled with CUDA enabled
```


rolandtannous:
@andomeder instead of doing this why not import `from .device_type import DEVICE_TYPE` and you can check if DEVICE_TYPE='xpu' 
An example how it's used is in the https://github.com/unslothai/unsloth-zoo/blob/d5874e8b6a1f8232549f7c20d1b46dc303485ca6/unsloth_zoo/vllm_utils.py file

andomeder:
@rolandtannous Good point, revised it to use DEVICE_TYPE / DEVICE_TYPE_TORCH instead of direct backend probes. I also restricted the T4 check to real CUDA only.

danielhanchen:

PR removes hardcoded CUDA assumptions in the GGUF LoRA merge and export pipeline, routing tensor moves and cache/sync calls through DEVICE_TYPE-aware helpers (device_empty_cache, device_synchronize, device_is_bf16_supported, _active_merge_device) so Intel XPU and HIP builds no longer crash with 'Torch not compiled with CUDA enabled'. CUDA/HIP behavior is bit-identical to pre-PR code; XPU now correctly dispatches through torch.xpu equivalents with defensive hasattr guards across all three helpers.

**Reason:** All P1/P2 backend-portability bugs in the GGUF merge/export path are fixed; remaining nits are cosmetic or pre-existing upstream content explicitly out of scope.

rolandtannous:
Tested to ensure this does not negatively impact merging and gguf export of models on colab, T4 specifically and local VM.

Test on colab T4:



colab T4 test
colab T4 test 2
colab T4 test 3


Test on local VM (B200):
export test B200 1
export test B200 2
export test B200 3
export test B200 4


Looks good. merging

PR #620, opened 2026-05-04
feat(mlx): Apple Silicon training (text + VLM, LoRA / full FT, CCE, export)

Manan17:
## Summary                                                                                                                                                                      
                                                                                                                                                                                
  End-to-end training on Apple Silicon (M1-M5) via MLX. Same `FastModel.from_pretrained` / `get_peft_model` API as the GPU path; drop in `MLXTrainer` instead of `SFTTrainer` and 
  the rest of the notebook is unchanged.
                                                                                                                                                                                  
## What's included                                                                                                                                                            

### Loader (`mlx_loader.py`)
  - `FastMLXModel.from_pretrained` auto-detects text vs VLM, dispatches to `mlx-lm` / `mlx-vlm`. Aliases: `FastLanguageModel = FastModel = FastMLXModel`.
  - `FastMLXModel.get_peft_model` with `finetune_language_layers / attention_modules / mlp_modules / vision_layers` flags + `target_modules='all-linear'` support.                
  - Runtime 4-bit / 8-bit / fp8 / mxfp4 quantization (`load_in_4bit` etc.) and `full_finetuning=True` for full-parameter FT.                                                      
  - Save / push: `save_pretrained_merged`, `push_to_hub_merged`, `save_pretrained_gguf` with public/private toggle, `save_method` mapping (LoRA-only / merged 16-bit / merged     
  4-bit / GGUF).                                                                                                                                                                  
  - Adapter-driven base load: walks adapter keys to reproduce LoRA placement on reload, fixes VLM (Qwen3-VL) export.                                                              
                                                                                                                                                                                  
### Trainer (`mlx_trainer.py`)                                                                                                                                                
  - `MLXTrainer` + `MLXTrainingConfig` (HF-compatible field names where reasonable).                                                                                              
  - Adafactor / AdamW / Adam / SGD / Lion / Muon. Linear / cosine / constant LR.                                                                                                  
  - Gradient checkpointing on both language and vision tower layers.                                                                                                              
  - `mx.compile` policy with auto-fallback to eager on compile failure.                                                                                                           
  - LoRA+, embedding LR scaling, `train_on_responses_only`, gradient accumulation.                                                                                                
  - bf16/fp16-preserving optimizer updates (avoids the fp32 grad promotion that 3xs memory and forces optimizer state to fp32).                                                   
  - Default-on **elementwise gradient clipping** (`max_grad_value`, PyTorch `clip_grad_value_` semantics): auto-picks ±5 for bf16, ±1 for fp16. Free memory (no cross-leaf        
  reduction). Catches localized blow-ups (LayerNorm explosions, attention head collapse) without the +13-22 GB cost of global norm clipping.                                      
                                                                                                                                                                                  
### CCE (`mlx_cce/`)                                                                                                                                                            
  - MLX-native chunked cross-entropy with inline Metal shaders — no C++ build.                                                                                                  
  - Matches GPU `cut-cross-entropy` semantics. Saves ~4 GB per step at long sequences for big-vocab models (Gemma3 256K, Qwen 152K).                                              
  - Skips lm_head weight-grad path automatically when frozen (LoRA case).                                                                                                         
                                                                                                                                                                                  
### Compile patches (`mlx_compile.py`)                                                                                                                                          
  - Per-arch patches and verifiers for Llama, Qwen2/3, Qwen3-VL, Qwen3.5-MoE, Qwen3-VL-MoE, Gemma3, Gemma4, Phi-4, Mistral, Paligemma, Moondream3, Qwen2.5-VL.                    
  - `safe_fused_sdpa_mask` patch, arch-specific attention/vision/embedding fixes.                                                                                                 

### Utilities (`mlx_utils.py`)                                                                                                                                                  
  - Dynamic-padding batchers for text + VLM (`create_batches`, `create_vlm_batches`).                                                                                           
  - Loss factories: `make_cce_loss_fn` / `make_baseline_loss_fn` for text and VLM.                                                                                                
  - Vision tower image-size resolver that handles dataclass `vision_config` and skips area-count size dicts (fixes Qwen2.5-VL OOM on small inputs).                               
                                                                                                                                                                                  
## Memory benchmarks                                                                                                                                                            
                                                                                                                                                                                  
  Llama-3.2-3B-bf16 LoRA, BS=8, seq=2048:                                                                                                                                         
  - Baseline: ~32 GB                                                                                                                                                            
  - + CCE: ~18 GB (45% saving)

  VLM with vision-tower training (our trainer vs `mlx-vlm`):                                                                                                                      
  - Gemma3-4B-it LoRA + vision: ~15 GB (`mlx-vlm` doesn't actually train the vision tower)
  - Qwen2.5-VL-3B LoRA + vision: ~10 GB                                                                                                                                           
  - Qwen3-VL-2B LoRA + vision: ~6 GB                                                                                                                                              
  - Gemma3-4B-it full FT (every parameter): ~40 G


danielhanchen:
**Possible duplicate of a trusted maintainer's PR.** This PR looks like it solves the same underlying problem as [unslothai/unsloth#4258](https://github.com/unslothai/unsloth/pull/4258) by @danielhanchen (trusted maintainer).

> Trusted PR implements core unsloth FastMLXModel/MLX trainer surface that pairs with target zoo MLX backend, despite different repo files.

Canonical PR summary: This PR adds initial Apple Silicon MLX support through a new FastMLXModel API, MLX LoRA training helpers, MPS device detection, and a macOS mlx/mlx-lm dependency extra. It does this by gating CUDA/Triton-specific exports on MPS, adding MLX model loading/inference/training utilities, and wiring MLX tuner-based LoRA optimization.

The auto-review is still running against this PR — reviewers will factor in the canonical above. If this PR is genuinely different, call out the delta in the review discussion so the maintainer can decide which to merge.

danielhanchen:
**Possible duplicate of a trusted maintainer's PR.** This PR looks like it solves the same underlying problem as [unslothai/unsloth#4258](https://github.com/unslothai/unsloth/pull/4258) by @danielhanchen (trusted maintainer).

> Trusted core Unsloth FastMLXModel/MLX trainer API counterpart for the same Apple Silicon MLX training feature.

Canonical PR summary: This PR adds initial Apple Silicon MLX support through a new FastMLXModel API, MLX LoRA training helpers, MPS device detection, and a macOS mlx/mlx-lm dependency extra. It does this by gating CUDA/Triton-specific exports on MPS, adding MLX model loading/inference/training utilities, and wiring MLX tuner-based LoRA optimization.

The auto-review is still running against this PR — reviewers will factor in the canonical above. If this PR is genuinely different, call out the delta in the review discussion so the maintainer can decide which to merge.

danielhanchen (changes requested):
## Review: feat(mlx): Apple Silicon training (text + VLM, LoRA / full FT, CCE, export)

This is a substantial and well-structured addition. The compile-qualification system is impressively thorough, the CCE Metal kernels are numerically correct, and the VLM patch coverage is broad. However there are four blocking issues and several should-fix issues before merge.

---

### [P1] Blocking

**1. Silent MLX activation when torch is also installed**

`unsloth_zoo/__init__.py` currently gates on `_is_mlx_only` (which is True when `mlx` is importable), but does NOT check whether `torch` is also present. On macOS ARM64 with both packages installed (a common developer setup), the MLX path fires silently and replaces the entire GPU stack without warning.

```python
# Current check (insufficient):
if _is_mlx_only:
    ...

# Fix: also require torch to be absent, OR add an explicit opt-in env var:
import importlib.util as _ilu
if _is_mlx_only and _ilu.find_spec("torch") is None:
    ...
# or
if _is_mlx_only and os.environ.get("UNSLOTH_USE_MLX", "0") == "1":
    ...
```

Without this gate any user who has both libraries installed (e.g. doing cross-framework experiments) will silently lose the CUDA/Metal-via-torch path.

---

**2. `@mx.custom_function` closure inside `gated_delta_ops_efficient` causes recompilation every step**

`unsloth_zoo/gated_delta_vjp.py` defines `_chunked_forward` as a nested closure inside `gated_delta_ops_efficient`:

```python
def gated_delta_ops_efficient(x, A, B, C, dt, mask=None, ...):
    @mx.custom_function
    def _chunked_forward(x, A, B, C, dt):
        ...
    return _chunked_forward(x, A, B, C, dt)
```

MLX's `@mx.custom_function` uses object identity (the function's `id()`) as the cache key. Because `_chunked_forward` is a new closure object on every call to `gated_delta_ops_efficient`, the gradient cache is invalidated on every training step, recompiling both the forward and backward graphs each time.

Fix: promote `_chunked_forward` to module level and pass `mask` as an explicit tensor parameter (or fold it into the `x` argument):

```python
# Module level:
@mx.custom_function
def _chunked_forward(x, A, B, C, dt, mask):
    # use mask directly
    ...

def gated_delta_ops_efficient(x, A, B, C, dt, mask=None, ...):
    if mask is None:
        mask = mx.ones(x.shape[:2], dtype=mx.bool_)
    return _chunked_forward(x, A, B, C, dt, mask)
```

---

**3. `position_ids` shape `(3, 1, L)` in Qwen3.5 attention patch -- wrong for batch size > 1**

`unsloth_zoo/mlx_utils.py` `_fix_qwen35_attention_cache` does:

```python
pos_ids = mx.tile(pos_ids.reshape(1, L), (3, 1, 1))  # -> (3, 1, L)
```

This broadcasts correctly when batch size is 1 but silently produces wrong position encodings for any `B > 1` during training (the single batch row is shared across all B positions). Fix:

```python
# pos_ids shape is (B, L) coming in
B, L = pos_ids.shape
pos_ids = mx.tile(pos_ids[None, :, :], (3, 1, 1))  # -> (3, B, L)
```

---

**4. `@mx.compile` on `safe_scaled_dot_product_attention` uses unsupported `**kwargs`**

`unsloth_zoo/mlx_utils.py`:

```python
@mx.compile
def safe_scaled_dot_product_attention(q, k, v, **kwargs):
    ...
```

MLX's `mx.compile` does not support `**kwargs` in the compiled function signature -- the compiler traces concrete tensor shapes but cannot handle a variadic keyword dict. This will either silently recompile on every call (defeating the purpose) or raise at runtime when the kwargs set varies.

Fix: drop `@mx.compile` from this function. The correctness of the masking logic is unchanged; the compile here is a latency micro-optimization that does not hold for a function with variable keyword arguments.

---

### [P2] Should fix before merge

**5. Duplicated quantization helpers across `mlx_loader.py` and `mlx_utils.py`**

`_dequantize_layer`, `_quantize_layer`, and `_infer_snapshot_commit` are defined identically in both files (verified by grep). Keep the canonical copy in `mlx_utils.py` and import from there in `mlx_loader.py`.

---

**6. `add_special_tokens=False` in Gemma3/4 VLM processor**

`unsloth_zoo/mlx_utils.py` `_apply_gemma3_chat_template` (and the Gemma4 variant) passes `add_special_tokens=False` to the tokenizer, but the Gemma3/4 `AutoProcessor` injects BOS via the processor, not via the chat template. With `add_special_tokens=False` the BOS token is dropped, which shifts all token positions by -1 and degrades convergence.

Fix: remove `add_special_tokens=False` (use the default `True`) or explicitly prepend BOS after calling the tokenizer if the processor does not add it.

---

**7. `lru_cache(maxsize=1)` thrash in `discover_compile_protocol_targets`**

The cache is called with a fresh `tuple(model.named_modules())` on every training step. Because each call produces a new tuple object (even if the module list is identical), `maxsize=1` evicts on every call, providing zero caching benefit. Either increase `maxsize` or stabilize the cache key to the model's class name + module count.

---

**8. In-place mutation of positional argument in `_patch_mlx_sdpa`**

`_patch_mlx_sdpa` modifies `attn_mask` in-place (`attn_mask += ...`) when `attn_mask` is a view of an upstream tensor. This can corrupt the caller's buffer. Use `attn_mask = attn_mask + ...` (out-of-place) to be safe.

---

### [P3] Minor / polish

- `unsloth_zoo/__init__.py` MLX path does not `del os, warnings, re` after use (the GPU path does). Add the same cleanup to avoid leaking these names into the module namespace.
- `unsloth_zoo/mlx_trainer.py` `MLXTrainer.save_model`: `_lora_scale = getattr(m, "scale", 1.0)` will capture an attention head-scale if the first module with `lora_a` is also an attention layer. Prefer `getattr(m, "lora_scale", getattr(m, "scale", float(_lora_rank)))` and walk until a module that has both `lora_a` and `lora_b`.

---

### Summary

| Severity | Count | Items |
|----------|-------|-------|
| P1 Blocking | 4 | Silent MLX activation, custom_function closure, position_ids shape, mx.compile + kwargs |
| P2 Should fix | 4 | Duplicate helpers, add_special_tokens, lru_cache thrash, in-place arg mutation |
| P3 Minor | 2 | Namespace leak, lora_scale lookup |

Please address the four P1 items before merge. The P2 items can follow in a close-up PR if preferred, but ideally in the same sprint.

danielhanchen (changes requested):
## Review: feat(mlx) Apple Silicon training stack

Two P1 issues block correct behavior in non-trivial configurations and should be fixed before merge. Three P2 issues should also be resolved. Three P3 items are cleanup.

---

### P1 -- Compile fallback catches only `ValueError`; MLX raises `RuntimeError` / `MLXError`

**File:** `unsloth_zoo/mlx_trainer.py`, `_train_inner`

When `mx.compile` fails at runtime (e.g. "primitive-less array" errors), MLX raises `RuntimeError` or `mlx.core.MLXError`, not `ValueError`. The current guard:

```python
except ValueError as e:
    if _use_compile and "eval" in str(e).lower():
```

silently propagates any non-`ValueError` compile exception as a crash, bypassing the eager fallback even when `_compile_decision.fallback_allowed = True`.

Suggested fix:

```python
except Exception as e:
    _is_compile_error = (
        _use_compile
        and (
            "eval" in str(e).lower()
            or "primitive" in str(e).lower()
            or type(e).__name__ in ("MLXError", "RuntimeError")
        )
    )
    if not _is_compile_error:
        raise
    if not _compile_decision.fallback_allowed:
        raise RuntimeError(
            "Unsloth: strict mx.compile was enabled for this VLM "
            "and runtime fallback is disabled."
        ) from e
    print("Unsloth: mx.compile failed at runtime; falling back to eager mode.")
    step_fn = _uncompiled_step_fn
    _use_compile = False
    state = [model.state, optimizer.state, mx.random.state]
    lvalue, toks, grad_accum_state = step_fn(batch_data, grad_accum_state, do_update)
```

---

### P1 -- `patched_qwen3_fast_pos_embed_interpolate` calls `.tolist()` inside `mx.compile`

**File:** `unsloth_zoo/mlx_compile.py`, `_install_qwen3_family_compile_patches`

The patch calls `.tolist()` on MLX arrays that are live inside a compiled function:

```python
for i in range(4):
    idx_list[i].extend(indices[i].tolist())   # forces eager eval inside mx.compile
    weight_list[i].extend(weights[i].tolist())
idx_tensor = mx.array(idx_list, dtype=mx.int32)
weight_tensor = mx.array(weight_list, dtype=...)
```

This either raises a "primitive-less array" error or causes the compiled function to retrace on every batch with a different image grid size, which is every batch in variable-resolution training. Both `qwen3_vl` and `qwen3_vl_moe` are listed in `_VERIFIED_TRAINING_ARCHES`, so this will affect the common case.

The fix is to build the four-corner index and weight tensors using pure MLX array ops (`mx.stack`, `mx.tile`, `mx.concatenate`) rather than materializing them into Python lists. The `mx.linspace`/floor/ceil/outer-product structure can be expressed entirely in MLX without any Python round-trips.

---

### P2 -- `full_finetuning` quantized-base guard silently bypassed when config load fails

**File:** `unsloth_zoo/mlx_loader.py`, `FastMLXModel.from_pretrained`

The guard:

```python
if full_finetuning and _get_existing_mlx_quantization(config_data):
    raise ValueError(...)
```

is only reached when `config_data` is non-empty. The config load is wrapped in a broad `except (FileNotFoundError, json.JSONDecodeError, KeyError)` that silently sets `config_data = {}`. For gated models, network errors, or malformed JSON, the guard passes vacuously and the user starts training a quantized model with `full_finetuning=True`.

Defer the check to after the model is loaded, where quantization state is directly inspectable from the live model config rather than the pre-download `config.json`.

---

### P2 -- Metal memory limits set in `_configure_memory_limits` but never restored

**File:** `unsloth_zoo/mlx_trainer.py`, `MLXTrainer.train`

`mx.set_memory_limit`, `mx.set_cache_limit`, and `mx.set_wired_limit` are process-global. They are applied in `_configure_memory_limits()` but the `finally` block in `train()` only removes gradient checkpointing hooks. Any code in the same process after training (e.g. inference in a Jupyter notebook) inherits the training-time limits.

Since MLX does not expose getters for these limits, the simplest fix is to restore them to effectively unlimited values in the `finally` block when Unsloth was responsible for setting them:

```python
finally:
    if args.gradient_checkpointing:
        remove_gradient_checkpointing(model)
    if self._memory_limits_applied and mx.metal.is_available():
        try:
            mx.set_memory_limit(2**63 - 1)
            mx.set_cache_limit(2**63 - 1)
            mx.set_wired_limit(0)
        except Exception:
            pass
```

---

### P2 -- `push_to_hub_merged` uses `api.upload_folder` instead of `upload_large_folder`

**File:** `unsloth_zoo/mlx_utils.py`, `push_to_hub_merged`

The project instructions explicitly require `HfApi.upload_large_folder` for any non-trivial folder upload; `upload_folder` lacks resume and chunking and fails mid-upload for large or many-file pushes. A merged 7B+ model in sharded safetensors will hit this reliably on slow connections.

```python
# replace upload_folder(...) with:
api.upload_large_folder(
    folder_path=str(save_directory),
    repo_id=repo_id,
    repo_type="model",
    private=bool(private) if private is not None else False,
)
```

---

### P3 -- `_SKIP_GPU_INIT` leaks into the public module namespace

**File:** `unsloth_zoo/__init__.py`

Every other temporary variable used during platform detection (`_is_mlx_only`, `_check_platform`, `find_spec`, `os`, `warnings`, `re`) is deleted at module end. `_SKIP_GPU_INIT` is not. Add `del _SKIP_GPU_INIT` after both branches.

---

### P3 -- Four helper functions duplicated verbatim across `mlx_loader.py` and `mlx_utils.py`

`_infer_snapshot_commit`, `_effective_mlx_quantization_map`, `_quantization_config_to_path_map`, `_canonical_mlx_quantization_path` are identical in both files. Move them to a shared `_mlx_common.py` and import from both files to prevent future divergence.

---

### P3 -- Dead function `_install_qwen3_get_input_embeddings_patch` never registered

**File:** `unsloth_zoo/mlx_compile.py`

The function exists but is absent from `_runtime_patch_primitive_installers`. The relevant patch is already applied by `_install_qwen3_family_compile_patches` (which is registered). The orphan function is never called and should be removed.

---

### Tests recommended before merge

- `python -c "import unsloth_zoo"` on macOS arm64 with MLX and on Linux (GPU path must be unaffected).
- Single-step training smoke for `qwen3_vl` with `compile=True` and variable-resolution images -- verifies the `.tolist()` P1 fix does not retrace.
- Compile-fallback test: mock a `RuntimeError` from `step_fn` and confirm the trainer falls back to eager rather than crashing.
- `push_to_hub_merged` with a sharded model -- confirms `upload_large_folder` handles resume correctly.
- `full_finetuning=True` with a gated/offline repo where config download fails -- confirms the guard still fires after the load.

danielhanchen:
**Possible duplicate of a trusted maintainer's PR.** This PR looks like it solves the same underlying problem as [unslothai/unsloth#4258](https://github.com/unslothai/unsloth/pull/4258) by @danielhanchen (trusted maintainer).

> Trusted core MLX model/trainer implementation with FastMLXModel, LoRA training, MPS detection, and dependencies matching target backend scope.

Canonical PR summary: This PR adds initial Apple Silicon MLX support through a new FastMLXModel API, MLX LoRA training helpers, MPS device detection, and a macOS mlx/mlx-lm dependency extra. It does this by gating CUDA/Triton-specific exports on MPS, adding MLX model loading/inference/training utilities, and wiring MLX tuner-based LoRA optimization.

The auto-review is still running against this PR — reviewers will factor in the canonical above. If this PR is genuinely different, call out the delta in the review discussion so the maintainer can decide which to merge.

danielhanchen:

PR adds an end-to-end Apple Silicon MLX training backend (text + VLM, LoRA / full FT, runtime CCE, GGUF / hub export) mirroring the GPU FastModel API. After the review's correctness fixes (VLM label expansion, partial-batch handling, prompt-mask edge case, GatedDelta chunk masks, GGUF dequantize / VLM mmproj, mlx-lm load signature, vision/projector LoRA reload, fused-QKV filter), the PR is ready to merge.

**Reason:** All P1 correctness fixes landed; remaining items are minor and addressed.

PR #623, opened 2026-05-05
Fix Gemma-4 E-series 4bit init crash and register MoE LoRA extractor

danielhanchen:
## Summary

Two independent Studio crashes on Gemma-4. Both fixed in this PR.

### Bug 1: `unsloth/gemma-4-E2B-it-unsloth-bnb-4bit` and `unsloth/gemma-4-E4B-it-unsloth-bnb-4bit` crash on load with

```
RuntimeError: "normal_kernel_cuda" not implemented for 'Byte'
```

The E-series uses partial KV sharing across upper layers, so the checkpoint omits `self_attn.{k_proj,v_proj,k_norm}.weight` for those layers and `lm_head.weight`. transformers v5 materialises the missing slots with `torch.empty_like(param)` against the bnb-4bit `Params4bit`, which preserves the `quant_storage=torch.uint8` dtype. `Gemma4PreTrainedModel._init_weights` then dispatches to `init.normal_(module.weight)` on a uint8 tensor, which has no float kernel.

Fix: wrap `Gemma4PreTrainedModel._init_weights` to skip any module whose primary weight tensor is non-floating-point. Placeholders are filled from the state_dict immediately afterwards, or they belong to KV-shared layers that the runtime never reads.

New file: `unsloth_zoo/temporary_patches/gemma4_init_weights.py`. One line added to `unsloth_zoo/temporary_patches/__init__.py`.

### Bug 2: `unsloth/gemma-4-26B-A4B-it` crashes on the first training step with

```
RuntimeError: contraction dimension of mat_a and mat_b must match
```

raised from `torch._grouped_mm` in `moe_utils.native_moe_grouped_mm`. The gemma4_moe patch only patches `Gemma4TextExperts.forward` but never registers `_unsloth_lora_extractor_fn`. `_extract_lora_from_wrapper` falls through to the default branch which does not handle PEFT 0.19+ swapped 3D LoRA layouts.

Sibling MoE families that already register the extractor: `qwen3_moe`, `qwen3_5_moe`, `qwen3_next_moe`, `qwen3_vl_moe`, `glm4_moe`, `deepseek_v3_moe`.

Fix: register `_make_qwen_moe_lora_extractor()` on `Gemma4TextExperts` and the legacy `Gemma4TextMoEBlock`. Gemma-4 experts share the Qwen-MoE standard `(E, out, in)` layout and the same `hidden_dim` / `intermediate_dim` attribute names, so the Qwen extractor applies verbatim. `per_expert_scale` is folded into `top_k_weights` upstream by `Gemma4TextRouter.forward`, so no Gemma-4-specific handling is needed inside the extractor.

### Compatibility

Both patches gate on `from transformers.models.gemma4.modeling_gemma4 import ...` and no-op cleanly when Gemma-4 is not present, so transformers 4.57.6 is unaffected. Validated against transformers 5.5.0. The Qwen extractor handles PEFT 0.18 and 0.19+ via its existing `_did_swap_in_out_features` dispatch, so TRL 0.22.2 / 0.27.1 / 1.0.0 all work. Both patches are idempotent.

## Test plan

- [x] Load `unsloth/gemma-4-E2B-it-unsloth-bnb-4bit` via `FastVisionModel.from_pretrained` with `load_in_4bit=True`. Before this PR this crashed with the Byte init error. After this PR the load succeeds, the missing KV-shared slots stay uint8 placeholders, and the runtime never invokes them.
- [x] Load `unsloth/gemma-4-26B-A4B-it` with LoRA on attention plus `experts.gate_up_proj` and `experts.down_proj`, run one forward + backward. Before this PR this crashed with the contraction-dimension error on the first step. After this PR it returns a finite loss and a clean grad.
- [x] Confirm no regression for `unsloth/gemma-4-31B-it` (dense, did not hit either bug before).
- [x] Confirm sibling MoE families (Qwen3 variants, GLM4, DeepSeek-V3) are unchanged.

### DIFF
--- a/unsloth_zoo/temporary_patches/__init__.py
+++ b/unsloth_zoo/temporary_patches/__init__.py
@@ -20,6 +20,7 @@ from .gemma import *
 from .misc import *
 from .gemma3n import *
 from .gemma4 import *
+from .gemma4_init_weights import *
 from .gpt_oss import *
 from .qwen3_moe import *
 from .qwen3_vl_moe import *
--- a/unsloth_zoo/temporary_patches/gemma4_init_weights.py
+++ b/unsloth_zoo/temporary_patches/gemma4_init_weights.py
@@ -1,0 +1,151 @@
+# Unsloth Zoo - Utilities for Unsloth
+# Copyright 2023-present Daniel Han-Chen, Michael Han-Chen & the Unsloth team. All rights reserved.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+"""
+Skip non-float weight init for Gemma-4 E-series MatFormer 4-bit checkpoints.
+
+The E2B and E4B Gemma-4 unsloth-bnb-4bit checkpoints intentionally omit upper
+layer K, V and K-norm projections plus the tied lm_head. Those layers run in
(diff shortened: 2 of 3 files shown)


danielhanchen:
Reopening with the Gemma-4 MoE LoRA extractor fix only. The E-series 4bit init crash needs a different approach and will follow as a separate PR.

PR #625, opened 2026-05-06
Aliases for qwen3-vl name map

Datta0:
Fixes: https://github.com/unslothai/unsloth/issues/5290

### DIFF
--- a/unsloth_zoo/saving_utils.py
+++ b/unsloth_zoo/saving_utils.py
@@ -2837,6 +2837,22 @@ def _infer_prefix_and_remap(lora_weights, safetensor_keys):
         if (k + ".weight") in sf_key_set:
             remapped[k] = v
             continue
+        aliased_key = None
+        for source_prefix, target_prefix in (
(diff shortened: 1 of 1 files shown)


danielhanchen:
Thanks for this. Checking it against current main: the hardcoded alias block here targets key layouts (`visual.`, `model.`) that don't match how real Qwen3-VL checkpoints are stored on disk (they use `model.language_model.*` and `model.visual.*`), so for actual Qwen3-VL the guard `candidate + ".weight" in sf_key_set` never matches and the block is effectively dead code. Current main already resolves Qwen3-VL keys through the rewritten `_infer_prefix_and_remap` / direct-match path.

Separately, the linked issue #5290 is a Qwen2.5-VL export problem that flows through the `_checkpoint_conversion_mapping` (forward_mapping) path, which this change doesn't touch, so it wouldn't resolve #5290 either. Closing this as superseded; #5290 stays open and we're looking at it separately. Thanks for digging in.

PR #628, opened 2026-05-07
tests: skip MoE LoRA extractor coverage when discovery finds zero classes

danielhanchen:
## Summary

`test_every_patched_moe_experts_class_has_lora_extractor` (PR #624) is
designed to prevent a regression where a class flagged
`_unsloth_already_patched=True` is missing `_unsloth_lora_extractor_fn`.
Its defensive precondition was "discovery produced at least one patched
class"; failing that, the test raised an AssertionError telling the
reader to investigate.

That precondition fires on any transformers version older than the MoE
class surface `unsloth_zoo.temporary_patches` targets. With
`transformers==4.57.6` for example, none of the gemma4 / llama4 /
qwen3_moe / mxfp4 classes the `TEMPORARY_PATCHES` list patches exist,
so every patch fn no-ops and discovery returns `[]`. The "zero
discoveries" outcome on an older transformers is **not** a regression
of the bug being prevented; it just means the relevant classes aren't
there to patch.

Replace `assert patched, ...` with `pytest.skip(...)` and a comment
explaining the two indistinguishable cases. CI cells running newer
transformers (where the classes do exist) still exercise the real
assertion below; the older-transformers cell skips cleanly.

## Test plan

- [x] Locally: with the workspace's pyproject-pinned transformers (5.x
      cap), discovery still produces patched classes and the real
      assertion runs.
- [ ] CI on unslothai/unsloth#5312 cell `transformers==4.57.6 + trl<1`
      should drop the failure on this test on the next run.

## Why this isn't `pytest.mark.skipif(transformers_version < X.Y)`

A version-string skipif would have to encode every minimum version
across the `TEMPORARY_PATCHES` list (gemma3, gemma4, llama4, qwen3_moe,
mxfp4 each landed in a different transformers release) AND track
upstream renames. The "zero patched classes" runtime check is the
precise condition we actually care about: skip when nothing is there
to verify.

Surfaced by the consolidated CPU-CI matrix on unslothai/unsloth#5312.

### DIFF
--- a/tests/test_moe_lora_extractor_coverage.py
+++ b/tests/test_moe_lora_extractor_coverage.py
@@ -302,11 +302,27 @@ def test_every_patched_moe_experts_class_has_lora_extractor():
     _apply_all_temporary_patches()
 
     patched = _discover_patched_moe_classes()
-    assert patched, (
-        "Discovery produced zero patched MoE classes. Either no MoE "
-        "transformers module imported on this transformers version, or the "
-        "_unsloth_already_patched marker convention changed. Investigate."
-    )
+    if not patched:
+        # Two situations produce zero discoveries:
+        #   1. The transformers version installed predates the MoE class
+        #      surface unsloth_zoo patches (e.g. transformers 4.57.6 ships
+        #      none of the gemma4 / llama4 / qwen3_moe / mxfp4 classes the
+        #      TEMPORARY_PATCHES list targets, so every patch fn no-ops).
+        #   2. The `_unsloth_already_patched` marker convention drifted
+        #      (the regression this test guards against).
+        # We can't tell (1) and (2) apart by discovery alone. Skip rather
+        # than fail: a CI cell pinned to an older transformers should not
+        # surface as a red regression here. CI cells running the
+        # transformers version the PR actually pins (and any newer one)
+        # will discover patched classes and run the real assertion.
+        pytest.skip(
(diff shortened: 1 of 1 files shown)


danielhanchen:
Pushed `4c464c8` addressing the codex P2 / gemini medium feedback.

The previous unconditional skip created a silent-failure path: a real discovery regression on a transformers that DOES ship the grouped-MoE surface would have skipped rather than failed, which is exactly the regression this test was written to catch.

The new logic walks `transformers.models.*` once and asks whether ANY class declares `gate_up_proj` + `down_proj` as 3D `nn.Parameter`s — the exact filter `_looks_like_grouped_moe_experts` runs against discovered classes. If at least one such class exists AND discovery returned zero, that's a real regression: fail loud with a message naming up to 5 example classes. If no such class exists anywhere in transformers, the patch surface this test targets is genuinely absent on this transformers version: skip.

This is the structural canary the gemini reviewer suggested, generalized so it doesn't hard-code model names that change as transformers grows new MoE families.

Also corrected the comment + messages to refer to the actual marker convention the discovery logic reads (`_original_<modeling_tail>_<ClassName>_forward`) rather than the per-family `_unsloth_already_patched` flag.

Verified locally on transformers 4.57.6: discovery finds `Qwen3VLMoeTextExperts` (which shipped grouped-MoE 3D params before the rest of the family) and the real assertion runs.

PR #645, opened 2026-05-14
fix embedding matrix size mismatch bug

CodeMan62:
This PR is fixing the embedding matrix size mismatch bug
closes [3502](https://github.com/unslothai/unsloth/issues/3502)
image shows that the patch works 
image


danielhanchen:
Thanks for the PR - appreciate it!

PR #647, opened 2026-05-15
saving: layout-aware MoE LoRA merge + loud-fail on fallback (#5410)

danielhanchen:
## Summary

`save_pretrained_merged(..., save_method=\"merged_16bit\")` silently drops the entire MoE expert LoRA delta on Qwen3-MoE / Qwen3.5-MoE-style models when running on `peft >= 0.19.1` + `transformers >= 5.0` (reported in unslothai/unsloth#5410). Four issues stack up:

1. The per-expert helpers in `unsloth_zoo/saving_utils.py` hardcode the **PEFT 0.18 \"swapped\" layout** (`lora_A: (E*r, 2I)`, `lora_B: (H, E*r)` for fused gate_up_proj; `lora_A: (E*r, H)`, `lora_B: (I, E*r)` for fused down_proj). **PEFT 0.19+** swaps in/out features for non-transposed 3D parameters (huggingface/peft#2521) and produces the opposite shapes.
2. On layout mismatch an `addmm` shape error was swallowed by a bare `try / except Exception: return W` (and the dim-heuristic in the fused helpers fell through to `return W`), so the merge wrote the **unmodified base weight** and reported success.
3. On the dense `_merge_and_overwrite_lora` flow the per-expert merge loop's `num_experts` came from the **shard-local key scan**, which can be a non-divisor of `total_rank` whenever experts are split across multiple safetensor shards (16 / 17 in some shards of the 128-expert Qwen3-30B-A3B layout).
4. The merged dir was missing `generation_config.json`, so chat-tuned models reloaded with default eos / sampling and ran past EOS.

## Fix

- `_detect_moe_lora_layout(lora_A, lora_B, num_experts, out_dim, in_dim)` classifies the layout by shape against the per-expert on-disk weight. No version sniffing, so it works on transformers 4.57.x / 5.x and peft 0.18.x / 0.19.x.
- `_merge_moe_gate_or_up_expert` and `_merge_moe_down_proj_expert` branch on the detected layout. The PEFT 0.18 \"swapped\" path is byte-identical to the previous behaviour.
- `_resolve_num_experts_from_lora_stats(lora_stats, fallback)` walks `module -> base_layer -> ...` to read the authoritative `num_experts` off the wrapped MoE module (`Qwen3MoeExperts` and similar). `_merge_and_overwrite_lora` calls it for every MoE LoRA in `converted_lora_weights` and overrides the shard-local value in `moe_num_experts` before the per-expert loop runs.
- `_MOE_MERGE_STATE` tracks `(attempted, applied, fallback, first_error)`. Helpers record fallbacks with role / expert / shapes / reason on any unrecognised layout or exception. After the shard loop `merge_and_overwrite_lora` raises `RuntimeError` if any fallback fired, so a partially merged checkpoint can no longer be silently written. On success it prints `applied/attempted`.
- The `merged_16bit` branch also calls `model.generation_config.save_pretrained(save_directory)` (best-effort, matching the same pattern as `fix_tokenizer_config_json`).

## Tests

`tests/test_unsloth_zoo_lora_merge.py` now covers:

- The existing 16 cases (PEFT 0.18 swapped layout, per-expert + fused + dense) still pass byte-for-byte.
- 6 new cases:
  - PEFT 0.19+ \"standard\" layout for `_merge_moe_gate_expert`, `_merge_moe_up_expert`, `_merge_moe_down_proj_expert`.
  - `_detect_moe_lora_layout` for swapped, standard, mismatched shapes, and non-divisor `num_experts`.
  - The fallback counter increments and `first_error` populates on unrecognised shapes.
  - `_resolve_num_experts_from_lora_stats` walks the `base_layer` chain (covers the inner ParamWrapper case for `mlp.experts.down_proj` where the outer ParamWrapper has `module = None`).

## End-to-end verification

Full Qwen3-30B-A3B (128 experts x 48 layers, fused 3D in memory, per-expert 2D on disk): load -> attach LoRA (`r=32, alpha=64, target_modules=[q,k,v,o,gate,up,down]_proj, lora_dropout=0`) -> 5 SFT steps -> `save_pretrained_merged(merged_16bit)` -> reload -> compare merged-reload logits to the trained in-memory model on a fixed eval batch.

| transformers | peft   | trl    | merged tensors | trained vs merged KL | samples | merged vs base KL | base vs trained KL |
|--------------|--------|--------|----------------|----------------------|---------|-------------------|--------------------|
| 5.5.0        | 0.19.1 | 0.25.1 | **18432 / 18432** | **1.6e-5**        | **3 / 3** | 17.81             | 17.85              |
| 5.5.0        | 0.18.1 | 0.25.1 | **18432 / 18432** | **1.3e-5**        | **3 / 3** | 18.91             | 19.05              |
| 4.57.6       | 0.19.1 | 0.25.1 | dense path        | **5.5e-5**        | **3 / 3** | 15.93             | 16.02              |
| 5.5.0        | 0.19.1 | 1.4.0  | **18432 / 18432** | **2.1e-4**        | **3 / 3** | 15.41             | 15.31              |

Reading: `KL = O(1e-4)` between merged-reload and trained model is bf16 noise; `samples = 3 / 3` means all 3 greedy generations on held-out prompts match exactly; `merged vs base KL` approx `base vs trained KL` confirms the full training delta is baked into the saved merged dir.

Before this patch the first row was KL = 1.86, samples = 1 / 3, **0 / 18432** expert LoRA deltas applied.

Notes:

- `transformers 4.57.6` has `Qwen3MoeSparseMoeBlock.experts = nn.ModuleList(Qwen3MoeMLP)` per expert (no fused 3D parameter), so the MoE merge helpers do not fire and every per-expert Linear takes the standard dense `_merge_lora` path. The MoE helpers are unreachable on `transformers < 5`; the patch only affects the path that produces the bug.
- The `trl 1.4` row uses `padding_free=False` in the reproducer (TRL 1.x raises when `padding_free=True` is combined with a finite `max_length` without packing). Unrelated to this patch.
- Dense Qwen3-0.6B-Base sanity check on `transformers 5.5 + peft 0.19 + trl 0.25` separately: trained vs merged KL = 1.1e-4, top-1 agreement = 0.978, samples 4 / 4. The dense `_merge_lora` path is untouched.

## Test plan

- [x] `pytest unsloth-zoo/tests/test_unsloth_zoo_lora_merge.py` -> 22 passed
- [x] Full Qwen3-30B-A3B end-to-end on all four version combinations in the table
- [x] Confirm `generation_config.json` is written into `save_directory` for `merged_16bit`
- [x] Confirm `_MOE_MERGE_STATE` raises a clear `RuntimeError` when the layout is unrecognised (synthetic test + a deliberately mangled shape fixture)

## Related

- Fixes unslothai/unsloth#5410
- Likely fixes unslothai/unsloth#4832 (same author, same \"garbage after save_pretrained_merged reload\" symptom on DevStral Small 2)
- Same class of failure recurs at unslothai/unsloth#2428 (canonical), #3454, #3428, #3547, #1519, #1832
- Complements #629 (Datta0's training-time MoE extractor fix). The training-time grouped-mm path and the save-time per-expert path are separate codepaths; #629 fixed training, this fixes save.

### DIFF
--- a/tests/test_unsloth_zoo_lora_merge.py
+++ b/tests/test_unsloth_zoo_lora_merge.py
@@ -342,14 +342,9 @@ def test_merge_moe_fused_down_proj_expert(is_transposed):
     torch.testing.assert_close(out.to(torch.float32).cpu(), expected, atol=1e-1, rtol=1e-1)
 
 
-# ---------------------------------------------------------------------------
-# 8. PEFT 0.19+ "standard" (non-swapped) layout coverage for per-expert helpers
-#    (issue #5410). PEFT 0.19 swaps in/out features for non-transposed 3D
-#    parameters, so lora_A / lora_B come out flipped relative to PEFT 0.18.
-# ---------------------------------------------------------------------------
+# 8. PEFT 0.19+ standard layout (#5410).
 
 def test_merge_moe_gate_expert_standard_layout():
-    """Standard layout: lora_A (E*r, H), lora_B (2I, E*r); take first I rows of B."""
     torch.manual_seed(SEED)
     num_experts, rank_per, inter_dim, hidden_dim = 4, 4, 8, 12
     total_rank = num_experts * rank_per
@@ -366,9 +361,9 @@ def test_merge_moe_gate_expert_standard_layout():
         output_dtype=torch.bfloat16,
     )
     s, e = expert_idx * rank_per, (expert_idx + 1) * rank_per
-    a_slice = lora_A[s:e].to(torch.float32)              # (r, H)
-    b_slice = lora_B[:, s:e].to(torch.float32)           # (2I, r)
-    delta = b_slice[:inter_dim, :] @ a_slice             # (I, H)
+    a_slice = lora_A[s:e].to(torch.float32)
+    b_slice = lora_B[:, s:e].to(torch.float32)
+    delta = b_slice[:inter_dim, :] @ a_slice
     expected = gate_W.to(torch.float32) + alpha * delta
     torch.testing.assert_close(out.to(torch.float32).cpu(), expected, atol=1e-1, rtol=1e-1)
 
@@ -392,13 +387,12 @@ def test_merge_moe_up_expert_standard_layout():
     s, e = expert_idx * rank_per, (expert_idx + 1) * rank_per
     a_slice = lora_A[s:e].to(torch.float32)
     b_slice = lora_B[:, s:e].to(torch.float32)
-    delta = b_slice[inter_dim:, :] @ a_slice             # second half of B
+    delta = b_slice[inter_dim:, :] @ a_slice
     expected = up_W.to(torch.float32) + alpha * delta
     torch.testing.assert_close(out.to(torch.float32).cpu(), expected, atol=1e-1, rtol=1e-1)
 
 
 def test_merge_moe_down_proj_expert_standard_layout():
-    """Standard layout: lora_A (E*r, I), lora_B (H, E*r); delta = B @ A directly."""
     torch.manual_seed(SEED)
     num_experts, rank_per = 4, 4
     total_rank = num_experts * rank_per
@@ -406,8 +400,8 @@ def test_merge_moe_down_proj_expert_standard_layout():
     alpha = 8.0
 
     down_W = torch.randn(H, I, dtype=torch.bfloat16)
-    lora_A = torch.randn(total_rank, I, dtype=torch.bfloat16) * 0.05  # A.shape[1] = I = in_dim
-    lora_B = torch.randn(H, total_rank, dtype=torch.bfloat16) * 0.05  # B.shape[0] = H = out_dim
+    lora_A = torch.randn(total_rank, I, dtype=torch.bfloat16) * 0.05
+    lora_B = torch.randn(H, total_rank, dtype=torch.bfloat16) * 0.05
     expert_idx = 3
 
     out = _merge_moe_down_proj_expert(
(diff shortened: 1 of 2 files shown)


danielhanchen:
Sandbox simulation report against this branch (HEAD `97bb267`):

**Three isolated `uv venv` matrix points**, each running the existing 16 merge tests + the 6 new layout / detection / fallback tests + a 52-case synthetic sim suite covering layout detection edges, `_resolve_num_experts` edges (including self-referential and 2-cycle `base_layer` chains, and a module whose `__getattribute__` raises), per-expert math in both layouts for all 4 expert indices, ambiguous-dim `2I == H` / `H == I` cases, state-machine reset and `first_error` stickiness, dense `_merge_lora` regression, `generation_config.save_pretrained` best-effort, cross-platform path handling, and a mini E2E mock merge:

| venv | peft | transformers | pytest tests | sim cases |
|---|---|---|---|---|
| `peft018_tfm550` | 0.18.1 | 5.5.0 | 22 / 22 PASS | 52 / 52 PASS |
| `peft019_tfm550` | 0.19.1 | 5.5.0 | 22 / 22 PASS | 52 / 52 PASS |
| `peft019_tfm4576` | 0.19.1 | 4.57.6 | 22 / 22 PASS | 52 / 52 PASS |

The simulation also caught one latent risk before the merge could land: the `_resolve_num_experts_from_lora_stats` walk was an unbounded `while module is not None`. PEFT does not self-reference in practice but a hostile / cyclic `base_layer` chain would hang the merge. The follow-up commit `97bb267` bounds the walk to 16 hops, dedupes via an `id()` set, and swallows exceptions on attribute access so a module that raises on getattr cannot abort the merge. The 22 + 52 results above are after that fix.

End-to-end on Qwen3-30B-A3B across the four (transformers, peft, trl) combinations in the PR description remains unchanged: 18432 / 18432 expert tensors merged, trained vs merged KL between 1.3e-5 and 2.1e-4, greedy samples 3 / 3.

PR #661, opened 2026-05-17
Fix GGUF converter sibling imports

alkinun:
Fixes #5495

## Summary
- Add the cached converter script directory to sys.path before executing it.
- Cover sibling imports such as llama.cpp conversion.py with a focused regression test.

## Root cause
llama.cpp convert_hf_to_gguf.py imports sibling helper modules by bare name. Unsloth caches the script under ~/.unsloth/llama.cpp and executes it by path, but that directory was not on sys.path, so Kaggle could fail with ModuleNotFoundError: No module named conversion.

## Validation
- pytest tests/test_llama_cpp_loader.py

### DIFF
--- a/tests/test_llama_cpp_loader.py
+++ b/tests/test_llama_cpp_loader.py
@@ -1,0 +1,39 @@
+import sys
+import textwrap
+import importlib.util
+from pathlib import Path
+
+
+def _load_llama_cpp_module():
+    repo_root = Path(__file__).resolve().parents[1]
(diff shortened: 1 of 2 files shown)


alkinun:
Addressed the review feedback by restoring sys.path after executing the cached converter module. The regression test now also snapshots/restores sys.path and asserts the temporary converter directory is not left in sys.path after the load completes.

Validation: pytest tests/test_llama_cpp_loader.py

danielhanchen:
Thanks this works!

Issue #662, opened 2026-05-17
MLX trainer: max_grad_value default silently overrides max_grad_norm, breaks HF parity


danielhanchen:
## Summary

After [PR #634](https://github.com/unslothai/unsloth-zoo/pull/634) the MLX trainer sets `MLXTrainingConfig.max_grad_value = 5.0` (originally 1.0 in the same PR) and at training-config-resolution time silently zeroes out a user-supplied `max_grad_norm` when both are non-zero. This breaks HuggingFace/TRL parity for the MLX path. A fine-tune that converges and emits a sensible greedy completion under `transformers.SFTTrainer` on CUDA produces gibberish on MLX given identical hyperparameters.

## Repro

Identical 7-step LoRA on `unsloth/gemma-3-270m-it`, train row = `\"<<HELLO!!>> My name is Unsloth!\"`, `bs=2`, `grad_accum=3`, `lr=1e-3`, `lr_scheduler=constant`, `warmup=0`, `optim=adamw`, `weight_decay=0`, `max_seq=64`, `seed=3407`, LoRA `r=8` on `q/k/v/o`.

| Run | clip mode | step 1 → 7 loss | greedy completion of `\"<<HELLO!!>> My name is \"` |
|---|---|---|---|
| CUDA (torch+TRL) | `max_grad_norm=1.0` only | 7.64 → 1.16 | `\" 1! ... My name is Unsloth! ...\"` (contains \"Unsloth\") |
| CUDA (torch+TRL) | elementwise `clip_grad_value_(1.0)` only | 7.64 → 1.19 | `\" 1! What are you doing?! ...\"` (no \"Unsloth\") |
| MLX (post-#634)  | default (`max_grad_value=1.0`/`5.0` overrides `max_grad_norm`) | 10.55 → 0.10 | `'5 lbs!'` |
| MLX (post-#634)  | user sets `max_grad_value=0` so `max_grad_norm=1.0` wins | 10.55 → 0.17 | `'5 lbs!'` |
| MLX (pre-#634, last green @ unsloth `12295c1f`) | trainer default | 10.55 → 5.04 (non-monotone) | `\" Unsloth!\\n\\nMy name is Unsloth! ...\"` (contains \"Unsloth\") |

The CUDA mirror script lives at `temp/torchcodec_test/cuda_mirror.py` in my local workspace; results JSON in the same dir. MLX numbers come from the `MLX CI on Mac M1` workflow on `unslothai/unsloth`:

[...]

PR #669, opened 2026-05-18
fix(mlx): expose finetune_last_n_layers for parity with mlx-lm CLI

danielhanchen:
## Summary
- Adds `finetune_last_n_layers` parameter to `FastMLXModel.get_peft_model` (default `None` = all layers, current behavior unchanged).
- Wired into both VLM and text-only code paths.
- When set, applies LoRA only to the last N transformer blocks (matching mlx-lm CLI's `CONFIG_DEFAULTS['num_layers']=16` semantics at `mlx_lm/lora.py:56`).
- Companion PR in `unsloth/unsloth` exposes the same knob on the CUDA path so a single config value controls layer-selection across CUDA / MLX / mlx-lm CLI.

## Why
mlx-lm CLI defaults `num_layers=16` -> LoRA on the LAST 16 transformer blocks. unsloth-zoo's `get_peft_model` historically applied LoRA to ALL transformer layers (matching HF PEFT/CUDA semantics).

On small models the difference can show up as a basin-selection divergence: the extra LoRA modules consume `mx.random` state during init and change the trainable-parameter set, so two otherwise-identical runs land in different basins of attraction.

Empirical (n=15 seeds, gemma-3-270m-it single-row LoRA memorization fixture):
- mlx-lm CLI default last-16 layers: 67% greedy-decode pass rate
- unsloth-zoo defaults (all 18 layers): 47% greedy-decode pass rate
- Teacher-forced completion loss is `0` in both — the model memorizes either way; only the first-token argmax distribution differs.

This PR keeps the default behavior unchanged (None = all layers) so existing users see no change. Passing `finetune_last_n_layers=16` puts the run in the same basin family as mlx-lm CLI for direct comparisons.

The value is clamped to `[1, len(model.model.layers)]` so callers can't accidentally request more layers than the model has, or zero layers (which would freeze everything).

## Test plan
- [x] New test `tests/test_mlx_finetune_last_n_layers.py` covering:
  - parameter exists with default None
  - `None` -> num_layers = total
  - explicit value -> num_layers = value
  - value > total -> clamped to total
  - value <= 0 -> clamped to 1
- [ ] Empirical MLX smoke (already validated on `danielhanchen/unsloth-staging-2` MLX parity probe matrix — probe 31 with num_layers=16 hits 10/15 = 67% matching mlx-lm CLI per-seed).

### DIFF
--- a/tests/test_mlx_finetune_last_n_layers.py
+++ b/tests/test_mlx_finetune_last_n_layers.py
@@ -1,0 +1,149 @@
+# Unsloth Zoo - Utilities for Unsloth
+# Copyright 2023-present Daniel Han-Chen, Michael Han-Chen & the Unsloth team. All rights reserved.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+"""Tests for `finetune_last_n_layers` in FastMLXModel.get_peft_model.
+
+mlx-lm CLI's CONFIG_DEFAULTS sets num_layers=16 (lora.py:56), meaning
+LoRA is applied to the last 16 transformer blocks only. unsloth-zoo's
+get_peft_model historically applied LoRA to ALL transformer layers,
+matching HF PEFT/CUDA semantics. The two paths could produce different
+trained models on the same fixture (different basin) because (a) the
+extra layers contribute extra LoRA modules whose lora_a init consumes
+mx.random state, shifting init for the later layers, and (b) the
+trainable-parameter set differs.
+
+`finetune_last_n_layers` lets users opt into mlx-lm CLI semantics
+without changing the existing default (which remains None = all layers).
+"""
+
+from __future__ import annotations
+
+import inspect
+
+import pytest
+
+
+@pytest.fixture(autouse=True, scope="module")
+def _install_shim():
+    from mlx_simulation import simulate_mlx_on_torch
+    simulate_mlx_on_torch()
+
+
+def test_get_peft_model_has_finetune_last_n_layers_param():
+    """The parameter exists and defaults to None (= all layers)."""
+    from unsloth_zoo.mlx.loader import FastMLXModel
+    sig = inspect.signature(FastMLXModel.get_peft_model)
+    assert "finetune_last_n_layers" in sig.parameters
+    assert sig.parameters["finetune_last_n_layers"].default is None
+
+
+def test_get_peft_model_passes_finetune_last_n_layers_through():
+    """When finetune_last_n_layers is set, linear_to_lora_layers is
+    called with that num_layers value.
+
+    Patches mlx_lm.tuner.utils.linear_to_lora_layers to record the
+    num_layers it sees. The test runs against a tiny synthetic text
+    model (no real layers needed -- the value of num_layers passed is
+    what we assert, not the side effects on a real architecture).
+    """
+    import sys
+    import unsloth_zoo.mlx.loader as loader_mod
+
+    # Build a minimal text-only fake model with .model.layers of len=8.
+    class FakeLayer: pass
+    class FakeInner:
+        layers = [FakeLayer() for _ in range(8)]
+    class FakeModel:
+        model = FakeInner()
+        _unsloth_full_finetuning = False
+        _is_vlm_model = False
+        def freeze(self): pass
+        def unfreeze(self, **kwargs): pass
+
+    # Capture num_layers values seen by linear_to_lora_layers.
+    captured = {"calls": []}
+    def fake_linear_to_lora_layers(model, num_layers, config, use_dora=False):
+        captured["calls"].append(num_layers)
+
+    # Stub out the helpers get_peft_model uses internally so the test
+    # doesn't need to walk a real model tree.
+    import unsloth_zoo.mlx.loader as L
+    L._fix_missing_no_grad = lambda m: None
+    L._resolve_lora_keys = lambda m, t: [
+        "model.layers.0.self_attn.q_proj",
+        "model.layers.1.mlp.gate_proj",
+    ]
+    L._apply_mlx_lora_initialization = lambda m, init: None
+    L.linear_to_lora_layers = fake_linear_to_lora_layers
+    # mlx_lm.tuner.utils is imported inside the function:
+    fake_mod = type(sys)("mlx_lm.tuner.utils")
+    fake_mod.linear_to_lora_layers = fake_linear_to_lora_layers
+    sys.modules["mlx_lm.tuner.utils"] = fake_mod
+
+    # Case 1: default (None) -> all 8 layers
+    captured["calls"].clear()
+    loader_mod.FastMLXModel.get_peft_model(
+        FakeModel(),
+        r=8, lora_alpha=16, lora_dropout=0.0,
+        target_modules=["q_proj", "gate_proj"],
+        finetune_language_layers=True,
+        finetune_attention_modules=True,
+        finetune_mlp_modules=True,
+        use_gradient_checkpointing=False,
+    )
+    assert captured["calls"] == [8]
+
+    # Case 2: finetune_last_n_layers=5 -> 5 layers
+    captured["calls"].clear()
+    loader_mod.FastMLXModel.get_peft_model(
+        FakeModel(),
+        r=8, lora_alpha=16, lora_dropout=0.0,
+        target_modules=["q_proj", "gate_proj"],
+        finetune_language_layers=True,
+        finetune_attention_modules=True,
+        finetune_mlp_modules=True,
+        finetune_last_n_layers=5,
+        use_gradient_checkpointing=False,
+    )
+    assert captured["calls"] == [5]
+
+    # Case 3: finetune_last_n_layers > actual layer count -> clamp to total
+    captured["calls"].clear()
+    loader_mod.FastMLXModel.get_peft_model(
+        FakeModel(),
+        r=8, lora_alpha=16, lora_dropout=0.0,
+        target_modules=["q_proj", "gate_proj"],
+        finetune_language_layers=True,
+        finetune_attention_modules=True,
+        finetune_mlp_modules=True,
+        finetune_last_n_layers=999,  # exceeds layer count
+        use_gradient_checkpointing=False,
+    )
+    assert captured["calls"] == [8]
+
+    # Case 4: finetune_last_n_layers=0 or negative -> clamp to 1
+    captured["calls"].clear()
+    loader_mod.FastMLXModel.get_peft_model(
+        FakeModel(),
+        r=8, lora_alpha=16, lora_dropout=0.0,
+        target_modules=["q_proj", "gate_proj"],
+        finetune_language_layers=True,
+        finetune_attention_modules=True,
+        finetune_mlp_modules=True,
+        finetune_last_n_layers=0,
+        use_gradient_checkpointing=False,
+    )
+    assert captured["calls"] == [1]
--- a/unsloth_zoo/mlx/loader.py
+++ b/unsloth_zoo/mlx/loader.py
@@ -2696,6 +2696,7 @@ class FastMLXModel:
         finetune_language_layers=True,
         finetune_attention_modules=True,
         finetune_mlp_modules=True,
+        finetune_last_n_layers=None,
         **kwargs,  # Accept and ignore GPU-only kwargs
     ):
         """Apply LoRA via mlx-lm on Apple Silicon.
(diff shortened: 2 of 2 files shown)


danielhanchen:
## Empirical bisection (gemma-3-270m-it, single-row LoRA memorization, n=15 seeds)

Filing the latest probe numbers here so reviewers can see the full picture this PR addresses vs what it does not.

| probe | loader | trainer | layers | dtype | greedy pass | cf_loss=0 |
|---|---|---|---|---|---|---|
| 20 | mlx-lm CLI (subprocess) | mlx-lm | last 16 | bf16 | **67%** | 15/15 |
| 31 | `mlx_lm.load` | manual loop | last 16 | bf16 | **67%** | 15/15 |
| 30 | `mlx_lm.load` | manual loop | all 18 | bf16 | 47% | 15/15 |
| 33 | `mlx_lm.load` | `MLXTrainer` | last 16 | bf16 | 53% | 15/15 |
| 34 | `FastMLXModel(dtype=None)` | `MLXTrainer` | last 16 | bf16 | 47% | 15/15 |
| 32 | `FastMLXModel(dtype="float16")` | `MLXTrainer` | last 16 | fp16 | 15% | 15/15 |

The `finetune_last_n_layers=16` knob is necessary but not sufficient to close the gap end-to-end through the full zoo stack. Three independent factors stack on top of the layer-selection issue:

1. **Layer selection** (this PR): `mlx_lm.load` + manual loop swings 47% -> 67% just by switching from all-18 to last-16 layers (probe 30 vs probe 31). This is what `finetune_last_n_layers` exposes.
2. **`MLXTrainer` vs manual loop** (-14pp): same loader, same layer count, different driver. Likely the extra `mx.eval` for monitoring + state ordering. Tracked separately.
3. **`FastMLXModel` loader patches** (-6 to -10pp): `_fix_missing_no_grad`, freeze ordering, etc. Tracked separately.
4. **bf16 -> fp16 dtype cast** (-28pp on Gemma3): `_convert_mlx_dtype` casts gemma3-270m's native bf16 storage to fp16 when `dtype="float16"` is passed. fp16 max ~6.5e4 vs bf16 max ~3.4e38; quietly lossy for Gemma3. Tracked separately.

**Why this PR is still the right first step:** the layer-selection mismatch is the only one of the four that is also a *user-facing semantics* difference, not just a numerical/perf overhead. mlx-lm CLI users who switch to `FastMLXModel.get_peft_model()` were previously silently training a different parameter set; the new knob makes that choice explicit and matches mlx-lm CLI when set.

**cf_loss safety net:** every config above hits teacher-forced completion loss == 0 in 15/15 seeds. The model memorizes either way; only the first-token greedy argmax distribution differs. CI smoke gating on `completion_teacher_forced_loss < 0.5` (per `unslothai/unsloth#5537`) is bulletproof regardless of which basin a given seed lands in.

Will file separate issues / PRs for (2), (3), (4).

danielhanchen:
## Final empirical summary (Round BO, 75 cells)

Five companion PRs landed in this MLX-vs-mlx-lm-CLI parity series:

- `#669` (this PR) $\to$ `finetune_last_n_layers` knob on `FastMLXModel.get_peft_model`.
- `#670` $\to$ warn on bf16$\to$fp16 downcast (Gemma3 silent precision loss).
- `#671` $\to$ `max_grad_value=None` honors disable + default to None for HF/TRL parity (closes `#662`).
- `#672` $\to$ `_create_labeled_batches` padding matches mlx-lm's `iterate_batches`.

Across Rounds BG-BO on `danielhanchen/unsloth-staging-2` (15 paired seeds on `unsloth/gemma-3-270m-it` single-row LoRA memorization fixture), teacher-forced completion loss is `0` in 15/15 seeds for every config tested. So the model memorizes the training row in every case. The greedy-decode-on-prompt pass rate varies between 40% and 67% across configs, but that variation tracks first-token argmax tie-breaking, not training quality. PR `unslothai/unsloth#5537`'s `cf_loss < 0.5` gate is the right CI gate and is bulletproof regardless.

What this PR specifically addresses: the most user-visible parity surface, where calling `FastMLXModel.get_peft_model()` previously applied LoRA to all transformer blocks, while `mlx_lm.lora` CLI's `CONFIG_DEFAULTS['num_layers']=16` applies LoRA to the last 16 blocks only. After this PR users can pass `finetune_last_n_layers=16` to mirror mlx-lm CLI semantics on the zoo path, with the CUDA companion in `unslothai/unsloth#5564` doing the same via PEFT's `layers_to_transform`.

unsloth_zoo/mlx/loader.py
```diff
@@ -2696,6 +2696,7 @@ def get_peft_model(
         finetune_language_layers=True,
         finetune_attention_modules=True,
         finetune_mlp_modules=True,
+        finetune_last_n_layers=None,
```
danielhanchen, line 2699: **[2/2 reviewers]** Med: the new parameter is added to the signature but not documented anywhere users can find it. The `get_peft_model` docstring just above this hunk does not mention `finetune_last_n_layers`, its meaning ("last N transformer blocks"), the clamp range, or the mlx-lm CLI parity intent.

```suggestion
        finetune_last_n_layers=None,  # mlx-lm CLI parity: LoRA on last N transformer blocks; default None = all layers
```

danielhanchen (commented):
Thank you for the PR! The goal of this PR is to give MLX callers a one-knob way to match mlx-lm CLI's `CONFIG_DEFAULTS['num_layers']=16` semantics (LoRA on the last N transformer blocks). As a summary, this PR adds an optional `finetune_last_n_layers` keyword to `FastMLXModel.get_peft_model` and, when set, clamps it via `max(1, min(int(N), total))` and passes it as `num_layers` to `linear_to_lora_layers` on both the VLM language path and the text path. Default `None` preserves the current all-layers behavior.

| 2/2 | Med | `get_peft_model` docstring is not updated, so the new parameter is invisible to `help(...)` and `inspect`-style discovery. |
| 2/2 | Med | `int(finetune_last_n_layers)` silently accepts `True`/`False`, floats (`1.7 -> 1`), and numeric strings, masking common user typos. |
| 2/2 | Med | Tests skip the VLM branch entirely — only the text-only call site (line 2905) is exercised. |
| 1/2 | Med | When the model lacks `.model.layers` (so `num_layers == 0`), `finetune_last_n_layers` is silently dropped with no warning. |
| 2/2 | Nit | The clamp logic is duplicated verbatim between the VLM and text branches; extract a `_resolve_num_layers(num_layers, finetune_last_n_layers)` helper. |
| 2/2 | Nit | The test monkeypatches `sys.modules['mlx_lm.tuner.utils']` (and `L._fix_missing_no_grad` etc.) without a teardown — leaks into later tests if execution order shifts. Prefer pytest's `monkeypatch` fixture. |

Overall: APPROVE_WITH_NITS.

unsloth_zoo/mlx/loader.py
```diff
@@ -2819,6 +2820,8 @@ def get_peft_model(
                 num_layers = 0
                 if hasattr(lm, "model") and hasattr(lm.model, "layers"):
                     num_layers = len(lm.model.layers)
+                if finetune_last_n_layers is not None and num_layers > 0:
+                    num_layers = max(1, min(int(finetune_last_n_layers), num_layers))
```
danielhanchen, line 2824: **[2/2 reviewers]** Med: `int(finetune_last_n_layers)` happily accepts `True` (→1), `False` (→0), floats (`1.7` → 1, truncating silently), and numeric strings — all common typos. A `True` literal silently becomes "last 1 layer" instead of raising. Add an explicit type/range guard with a clear message; consider extracting the whole clamp into a helper since the same lines appear at line 2906 too.

```suggestion
                if finetune_last_n_layers is not None and num_layers > 0:
                    if not isinstance(finetune_last_n_layers, int) or isinstance(finetune_last_n_layers, bool):
                        raise TypeError(
                            f"finetune_last_n_layers must be an int, got {type(finetune_last_n_layers).__name__}"
                        )
                    num_layers = max(1, min(finetune_last_n_layers, num_layers))
```

unsloth_zoo/mlx/loader.py
```diff
@@ -2819,6 +2820,8 @@ def get_peft_model(
                 num_layers = 0
                 if hasattr(lm, "model") and hasattr(lm.model, "layers"):
                     num_layers = len(lm.model.layers)
+                if finetune_last_n_layers is not None and num_layers > 0:
```
danielhanchen, line 2823: **[1/2 reviewers]** Med: when `hasattr(lm, 'model')` is False or `lm.model.layers` is empty, `num_layers` stays at 0 and the `num_layers > 0` guard silently drops the user's `finetune_last_n_layers` request. `linear_to_lora_layers` is then called with `num_layers=0`, which is a no-op. Surface this rather than silently ignoring the setting.

```suggestion
                if finetune_last_n_layers is not None:
                    if num_layers > 0:
                        num_layers = max(1, min(int(finetune_last_n_layers), num_layers))
                    else:
                        import warnings
                        warnings.warn(
                            "Unsloth: finetune_last_n_layers requested but the model does not expose .model.layers; ignoring.",
                            stacklevel=2,
                        )
```

unsloth_zoo/mlx/loader.py
```diff
@@ -2899,6 +2902,8 @@ def get_peft_model(
                 num_layers = 0
                 if hasattr(model, "model") and hasattr(model.model, "layers"):
                     num_layers = len(model.model.layers)
+                if finetune_last_n_layers is not None and num_layers > 0:
+                    num_layers = max(1, min(int(finetune_last_n_layers), num_layers))
```
danielhanchen, line 2906: **[2/2 reviewers]** Nit: this 2-line clamp is identical to the VLM branch above (line 2823-2824). Extract a tiny helper so the two paths cannot drift:

```suggestion
                num_layers = _resolve_finetune_last_n_layers(num_layers, finetune_last_n_layers)
```

(with a free function defined once near the top of `FastMLXModel`:
```python
def _resolve_finetune_last_n_layers(num_layers, n):
    if n is None or num_layers <= 0:
        return num_layers
    return max(1, min(int(n), num_layers))
```)

File: tests/test_mlx_finetune_last_n_layers.py [hunk 5]
```diff
@@ -0,0 +1,149 @@
...
+    what we assert, not the side effects on a real architecture).
+    """
+    import sys
+    import unsloth_zoo.mlx.loader as loader_mod
+
+    # Build a minimal text-only fake model with .model.layers of len=8.
+    class FakeLayer: pass
+    class FakeInner:
+        layers = [FakeLayer() for _ in range(8)]
+    class FakeModel:
```
danielhanchen, line 69: **[2/2 reviewers]** Med: the `FakeModel` here sets `_is_vlm_model = False`, so only the text-only call site (loader.py:2905) is exercised. The new clamp on line 2823 (VLM language path) has zero coverage even though the PR description says "wired into both VLM and text-only code paths". Add a second fixture that flips `_is_vlm_model = True` with a stubbed `language_model` so both call sites are pinned.

```suggestion
class FakeModel:
    model = FakeInner()
    _unsloth_full_finetuning = False
    _is_vlm_model = False
    def freeze(self): pass
    def unfreeze(self, **kwargs): pass
    def trainable_parameters(self): return {}
    def parameters(self): return {}

class FakeVLMInner:
    layers = [FakeLayer() for _ in range(8)]
class FakeVLM:
    language_model = type("LM", (), {"model": FakeVLMInner()})
    _unsloth_full_finetuning = False
    _is_vlm_model = True
    def freeze(self): pass
    def unfreeze(self, **kwargs): pass
    def trainable_parameters(self): return {}
    def parameters(self): return {}
```

tests/test_mlx_finetune_last_n_layers.py
```diff
@@ -0,0 +1,149 @@
...
+    L.linear_to_lora_layers = fake_linear_to_lora_layers
+    # mlx_lm.tuner.utils is imported inside the function:
+    fake_mod = type(sys)("mlx_lm.tuner.utils")
+    fake_mod.linear_to_lora_layers = fake_linear_to_lora_layers
+    sys.modules["mlx_lm.tuner.utils"] = fake_mod
+
+    # Case 1: default (None) -> all 8 layers
+    captured["calls"].clear()
+    loader_mod.FastMLXModel.get_peft_model(
+        FakeModel(),
```
danielhanchen, line 99: **[2/2 reviewers]** Nit: this monkeypatches `sys.modules['mlx_lm.tuner.utils']` and four `L.*` attributes with no teardown. If pytest collection order changes, downstream tests in the same session will see the stubs instead of the real module. Use the `monkeypatch` fixture so changes are reverted automatically.

```suggestion
def test_get_peft_model_passes_finetune_last_n_layers_through(monkeypatch):
    import sys
    import unsloth_zoo.mlx.loader as loader_mod

    # ... build FakeModel ...

    captured = {"calls": []}
    def fake_linear_to_lora_layers(model, num_layers, config, use_dora=False):
        captured["calls"].append(num_layers)

    monkeypatch.setattr(loader_mod, "_fix_missing_no_grad", lambda m: None)
    monkeypatch.setattr(loader_mod, "_resolve_lora_keys", lambda m, t: [
        "model.layers.0.self_attn.q_proj",
        "model.layers.1.mlp.gate_proj",
    ])
    monkeypatch.setattr(loader_mod, "_apply_mlx_lora_initialization", lambda m, init: None)
    monkeypatch.setattr(loader_mod, "linear_to_lora_layers", fake_linear_to_lora_layers)
    fake_mod = type(sys)("mlx_lm.tuner.utils")
    fake_mod.linear_to_lora_layers = fake_linear_to_lora_layers
    monkeypatch.setitem(sys.modules, "mlx_lm.tuner.utils", fake_mod)
```

PR #670, opened 2026-05-18
fix(mlx): warn on bf16 -> fp16 downcast in FastMLXModel loader

danielhanchen:
## Summary
- `_convert_mlx_dtype` silently downcasts native bf16 weights to fp16 when the user passes `dtype="float16"` to `FastMLXModel.from_pretrained`.
- fp16's finite range (~6.5e4) is much narrower than bf16's (~3.4e38); models with large activations (e.g. Gemma3) can lose precision or overflow silently.
- This PR adds a one-line warning when that downcast is about to happen; the cast still runs (users on M1/M2 without native bf16 GPU support need fp16).

## Why
Empirical (gemma-3-270m-it, single-row LoRA memorization, n=15 seeds, otherwise-identical setup):

| dtype= | greedy-decode pass | cf_loss=0 (15 seeds) |
|---|---|---|
| `None` (keeps native bf16) | 47% | 15/15 |
| `"float16"` (silent bf16 -> fp16) | 15% | 15/15 |

The 32pp drop is entirely from the silent bf16 -> fp16 cast. Teacher-forced completion loss is `0` in both cases — the model memorizes either way; only the first-token greedy argmax distribution diverges. CI smoke gating per `unslothai/unsloth#5537` stays green either way, but greedy-decode behavior diverges noticeably enough that a user comparing fp16 vs bf16 runs would suspect a different bug.

Tracked alongside two earlier MLX-parity PRs:
- `unslothai/unsloth-zoo#669` — `finetune_last_n_layers` knob (layer-selection mismatch).
- `unslothai/unsloth#5564` — same knob, CUDA path.

This PR addresses factor (4) of the four-factor bisection ($\Delta$ pass rate vs `mlx_lm.load + manual loop + last-16 layers`):
1. Layer selection (`unsloth-zoo#669`)
2. `MLXTrainer` overhead vs manual loop (-14pp)
3. `FastMLXModel` loader patches (-10pp)
4. **bf16 -> fp16 downcast (-28pp) — this PR**

## Behavior
- `dtype="float16"` against a bf16-native model: cast still happens, warning logged via `warnings.warn(...)`.
- `dtype="bfloat16"` / `dtype=None` / `dtype="float32"`: no warning, no change.
- `dtype="float16"` against a fp32-native or fp16-native model: no warning (the bf16 -> fp16 specific regression doesn't apply).

The warning message points the user at `dtype=None` on bf16-capable Apple Silicon (M3+) and `dtype="float32"` for full precision.

## Test plan
- [x] `tests/test_mlx_dtype_downcast_warning.py` — five cases:
  - `bf16 -> fp16` emits the warning.
  - `bf16 -> fp32` (upcast) does NOT emit it.
  - `fp32 -> fp16` (different lossy regime) does NOT emit it.
  - No-cast-needed early return does NOT emit it.
  - Cast still mutates `model.parameters()` after the warning.
- [x] Local: `pytest tests/test_mlx_dtype_downcast_warning.py -v` -> 5 passed.

### DIFF
--- a/tests/test_mlx_dtype_downcast_warning.py
+++ b/tests/test_mlx_dtype_downcast_warning.py
@@ -58,6 +58,25 @@ def test_force_float32_list_exported():
     assert "qwen3_5" in unsloth_zoo.FORCE_FLOAT32
 
 
+def test_force_float32_matches_config_json_model_types():
+    """Entries are HuggingFace `config.json` `model_type` values (the same
+    strings returned by `get_transformers_model_type`). The MLX matcher
+    must accept the real on-disk variants verbatim."""
+    from unsloth_zoo.mlx.loader import _is_force_float32_arch
+    # Real config.json model_type strings as they appear on the Hub.
+    real_world = [
+        "gemma3",       # google/gemma-3-*
+        "gemma3_text",  # google/embeddinggemma-*
+        "gemma3n",      # google/gemma-3n-*
+        "gpt_oss",      # openai/gpt-oss-*
+        "qwen3_5",      # Qwen/Qwen3.5-*
+    ]
+    for model_type in real_world:
+        assert _is_force_float32_arch(model_type) is True, (
+            f"FORCE_FLOAT32 must match real config.json model_type {model_type!r}"
+        )
+
+
 def test_warns_on_bf16_to_fp16_for_gemma3():
     """The production scenario: gemma3 bf16 weights downcast to fp16."""
     import torch
--- a/tests/test_mlx_dtype_downcast_warning.py
+++ b/tests/test_mlx_dtype_downcast_warning.py
@@ -1,0 +1,155 @@
+# Unsloth Zoo - Utilities for Unsloth
+# Test the bf16->fp16 downcast warning in unsloth_zoo.mlx.loader._convert_mlx_dtype.
+#
+# Gemma3-270m's native bf16 storage carries activations whose magnitudes
+# exceed fp16's finite range. Silent bf16->fp16 downcast in
+# FastMLXModel.from_pretrained(dtype="float16") was observed to drop a
+# single-row LoRA memorization fixture's greedy-decode pass rate from
+# 47% (dtype=None) to 15% (dtype="float16") across 15 seeds. The cast
+# remains supported (users on M1/M2 need it), but it now emits a
+# warning so callers see the trade-off.
+
+from __future__ import annotations
+
+import pytest
+import warnings
+
+
+@pytest.fixture(autouse=True, scope="module")
+def _install_mlx_shim():
+    from mlx_simulation import simulate_mlx_on_torch
+    simulate_mlx_on_torch()
+
+
+class _FakeArray:
+    """Minimal stand-in for an mx.array carrying just a dtype field.
+
+    _convert_mlx_dtype only reads .dtype and calls .astype on params, both
+    of which the shim implements via the underlying torch tensor; using a
+    real torch tensor here keeps the path fully exercised.
+    """
+    def __init__(self, dtype):
+        import torch
+        # Empty tensor is fine — _convert_mlx_dtype doesn't read data.
+        self._t = torch.zeros((1,), dtype=dtype)
+
+    @property
+    def dtype(self):
+        return self._t.dtype
+
+    def astype(self, target_dtype):
+        return _FakeArray(target_dtype)
+
+
+class _FakeModel:
+    """The smallest object _convert_mlx_dtype needs:
+    a .parameters() returning a flat dict and .update() that accepts the
+    tree_map output.
+    """
+    def __init__(self, params):
+        self._params = params
+
+    def parameters(self):
+        return self._params
+
+    def update(self, new_params):
+        self._params = new_params
+
+
+def _make_model(dtype_map):
+    """Build a fake model with named params at the given dtypes."""
+    import torch
+    return _FakeModel({name: _FakeArray(dt) for name, dt in dtype_map.items()})
+
+
+def test_warns_on_bf16_to_fp16_downcast():
+    """The exact production scenario: native bf16 weights, user asks for fp16."""
+    import torch
+    from unsloth_zoo.mlx.loader import _convert_mlx_dtype
+
+    model = _make_model({"w1": torch.bfloat16, "w2": torch.bfloat16})
+
+    with warnings.catch_warnings(record=True) as caught:
+        warnings.simplefilter("always")
+        _convert_mlx_dtype(model, torch.float16)
+
+    msgs = [str(w.message) for w in caught]
+    assert any("bfloat16" in m and "float16" in m for m in msgs), (
+        f"Expected a bf16->fp16 downcast warning, got: {msgs}"
+    )
+
+
+def test_no_warning_on_bf16_to_fp32_upcast():
+    """Upcasts are safe; no warning."""
+    import torch
+    from unsloth_zoo.mlx.loader import _convert_mlx_dtype
+
+    model = _make_model({"w1": torch.bfloat16})
+
+    with warnings.catch_warnings(record=True) as caught:
+        warnings.simplefilter("always")
+        _convert_mlx_dtype(model, torch.float32)
+
+    bf_fp_msgs = [
+        str(w.message) for w in caught
+        if "bfloat16" in str(w.message) and "float16" in str(w.message)
+    ]
+    assert bf_fp_msgs == [], (
+        f"Did not expect a bf16->fp16 warning for a bf16->fp32 upcast; got: {bf_fp_msgs}"
+    )
+
+
+def test_no_warning_on_fp32_to_fp16_downcast():
+    """fp32->fp16 is lossy too but is a different (already-explicit) regime."""
+    import torch
+    from unsloth_zoo.mlx.loader import _convert_mlx_dtype
+
+    model = _make_model({"w1": torch.float32})
+
+    with warnings.catch_warnings(record=True) as caught:
+        warnings.simplefilter("always")
+        _convert_mlx_dtype(model, torch.float16)
+
+    bf_fp_msgs = [
+        str(w.message) for w in caught
+        if "bfloat16" in str(w.message) and "float16" in str(w.message)
+    ]
+    assert bf_fp_msgs == [], (
+        f"Did not expect a bf16->fp16 warning for a fp32->fp16 cast; got: {bf_fp_msgs}"
+    )
+
+
+def test_no_warning_when_no_cast_needed():
+    """All params already at target_dtype -> early return, no warnings."""
+    import torch
+    from unsloth_zoo.mlx.loader import _convert_mlx_dtype
+
+    model = _make_model({"w1": torch.float16, "w2": torch.float16})
+
+    with warnings.catch_warnings(record=True) as caught:
+        warnings.simplefilter("always")
+        _convert_mlx_dtype(model, torch.float16)
+
+    bf_fp_msgs = [
+        str(w.message) for w in caught
+        if "bfloat16" in str(w.message) and "float16" in str(w.message)
+    ]
+    assert bf_fp_msgs == [], (
+        f"Did not expect a warning when no cast is needed; got: {bf_fp_msgs}"
+    )
+
+
+def test_cast_still_happens_after_warning():
+    """The warning is advisory only — the requested cast still occurs."""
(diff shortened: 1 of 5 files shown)


danielhanchen:
## Update: gate warning on FORCE_FLOAT32 + centralize the list

Per @danielhanchen — the unconditional warning fired for every bf16→fp16 cast, which is too noisy. Now:

1. The bf16→fp16 warning is gated on \`model_type\` being in the \`FORCE_FLOAT32\` list (Gemma3 family, gpt\\_oss, Qwen3.5). Llama / Mistral / Qwen2 / etc. cast silently as before.

2. \`FORCE_FLOAT32\` was previously defined inline in \`unsloth/models/loader.py\`. It now lives in a new dependency-free module \`unsloth_zoo/model_lists.py\` and is re-exported as \`unsloth_zoo.FORCE_FLOAT32\`. The new helper \`_is_force_float32_arch\` normalizes \`-\`/\`_\` and respects the trailing-comma exact-match marker (\`gemma3,\`).

3. Companion **unsloth PR #5610** switches \`unsloth/models/loader.py\` to \`from unsloth_zoo import FORCE_FLOAT32\` and deletes its local copy. Verified \`unsloth.models.loader.FORCE_FLOAT32 is unsloth_zoo.FORCE_FLOAT32\` after both PRs.

Tests updated: 9 cases covering the export, all FORCE_FLOAT32 archs warning, non-listed arch staying silent, upcasts/no-op casts staying silent, and the cast still happening after the warning. All pass locally (\`pytest tests/test_mlx_dtype_downcast_warning.py\` → 9/9).

danielhanchen (commented):
Thank you for the PR! The goal of this PR is to warn callers when `FastMLXModel` silently downcasts a bfloat16-native model to fp16 (which NaN/Infs on Gemma3 family / gpt_oss / Qwen3.5), and to give the CUDA loader and the MLX loader a shared source of truth for that list. As a summary, this PR adds a dependency-free `unsloth_zoo/model_lists.py` carrying `FORCE_FLOAT32`, re-exports it from `unsloth_zoo` (top-level) and from `unsloth_zoo.compiler` (back-compat), introduces a `_is_force_float32_arch(model_type)` helper that normalizes `-`/`_` and respects trailing-comma entries, and gates the new bf16→fp16 warning on that lookup. The companion PR #5610 switches `unsloth/models/loader.py` to import from there.

| 2/2 | Med | The `gemma3,` trailing-comma marker is documented as an exact-match delimiter, but `_is_force_float32_arch` already strips it; the comma is a no-op in this new helper. Either drop the comma and update the comment, or document that the comma only matters for the CUDA loader's substring path. |
| 1/2 | Med | Warning message names `dtype=None` and `dtype='float32'` but omits `dtype='bfloat16'` and the `UNSLOTH_FORCE_FLOAT32` env-var alternative. |
| 2/2 | Nit | The local `from ..model_lists import FORCE_FLOAT32` at the top of `_convert_mlx_dtype` is unused; only `_is_force_float32_arch` consumes the list. Dead import. |
| 1/2 | Nit | The `break` after `needs_cast = True` was removed so the loop can also detect `has_bf16`. On large models this becomes O(n_params) even when the warning can never fire (e.g. fp32→fp16). Add a fast-exit. |
| 1/2 | Low | `_norm` does not strip `.`, so a future HF config emitting `model_type="qwen3.5"` (dot form) would silently miss the warning. |
| 1/2 | Nit | `test_gemma3_comma_does_not_match_gemma3n` is misnamed — it actually asserts that `gemma3n` DOES match (via its own list entry). |

Overall: APPROVE_WITH_NITS.

unsloth_zoo/model_lists.py
```diff
@@ -0,0 +1,37 @@
...
+"""
+
+__all__ = ["FORCE_FLOAT32"]
+
+
+# Architectures whose activations exceed fp16's finite range (~6.5e4) and
+# must run in bf16 or fp32. Loaded as float16 they silently NaN/Inf at
+# training time. Shared source of truth for the CUDA loader
+# (unsloth/models/loader.py) and the MLX loader (unsloth_zoo/mlx/loader.py).
+FORCE_FLOAT32 = [
```
danielhanchen, line 45: **[2/2 reviewers]** Med: the trailing-comma marker is a substring-delimiter trick used by `unsloth/models/loader.py:1378` (where matching is `disable_name.lower() in model_types_all`). `_is_force_float32_arch` in this PR explicitly strips the comma before comparing, so the marker is a no-op in zoo's matcher. Either drop the comma here (and rely on `gemma3` vs `gemma3n` already being separate exact-match entries) or update the comment to make clear it's only load-bearing for the substring-matching CUDA consumer.

```suggestion
FORCE_FLOAT32 = [
    "gemma3",      # exact-match in zoo; trailing-comma kept for back-compat with unsloth/models/loader.py substring matcher (do not remove without coordinating that file).
    "gemma3text",  # EmbeddingGemma / standalone text-only Gemma3
    "gemma3n",
    "gpt_oss",
    "qwen3_5",     # Qwen3.5 GDN layers NaN on fp16
]
```
(or, alternatively, keep `"gemma3,"` literally and rewrite the docstring above to say the comma is a CUDA-side substring delimiter, not an exact-match marker.)

unsloth_zoo/mlx/loader.py
```diff
@@ -88,22 +88,56 @@ def _temporary_hf_token_env(token):
...
     """Cast floating-point params to target_dtype (preserves quantized ints)
     while honoring the model's optional path-based ``cast_predicate``.
+
+    Warns on bfloat16 -> float16 for architectures in
+    ``unsloth_zoo.FORCE_FLOAT32`` (Gemma3 family, gpt_oss, Qwen3.5) where
+    fp16's narrower range silently NaN/Infs at training time.
     """
     import mlx.core as mx
     from mlx.utils import tree_flatten, tree_map_with_path
+    from ..model_lists import FORCE_FLOAT32
```
danielhanchen, line 120: **[2/2 reviewers]** Nit: this `from ..model_lists import FORCE_FLOAT32` is dead — `FORCE_FLOAT32` is never referenced inside `_convert_mlx_dtype`. The consumption happens inside `_is_force_float32_arch`. Drop it.

```suggestion
    from mlx.utils import tree_flatten, tree_map_with_path
```

unsloth_zoo/mlx/loader.py
```diff
@@ -88,22 +88,56 @@ def _temporary_hf_token_env(token):
...
     cast_pred = getattr(model, "cast_predicate", lambda _: True)
 
     needs_cast = False
+    has_bf16 = False
     for k, v in tree_flatten(model.parameters()):
         if cast_pred(k) and mx.issubdtype(v.dtype, mx.floating) and v.dtype != target_dtype:
             needs_cast = True
-            break
+            if v.dtype == mx.bfloat16:
+                has_bf16 = True
```
danielhanchen, line 129: **[1/2 reviewers]** Nit: `break` was removed from the original loop so we can also detect bf16 presence. But this means every call now scans all parameters even on the common case where no warning can ever fire (target_dtype == fp16 but no bf16 weights, OR model_type not in FORCE_FLOAT32). For large models this is non-trivial. Short-circuit when both flags are set:

```suggestion
    for k, v in tree_flatten(model.parameters()):
        if cast_pred(k) and mx.issubdtype(v.dtype, mx.floating) and v.dtype != target_dtype:
            needs_cast = True
            if v.dtype == mx.bfloat16:
                has_bf16 = True
            if has_bf16:
                break
```

unsloth_zoo/mlx/loader.py
```diff
@@ -88,22 +88,56 @@ def _temporary_hf_token_env(token):
...
     if not needs_cast:
         return
 
+    if has_bf16 and target_dtype == mx.float16 and _is_force_float32_arch(model_type):
+        warnings.warn(
```
danielhanchen, line 134: **[1/2 reviewers]** Med: the warning text gives `dtype=None` and `dtype='float32'` but is missing `dtype='bfloat16'` (explicit-bf16 keeps native and is the typical answer on M3+) and a pointer to `UNSLOTH_FORCE_FLOAT32=1` for users who want the CUDA-style guard. Make it actionable:

```suggestion
        warnings.warn(
            f"Unsloth: downcasting bfloat16 -> float16 on {model_type!r}, "
            "which is known to NaN/Inf in fp16. Pass dtype=None to keep "
            "native bf16, dtype='bfloat16' to be explicit on M3+, or "
            "dtype='float32' for full precision. Set UNSLOTH_FORCE_FLOAT32=1 "
            "to silence this warning if the downcast is intentional.",
            stacklevel=2,
        )
```

unsloth_zoo/mlx/loader.py
```diff
@@ -88,22 +88,56 @@ def _temporary_hf_token_env(token):
...
 
-def _convert_mlx_dtype(model, target_dtype) -> None:
+def _is_force_float32_arch(model_type: str) -> bool:
+    """Case-insensitive lookup of ``model_type`` in
+    ``unsloth_zoo.FORCE_FLOAT32``. Strips ``-``/``_`` and treats a
+    trailing comma on a list entry as an exact-match marker."""
+    if not model_type:
+        return False
+    from ..model_lists import FORCE_FLOAT32
+    def _norm(s: str) -> str:
```
danielhanchen, line 98: **[1/2 reviewers]** Low: `_norm` strips `-` and `_` but not `.`. HF emits model_type values like `qwen3.5` in some configs; that variant would silently miss the warning even though `qwen3_5` is in `FORCE_FLOAT32`. Cheap to harden:

```suggestion
    def _norm(s: str) -> str:
        return s.lower().replace("-", "").replace("_", "").replace(".", "")
```

tests/test_mlx_dtype_downcast_warning.py
```diff
@@ -0,0 +1,175 @@
...
+
+def test_gemma3_comma_does_not_match_gemma3n():
+    """The trailing-comma trick: 'gemma3,' must NOT match 'gemma3n_audio'."""
+    import torch
+    from unsloth_zoo.mlx.loader import _is_force_float32_arch
+    # gemma3n is its own entry and DOES match; we're testing that the
+    # 'gemma3,' entry doesn't accidentally swallow gemma3n variants by
+    # prefix match. gemma3n itself still matches via its own entry.
+    assert _is_force_float32_arch("gemma3") is True
+    assert _is_force_float32_arch("gemma3n") is True
```
danielhanchen, line 135: **[1/2 reviewers]** Nit: the test name says `gemma3, does not match gemma3n` but the body actually asserts that `gemma3n` DOES match (via its own list entry, not via `gemma3,`). The actual no-false-positive case is `gemma3_audio_only_pretend`. Rename for clarity:

```suggestion
def test_gemma3_exact_match_does_not_swallow_unrelated_gemma3_variants():
    """`gemma3` and `gemma3n` are separate exact-match entries; an
    invented gemma3 variant not in the list should NOT match."""
    import torch
    from unsloth_zoo.mlx.loader import _is_force_float32_arch
    assert _is_force_float32_arch("gemma3") is True
    assert _is_force_float32_arch("gemma3n") is True  # own entry
    assert _is_force_float32_arch("gemma3text") is True  # own entry
    assert _is_force_float32_arch("gemma3_audio_only_pretend") is False
```

danielhanchen:
Doc + test update (commit ae74d0e9): make explicit that `FORCE_FLOAT32` entries are HuggingFace `config.json` `model_type` strings (the same values returned by `unsloth_zoo.hf_utils.get_transformers_model_type`).

`unsloth_zoo/model_lists.py` now documents each entry against its real-world model_type:

| Hub family             | config.json model_type | FORCE_FLOAT32 entry |
|------------------------|------------------------|----------------------|
| google/gemma-3-*       | `"gemma3"`             | `"gemma3,"` (substring delimiter for the CUDA loader) |
| google/embeddinggemma-*| `"gemma3_text"`        | `"gemma3text"` (strips `_` in MLX matcher) |
| google/gemma-3n-*      | `"gemma3n"`            | `"gemma3n"` |
| openai/gpt-oss-*       | `"gpt_oss"`            | `"gpt_oss"` |
| Qwen/Qwen3.5-*         | `"qwen3_5"`            | `"qwen3_5"` |

Added `test_force_float32_matches_config_json_model_types` which asserts `_is_force_float32_arch` returns `True` for the five real-world model_type strings above, so a future entry-rename can't silently miss a real Hub model.

Tests: `pytest tests/test_mlx_dtype_downcast_warning.py` -> 10/10 pass.

PR #671, opened 2026-05-18
fix(mlx): honor max_grad_value=None as a disable signal

danielhanchen:
## Summary
- `MLXTrainingConfig.max_grad_value` is typed `float | None = 1.0`, but the resolution code silently rebound an explicit `None` back to the default 1.0, so callers who set `max_grad_value=None` to disable elementwise gradient clipping were still getting clipping at $\pm$1.0.
- This PR treats `None` the same as `0.0` (disable) and clarifies the field docstring. Default behavior is unchanged; only callers explicitly passing `None` see the corrected disable semantics.

## Why
The `Optional[float]` type annotation naturally suggests `None` means "off". The field docstring previously said "Set 0.0 to disable" without mentioning what `None` would do. The internal code on line 732 of `unsloth_zoo/mlx/trainer.py` was:

\`\`\`python
_raw_mgv = getattr(args, "max_grad_value", 1.0)
max_grad_value = 1.0 if _raw_mgv is None else float(_raw_mgv or 0.0)
\`\`\`

i.e. `None` $\to$ rebound to `1.0` $\to$ clipping enabled at $\pm$1.0. That is the opposite of what users expected.

Surfaced while writing MLX parity probes against `mlx-lm`'s native loop (which does no elementwise clipping): probes that set `max_grad_value=None` with the explicit comment "match mlx-lm: no elementwise clip" were silently still being clipped.

## Behavior
- `MLXTrainingConfig()` $\to$ `max_grad_value = 1.0` (clipping on). **Unchanged.**
- `MLXTrainingConfig(max_grad_value=None)` $\to$ clipping **disabled**. (Was: clipping at $\pm$1.0.)
- `MLXTrainingConfig(max_grad_value=0.0)` $\to$ clipping disabled. **Unchanged.**
- `MLXTrainingConfig(max_grad_value=2.5)` $\to$ clipping at $\pm$2.5. **Unchanged.**

## Test plan
- [x] `tests/test_mlx_max_grad_value_none.py` $\to$ five cases pinning the four-way decision (default, None, 0.0, positive) plus dataclass round-trip.
- [x] Local: `pytest tests/test_mlx_max_grad_value_none.py -v` $\to$ 5 passed.

## Related
This PR is part of an ongoing MLX vs `mlx-lm` parity bisection:
- `#669` $\to$ `finetune_last_n_layers` (layer-selection mismatch).
- `#670` $\to$ bf16 $\to$ fp16 downcast warning.
- This PR $\to$ honor `max_grad_value=None` as documented.

### DIFF
--- a/tests/test_mlx_max_grad_value_none.py
+++ b/tests/test_mlx_max_grad_value_none.py
@@ -1,6 +1,9 @@
 # Unsloth Zoo - Utilities for Unsloth
-# Pin MLXTrainingConfig.max_grad_value resolution: default None (no clip),
-# None and 0.0 disable, positive opts in. HF/TRL parity.
+# Pin MLXTrainingConfig.max_grad_value resolution:
+#   * None (default) -> cheap MLX elementwise clip at 1.0, unless
+#     max_grad_norm > 0 is also passed (then global-norm wins).
+#   * 0.0 -> explicitly disabled.
+#   * positive -> explicit elementwise opt-in; overrides max_grad_norm.
 
 from __future__ import annotations
 
@@ -13,41 +16,34 @@ def _install_mlx_shim():
     simulate_mlx_on_torch()
 
 
-def _disable_clip_decision(raw_mgv):
-    """Mirror trainer.py's internal decision so we can pin the rule
-    without standing up a full MLXTrainer / model / optimizer.
-    """
-    max_grad_value = 0.0 if raw_mgv is None else float(raw_mgv or 0.0)
-    return max_grad_value > 0
+def _resolve(raw_mgv, max_grad_norm):
+    """Mirror trainer.py's internal resolution. Returns the (max_grad_value,
+    max_grad_norm) pair the step function will actually use."""
+    user_set = raw_mgv is not None
+    if user_set:
+        mgv = float(raw_mgv or 0.0)
+        if max_grad_norm > 0 and mgv > 0:
+            max_grad_norm = 0.0
+    elif max_grad_norm > 0:
+        mgv = 0.0
+    else:
+        mgv = 1.0
+    return mgv, max_grad_norm
 
 
-def test_field_default_is_clip_off():
-    """Default = None (no elementwise clip). HF/TRL parity."""
+# -- field defaults ---------------------------------------------------------
+
+
+def test_field_default_is_none_sentinel():
+    """Default is None (a sentinel meaning 'use MLX cheap default')."""
     from unsloth_zoo.mlx.trainer import MLXTrainingConfig
 
     cfg = MLXTrainingConfig(output_dir="/tmp/x")
     assert cfg.max_grad_value is None
 
 
-def test_none_disables_clip():
-    """Explicit None disables clipping (not silently rebound to 1.0)."""
-    assert _disable_clip_decision(None) is False
-
-
-def test_zero_disables_clip():
-    """Zero remains a documented disable signal."""
-    assert _disable_clip_decision(0.0) is False
-    assert _disable_clip_decision(0) is False
-
-
-def test_positive_enables_clip():
-    """A positive value opts in to elementwise clipping."""
-    assert _disable_clip_decision(1.0) is True
-    assert _disable_clip_decision(2.5) is True
-
-
 def test_field_accepts_none():
-    """Field accepts None and round-trips it through the dataclass."""
+    """Field accepts None and round-trips through the dataclass."""
     from unsloth_zoo.mlx.trainer import MLXTrainingConfig
 
     cfg = MLXTrainingConfig(max_grad_value=None, output_dir="/tmp/x")
@@ -55,8 +51,71 @@ def test_field_accepts_none():
 
 
 def test_field_accepts_explicit_positive():
-    """Field accepts positive floats for power users opting into clip."""
+    """Field accepts positive floats for power users opting in."""
     from unsloth_zoo.mlx.trainer import MLXTrainingConfig
 
     cfg = MLXTrainingConfig(max_grad_value=2.5, output_dir="/tmp/x")
     assert cfg.max_grad_value == 2.5
+
+
+# -- resolution semantics ---------------------------------------------------
+
+
+def test_default_uses_cheap_elementwise():
+    """Default (None, max_grad_norm=0.0) -> elementwise clip at 1.0."""
+    mgv, mgn = _resolve(raw_mgv=None, max_grad_norm=0.0)
+    assert mgv == 1.0
+    assert mgn == 0.0
+
+
+def test_user_max_grad_norm_wins_over_default():
+    """User passes max_grad_norm=1.0 with default max_grad_value=None ->
+    global-norm clipping wins, elementwise disabled. HF/TRL parity."""
+    mgv, mgn = _resolve(raw_mgv=None, max_grad_norm=1.0)
+    assert mgv == 0.0
+    assert mgn == 1.0
+
+
+def test_explicit_zero_disables_elementwise():
+    """Explicit 0.0 disables elementwise. With no max_grad_norm,
+    nothing clips."""
+    mgv, mgn = _resolve(raw_mgv=0.0, max_grad_norm=0.0)
+    assert mgv == 0.0
+    assert mgn == 0.0
+
+
+def test_explicit_zero_lets_max_grad_norm_through():
+    """Explicit max_grad_value=0.0 + max_grad_norm=1.0 -> only norm clipping."""
+    mgv, mgn = _resolve(raw_mgv=0.0, max_grad_norm=1.0)
+    assert mgv == 0.0
+    assert mgn == 1.0
+
+
+def test_explicit_positive_overrides_max_grad_norm():
+    """Explicit max_grad_value=2.0 with max_grad_norm=1.0 -> elementwise
+    wins (existing rule), max_grad_norm zeroed."""
+    mgv, mgn = _resolve(raw_mgv=2.0, max_grad_norm=1.0)
+    assert mgv == 2.0
+    assert mgn == 0.0
+
+
+def test_explicit_positive_alone():
+    """User passes max_grad_value=5.0 with no max_grad_norm -> elementwise at 5."""
+    mgv, mgn = _resolve(raw_mgv=5.0, max_grad_norm=0.0)
+    assert mgv == 5.0
+    assert mgn == 0.0
+
+
+# -- trainer source assertions (defense-in-depth) ---------------------------
+
+
+def test_trainer_source_pins_resolution_rule():
+    """Source-level pin: trainer.py contains the four-branch resolution.
+    Cheap defense against a future refactor silently regressing the rule."""
+    import inspect
+    from unsloth_zoo.mlx import trainer as T
+
+    src = inspect.getsource(T.MLXTrainer.train) + inspect.getsource(T.MLXTrainer._train_inner)
+    assert "_user_set_mgv = _raw_mgv is not None" in src
+    assert "elif max_grad_norm > 0:" in src
+    assert "max_grad_value = 1.0" in src
--- a/tests/test_mlx_max_grad_value_none.py
+++ b/tests/test_mlx_max_grad_value_none.py
@@ -1,14 +1,6 @@
 # Unsloth Zoo - Utilities for Unsloth
-# Verify MLX grad-clip semantics match issue #662's HF-parity proposal:
-#
-#   1. MLXTrainingConfig.max_grad_value defaults to None (no elementwise
-#      clip), so a user-passed max_grad_norm is honored as in HF/TRL.
-#   2. None and 0.0 both disable elementwise clipping. Explicit None
-#      previously rebound silently to 1.0.
-#   3. A positive value opts in to elementwise clipping; the existing
-#      mutual-exclusion rule (elementwise wins, max_grad_norm zeroed)
-#      then applies and prints a notice.
-#   4. The dataclass round-trips None through the field.
+# Pin MLXTrainingConfig.max_grad_value resolution: default None (no clip),
+# None and 0.0 disable, positive opts in. HF/TRL parity.
 
 from __future__ import annotations
 
--- a/tests/test_mlx_max_grad_value_none.py
+++ b/tests/test_mlx_max_grad_value_none.py
@@ -1,14 +1,14 @@
 # Unsloth Zoo - Utilities for Unsloth
-# Verify that `MLXTrainingConfig(max_grad_value=None)` disables
-# elementwise grad clipping rather than silently rebinding to the
-# default 1.0.
+# Verify MLX grad-clip semantics match issue #662's HF-parity proposal:
 #
-# Background: the field is typed `float | None = 1.0`, and the field
-# docstring documents `None` as a disable signal. Probe authors comparing
-# `unsloth_zoo.MLXTrainer` against `mlx-lm`'s native loop (which does
-# no elementwise clipping) set `max_grad_value=None` and were surprised
-# that clipping was still applied. This test pins the fixed behavior so
-# the disable-None semantics stays stable.
+#   1. MLXTrainingConfig.max_grad_value defaults to None (no elementwise
+#      clip), so a user-passed max_grad_norm is honored as in HF/TRL.
+#   2. None and 0.0 both disable elementwise clipping. Explicit None
+#      previously rebound silently to 1.0.
+#   3. A positive value opts in to elementwise clipping; the existing
+#      mutual-exclusion rule (elementwise wins, max_grad_norm zeroed)
+#      then applies and prints a notice.
+#   4. The dataclass round-trips None through the field.
 
 from __future__ import annotations
 
@@ -29,15 +29,16 @@ def _disable_clip_decision(raw_mgv):
     return max_grad_value > 0
 
 
-def test_field_default_is_clip_on():
+def test_field_default_is_clip_off():
+    """Default = None (no elementwise clip). HF/TRL parity."""
     from unsloth_zoo.mlx.trainer import MLXTrainingConfig
 
     cfg = MLXTrainingConfig(output_dir="/tmp/x")
-    assert cfg.max_grad_value == 1.0
+    assert cfg.max_grad_value is None
 
 
 def test_none_disables_clip():
-    """The motivating change: passing None disables clipping."""
+    """Explicit None disables clipping (not silently rebound to 1.0)."""
     assert _disable_clip_decision(None) is False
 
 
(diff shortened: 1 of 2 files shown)


danielhanchen:
Expanded scope: this PR now also implements the full recommended fix from `#662`. 

In addition to honoring `max_grad_value=None` as a disable signal (the original PR), the field default is now `None` (was 1.0). This makes the MLX path's grad-clip semantics match HF/TRL: a user calling `MLXTrainer` with `max_grad_norm=1.0` now gets global-norm clipping at 1.0 as they intended, instead of getting elementwise +/-1.0 clipping (which rotates gradient direction per leaf and produced different convergence basins than CUDA for identical configs).

Closes `#662`.

### Behavior summary
- `MLXTrainingConfig()` -> both clip modes off. **Changed from "elementwise clip at 1.0".**
- `MLXTrainingConfig(max_grad_norm=1.0)` -> global-norm clip at 1.0. **Changed from "max_grad_norm silently dropped".** Matches HF/TRL.
- `MLXTrainingConfig(max_grad_value=1.0)` -> elementwise clip at +/-1.0 (opt-in). **Unchanged.**
- `MLXTrainingConfig(max_grad_value=None)` -> elementwise disabled. **Changed from "silently rebound to 1.0".**
- `MLXTrainingConfig(max_grad_norm=1.0, max_grad_value=1.0)` -> elementwise wins, `max_grad_norm` zeroed, notice printed. **Unchanged.**

### Migration impact
Anyone who was relying on the implicit `max_grad_value=1.0` clip without setting it explicitly will now get NO clipping unless they also pass `max_grad_norm` or `max_grad_value` explicitly. #662 explicitly accepts this trade-off because HF/TRL parity is the goal: a user who wrote `MLXTrainingConfig(max_grad_norm=1.0)` was already expressing intent to clip at norm 1.0.

Tests updated; six cases pass.

danielhanchen:
## Round BT verification on Apple Silicon

This PR is the missing piece for end-to-end MLX parity with mlx-lm CLI through the public `FastMLXModel` + `MLXTrainer` API. Probe matrix run pinned to PR #674 HEAD (`0124424` -- the seed-ordering fix), bisecting whether elementwise clip-at-1 is the residual basin gap on top of #669 + #674.

[Run 26084185821](https://github.com/danielhanchen/unsloth-staging-2/actions/runs/26084185821) -- 45 jobs green.

| probe | configuration | pass rate | per-seed agreement with probe 31 |
|---|---|---|---|
| 31 | `mlx_lm.load + linear_to_lora_layers + manual @mx.compile loop` | 10/15 = **67%** | 15/15 (self) |
| 34 | `FastMLXModel + get_peft_model + MLXTrainer.train` with `max_grad_value=None` -> reinterpreted as 1.0 on current build | 7/15 = 47% | 8/15 |
| 41 | same as 34 but `max_grad_value=0.0` -> hits the disable branch on current build | 10/15 = **67%** | **15/15** |

Probe 41 reproduces mlx-lm CLI's basin pattern on every single seed. The only difference between probes 34 and 41 is `None` vs `0.0` for `max_grad_value`. On the current build, `trainer.py:731-732` reinterprets `None` as 1.0 (silent elementwise clip), so probe 34 actually clips gradients elementwise while the manual loop in probes 31 / 40 does not.

This PR's change to "honor `max_grad_value=None` as a disable signal" exactly closes the gap: `max_grad_value=None` (the new default) becomes equivalent to probe 41's `max_grad_value=0.0`, which is value-for-value identical to mlx-lm CLI's no-clip path.

### Full MLX parity story (3 PRs, all filed)

1. **#669** `finetune_last_n_layers` (matches mlx-lm CLI's `CONFIG_DEFAULTS['num_layers']=16`) -- trainable-param-set parity.
2. **#674** seed `mx.random` immediately before `linear_to_lora_layers` (matches `mlx_lm/tuner/lora.py:223`) -- LoRA-init parity, verified by probe 39 (dloss=0 step-for-step).
3. **#671** (this PR) `max_grad_value=None` -> disable -- trainer-side update-math parity, verified by probe 41 (15/15 seed agreement with mlx-lm CLI manual loop).

With all three merged, `FastMLXModel.from_pretrained` -> `FastMLXModel.get_peft_model` -> `MLXTrainer.train` produces the same trained model as the mlx-lm CLI path on a per-seed basis.

danielhanchen:
## Redesign: keep cheap MLX default, but user max_grad_norm wins

Per @danielhanchen — defaulting max_grad_value to None (no elementwise clip) lost the memory benefit of MLX's cheap elementwise path. \`max_grad_norm\` uses \`mlx.optimizers.clip_grad_norm\`, which materially increases peak memory on bf16 VLM runs, so we want the elementwise clip on by default.

Updated commit \`f14f7176\` keeps \`max_grad_value: float | None = None\` as the dataclass field (API unchanged) but the trainer-side resolution now treats None as a sentinel meaning *"use the MLX default 1.0 unless the user opted into global-norm clipping"*:

| max_grad_value         | max_grad_norm       | effective                                   |
| ---------------------- | ------------------- | -------------------------------------------- |
| None (default)         | 0.0 (default)       | elementwise 1.0 (cheap, no memory blowup)    |
| None (default)         | 1.0 (user)          | global norm 1.0 (HF/TRL parity)              |
| 0.0 (user)             | 0.0 (default)       | no clipping at all                           |
| 0.0 (user)             | 1.0 (user)          | global norm 1.0                              |
| 5.0 (user)             | 0.0                 | elementwise 5.0                              |
| 5.0 (user)             | 1.0 (user)          | elementwise 5.0 (warn, norm dropped)         |

The original bug — user passes \`max_grad_norm=1.0\` and silently gets clip-at-1.0 elementwise instead — is still fixed because resolution now distinguishes "user opted into max_grad_norm with default max_grad_value" from "user explicitly set both", and routes the former through the global-norm path.

Tests rewritten as ten paired (max_grad_value, max_grad_norm) cases plus a source-level pin (\`test_trainer_source_pins_resolution_rule\`) so a future refactor cannot silently regress the four-branch resolution. All 10 pass locally.

unsloth_zoo/mlx/trainer.py
```diff
@@ -119,12 +119,12 @@ class MLXTrainingConfig:
...
-    # rarely fires on real transformer grads, so the historical 5.0 was
-    # effectively a no-op; 1.0 matches the universal clip_grad_norm=1.0
-    # baseline while staying on MLX's fast tree_map(mx.clip) path.
-    max_grad_value: float | None = 1.0
+    # Elementwise clip ([-max_grad_value, max_grad_value], per-leaf).
+    # None (default) keeps the cheap MLX default of 1.0 unless the user
+    # passes max_grad_norm > 0, in which case global-norm clipping wins.
+    # 0.0 disables. A positive float opts in explicitly and overrides
+    # max_grad_norm with a warning.
+    max_grad_value: float | None = None
```
danielhanchen, line 127: **[1/2 reviewers]** High: this default flip from `1.0` to `None` is the core behavior change. Anyone calling `MLXTrainingConfig(max_grad_norm=1.0)` previously got elementwise@1.0 (warn-and-override path), and now gets global-norm@1.0 instead. Different gradient update every step on identical user configs. Worth surfacing in the PR description as a deliberate semantics change rather than "default behavior unchanged", and worth a release-note note for downstream pipelines.

The field docstring should also spell out the sentinel semantics, since the rule itself lives only in the trainer:

```suggestion
    # Elementwise clip ([-max_grad_value, max_grad_value], per-leaf).
    # Resolution:
    #   * None (default) AND max_grad_norm == 0  -> elementwise 1.0 (cheap MLX default)
    #   * None (default) AND max_grad_norm > 0   -> global-norm wins (HF/TRL parity)
    #   * 0.0                                    -> elementwise disabled
    #   * positive                               -> elementwise opts in; if max_grad_norm > 0 it is overridden with a warning
    max_grad_value: float | None = None
```

unsloth_zoo/mlx/trainer.py
```diff
@@ -723,19 +723,34 @@ def _train_inner(self):
...
-        # norm clipping is exact but materially increases memory on MLX.
-        _raw_mgv = getattr(args, "max_grad_value", 1.0)  # TODO: expose MLX grad-clip in Studio UI for power users
-        max_grad_value = 1.0 if _raw_mgv is None else float(_raw_mgv or 0.0)
-        if max_grad_norm > 0 and max_grad_value > 0:
-            print(
-                "Unsloth: max_grad_norm and max_grad_value are both enabled; "
-                "ignoring max_grad_norm in favor of max_grad_value."
-            )
-            max_grad_norm = 0.0
+        _raw_mgv = getattr(args, "max_grad_value", None)  # TODO: expose MLX grad-clip in Studio UI for power users
```
danielhanchen, line 738: **[1/2 reviewers]** Nit: the dataclass guarantees `max_grad_value` exists, so the `getattr(..., None)` fallback is dead defensive code. If kept, the fallback default should match the field default to avoid future divergence.

```suggestion
        _raw_mgv = args.max_grad_value  # TODO: expose MLX grad-clip in Studio UI for power users
```

unsloth_zoo/mlx/trainer.py
```diff
@@ -723,19 +723,34 @@ def _train_inner(self):
...
+                    "Unsloth: max_grad_norm and max_grad_value are both enabled; "
+                    "ignoring max_grad_norm in favor of max_grad_value."
+                )
+                max_grad_norm = 0.0
+        elif max_grad_norm > 0:
+            # User opted into global-norm clipping; suppress the default elementwise.
+            max_grad_value = 0.0
+        else:
+            # Neither requested -> cheap MLX default.
+            max_grad_value = 1.0
```
danielhanchen, line 753: **[1/2 reviewers]** High: the test file's `_resolve` helper is a pure-Python copy of this resolution logic and is validated against itself in `test_mlx_max_grad_value_none.py`. If you rename `_user_set_mgv`, extract the four-branch logic into a helper, or otherwise refactor without changing semantics, the source-pin test (`test_trainer_source_pins_resolution_rule`) breaks while the unit tests stay green — and vice versa, semantic-breaking refactors that preserve the substrings slide through. Couple them by lifting the resolution into a named function and importing it in the tests.

```suggestion
        else:
            # Neither requested -> cheap MLX default.
            max_grad_value = 1.0
        # NOTE: keep this four-branch resolution in sync with
        # tests/test_mlx_max_grad_value_none.py::_resolve. Prefer extracting
        # into a module-level helper so the tests can import it directly.
```

danielhanchen (commented):
Thank you for the PR! The goal of this PR is to fix a parity bug where a user passing `max_grad_norm=1.0` to `MLXTrainingConfig` was silently rebound to elementwise clip at +/-1.0 (HF/TRL parity break). As a summary, this PR keeps the field signature `max_grad_value: float | None = None` but reinterprets None inside `MLXTrainer.train`'s resolution as a sentinel meaning "cheap MLX elementwise clip 1.0 unless user opted into `max_grad_norm > 0` (then global-norm wins)". Explicit `0.0` disables; explicit positive overrides `max_grad_norm` with a warning.

| 1/2 | High | PR description and code drifted: the body says "default behavior unchanged" and the title says "honor None as a disable signal," but the diff also flips the field default from `1.0` to `None`, adds a new `max_grad_norm > 0` precedence branch, and silently changes behavior for pre-existing `MLXTrainingConfig(max_grad_norm=1.0)` callers (was elementwise@1 after warn-and-override; now global-norm@1). |
| 1/2 | High | The test's `_resolve(...)` helper is a hand-written Python copy of the trainer's resolution logic, validated against itself. The source-pin test is the only thing connecting them, and it grep-matches three literal substrings — a benign refactor inside `_train_inner` (e.g. extracting the resolution into a helper, renaming `_user_set_mgv`) breaks the pin without changing semantics. Couple them or instantiate a real `MLXTrainer` and inspect the resolved state. |
| 2/2 | Med | The resolution rule lives only in a trainer-side comment; the dataclass field docstring describes None/0.0/positive but does not say None silently becomes 0.0 when `max_grad_norm > 0`. Hoist into the field docstring. |
| 1/2 | Med | No runtime log of which clipping mode is active. Three different behaviors (`elementwise@1.0`, `norm@N`, both-disabled) run silently with `max_grad_value=None` — and they produce visibly different `grad_norm` numbers in callbacks. |
| 1/2 | Med | No end-to-end behavior test — only the resolver mirror. A future refactor that flips a branch inside the live `_train_inner` is not caught. |
| 1/2 | Low | Negative `max_grad_value` and `NaN` both pass `_user_set_mgv` but then fail the `> 0` gate silently. Old code had the same gap. |
| 1/2 | Low | The both-enabled mutual-exclusion warning is `print`, not `warnings.warn` — cannot be filtered or captured by `pytest.warns`. |
| 1/2 | Nit | `_user_set_mgv` is computed but unused outside the resolver. Drop or surface as a state field. |
| 1/2 | Nit | `getattr(args, "max_grad_value", None)` defensive default is unnecessary — the dataclass guarantees the attribute exists. |

Overall: REQUEST_CHANGES (PR body/title vs. actual semantics mismatch and silent regression of `max_grad_norm=1.0`-only configs must be addressed first; the rest are improvements.)

unsloth_zoo/mlx/trainer.py
```diff
@@ -723,19 +723,34 @@ def _train_inner(self):
...
+        _user_set_mgv = _raw_mgv is not None
+        if _user_set_mgv:
+            max_grad_value = float(_raw_mgv or 0.0)
+            if max_grad_norm > 0 and max_grad_value > 0:
+                print(
+                    "Unsloth: max_grad_norm and max_grad_value are both enabled; "
+                    "ignoring max_grad_norm in favor of max_grad_value."
+                )
+                max_grad_norm = 0.0
+        elif max_grad_norm > 0:
```
danielhanchen, line 748: **[1/2 reviewers]** Med: with `max_grad_value=None` and `max_grad_norm=0.0` we silently take the elementwise@1.0 branch; with `None` and `max_grad_norm>0` we silently take the global-norm branch; with `0.0` we silently disable. Print one line at step-build time so users debugging `grad_norm` curves know which clipping mode is active.

```suggestion
        elif max_grad_norm > 0:
            # User opted into global-norm clipping; suppress the default elementwise.
            max_grad_value = 0.0
            print(f"Unsloth: gradient clipping = global-norm at {max_grad_norm}.")
        else:
            # Neither requested -> cheap MLX default.
            max_grad_value = 1.0
            print("Unsloth: gradient clipping = elementwise at +/-1.0 (MLX default).")
```

unsloth_zoo/mlx/trainer.py
```diff
@@ -723,19 +723,34 @@ def _train_inner(self):
...
-        _raw_mgv = getattr(args, "max_grad_value", 1.0)  # TODO: expose MLX grad-clip in Studio UI for power users
-        max_grad_value = 1.0 if _raw_mgv is None else float(_raw_mgv or 0.0)
-        if max_grad_norm > 0 and max_grad_value > 0:
-            print(
-                "Unsloth: max_grad_norm and max_grad_value are both enabled; "
-                "ignoring max_grad_norm in favor of max_grad_value."
-            )
-            max_grad_norm = 0.0
+        _raw_mgv = getattr(args, "max_grad_value", None)  # TODO: expose MLX grad-clip in Studio UI for power users
+        _user_set_mgv = _raw_mgv is not None
```
danielhanchen, line 739: **[1/2 reviewers]** Nit: `_user_set_mgv` is computed but never used outside the resolver block. Inline it to keep the resolver self-contained, or surface it as `self._mgv_user_opt_in` for downstream introspection / callback reporting.

```suggestion
        if _raw_mgv is not None:
```
(removing the standalone `_user_set_mgv` binding and using the predicate inline in `if`/`elif`.)

tests/test_mlx_max_grad_value_none.py
```diff
@@ -0,0 +1,121 @@
...
+    user_set = raw_mgv is not None
+    if user_set:
+        mgv = float(raw_mgv or 0.0)
+        if max_grad_norm > 0 and mgv > 0:
+            max_grad_norm = 0.0
+    elif max_grad_norm > 0:
+        mgv = 0.0
+    else:
+        mgv = 1.0
+    return mgv, max_grad_norm
```
danielhanchen, line 31: **[1/2 reviewers]** High: `_resolve` here is a hand-written copy of `MLXTrainer._train_inner`'s resolution. Every "semantics" test (rows 1-5) validates this helper against itself — not the real trainer. The only thing connecting them is `test_trainer_source_pins_resolution_rule`, which is a substring grep. A semantic-breaking refactor that preserves the substrings passes; a benign refactor that breaks the substrings fails. Better: lift the resolver into `unsloth_zoo.mlx.trainer._resolve_grad_clip(raw_mgv, max_grad_norm)` and import it here.

```suggestion
from unsloth_zoo.mlx.trainer import _resolve_grad_clip as _resolve  # type: ignore[attr-defined]
```
(then the rest of the test file uses the real function rather than the local mirror.)

tests/test_mlx_max_grad_value_none.py
```diff
@@ -0,0 +1,121 @@
...
+def test_trainer_source_pins_resolution_rule():
+    """Source-level pin: trainer.py contains the four-branch resolution.
+    Cheap defense against a future refactor silently regressing the rule."""
+    import inspect
+    from unsloth_zoo.mlx import trainer as T
+
+    src = inspect.getsource(T.MLXTrainer.train) + inspect.getsource(T.MLXTrainer._train_inner)
+    assert "_user_set_mgv = _raw_mgv is not None" in src
+    assert "elif max_grad_norm > 0:" in src
+    assert "max_grad_value = 1.0" in src
```
danielhanchen, line 121: **[1/2 reviewers]** Med: this test pins three literal substrings (`_user_set_mgv = _raw_mgv is not None`, `elif max_grad_norm > 0:`, `max_grad_value = 1.0`). Two of them (`elif max_grad_norm > 0:` and `max_grad_value = 1.0`) are generic and would survive a branch-flip regression as long as the strings still appear somewhere in the function. Also, it concatenates `MLXTrainer.train` AND `MLXTrainer._train_inner` source even though `train` contains none of the resolution code — extra surface area for false positives.

```suggestion
def test_trainer_source_pins_resolution_rule():
    """Pin the four-branch resolution against silent regression. Prefer
    extracting the resolver into a helper and importing it directly; this
    source-pin is a last-resort tripwire."""
    import inspect
    from unsloth_zoo.mlx import trainer as T

    src = inspect.getsource(T.MLXTrainer._train_inner)
    # 'norm wins' branch must zero max_grad_value
    assert "max_grad_value = 0.0" in src
    # 'cheap default' branch must materialize 1.0
    assert "max_grad_value = 1.0" in src
    # The sentinel detection must distinguish 'user passed' from 'default'
    assert "is not None" in src
```

PR #673, opened 2026-05-19
fix(mlx): make_baseline_loss_fn byte-identical to mlx-lm default_loss when labels=None

danielhanchen:
## Summary
`make_baseline_loss_fn` previously routed both the `labels=None` case and the labels-aware case through the same code, with three small differences from `mlx_lm.tuner.trainer.default_loss`:

1. mask cast to fp32 (`length_mask.astype(mx.float32)`) instead of leaving the bool mask alone.
2. `safe_targets = mx.where(targets == -100, 0, targets)` even when there are no -100 entries in the `labels=None` path.
3. division by `_safe_token_denominator(ntoks)` (fp32 cast + maximum with 1.0) instead of raw `ce.sum() / ntoks`.

This PR restructures the function so the `labels=None` branch is byte-for-byte identical to mlx-lm's `default_loss`; the labels-aware branch keeps the existing -100 / safe_targets / fp32-mask machinery (it needs all three).

## Why
Mathematically all three differences are no-ops when there are no -100 entries (i.e. on the `labels=None` path), but each adds an extra node to the MLX autodiff graph and changes rounding inside the backward pass.

Empirical (`unsloth/gemma-3-270m-it` single-row LoRA memorization, paired-seed data from `danielhanchen/unsloth-staging-2` Round BO probes 31 (mlx-lm manual loop) vs 35 / 37 (zoo MLXTrainer), 5 representative seeds):

| step | probe 31 (mlx-lm) | probe 35 (zoo, compile=True) | probe 37 (zoo, compile=False, no clip) | $\Delta$ probe 31 - probe 35 |
|---|---|---|---|---|
| 1 | 9.769231 | 9.769231 | 9.769231 | 0 |
| 2 | 5.254807 | 5.276443 | 5.276443 | -0.021635 |
| 3 | 3.206731 | 3.221154 | 3.221154 | -0.014423 |
| 4 | 1.894231 | 1.947115 | 1.947115 | -0.052885 |
| 5 | 1.163462 | 1.168269 | 1.180288 | -0.004808 |
| 28+ | 0 | 0 | 0 | 0 |

Step 1's forward loss is identical (same initial weights, same data), but the gradient applied at step 1 differs between mlx-lm and zoo — so step 2 onward the loss curves are 0.01-0.06 apart until both converge to 0. The model still memorizes either way (`cf_loss = 0` in 15/15 seeds for every config). But callers cannot reproduce mlx-lm CLI's loss curve numerically, which makes diagnostic side-by-side comparisons useless.

After this PR the `labels=None` fast path produces an autodiff graph value-for-value identical to `mlx_lm.tuner.trainer.default_loss`, so probes against the same fixture should converge value-for-value with mlx-lm CLI.

## Behavior
- `loss_fn(model, batch, lengths)` (no labels) $\to$ identical numerical path to `mlx_lm.tuner.trainer.default_loss`. **Changed (parity).**
- `loss_fn(model, batch, lengths, labels)` (labels-aware, e.g. `train_on_responses_only`) $\to$ existing path with fp32 mask, `safe_targets`, `_safe_token_denominator`. **Unchanged.**

## Test plan
- [x] `tests/test_mlx_baseline_loss_parity.py` $\to$ 5 source-level assertions pinning the fast-path code patterns. Guards against future refactors silently re-introducing `length_mask.astype(mx.float32)`, `mx.where(safe_targets)`, or `_safe_token_denominator` on the `labels=None` branch. Verifies the labels-aware path still uses `safe_targets`.
- [x] Local: `pytest tests/test_mlx_baseline_loss_parity.py -v` $\to$ 5 passed.
- [x] Broader regression: `pytest tests/test_pr_a_*.py tests/test_mlx_*.py` $\to$ 58 passed, 1 skipped, no regressions.
- [ ] Round BP probe 38 on Apple Silicon (in flight) captures per-step loss + grad_norm for both paths to confirm value-for-value parity at all steps after this fix.

## Related
Sixth PR in the MLX vs `mlx-lm` parity series:
- `#669` $\to$ `finetune_last_n_layers` knob (zoo).
- `unslothai/unsloth#5564` $\to$ same knob, CUDA path.
- `#670` $\to$ warn on bf16$\to$fp16 downcast.
- `#671` $\to$ `max_grad_value=None` default + HF/TRL parity (closes `#662`).
- `#672` $\to$ `_create_labeled_batches` padding matches mlx-lm.
- This PR $\to$ `make_baseline_loss_fn` `labels=None` fast path matches mlx-lm.

### DIFF
--- a/tests/test_mlx_baseline_loss_parity.py
+++ b/tests/test_mlx_baseline_loss_parity.py
@@ -1,0 +1,124 @@
+# Unsloth Zoo - Utilities for Unsloth
+# Pin `make_baseline_loss_fn` source so the labels=None fast path stays
+# byte-for-byte equivalent to mlx_lm.tuner.trainer.default_loss.
+#
+# Why pin source rather than run a numerical comparison: the test
+# harness uses a torch-based MLX shim that doesn't faithfully reproduce
+# MLX's autodiff graph or its rounding; an apples-to-apples numerical
+# parity check requires a real MLX runtime (Apple Silicon), so it's
+# done in the Round BP probe matrix on
+# danielhanchen/unsloth-staging-2. Locally we guard against future
+# refactors silently re-introducing the divergent code patterns
+# (fp32-cast mask, mx.where(safe_targets), _safe_token_denominator)
+# that mlx_lm.tuner.trainer.default_loss does NOT do.
+
+from __future__ import annotations
+
+import inspect
+import re
+import textwrap
+
+import pytest
+
+
+@pytest.fixture(autouse=True, scope="module")
+def _install_mlx_shim():
+    from mlx_simulation import simulate_mlx_on_torch
+    simulate_mlx_on_torch()
+
+
+def _labels_none_block():
+    """Return CODE LINES (comments stripped) for the labels=None fast path."""
+    from unsloth_zoo.mlx import utils
+    src = inspect.getsource(utils.make_baseline_loss_fn)
+    m = re.search(
+        r"if labels is None:\s*\n(.*?)(?:# labels-aware path|else:\s*\n)",
+        src,
+        flags=re.DOTALL,
+    )
+    assert m, "make_baseline_loss_fn must keep a `labels is None` fast path"
+    raw = textwrap.dedent(m.group(1))
+    # Strip whole-line comments so test assertions on code don't trip on
+    # explanatory prose like "no safe_targets mx.where" in docstrings.
+    code_lines = []
+    for line in raw.splitlines():
+        stripped = line.strip()
+        if stripped.startswith("#"):
+            continue
+        code_lines.append(line)
+    return "\n".join(code_lines)
+
+
+def test_no_fp32_mask_cast_in_fast_path():
+    """The labels=None path must NOT cast the bool mask to fp32. mlx-lm's
+    default_loss multiplies the cross-entropy result by a raw bool mask;
+    casting to fp32 produces a different MLX autodiff graph and shifts
+    gradients by ~1e-2 per step on small fixtures."""
+    block = _labels_none_block()
+    # Anything that would cast a mask to fp32: `.astype(mx.float32)` or
+    # `astype(float32)` immediately on a `mask` / `length_mask` name.
+    bad_patterns = (
+        r"length_mask\.astype\(mx\.float32\)",
+        r"mask\s*=\s*[^=]*\.astype\(mx\.float32\)",
+    )
+    for pat in bad_patterns:
+        assert not re.search(pat, block), (
+            f"labels=None fast path must not contain `{pat}`; "
+            "matches mlx-lm requires a bool mask."
+        )
+
+
+def test_no_safe_targets_where_in_fast_path():
+    """The labels=None path has no -100 to worry about, so no mx.where
+    is needed on targets. The where node was empirically the cause of
+    step-2 loss divergence vs mlx-lm CLI (Round BO probe_31 vs probe_37)."""
+    block = _labels_none_block()
+    assert "mx.where" not in block, (
+        "labels=None fast path must not include mx.where on targets; "
+        "mlx-lm's default_loss does not call mx.where."
+    )
+    assert "safe_targets" not in block, (
+        "labels=None fast path must use `targets` directly; "
+        "`safe_targets = mx.where(...)` belongs only to the labels-aware path."
+    )
+
+
+def test_no_safe_token_denominator_in_fast_path():
+    """mlx-lm's default_loss divides by raw `ntoks` (int). The fast path
+    must match that to preserve the autodiff graph; the safety wrapper
+    `_safe_token_denominator` introduces a fp32 cast + maximum() that
+    changes rounding in MLX."""
+    block = _labels_none_block()
+    assert "_safe_token_denominator" not in block, (
+        "labels=None fast path must divide by raw ntoks for mlx-lm parity. "
+        "_safe_token_denominator is fine on the labels-aware path."
+    )
+
+
+def test_fast_path_returns_ce_and_ntoks_in_that_order():
+    """Match the (loss, ntoks) return signature mlx-lm uses; the test
+    pins return-order so a future refactor doesn't accidentally swap."""
+    block = _labels_none_block()
+    # Look for a `return X, Y` somewhere in the fast path. The variable
+    # names are loose (mlx-lm uses `ce`; zoo previously used `loss`),
+    # but the order matters.
+    m = re.search(r"return\s+(\w+),\s*(\w+)", block)
+    assert m, "labels=None fast path must return a (loss, ntoks) tuple"
+    loss_name, ntoks_name = m.group(1), m.group(2)
+    assert ntoks_name in ("ntoks", "n_toks", "ntokens"), (
+        f"second return value name should look like a token count, got "
+        f"{ntoks_name!r}"
+    )
+
+
+def test_labels_aware_path_still_uses_safe_targets():
+    """The labels-aware path (train_on_responses_only) DOES need
+    `safe_targets` and the fp32 mask because labels can contain -100."""
+    from unsloth_zoo.mlx import utils
+    src = inspect.getsource(utils.make_baseline_loss_fn)
+    # The labels-aware path lives after the fast path's `return`. Look
+    # at the full source to verify the machinery still exists somewhere.
+    assert "mx.where" in src, (
+        "make_baseline_loss_fn must still call mx.where on the labels-aware path"
+    )
+    assert "safe_targets" in src, "labels-aware path must keep safe_targets"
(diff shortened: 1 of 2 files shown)


danielhanchen:
**Correction: this PR is a code simplification, NOT a parity fix.**

Round BP probe 38 ran the mlx-lm manual loop and zoo `MLXTrainer` back-to-back in the same process, capturing per-step loss with the OLD `make_baseline_loss_fn` (fp32 mask cast + `mx.where(safe_targets)` + `_safe_token_denominator`). Result on 5 seeds (1, 42, 999, 3407, 22222):

| step | mlxlm loss | zoo loss | $\Delta$ loss |
|---|---|---|---|
| 1 | 9.769231 | 9.769231 | 0 |
| 2-30 | converging | converging | **0 (every step, every seed)** |

So the three "no-op math" differences (fp32 mask, `mx.where(safe_targets)`, `_safe_token_denominator`) ARE truly no-ops in MLX's autodiff. They add graph nodes but produce value-for-value identical gradients. My earlier diagnosis from Round BO data was wrong.

What ACTUALLY explains the Round BO step-2 divergence between `probe_31` (manual loop) and `probe_33/35/37` (zoo MLXTrainer) is the `mx.random.seed(seed)` order: probe 31 reseeds AFTER `mlx_load`; probes 33/35/37 didn't reseed and so their LoRA init used advanced RNG state. Probe 38's zoo path reseeds after load (matching mlx-lm CLI's `lora.py:223`) and the basins align step-for-step.

This PR is still defensible as a cleanup: the labels=None fast path now mirrors `mlx_lm.tuner.trainer.default_loss` line-for-line, removes three unused graph nodes per forward/backward, and keeps the `train_on_responses_only` path's -100 / safe_targets machinery on its own branch. No behavior change but a clearer contract. Re-framing the PR title + body to match — it's not closing a parity defect, just simplifying the labels=None path.

unsloth_zoo/mlx/utils.py
```diff
@@ -412,21 +412,28 @@ def make_baseline_loss_fn():
...
         When labels is provided, uses labels[:,1:] for targets with
-        (targets != -100) as the loss mask.
+        (targets != -100) as the loss mask. The labels=None branch is
+        byte-identical to ``mlx_lm.tuner.trainer.default_loss``.
     """
     def loss_fn(model, batch, lengths, labels=None):
         if labels is None:
-            inputs, targets = batch[:, :-1], batch[:, 1:]
-        else:
+            # byte-identical to mlx_lm.tuner.trainer.default_loss
```
danielhanchen, line 420: **[2/2 reviewers]** Med: `ce.astype(mx.float32).sum() / ntoks` matches mlx-lm exactly (no `_safe_token_denominator` guard), so byte parity is preserved — but an empty-mask batch (`lengths[:,0] > lengths[:,1]` for every row) now produces NaN/Inf instead of being smoothed by `max(1, ntoks)`. Realistic only with pathological inputs; mlx-lm has the same hazard. Add a one-line docstring note so the contract is explicit:

```suggestion
            # byte-identical to mlx_lm.tuner.trainer.default_loss
            # NOTE: caller must ensure ntoks > 0 (empty-mask batches
            # divide by zero). mlx-lm has the same precondition.
            inputs = batch[:, :-1]
```

danielhanchen (commented):
Thank you for the PR! The goal of this PR is to make `make_baseline_loss_fn`'s `labels=None` branch byte-identical to `mlx_lm.tuner.trainer.default_loss` so callers of `MLXTrainer` with no per-token label mask see value-for-value identical loss/grad sequences to the mlx-lm CLI. As a summary, this PR splits the previously-shared loss body into a fast path (`labels=None`: bool mask, raw `ntoks` division, no `safe_targets` `mx.where`) and a labels-aware path (unchanged — keeps `safe_targets`, fp32 mask cast, and `_safe_token_denominator` for `train_on_responses_only`).

| 2/2 | Med | Test is source-pin only (regex over the function body). No numerical/shape smoke. A semantic-breaking refactor that preserves the regex slides through. |
| 2/2 | Med | The new `labels=None` branch divides by raw `ntoks` (no `_safe_token_denominator` guard). mlx-lm has the same divide-by-zero hazard on empty-mask batches, so byte parity is preserved — but the previous code defensively handled it. Worth a docstring note. |
| 1/2 | Med | Existing `labels=None` callers see ~0.005-0.05 step-level loss drift vs prior versions. Per the PR's own table this only converges at the end; checkpoint-resume / regression-baseline comparisons will pick up the drift. Worth a release-note mention. |
| 2/2 | Low | The `_labels_none_block` regex in the test depends on the literal comment `# labels-aware path:` to delimit the block. Rename or reword the comment and all five tests silently extract the wrong block (or empty). |
| 1/2 | Nit | The "byte-identical to mlx_lm.tuner.trainer.default_loss" claim is true today but ties to mlx-lm's `main`. Pin a mlx-lm version range or commit SHA in the comment so a future maintainer can re-verify deterministically. |
| 1/2 | Nit | `test_fast_path_returns_ce_and_ntoks_in_that_order` accepts `ntoks`/`n_toks`/`ntokens`; tighten to `ntoks` since byte-identity is the explicit goal. |

Overall: APPROVE_WITH_NITS.

Verified: `make_cce_loss_fn` (line 266) and `make_vlm_cce_loss_fn` (line 1456) are not touched by this PR and share no state with the modified function — the MLX CCE kernel path is unaffected.

unsloth_zoo/mlx/utils.py
```diff
@@ -412,21 +412,28 @@ def make_baseline_loss_fn():
     Returns:
         A function (model, batch, lengths, labels=None) -> (loss, ntoks).
         When labels is provided, uses labels[:,1:] for targets with
-        (targets != -100) as the loss mask.
+        (targets != -100) as the loss mask. The labels=None branch is
+        byte-identical to ``mlx_lm.tuner.trainer.default_loss``.
```
danielhanchen, line 416: **[1/2 reviewers]** Nit: "byte-identical to mlx_lm.tuner.trainer.default_loss" is true today but tied to mlx-lm's `main`. Pin a known-good commit so a future maintainer can re-verify deterministically:

```suggestion
        byte-identical to ``mlx_lm.tuner.trainer.default_loss`` (verified
        against mlx-lm 0.x.y commit abc1234; rerun the byte-parity check
        if upstream changes).
```

tests/test_mlx_baseline_loss_parity.py
```diff
@@ -0,0 +1,124 @@
...
+import pytest
+
+
+@pytest.fixture(autouse=True, scope="module")
+def _install_mlx_shim():
+    from mlx_simulation import simulate_mlx_on_torch
+    simulate_mlx_on_torch()
+
+
+def _labels_none_block():
```
danielhanchen, line 30: **[2/2 reviewers]** Low: this helper's regex relies on either an `else:` or the literal comment `# labels-aware path` to mark the end of the fast-path block. If a future refactor renames that comment, all five tests below silently extract an empty or wrong block — `assert m` still fires (or the substring checks pass vacuously). Pin with a sentinel comment that's hard to rename casually:

```suggestion
def _labels_none_block():
    """Return the labels=None code block. Relies on the sentinel comment
    `# --- baseline-loss-fast-path-end ---` in the source. If you rename
    that comment, update this regex AND the sentinel in utils.py."""
    from unsloth_zoo.mlx import utils
    src = inspect.getsource(utils.make_baseline_loss_fn)
    m = re.search(
        r"if labels is None:\s*\n(.*?)# --- baseline-loss-fast-path-end ---",
        src,
        flags=re.DOTALL,
    )
    assert m, "sentinel comment missing from make_baseline_loss_fn fast path"
    return m.group(1)
```
(and add `# --- baseline-loss-fast-path-end ---` immediately before the labels-aware branch in `utils.py`.)

tests/test_mlx_baseline_loss_parity.py
```diff
@@ -0,0 +1,124 @@
...
+    """The labels-aware path (train_on_responses_only) DOES need
+    `safe_targets` and the fp32 mask because labels can contain -100."""
+    from unsloth_zoo.mlx import utils
+    src = inspect.getsource(utils.make_baseline_loss_fn)
+    # The labels-aware path lives after the fast path's `return`. Look
+    # at the full source to verify the machinery still exists somewhere.
+    assert "mx.where" in src, (
+        "make_baseline_loss_fn must still call mx.where on the labels-aware path"
+    )
+    assert "safe_targets" in src, "labels-aware path must keep safe_targets"
```
danielhanchen, line 124: **[2/2 reviewers]** Med: every test in this file is source-pin only — none actually executes the loss function on a tensor. The `mlx_simulation` shim is available; even a shape-only smoke (call `loss_fn(stub_model, batch, lengths, labels=None)` and assert the return is a 2-tuple of `(ce_scalar, ntoks_scalar)`) would catch refactors that preserve the regex but break the math. Add at the bottom of the file:

```suggestion
def test_fast_path_runtime_shape_smoke():
    """Shape-only end-to-end smoke. Catches refactors that pass every
    regex test but produce the wrong tensors at runtime."""
    import torch
    from unsloth_zoo.mlx.utils import make_baseline_loss_fn
    # ... build stub model whose __call__ returns logits of shape (B, T, V)
    # ... call loss_fn(stub, batch, lengths, labels=None)
    # ... assert tuple of length 2, both 0-dim tensors
```

PR #674, opened 2026-05-19
fix(mlx): seed mx.random immediately before linear_to_lora_layers

danielhanchen:
## Summary
Move `_seed_mlx_random_state(random_state)` inside `FastMLXModel.get_peft_model` to **immediately before** each `linear_to_lora_layers(...)` call (one for the VLM language branch, one for the text-only branch), instead of leaving it at the top of the method.

The current code seeds at the top, then runs ~165 source lines of target-module normalization, `_fix_missing_no_grad`, `_resolve_lora_keys`, etc. before the actual LoRA construction. That window lets lazy MLX state mutations or implicit `mx.random` consumption slip in, so the lora_a matrices that `linear_to_lora_layers` initializes end up *different* from what `mlx-lm`'s CLI gets — even though both paths re-seed to the same integer.

Stacked on `#669` (the `finetune_last_n_layers` PR) because probe 39 below needs that parameter to land.

## Why
`mlx-lm` CLI does the seeding **last** before LoRA: `mx.random.seed(args.seed)` is at `mlx_lm/lora.py:223`, with `linear_to_lora_layers` on the very next line. So the "right" mlx-lm-parity guarantee is "seed is the last thing before LoRA construction", and zoo's earlier-and-far-above placement broke that.

Empirical evidence from `danielhanchen/unsloth-staging-2` Round BQ probe 39 (5 paired seeds; both paths use `mlx_lm.load`-loaded weights to keep model state identical; only the LoRA-init pipeline differs):

| seed | max abs $\Delta$ loss | max abs $\Delta$ grad_norm |
|---|---|---|
| 1 | 1.019231 | 133.93 |
| 42 | 1.432692 | 60.92 |
| 999 | 0.168269 | 3.23 |
| 3407 | 0.110577 | 2.35 |
| 22222 | 0.105769 | 5.83 |

Step-1 forward loss is identical across both paths (`9.769231` in both — base weights load the same). Step-1 grad_norm differs by 2-3 units at low seeds (33.36 vs 36.11 at seed=1). The only thing that can produce that delta is different lora_a values, since at step 1 `lora_b = 0` and the only nonzero gradient flows through `dL/dlora_b = scale * dL/dout @ (lora_a @ input)^T`.

After this PR, the seed is the immediately-preceding statement to `linear_to_lora_layers`, so the lora_a initialization sees a fresh `mx.random.state == random_state` and matches `mlx-lm` CLI value-for-value (Round BQ rerun pending to confirm on Apple Silicon).

## Behavior
- Default API surface and behavior **unchanged** for callers. The seed move is purely internal.
- `random_state` parameter and its default of `3407` are unchanged.
- The VLM language LoRA branch (line 2842) and text-only branch (line 2933) each get the seed call moved right above them.

## Test plan
- [x] `tests/test_mlx_get_peft_model_seed_ordering.py` pins:
  1. Every `linear_to_lora_layers` call inside `get_peft_model` is preceded by `_seed_mlx_random_state` within 20 source lines.
  2. The `random_state=3407` default stays.
  3. The seed-immediately-precedes pattern matches both LoRA call sites (VLM + text).
- [x] `tests/test_mlx_finetune_last_n_layers.py` `FakeModel` gets `trainable_parameters()` / `parameters()` stubs so the param-count log at the tail of `get_peft_model` doesn't crash the unit test (pre-existing failure on `fix-mlx-num-layers-parity`, fixed here).
- [x] Local: `pytest tests/test_mlx_finetune_last_n_layers.py tests/test_mlx_get_peft_model_seed_ordering.py tests/test_pr_a_*.py` $\to$ 58 passed.
- [ ] Round BQ rerun on Apple Silicon with this branch pinned will confirm `dloss = 0` step-for-step against the mlx-lm-CLI style path.

## Related
Seventh PR in the MLX vs `mlx-lm` parity series, now addressing the user-reported asymmetry: "the seed fix should be on Unsloth's side, not require manual `mx.random.seed(seed)` from users":

- `#669` $\to$ `finetune_last_n_layers` knob (base for this PR).
- `unslothai/unsloth#5564` $\to$ same knob on CUDA path.
- `#670` $\to$ warn on bf16 $\to$ fp16 downcast.
- `#671` $\to$ `max_grad_value=None` default + HF/TRL parity (closes `#662`).
- `#672` $\to$ `_create_labeled_batches` padding matches `mlx-lm`.
- `#673` $\to$ `make_baseline_loss_fn` labels=None fast-path simplification.
- This PR $\to$ seeding ordered immediately before `linear_to_lora_layers`, closing the basin-family gap end-to-end on the user-facing `FastMLXModel` API.

### DIFF
--- a/tests/test_mlx_get_peft_model_seed_ordering.py
+++ b/tests/test_mlx_get_peft_model_seed_ordering.py
@@ -1,0 +1,98 @@
+# Unsloth Zoo - Utilities for Unsloth
+# Pin the seed-immediately-before-linear_to_lora_layers ordering in
+# FastMLXModel.get_peft_model so we match mlx-lm CLI's basin family.
+#
+# The actual numerical test (lora_a values bit-identical to
+# `mlx_lm.tuner.utils.linear_to_lora_layers` after `mx.random.seed`)
+# needs real MLX runtime; that's covered by probe 39 on
+# danielhanchen/unsloth-staging-2. Locally we guard against a future
+# refactor reverting to a single `_seed_mlx_random_state` call far
+# above `linear_to_lora_layers` again.
+
+from __future__ import annotations
+
+import inspect
+import re
+
+import pytest
+
+
+@pytest.fixture(autouse=True, scope="module")
+def _install_mlx_shim():
+    from mlx_simulation import simulate_mlx_on_torch
+    simulate_mlx_on_torch()
+
+
+def _get_peft_model_source():
+    from unsloth_zoo.mlx.loader import FastMLXModel
+    return inspect.getsource(FastMLXModel.get_peft_model)
+
+
+def test_seed_immediately_precedes_each_linear_to_lora_layers_call():
+    """Every `linear_to_lora_layers(...)` call inside get_peft_model
+    must be preceded -- within ~20 source lines, with no other MLX op
+    or model-tree walk in between -- by a `_seed_mlx_random_state` call.
+
+    Background: empirically (Round BQ probe 39 on Apple Silicon), having
+    the seed call >100 lines above the LoRA construction allows lazy MLX
+    state advances to slip in between seeding and lora_a init, producing
+    lora_a matrices different from mlx-lm CLI's despite both paths
+    re-seeding to the same int. The fix is to seed immediately above
+    each `linear_to_lora_layers` call.
+    """
+    src = _get_peft_model_source()
+    lines = src.splitlines()
+
+    # Indices of every linear_to_lora_layers invocation (filter to actual
+    # call sites, not comments).
+    call_lines = [
+        i for i, line in enumerate(lines)
+        if "linear_to_lora_layers(" in line and not line.strip().startswith("#")
+    ]
+    assert call_lines, "expected at least one linear_to_lora_layers call in get_peft_model"
+
+    for call_idx in call_lines:
+        # Walk back up to 20 lines looking for a _seed_mlx_random_state call.
+        window_start = max(0, call_idx - 20)
+        window = "\n".join(lines[window_start:call_idx])
+        assert "_seed_mlx_random_state" in window, (
+            f"linear_to_lora_layers at line {call_idx+1} of get_peft_model "
+            f"is not preceded by `_seed_mlx_random_state` within the prior "
+            f"20 lines. Move the seed call closer to the LoRA construction "
+            f"to keep mlx-lm CLI parity."
+        )
+
+
+def test_get_peft_model_still_accepts_random_state_arg():
+    """Ensure the API surface didn't change while moving the seed."""
+    import inspect
+    from unsloth_zoo.mlx.loader import FastMLXModel
+    sig = inspect.signature(FastMLXModel.get_peft_model)
+    assert "random_state" in sig.parameters
+    assert sig.parameters["random_state"].default == 3407
+
+
+def test_no_other_mlx_op_between_seed_and_linear_to_lora_layers():
+    """Between the seed-just-above-LoRA call and the linear_to_lora_layers
+    call itself, the only allowed code is the LoRA call. Anything else --
+    even pure Python -- is a tripwire because it makes it easy to slip
+    in an mx.eval/random call later. Enforced as a regex tripwire."""
+    src = _get_peft_model_source()
+    # Match: _seed_mlx_random_state(random_state), then ONLY whitespace + a
+    # comment block + the linear_to_lora_layers call. Comments are fine.
+    # Anything else (an `mx.` op, a module walk, an assignment) is not.
+    pattern = re.compile(
+        r"_seed_mlx_random_state\(random_state\)\s*"
+        r"(?:\n\s*#[^\n]*)*"  # any number of comment-only lines
+        r"\s*\n\s*linear_to_lora_layers\(",
+    )
+    # We have TWO linear_to_lora_layers calls (VLM language path + text path).
+    # Both need this tight coupling.
+    matches = pattern.findall(src)
+    assert len(matches) >= 2, (
+        "Expected at least two seed+LoRA-call tight pairings in "
+        "get_peft_model (one for VLM language LoRA, one for text). "
+        f"Found {len(matches)}. The seed call must immediately precede "
+        "each linear_to_lora_layers invocation with only comment lines "
+        "in between."
+    )
--- a/unsloth_zoo/mlx/loader.py
+++ b/unsloth_zoo/mlx/loader.py
@@ -2741,9 +2741,19 @@ class FastMLXModel:
                 "Install via: pip install unsloth-zoo[mlx]"
             )
 
-        # Seed mlx random state so LoRA matrix init (lora_b is zero, lora_a
-        # is random Gaussian) is reproducible across runs.
-        _seed_mlx_random_state(random_state)
+        # Note: the actual `mx.random.seed(random_state)` is moved closer to
+        # each `linear_to_lora_layers` call below. Calling it here -- and then
+        # running >100 lines of target-module normalization, `_fix_missing_no_grad`,
+        # `_resolve_lora_keys`, and (on VLM) `_fix_gemma4_kv_sharing` before the
+        # actual LoRA construction -- leaves a window in which lazy MLX
+        # evaluations can advance `mx.random.state` so that lora_a values end up
+        # different from what `mlx-lm`'s CLI gets after `mx.random.seed(seed)` at
+        # `mlx_lm/lora.py:223` (the seed is the last thing CLI does before
+        # `linear_to_lora_layers`). Empirically (probe 39 on
+        # danielhanchen/unsloth-staging-2) step-1 grad_norm diverged by 2-3 units
+        # vs the mlx-lm-direct path on the same seed even though the model
+        # weights loaded identically; seeding immediately before
+        # `linear_to_lora_layers` closes that window.
 
         # finetune_vision_layers (None = use train_vision arg; bool overrides it)
         if finetune_vision_layers is not None:
@@ -2829,6 +2839,10 @@ class FastMLXModel:
                     num_layers = max(1, min(int(finetune_last_n_layers), num_layers))
                 language_lora_keys = _resolve_lora_keys(lm, target_modules)
                 if language_lora_keys is None or len(language_lora_keys) > 0:
+                    # Reseed mx.random immediately before LoRA construction to
+                    # match mlx-lm CLI's lora.py:223 ordering exactly. See
+                    # rationale on the comment block at the top of this method.
+                    _seed_mlx_random_state(random_state)
                     linear_to_lora_layers(
                         lm,
                         num_layers=num_layers,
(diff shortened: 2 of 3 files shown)


danielhanchen:
## Round BR verification on Apple Silicon

Re-ran the parity probe matrix on `danielhanchen/unsloth-staging-2` with `ZOO_SPEC` pinned to this branch's HEAD (`0124424`):
[Run 26078606176](https://github.com/danielhanchen/unsloth-staging-2/actions/runs/26078606176) — 50 / 50 jobs green.

### Probe 39 (strict per-step diagnostic)

`FastMLXModel.from_pretrained + FastMLXModel.get_peft_model(finetune_last_n_layers=16)` vs `mlx_lm.load + linear_to_lora_layers(num_layers=16)`, both feeding the same manual `@mx.compile` training loop.

| seed | max \|dloss\| | max \|dgrad_norm\| |
|---|---|---|
| 1 | 0.0 | 0.0 |
| 42 | 0.0 | 0.0 |
| 999 | 0.0 | 0.0 |
| 3407 | 0.0 | 0.0 |
| 22222 | 0.0 | 0.0 |

Bit-identical losses AND gradient norms across all 30 steps x 5 seeds. Before this PR, lora_a values diverged from step 1 because the `_seed_mlx_random_state(random_state)` call sat about 165 lines above `linear_to_lora_layers`, leaving a window for lazy MLX state advances between seeding and LoRA init. Moving the seed call to immediately above each `linear_to_lora_layers(...)` (matching `mlx_lm/tuner/lora.py:223`) closes the gap.

### Side observation (not addressed by this PR)

Probes 34 / 36 (`FastMLXModel + MLXTrainer`) still hit 47% greedy pass vs probe 31's (`mlx-lm CLI manual loop`) 67% on the same 15 seeds, with probes 34 and 36 sharing an identical seed pattern (so `compile=True/False` is a no-op for the basin). The residual gap is downstream of `get_peft_model` and is separate from what this PR fixes -- planning a follow-up bisection (Round BS).

unsloth_zoo/mlx/loader.py
```diff
@@ -2820,15 +2816,14 @@ def get_peft_model(
...
-                # (= all layers, matching PEFT/CUDA semantics); pass an int
-                # to mirror mlx-lm CLI behavior.
                 if finetune_last_n_layers is not None and num_layers > 0:
                     num_layers = max(1, min(int(finetune_last_n_layers), num_layers))
                 language_lora_keys = _resolve_lora_keys(lm, target_modules)
                 if language_lora_keys is None or len(language_lora_keys) > 0:
+                    # Match mlx_lm/lora.py:223 — seed mx.random immediately
+                    # before LoRA init; lazy MLX state advances otherwise
+                    # leak into lora_a sampling.
+                    _seed_mlx_random_state(random_state)
```
danielhanchen, line 2826: **[2/2 reviewers]** Med — scope-of-fix documentation. This tight seed→LoRA pairing only covers the language-LoRA via `linear_to_lora_layers`. Further down (VLM branch), `_lora_walk_module` is called for `vision_tower`, `vision_model`, `vision_encoder` and `projector`/`connector` — each of those constructs `LoRALinear.from_base` which consumes `mx.random.uniform` for its own `lora_a` init with NO re-seed. So a user training a VLM with `train_vision=True` and `random_state=3407` still ends up with non-deterministic-vs-mlx-lm-CLI vision `lora_a` values. mlx-lm CLI doesn't LoRA vision, so "parity" is undefined — but the comment should make the scope explicit:

```suggestion
                    # Match mlx_lm/tuner/lora.py (def train) -- seed
                    # mx.random immediately before LoRA init so lazy MLX
                    # state advances do not leak into lora_a sampling.
                    # Scope: language LoRA only. Vision tower and
                    # projector LoRA inits via _lora_walk_module below
                    # are NOT covered (mlx-lm CLI does not LoRA vision).
                    _seed_mlx_random_state(random_state)
```

danielhanchen (commented):
Thank you for the PR! The goal of this PR is to make `FastMLXModel.get_peft_model`'s `lora_a` init match `mlx_lm/tuner/lora.py:223` value-for-value by seeding `mx.random` immediately above each `linear_to_lora_layers` call, eliminating a window in which MLX's lazy evaluation can advance the global state between the previous early seed call and the actual LoRA construction. As a summary, this PR deletes the seed call at the top of `get_peft_model` and inserts `_seed_mlx_random_state(random_state)` immediately above both the VLM-language and text-only `linear_to_lora_layers` invocations. Verified empirically (probe 39 on `danielhanchen/unsloth-staging-2`) to produce `dloss = 0` step-for-step against the mlx-lm CLI path.

| 2/2 | Med | The fix's parity claim only holds for the language-LoRA via `linear_to_lora_layers`. Vision-tower / multimodal-projector LoRA inits (via `_lora_walk_module` -> `LoRALinear.from_base`) run after `linear_to_lora_layers` consumes RNG and are NOT seeded immediately, so VLM users with `train_vision=True` still have non-deterministic-relative-to-mlx-lm-CLI vision `lora_a` values. mlx-lm CLI itself does not LoRA vision, so "parity" is technically undefined there, but the scope should be documented. |
| 1/2 | Med | Non-default `init_lora_weights="gaussian"` / `False` paths run `_apply_mlx_lora_initialization` after `linear_to_lora_layers` with no re-seed, so those modes inherit whatever state the LoRA construction left behind. Worth a TODO or a comment. |
| 1/2 | Med | Stacked-PR merge order matters: this branch contains #669's commit. If #674 merges before #669, the deletion of the top-of-method seed lands without the `finetune_last_n_layers` param it's stacked on. Confirm merge order #669 → #674. |
| 1/2 | Med | Fix depends on `linear_to_lora_layers` consuming `mx.random.uniform`. If mlx-lm migrates to explicit-state RNG (`mx.random.key(...)`), the global seed becomes a no-op silently. Suggest a numerical canary test that pins `mx.random.uniform(shape=(2,))` after `_seed_mlx_random_state(3407)`. |
| 2/2 | Low | The seed-ordering tripwire test uses regex pinning on `_seed_mlx_random_state(random_state)` literal — a benign refactor (`seed=random_state` kwarg form, helper wrapper) breaks it without an actual ordering regression. |
| 2/2 | Nit | `mlx_lm/lora.py:223` line-number reference will go stale fast; cite function name (`def train`) instead. |

Overall: APPROVE_WITH_NITS. The seed-ordering fix is correctly placed for the explicit scope of the PR (language LoRA via `linear_to_lora_layers`); the [Med] findings are pre-existing scope-of-fix limits that this PR doesn't claim to address but should document.

unsloth_zoo/mlx/loader.py
```diff
@@ -2907,16 +2902,15 @@ def get_peft_model(
...
-                # to mirror mlx-lm CLI behavior.
                 if finetune_last_n_layers is not None and num_layers > 0:
                     num_layers = max(1, min(int(finetune_last_n_layers), num_layers))
                 language_lora_keys = _resolve_lora_keys(model, target_modules)
                 if language_lora_keys is not None and len(language_lora_keys) == 0:
                     _raise_no_lora_targets(target_modules)
+                # Match mlx_lm/lora.py:223 — seed mx.random immediately
+                # before LoRA init; lazy MLX state advances otherwise
+                # leak into lora_a sampling.
+                _seed_mlx_random_state(random_state)
```
danielhanchen, line 2913: **[2/2 reviewers]** Nit: the comment `# Match mlx_lm/lora.py:223` is brittle to upstream renumber. Cite the function instead:

```suggestion
                # Match mlx_lm/tuner/lora.py (def train) -- seed
                # mx.random immediately before LoRA init; lazy MLX state
                # advances otherwise leak into lora_a sampling.
                _seed_mlx_random_state(random_state)
```

tests/test_mlx_get_peft_model_seed_ordering.py
```diff
@@ -0,0 +1,98 @@
...
+def test_no_other_mlx_op_between_seed_and_linear_to_lora_layers():
+    """Between the seed-just-above-LoRA call and the linear_to_lora_layers
+    call itself, the only allowed code is the LoRA call. Anything else --
+    even pure Python -- is a tripwire because it makes it easy to slip
+    in an mx.eval/random call later. Enforced as a regex tripwire."""
+    src = _get_peft_model_source()
+    # Match: _seed_mlx_random_state(random_state), then ONLY whitespace + a
+    # comment block + the linear_to_lora_layers call. Comments are fine.
+    # Anything else (an `mx.` op, a module walk, an assignment) is not.
+    pattern = re.compile(
```
danielhanchen, line 84: **[2/2 reviewers]** Low: this regex requires the literal token `_seed_mlx_random_state(random_state)` to live on a line directly above `linear_to_lora_layers(`. Benign refactors that don't change ordering (`_seed_mlx_random_state(int(random_state))`, kwarg form `_seed_mlx_random_state(seed=random_state)`, or wrapping in a one-liner helper `_seed_then_lora(...)`) would break the test without an actual ordering regression. Also, `len(matches) >= 2` could pass with one good pair and one bad — pin each call site explicitly.

```suggestion
    # Collect each linear_to_lora_layers call site, then verify each one
    # is preceded within ~5 lines by a seed call. This pin is structural,
    # not regex-literal, so benign refactors of the seed expression are OK.
    src = _get_peft_model_source()
    lines = src.splitlines()
    call_sites = [
        i for i, line in enumerate(lines)
        if "linear_to_lora_layers(" in line and not line.strip().startswith("#")
    ]
    assert len(call_sites) >= 2, (
        f"expected at least two linear_to_lora_layers call sites; got {len(call_sites)}"
    )
    for idx in call_sites:
        window = "\n".join(lines[max(0, idx - 5):idx])
        assert "_seed_mlx_random_state" in window, (
            f"linear_to_lora_layers at relative line {idx+1} is not preceded by "
            f"_seed_mlx_random_state within 5 lines. Window:\n{window}"
        )
```

tests/test_mlx_get_peft_model_seed_ordering.py
```diff
@@ -0,0 +1,98 @@
...
+        window_start = max(0, call_idx - 20)
+        window = "\n".join(lines[window_start:call_idx])
+        assert "_seed_mlx_random_state" in window, (
+            f"linear_to_lora_layers at line {call_idx+1} of get_peft_model "
+            f"is not preceded by `_seed_mlx_random_state` within the prior "
+            f"20 lines. Move the seed call closer to the LoRA construction "
+            f"to keep mlx-lm CLI parity."
+        )
+
+
```
danielhanchen, line 65: **[1/2 reviewers]** Med: this test file currently has no NUMERICAL canary. If mlx-lm changes their LoRA init to use explicit-state RNG (`mx.random.key(...)`) instead of the global state, the global-seed-call-then-LoRA pattern silently becomes a no-op and the parity claim regresses. The source-pinning tests would still pass. Add a tiny canary:

```suggestion
def test_seed_actually_resets_mx_random_state():
    """Numerical canary: confirm `_seed_mlx_random_state(seed)` causes
    the next `mx.random.uniform(...)` to return a deterministic value.
    If mlx-lm migrates to explicit RNG state, this fires before the
    source-pin tests do."""
    import mlx.core as mx
    from unsloth_zoo.mlx.loader import _seed_mlx_random_state
    _seed_mlx_random_state(3407)
    a = float(mx.random.uniform(shape=(1,)).item())
    _seed_mlx_random_state(3407)
    b = float(mx.random.uniform(shape=(1,)).item())
    assert a == b, (
        "Re-seeding mx.random produced different uniform draws -- the "
        "global-state seeding contract may have broken. Investigate before "
        "trusting the seed-ordering parity claim in get_peft_model."
    )
```

PR #676, opened 2026-05-19
tests/offline-cross-sync: realign with #675 broadened truthy parse

shimmyshimmer:
## Summary

[#675](https://github.com/unslothai/unsloth-zoo/pull/675) (Honour offline env vars when enabling hf_transfer) replaced the "exact `1` only" trigger with a truthy set (`{\"1\",\"true\",\"yes\",\"on\"}`, case-insensitive) and made the cross-sync block write literal `\"1\"` to all three vars instead of leaving the source string in place. `tests/test_offline_env_cross_sync.py::test_only_value_1_triggers` still asserted the old contract (`HF_HUB_OFFLINE=\"true\"` left as-is, no cross-sync), so it broke on the very first CI run after #675 merged and is currently red on every downstream PR's `Core (HF=latest + TRL=latest)` matrix in `unslothai/unsloth`.

## Changes

- Drop the stale `test_only_value_1_triggers` (its name and assertions both contradict the new contract).
- Add a parametrised replacement that exercises every spelling the new detector accepts and asserts the documented normalisation:
  - `true`, `True`, `TRUE`, `yes`, `on` all trigger cross-sync.
  - All three vars come back as the literal `\"1\"`.
- Add `test_falsy_value_does_not_trigger` (`HF_HUB_OFFLINE=\"0\"`) to lock in the negative path.

## Test plan

\`\`\`
$ pytest tests/test_offline_env_cross_sync.py -v
...
10 passed in 0.07s
\`\`\`

Verified locally on Python 3.12 with \`unsloth_zoo @ main\` (post-#675).

### DIFF
--- a/tests/test_offline_env_cross_sync.py
+++ b/tests/test_offline_env_cross_sync.py
@@ -71,11 +71,20 @@ class TestOfflineCrossSync:
         assert os.environ.get("TRANSFORMERS_OFFLINE") == "1"
         assert os.environ.get("HF_DATASETS_OFFLINE") == "1"
 
-    def test_only_value_1_triggers(self, monkeypatch, reload_zoo):
-        # The cross-sync only matches the literal "1" (existing convention).
-        # Other truthy spellings are left as-is; documented behaviour.
-        monkeypatch.setenv("HF_HUB_OFFLINE", "true")
+    @pytest.mark.parametrize("value", ["true", "True", "TRUE", "yes", "on"])
+    def test_truthy_spellings_trigger_and_normalize(self, monkeypatch, reload_zoo, value):
+        # Since #675, any value in {"1","true","yes","on"} (case-insensitive)
+        # triggers cross-sync and normalizes all three to "1".
+        monkeypatch.setenv("HF_HUB_OFFLINE", value)
         reload_zoo()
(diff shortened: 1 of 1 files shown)


danielhanchen:
Superseded by 2da104d on main (`tests: update offline cross-sync test to match widened truthy parsing`), which landed first and already realigns `tests/test_offline_env_cross_sync.py` with the broadened truthy parse from #675:

- Drops `test_only_value_1_triggers`.
- Adds `test_truthy_spellings_trigger_cross_sync_and_normalize` covering `true`, `yes`, `on`, `TRUE`, `Yes`, `On`.
- Adds `test_non_truthy_value_leaves_env_unchanged` pinning the `"0"` pass-through.

CI on main is green for this file, so the breakage this PR was opened to fix no longer exists. The remaining delta vs. this PR is cosmetic (a `pytest.mark.parametrize` vs. an inline `for` loop), and main's version actually covers two extra mixed-case spellings (`Yes`, `On`).

Closing as superseded. The Gemini suggestions (parametrise truthy with whitespace and `"1"`, parametrise the falsy test) are worth keeping in mind, but they apply to the current main tests too and can be a separate cleanup PR if we want them.

PR #679, opened 2026-05-19
fix(mlx): persist LoRA adapter metadata on save

Lyxot:
## Summary

Fix MLX LoRA adapter saves so `adapter_config.json` contains the metadata needed to recreate adapters faithfully on reload.

- Persist live LoRA `rank`, `scale`, and `dropout` into both `lora_parameters` and top-level compatibility fields
- Read MLX dropout correctly from either `.p` or MLX's `_p_1` keep-probability storage
- Infer LoRA rank from the correct tensor axis, including switch/MoE LoRA tensors shaped like `(num_experts, r, input_dims)`
- Preserve explicit `unsloth_mlx_lora_module_paths` instead of overwriting caller-provided topology metadata
- Re-apply eval mode after adapter reload so nonzero-dropout adapters do not run dropout during default inference
- Share the same rank/dropout inference helpers between adapter config enrichment and `MLXTrainer.save_model`

## Why

Several MLX save paths call `save_lora_adapters()` without a complete explicit adapter config. Missing or stale LoRA metadata can make reload reconstruct adapters with different rank, scale, or dropout settings, which changes post-reload behavior.

This also fixes two edge cases:
- MLX `Dropout` stores keep probability as `_p_1`, so reading only `.p` writes stale `dropout=0.0`.
- `LoRASwitchLinear` stores rank on `lora_a.shape[-2]`, not `shape[-1]`.


danielhanchen:

This PR makes MLX adapter saves faithfully persist live rank, scale, and dropout (correctly reading MLX's _p_1 keep-probability and MoE switch tensor axis), extends adapter reload to Linear/Switch/Embedding LoRA classes with model.eval() after reload, and adds compatibility-fallback attribute restoration. After review-driven hardening (explicit-filter scoping, dropout restoration in older mlx-lm fallback, None-safety in rank inference, dead-code removal), the change cleanly delivers correct save/reload round-trip behavior for MLX LoRA adapters.

**Reason:** Core MLX LoRA metadata fixes are correct; all real issues found across three review iterations were fixed.

danielhanchen:

PR adds rank/scale/dropout/path/num_layers persistence and is_layer-aware rank inference for MLX LoRA adapter save+reload, plus reload-time wrapping for Switch/Embedding LoRA and a compat fallback for older mlx-lm from_base signatures; without it, reloaded adapters silently default to scale=1.0 and stale dropout, changing post-reload logits.

**Reason:** Real metadata-persistence bug fixed, reviewer-flagged trainer asymmetry resolved, all loop 2 findings false-positive/hypothetical, 21 regression tests pass.

PR #682, opened 2026-05-20
fix(mlx): handle zero-token and invalid labels in CCE

Lyxot:
## Summary
Hardens the MLX fused CCE runtime against two edge cases:
- zero-token inputs
- invalid target labels outside `[0, vocab_size)` except `ignore_index`

## Behavior
Before this change:
- zero-token inputs could reach the Metal kernel with an empty grid and crash the process.
- invalid labels such as `-1` or `target == vocab_size` could produce finite-looking losses instead of surfacing bad input.

## Fix
- Return empty loss/LSE tensors for zero-token forward passes before launching Metal kernels.
- Return zero gradients for zero-token dense and quantized backward paths.
- Mark invalid-label rows with `NaN` loss/LSE so bad labels cannot silently train with plausible loss values.
- Keep invalid-label handling compile-safe under `mx.compile`.

## Validation
```bash
python -m pytest tests/test_mlx_runtime_cce_compile.py tests/test_mlx_baseline_loss_parity.py -q
```


danielhanchen:
**Auto-review verdict: Changes requested**

**Reason:** Deterministic gate: accumulated test(s) regressed after post_sim_commit at iteration 2 (1 regression event(s) total)

danielhanchen:

Hardens unsloth_zoo/mlx/cce/runtime_cce.py against (a) zero-token inputs that crash the Metal kernel via an empty grid and (b) out-of-vocab labels that previously produced finite-garbage losses and gradients; adds upfront shape validation, NaN poisoning of loss/lse for invalid labels in both fallback and Metal-finalize paths, an isnan guard in the dlogits backward kernel, and 7+ tests covering eager, compiled, and quantized paths.

**Reason:** All real review findings fixed; no remaining bugs; PR correctly addresses two concrete bugs in MLX CCE.

PR #687, opened 2026-05-22
Fix llama.cpp build failure when saving model to GGUF

harriebird:
Building `llama.cpp` using `make` is already replaced by `cmake`, which causes an error. Building via `make` in the `install_llama_cpp()` was removed for Linux/macOS systems. For more information about this, here's the latest `Makefile` content:
https://github.com/ggml-org/llama.cpp/blob/master/Makefile

Another issue is that deprecation warnings when calling `try_execute()` raises an `RuntimeError` by default. This causes the building of `llama.cpp` from source to stop. To fix this,  `ignore_deprecation` was set to `True` when calling `try_execute()` to build the `llama.cpp`.

This solves issue unslothai/unsloth#5832.

Screenshot for the llama.cpp build after cloning its repository:

Screenshot 2026-05-22 at 12-31-01 fine-tune ipynb (4) - JupyterLab(1)

Screenshot for the llama.cpp build using the existing cloned source code:

Screenshot 2026-05-22 at 12-22-26 fine-tune ipynb (4) - JupyterLab(1)

### DIFF
--- a/unsloth_zoo/llama_cpp.py
+++ b/unsloth_zoo/llama_cpp.py
@@ -802,75 +802,65 @@ def install_llama_cpp(
             build_errors.append(f"Windows cmake build failed: {str(e)}")
 
     else:
-        # Linux/macOS: Try make first, then cmake
+        # Linux/macOS: Use cmake to build (build using make is already removed from llama.cpp)
         try:
-            if print_output: print("Trying to build with make...")
-            try_execute("make clean", cwd = llama_cpp_folder, **kwargs)
-            try_execute(f"make all -j{cpu_count}", cwd = llama_cpp_folder, **kwargs)
+            # Clean up any partial build
+            try_execute(f"rm -rf build", cwd = llama_cpp_folder, **kwargs)
+
+            # Build cmake configure command with library detection
+            cmake_configure = (
+                f"cmake . -B build "
+                f"-DBUILD_SHARED_LIBS=OFF -DGGML_CUDA={gpu_support}"
+            )
+
+            # Detect OpenMP library path (fixes GOMP linker errors)
+            gomp_path = _find_lib_path('libgomp.so')
+            if gomp_path:
+                cmake_configure += (
+                    f" -DOpenMP_C_LIB_NAMES=gomp"
+                    f" -DOpenMP_CXX_LIB_NAMES=gomp"
+                    f" -DOpenMP_gomp_LIBRARY={gomp_path}"
+                )
+
+            # Detect OpenSSL library paths
+            ssl_path = _find_lib_path('libssl.so')
+            crypto_path = _find_lib_path('libcrypto.so')
+            if ssl_path and crypto_path:
+                cmake_configure += (
+                    f" -DOPENSSL_ROOT_DIR=/usr"
+                    f" -DOPENSSL_SSL_LIBRARY={ssl_path}"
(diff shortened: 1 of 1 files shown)


unsloth_zoo/llama_cpp.py
```diff
@@ -809,75 +809,66 @@ def install_llama_cpp(
             build_errors.append(f"Windows cmake build failed: {str(e)}")
 
     else:
-        # Linux/macOS: Try make first, then cmake
+        # Linux/macOS: Use cmake to build (build using make is already removed from llama.cpp)
```
Datta0, line 813: We need to ensure that cmake dependency is installed already and is the right verion for this
I'm not entirely sure why we were preferring make over cmake previously @danielhanchen any idea?

harriebird: @Datta0, llama.cpp build indeed fails when cmake is not yet installed. I will try to improve the code for this PR to cater such scenario. There's already a check but it leads to error.

harriebird: Hi @Datta0 and @danielhanchen, I updated the implementation regarding the installation of system packages required to build `llama.cpp` from source. I updated the implementation of `do_we_need_sudo()` and `install_package()`. It can now ask for a password if sudo is required for the installation of the system packages.

unsloth_zoo/llama_cpp.py
```diff
@@ -284,11 +285,14 @@ def install_package(package, sudo = False, print_output = False, print_outputs =
                 f"Unsloth: Execution of `{install_cmd}` was cancelled!\n"\
                 "Please install llama.cpp manually via https://docs.unsloth.ai/basics/troubleshooting-and-faqs#how-do-i-manually-save-to-gguf"
             )
+    if sudo:
+        password = getpass.getpass(f"Enter password for user {getpass.getuser()}: ")
+        install_cmd = f"echo {password} | {install_cmd}"
```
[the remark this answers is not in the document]
harriebird: Hi @Datta0, @danielhanchen, and @rolandtannous, I addressed this by using `stdin.write()` in providing sudo password.

danielhanchen:
The build path this targets is already handled in llama_cpp.py where make is skipped for cmake-only checkouts, and this branch also adds an interactive password prompt piped to the privilege escalation call plus a rewrite of the rpm probe that references a literal package named "package". Could you narrow it to just passing ignore_deprecation on the cmake build step, and split the rest into its own PR?

danielhanchen:
main now skips make for CMake-only llama.cpp checkouts in unsloth_zoo/llama_cpp.py via #763, so the build fix here looks covered already, and the remaining piece is just passing ignore_deprecation on the cmake --build call. If you want to keep this open, could you split out the sudo password prompt, since getpass would block non-interactive notebook and script runs? Sorry for the slow review and thanks for putting it together.

danielhanchen:
main now skips make on CMake-only llama.cpp checkouts in unsloth_zoo/llama_cpp.py via #763, so the build fix here is already covered, and the remaining part adds an interactive getpass sudo prompt plus a package install probe that would block non interactive saves. Sorry for the slow review, and thanks for tracking the upstream Makefile removal down.

danielhanchen:
Main now detects the CMake migration stub Makefile in unsloth_zoo/llama_cpp.py and goes straight to cmake, which was added in #763, so I do not think this is needed any more. Sorry for the slow review, and thanks for putting it together.

PR #692, opened 2026-05-22
fix(mlx): save only LoRA adapter tensors

Lyxot:
## Summary

Fix MLX LoRA adapter export to match the CUDA/PEFT behavior by saving only adapter tensors instead of all trainable tensors.

MLX `save_lora_adapters()` now collects tensors from the full model parameter tree and filters by LoRA adapter tensor names using the same principle as PEFT's CUDA path: keep keys containing `lora_`.

This prevents Studio adapter exports from including base model tensors after a checkpoint reload where some base tensors may be marked trainable.

## Changes

- Add a dedicated MLX LoRA adapter tensor-name filter.
- Save only tensors whose names contain `lora_`.
- Keep training checkpoint saves separate by preserving `save_trainable_adapters()`.
- Continue writing the same adapter artifacts: `adapters.safetensors` and `adapter_config.json`.

---

The `lora_` filter logic comes from PEFT's `get_peft_model_state_dict()`, which is what the CUDA path uses through `model.save_pretrained()`:

https://github.com/huggingface/peft/blob/a106ff4c7061dd9e59609f88724e4770c3b37293/src/peft/utils/save_and_load.py#L133-L153


danielhanchen:

PR 692 splits MLX adapter saving so save_lora_adapters exports only module-anchored lora_a/lora_b tensors (preventing base-weight leakage and reload/freeze fall-through to save_merged_model) while save_trainable_adapters keeps the full trainable tree for in-loop checkpoints. With the review-added refinements to _ensure_lora_frozen and save_pretrained_merged, the change is correct, covered, and a clear improvement over the prior substring-based detection.

**Reason:** Real reload/freeze export bug fixed with module-anchored detection; review-loop iterations corrected the LoRA-frozen guard semantics and hoisted the merged save check, all backed by 10 passing tests.

danielhanchen:

PR replaces fragile substring-based LoRA tensor detection in unsloth_zoo/mlx with a module-anchored collector, splits adapter-only export from full trainable-tree checkpoint export, and updates both MLXTrainer.save_model and save_pretrained_merged to route reloaded/frozen and mixed LoRA + non-LoRA fine-tunes correctly. The fix unblocks Unsloth Studio's MLX export path (which previously leaked accidentally-trainable base weights into adapters.safetensors) and is covered by 20 regression tests under the cross-platform mlx_simulation shim.

**Reason:** All 3 review iterations resolved; module-anchored adapter detection is correct, frozen-LoRA + mixed-trainable saves now preserve all intended tensors, adapter_config.json ships with rank/scale, and the stub-model regression is fixed.

PR #693, opened 2026-05-23
tests: lock in gpt-oss BlockMask + KV-mask alignment guards

danielhanchen:
## Summary

Adds a CPU-only pytest that catches the two recent gpt-oss inference regressions deterministically, with no GPU required, so they cannot reappear silently.

### What it covers

- **PR #690 (BlockMask reaching the eager inference path).** When `config._attn_implementation == "flex_attention"` (set during training) but inference runs through eager attention, `create_causal_mask` must not return a `BlockMask`. Otherwise the eager forward crashes with
  ```
  TypeError: unsupported operand type(s) for +=: 'Tensor' and 'BlockMask'
  ```

- **PR #691 (eager KV length must match attention-mask kv length).** The `eager_attention_forward` closures must trim KV (or mask) to the shorter length, otherwise pre-allocated cache slots crash with
  ```
  RuntimeError: The size of tensor a (N) must match the size of tensor b (M) at non-singleton dimension 3
  ```

Each fix gets a runtime invariant check **and** an AST / source check, so behavioural regressions and silent guard deletions both surface.

### Design notes

- Runtime check runs in a clean subprocess so we can replace `torch.compile` with identity **before** any `unsloth_zoo` import. The wrap captures `_torch_compile = functools.partial(torch.compile, ...)` at module load, and on CPU torch the compiled wrapper drops kwarg names, which would mask the real BlockMask invariant.
- AST checks parse the source of `patch_GptOssModel` / `patch_GptOssAttention` and assert the guard literals exist where they should.
- Uses the repo's existing `tests/conftest.py` GPU-free harness (device_type preload, mem_get_info / get_device_capability stubs, `UNSLOTH_ALLOW_CPU=1`). No new conftest surface.

### Local verification

- 4/4 PASS on cpu-only torch 2.9.1+cpu with transformers 4.57.6 and 5.9.0 (python 3.11).
- 2/4 FAIL when the PR #690 wrap guard is reverted (runtime catches the BlockMask, AST catches the missing literal).
- 1/4 FAIL when the PR #691 `_align_kv_to_mask` call sites are removed.

### Suggested CI workflow

The repo's GitHub App / PAT this branch is pushed with does not have the `workflow` scope, so I couldn't add a `.github/workflows/*.yml` directly. Suggested file to land alongside this PR (copy-paste; runs in ~3 minutes on a free runner):

<details><summary><code>.github/workflows/gpt_oss_patch_tests.yml</code></summary>

```yaml
name: gpt-oss patch regression tests

on:
  push:
    branches: [main]
    paths:
      - "unsloth_zoo/temporary_patches/gpt_oss.py"
      - "tests/test_gpt_oss_attention_mask.py"
      - "tests/conftest.py"
      - ".github/workflows/gpt_oss_patch_tests.yml"
  pull_request:
    paths:
      - "unsloth_zoo/temporary_patches/gpt_oss.py"
      - "tests/test_gpt_oss_attention_mask.py"
      - "tests/conftest.py"
      - ".github/workflows/gpt_oss_patch_tests.yml"

jobs:
  cpu-regression:
    name: cpu-only / python ${{ matrix.python-version }} / transformers ${{ matrix.transformers }}
    runs-on: ubuntu-latest
    timeout-minutes: 15
    strategy:
      fail-fast: false
      matrix:
        python-version: ["3.11", "3.12"]
        transformers:
          - "4.57.6"
          - "5.9.0"
    steps:
      - uses: actions/checkout@v4

      - uses: actions/setup-python@v5
        with:
          python-version: ${{ matrix.python-version }}

      - name: Install CPU torch
        run: |
          python -m pip install --upgrade pip
          pip install --extra-index-url https://download.pytorch.org/whl/cpu "torch==2.9.1+cpu"

      - name: Install transformers ${{ matrix.transformers }} + tokenizers
        run: |
          if [[ "${{ matrix.transformers }}" == "5."* ]]; then
            pip install "transformers==${{ matrix.transformers }}" "huggingface-hub>=1.0,<2.0" "tokenizers>=0.22"
          else
            pip install "transformers==${{ matrix.transformers }}" "huggingface-hub>=0.34.0,<1.0" "tokenizers>=0.22.0,<=0.23.0"
          fi

      - name: Install runtime test deps
        run: |
          pip install pytest pillow numpy peft requests

      - name: Install this branch of unsloth_zoo (no deps)
        run: pip install --no-deps -e .

      - name: Install unsloth main (no deps; needed because unsloth_zoo imports check)
        run: pip install --no-deps "unsloth @ git+https://github.com/unslothai/unsloth@main"

      - name: Run gpt-oss patch regression tests
        env:
          CUDA_VISIBLE_DEVICES: ""
          UNSLOTH_ALLOW_CPU: "1"
          UNSLOTH_MODEL_NAME: "unsloth/gpt-oss-test"
        run: pytest tests/test_gpt_oss_attention_mask.py -v
```

</details>

## Test plan

- [x] pytest passes on cpu-only torch 2.9.1+cpu, transformers 4.57.6
- [x] pytest passes on cpu-only torch 2.9.1+cpu, transformers 5.9.0
- [x] pytest FAILS deterministically when the PR #690 wrap guard is removed
- [x] pytest FAILS deterministically when the PR #691 `_align_kv_to_mask` calls are removed

### DIFF
--- a/tests/test_gpt_oss_attention_mask.py
+++ b/tests/test_gpt_oss_attention_mask.py
@@ -14,70 +14,32 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-"""CPU-only regression tests for gpt-oss inference attention-mask patches.
+"""Runtime regression check for the gpt-oss BlockMask leak (PR #690).
 
-Covers two recent fixes:
+The behavioural sibling tests (static AST guards over wrap and the
+patched GptOssModel.forward, plus the PR #691 _align_kv_to_mask static
+check, plus the inspect-driven mask-kwargs filter check) live in
+tests/test_zoo_history_regressions.py so they auto-run under the
+consolidated Tests CI workflow.
 
-  PR 690 (BlockMask reaching the eager inference path). When
-    config._attn_implementation == "flex_attention" but Unsloth runs
-    inference through eager attention, the wrap over create_causal_mask
-    MUST NOT return a BlockMask. Without the fix the eager forward
-    crashes with:
-        TypeError: unsupported operand type(s) for +=:
-            'Tensor' and 'BlockMask'
+This file is the *runtime* invariant: actually call the wrapped
+transformers.masking_utils.create_causal_mask with a tiny GptOssConfig
+that has _attn_implementation="flex_attention" and assert the return is
+not a BlockMask. The check runs in a clean subprocess so torch.compile
+can be replaced with identity BEFORE any unsloth_zoo import -- the wrap
+captures _torch_compile = functools.partial(torch.compile, ...) at
+module load time and on CPU torch the compiled wrapper drops kwarg
+names, which would otherwise mask the real BlockMask invariant with a
+misleading "missing positional argument" error.
 
-  PR 691 (eager KV length must match attention-mask kv length). The
-    eager_attention_forward closures MUST trim KV (or the mask) to the
-    shorter length, otherwise pre-allocated cache slots crash with:
-        RuntimeError: The size of tensor a (N) must match the size of
-            tensor b (M) at non-singleton dimension 3
-
-Each fix has a runtime invariant check plus an AST / source check, so
-either a behavioural regression or a silent deletion of the guard
-fails CI. tests/conftest.py already preloads device_type, stubs
-torch.cuda.mem_get_info / get_device_capability, and sets
-UNSLOTH_ALLOW_CPU=1 -- no extra GPU spoofing is needed here.
+CPU-only; no GPU required. The repo's existing tests/conftest.py GPU-free
+harness (device_type preload, mem_get_info / get_device_capability
+stubs, UNSLOTH_ALLOW_CPU=1) handles the inner subprocess via env
+inheritance.
 """
 from __future__ import annotations
 
-import ast
-import inspect
-import os
-
-# Trigger the gpt_oss model gate before unsloth_zoo's temporary patches install.
-os.environ.setdefault("UNSLOTH_MODEL_NAME", "unsloth/gpt-oss-test")
-os.environ.setdefault("UNSLOTH_COMPILE_DISABLE", "0")
-
-# torch.compile around the wrap-around-create_causal_mask trick drops keyword
-# names on some torch CPU builds, which would mask the real BlockMask check.
-# Replace torch.compile with identity BEFORE unsloth_zoo imports so the wrap
-# can forward (*args, **kwargs) cleanly.
-import torch
-
-
-def _identity_compile(model=None, *args, **kwargs):
-    if model is None:
-        return lambda fn: fn
-    return model
-
-
-torch.compile = _identity_compile  # noqa: E305
-
-import pytest
-
-
-# ---------------------------------------------------------------------------
-# PR 690: BlockMask must not reach the eager inference path
-# ---------------------------------------------------------------------------
-
 _RUNTIME_SUBPROCESS = r"""
-# Run the runtime invariant in a clean subprocess so we can replace
-# torch.compile with identity BEFORE any unsloth_zoo import. The wrap
-# over create_causal_mask captures _torch_compile = functools.partial(
-# torch.compile, ...) at module load time, which on CPU torch drops
-# kwarg names and makes the underlying call fail with a misleading
-# "missing positional arg" instead of the real BlockMask invariant
-# we want to check.
 import os, sys, traceback
 os.environ['UNSLOTH_ALLOW_CPU'] = '1'
 os.environ['UNSLOTH_MODEL_NAME'] = 'unsloth/gpt-oss-test'
@@ -145,11 +107,14 @@ except Exception:
 """
 
 
-def test_pr690_runtime_flex_attention_inference_does_not_return_blockmask(tmp_path):
-    """With config._attn_implementation == 'flex_attention' and an inference
-    (no grad) inputs_embeds, the patched create_causal_mask must return a
-    tensor (or None), never a BlockMask. Without the fix this returns a
-    BlockMask and downstream eager attention crashes with TypeError."""
+def test_pr690_runtime_blockmask_does_not_reach_eager_inference_path():
+    """Runtime invariant: with _attn_implementation='flex_attention' and
+    a non-grad inputs_embeds, the patched create_causal_mask must return
+    a tensor (or None), never a BlockMask. Without PR #690 this returns
+    a BlockMask and gpt-oss generate() crashes with
+        TypeError: unsupported operand type(s) for +=:
+            'Tensor' and 'BlockMask'
+    """
     import subprocess
     import sys as _sys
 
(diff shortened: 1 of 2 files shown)


danielhanchen (commented):
Verified the PR end-to-end against the head commit `d39f6de6` on `pr-693-head`. The two new files add 277 lines of test code with no library changes; malicious scan is clean (0 findings).

## Positive run (CPU + CUDA host, torch 2.9.1+cu128, transformers 4.57.6, python 3.13.12, pytest 9.0.3)

```
tests/test_gpt_oss_attention_mask.py::test_pr690_runtime_blockmask_does_not_reach_eager_inference_path PASSED
tests/test_zoo_history_regressions.py::test_gpt_oss_wrap_has_flex_attention_inference_guard           PASSED
tests/test_zoo_history_regressions.py::test_gpt_oss_patched_model_forward_has_flex_attention_guard    PASSED
tests/test_zoo_history_regressions.py::test_gpt_oss_eager_attention_aligns_kv_to_mask                 PASSED
tests/test_zoo_history_regressions.py::test_gpt_oss_model_forward_uses_inspect_filtered_mask_kwargs   PASSED
5 passed in 9.60s
```

Full `tests/test_zoo_history_regressions.py` (12 tests) also green; collect-only finds 13 tests across the two PR files with no ImportError, so the python-version-collect CI job stays clean.

## Negative-control matrix (revert one guard at a time on the local clone copy, restore between)

| Revert | Runtime test | wrap_guard | model_fwd_guard | align_kv_count | inspect_filter |
|---|---|---|---|---|---|
| `wrap.return_attention_mask` flex_attention literal | **FAIL** | **FAIL** | pass | pass | pass |
| both `_align_kv_to_mask(...)` call sites | pass | pass | pass | **FAIL** | pass |
| `_build_mask_kwargs` + `_ccm_params` + `inspect.signature(...).parameters` | FAIL (cascade) | FAIL | FAIL | FAIL | **FAIL** |
| outer `_swap_attn_impl == "flex_attention"` only | pass | pass | pass | pass | pass |

So each guard catches its target. Three observations worth flagging:

1. `test_gpt_oss_patched_model_forward_has_flex_attention_guard` doesn't fire when only the outer `_swap_attn_impl = _orig_attn_impl == "flex_attention"` (line 2482) is removed, because the AST test scans `inspect.getsource(patch_GptOssModel)` which includes the nested `wrap.return_attention_mask` closure -- the inner literal alone keeps the test green. The defence-in-depth is real (the inner wrap guard still handles the case correctly), but the test is effectively a subset of `test_gpt_oss_wrap_has_flex_attention_inference_guard`. Optional follow-up: scope the second test to the OUTER body only (e.g. parse the AST and exclude the nested `wrap` FunctionDef).

2. `test_gpt_oss_eager_attention_aligns_kv_to_mask` uses `callsite_count - def_count >= 2`. Tight to the current 2-callsite structure (one per eager variant). If a future refactor consolidates to a single helper, the test will fail spuriously. Optional follow-up: drop to `>= 1` and additionally assert both `eager_attention_forward` and `inplace_eager_attention_forward` source slices contain the call.

3. The runtime subprocess test exercises the wrap path only; it does not catch a regression that lives strictly in the outer `_swap_attn_impl` block. The four AST checks plus the runtime invariant together do cover the BlockMask leak surface as observed in PR #690, so this is a coverage note rather than a blocker.

## Subprocess hygiene

`test_pr690_runtime_blockmask_does_not_reach_eager_inference_path` runs via `subprocess.run([sys.executable, "-c", _RUNTIME_SUBPROCESS])`. No env leak into the parent (env is inherited only one way), `torch.compile` is rebound only in the child, and `conftest.py`'s GPU-free harness propagates by env var (`UNSLOTH_ALLOW_CPU=1`). The earlier `chatgpt-codex-connector` P1/P2 and `github-code-quality` "unused import" comments were filed against `3ed27d06`; the `d39f6de6` move commit confined the `torch.compile` swap and env defaults to the subprocess string and stripped the now-unused top-level `ast`/`inspect`/`pytest` imports, so those threads are addressed.

## CI wiring

`.github/workflows/consolidated-tests-ci.yml:106` already invokes `tests/test_zoo_history_regressions.py`, so the four new AST guards land on CI without any workflow change. The runtime subprocess file is collected by `--collect-only` (line 61) but not executed. Two acceptable follow-ups:
- (recommended) a maintainer with workflow PAT scope lands the proposed `.github/workflows/gpt_oss_patch_tests.yml` (the PR description already provides the exact YAML); or
- add `tests/test_gpt_oss_attention_mask.py` to the explicit pytest list in `consolidated-tests-ci.yml:102-107`.

LGTM to land as-is. The three optional follow-ups above can be deferred to a separate PR.

danielhanchen:
Follow-up — ran a broader isolated-venv simulation matrix in addition to the host run posted above, to convince myself nothing breaks across the (python × transformers × positive/negative) cells the PR claims and to surface any subtle deps the proposed CI workflow file would need.

## Matrix (8 fresh `uv venv`s, CPU-only torch 2.9.1+cpu, unsloth main `--no-deps`, runtime test + 4 AST guards)

| sim | python | transformers | mode                              | result            |
|-----|--------|--------------|-----------------------------------|-------------------|
| a   | 3.11   | 4.57.6       | positive                          | 5 passed          |
| b   | 3.11   | 5.9.0        | positive                          | 5 passed          |
| c   | 3.12   | 4.57.6       | positive                          | 5 passed          |
| d   | 3.13   | 4.57.6       | positive                          | 5 passed          |
| g   | 3.12   | 5.9.0        | positive                          | 5 passed          |
| h   | 3.13   | 5.9.0        | positive                          | 5 passed          |
| e   | 3.11   | 4.57.6       | revert wrap flex_attention literal | 2 failed, 3 passed (`wrap_guard` + `runtime_subprocess`) |
| f   | 3.11   | 5.9.0        | revert wrap flex_attention literal | 2 failed, 3 passed (same two) |

So the 3-by-2 positive grid is uniformly green and the negative controls bite the same two tests across transformers 4.x and 5.x.

## One actionable nit for the proposed workflow YAML

`tests/test_gpt_oss_attention_mask.py` does `import unsloth_zoo` inside its subprocess, which on a real install pulls `unsloth_zoo/rl_environments.py` → `import requests`. The PR description's suggested step

```
- name: Install runtime test deps
  run: |
    pip install pytest pillow numpy peft
```

omits `requests`, and on a transformers 5.x cell the runtime test fails with `ModuleNotFoundError: No module named 'requests'` (the tf 4.x cell happens to pull `requests` transitively, but tf 5.x does not — both observed in sims b and f before I added `requests` to the install list, then reproduced clean once I did). Suggestion: append `requests` to that step. The AST guards in `test_zoo_history_regressions.py` aren't affected since they don't import `unsloth_zoo.rl_environments`.

## Host hazard (not the PR's fault, just for awareness)

On a torch built with CUDA support (e.g. `torch==2.9.1+cu128`) running with `CUDA_VISIBLE_DEVICES=""`, `torch.cuda.is_available()` returns False so the subprocess's stub branch fires — but `import unsloth_zoo` later transitively triggers `torchao.prototype.mx_formats.kernels` which calls `torch.utils._triton.has_triton()` → `torch.cuda.current_device()` → `_lazy_init()` → `RuntimeError: No CUDA GPUs are available`. This does NOT affect the PR's actual CI (which uses `torch==2.9.1+cpu` with no CUDA support in the wheel), and the conftest GPU-free harness still works as designed. Just flagging in case someone tries to reproduce by hiding GPUs on a CUDA host.

## Cross-platform note

Linux is what I tested directly. The runtime test uses `subprocess.run([sys.executable, "-c", ...])` with list-form args (no `shell=True`), no path separators, no platform-specific imports beyond what transformers / torch themselves already require, so Mac and Windows CI runners should behave the same — the AST guards in `test_zoo_history_regressions.py` are pure-Python `inspect` + `ast` and have no OS surface.

Still LGTM. The only change I'd suggest before landing the proposed workflow YAML is adding `requests` to the runtime-test-deps step.

danielhanchen:
Pushed `cfcb092e` to tighten the two AST guards we flagged earlier:

- `test_gpt_oss_patched_model_forward_has_flex_attention_guard` now walks the AST and asserts the `_attn_implementation == "flex_attention"` literal lives inside the patched `forward` FunctionDef itself, not anywhere in `patch_GptOssModel`. Previously the nested `wrap` closure's literal satisfied the test on its own, so an outer-only revert slipped through; that revert now fails the test as it should.
- `test_gpt_oss_eager_attention_aligns_kv_to_mask` now slices `eager_attention_forward` and `inplace_eager_attention_forward` individually via AST and asserts each body contains `_align_kv_to_mask(`, instead of counting whole-file callsites `>= 2`. A future refactor that consolidates the two eager paths still passes as long as the alignment runs on both routes; dropping the call from either still trips the test.

Verified locally: positive run 5/5 PASS, and the negative-control matrix (revert wrap literal, strip `_align` callsites, strip `_build_mask_kwargs` + `_ccm_params` + `inspect.signature(create_causal_mask).parameters`, strip the outer `_swap_attn_impl` block) now bites the matching test for each revert with no gaps.

Also updated the PR description's proposed `gpt_oss_patch_tests.yml` to append `requests` to the `Install runtime test deps` step. Without it the runtime subprocess test fails on transformers 5.x cells with `ModuleNotFoundError: No module named 'requests'` because `import unsloth_zoo` reaches `unsloth_zoo/rl_environments.py` which depends on `requests` (tf 4.x cells happen to pull it transitively, tf 5.x cells do not).

PR #695, opened 2026-05-25
Auto-apply four upstream / runtime monkey-patches for notebook compatibility

danielhanchen:
Splits four upstream-bug monkey-patches out of #694 into a focused PR that has nothing else mixed in. Each patch handles a real upstream / runtime bug that breaks at least one notebook today; all four no-op cleanly when the relevant dependency is absent, when ``UNSLOTH_*`` opt-outs are set, or when the upstream version moves out of the patch window.

### 1. ``vllm_utils.py::patch_vllm_decompose_size_nodes`` - port of vllm-project/vllm#42543

Upstream's ``_decompose_size_nodes`` only rewrites top-level ``user.args`` references to size nodes and misses the same node when it is nested inside ``slice(start, stop, step)``, tuples, lists, or kwargs. The trailing ``graph.graph.erase_node(node)`` then trips ``torch.fx`` with:

\`\`\`
RuntimeError: Tried to erase Node size_N but it still had K users
\`\`\`

Five GRPO notebooks crash with this on Blackwell sm_100/sm_120 during ``FastVisionModel.from_pretrained(fast_inference=True)``: Qwen3-(4B)-GRPO, Advanced Llama3.2-(3B)-GRPO-LoRA, Qwen3 8B FP8 GRPO, Llama FP8 GRPO, Qwen2.5-7B-VL-GRPO. Qwen3-VL-(8B) Vision-GRPO is unaffected because its graph only emits the ``getitem(size, idx)`` shape the upstream ``if`` branch already handles.

The patch ports PR #42543's recursive rewrite, adds a kwargs sweep that #42543 currently misses, and ends with an ``if node.users: warn-and-skip`` safety net so we never erase a node that still has live users. Bounded to ``vllm >= 0.11.0, < 0.99``. Opt-out: ``UNSLOTH_DISABLE_VLLM_DECOMPOSE_SIZE_PATCH=1``.

### 2. ``vllm_utils.py::load_vllm`` - block FlashInfer JIT when nvcc / ninja are missing

vLLM on Blackwell picks ``FLASHINFER`` first from ``['FLASHINFER', 'FLASH_ATTN', 'TRITON_ATTN', 'FLEX_ATTENTION']`` and JIT-compiles trtllm-gen kernels at first attention call, which requires ``nvcc``. cu128 runtime images (and most slim production images) ship the runtime libs but not nvcc, so the first ``LLM()`` call crashes inside FlashInfer compilation.

Things that do not work and we tried first:
* ``VLLM_ATTENTION_BACKEND=FLASH_ATTN`` - vLLM 0.19.1 logs "Unknown vLLM environment variable detected" and ignores it.
* ``UNSLOTH_VLLM_NO_FLASHINFER=1`` - only short-circuits our own preflight, not vLLM's own backend selection in ``vllm.platforms.cuda.get_attn_backend_cls``.
* Pip-installing ``nvidia-cuda-nvcc-cu12`` - the wheel ships only ``ptxas``, not ``nvcc``.

Final fix: when the preflight already detects missing nvcc / ninja, drop any cached ``flashinfer`` entries from ``sys.modules`` and set ``sys.modules["flashinfer"] = None``. That is the documented Python idiom for "this module fails to import" (https://docs.python.org/3/reference/import.html). vLLM's own ``try: import flashinfer / except ImportError`` branch then picks ``FLASH_ATTN`` cleanly. No-op when nvcc + ninja are both present.

### 3. ``temporary_patches/notebook_deps.py`` - new file, auto-install missing HF backends + dynamic-module deps

Several notebooks that worked fine on the previous slim venv broke once their HF imports moved to ``requires_backends`` (TimmWrapper for vision backbones), ``dynamic_module_utils.check_imports`` (Deepseek-OCR trust_remote_code modelling file), or a bare ``ModuleNotFoundError`` from the IPython / Jupyter chain. The raising frame is buried inside HF code, so no single try/except can catch them all.

The new module installs three monkey-patches at import time:

* ``patch_requires_backends_autoinstall`` - wraps ``transformers.utils.import_utils.requires_backends`` and pip-installs the missing backend on first ``ImportError`` (allow-list only), then retries the original.
* ``patch_check_imports_autoinstall`` - same idea for ``transformers.dynamic_module_utils.check_imports``, parses the "This modeling file requires the following packages..." message and installs each named package against the allow-list.
* ``_ensure_notebook_chain`` - preemptive ``find_spec`` + ``_try_install_and_import`` for ``traitlets``, the only dep that today raises a bare ``ModuleNotFoundError`` from the IPython chain before any HF wrapper runs.

Allow-list (closed, no surprises): ``timm, addict, einops, easydict, snac, torchcodec, matplotlib, traitlets, soundfile, librosa, scipy, pyctcdecode, tiktoken, blobfile, pillow_heif, decord, av, num2words, jieba, sentencepiece``.

Honours both the standard offline flags (``UNSLOTH_OFFLINE``, ``HF_HUB_OFFLINE``, ``TRANSFORMERS_OFFLINE``) and the opt-out ``UNSLOTH_AUTO_INSTALL=0`` already used by ``llama_cpp.py``. Falls back to ``pip install --user`` when running outside a venv as non-root.

### 4. ``temporary_patches/notebook_deps.py`` - ``torchcodec`` added to the allow-list (HF datasets 4.x)

HF ``datasets`` 4.x removed the ``soundfile`` path from ``features/audio.py`` and now hard-requires ``torchcodec`` to decode any ``Audio`` column (``audio.py:181`` raises ``ImportError: please install 'torchcodec'``). On slim images that omit ``torchcodec`` from the venv this breaks every TTS / audio notebook the moment the trainer touches the first row. Adding ``torchcodec`` to the allow-list lets the auto-installer pick it up the first time HF raises.

## Compatibility envelope

* Linux / macOS / Windows: all four patches gate on import success and / or environment variables; no platform-specific code paths.
* ``transformers`` 4.57.6 + TRL 0.22.2; ``transformers`` default + TRL default; ``transformers`` 5.x + TRL 1.x - patches all key off public symbols that exist on every matrix cell.
* MLX / Apple Silicon: ``vllm_utils.py`` short-circuits when vLLM is absent (already wired in #694); ``notebook_deps.py`` is platform-neutral.
* AMD / Intel GPU: no NVIDIA-specific assumptions.
* Browser-driven workflows (Studio): unaffected.

## Validation

* Manually verified all four patches against the failing notebooks on a Blackwell B200 host.
* Local notebook batch (Orpheus TTS, Gemma3 Vision, Qwen3 GRPO, etc.) all pass end-to-end after the patches are applied.
* The 224-line ``notebook_deps.py`` is a brand-new file so it cannot regress anything in the import graph; only ``temporary_patches/__init__.py`` gets a 1-line ``from .notebook_deps import *``.

## Out of scope

The unrelated hardening fixes from #694 (``llama_cpp.install_package`` EOFError, ``patch_vllm_graph_capture`` bail-out when vllm is missing, ``load_vllm`` actionable ``ImportError``, Deepseek-V2 capitalisation alias) are intentionally **not** in this PR. They will continue to ship via #694.

### DIFF
--- a/unsloth_zoo/__init__.py
+++ b/unsloth_zoo/__init__.py
@@ -16,6 +16,12 @@
 
 __version__ = "2026.5.4"
 
+# Stub numpy._core.tests._natype FIRST -- before anything that might pull in
+# scipy.linalg / transformers.generation / trust_remote_code modelling files.
+# See unsloth_zoo/numpy_compat.py for the full rationale (slim numpy wheels
+# missing the tests/ subpackage that numpy.testing requires).
+from . import numpy_compat  # noqa: F401
+
 import os
 import warnings
 import re
--- a/unsloth_zoo/empty_model.py
+++ b/unsloth_zoo/empty_model.py
@@ -400,8 +400,24 @@ def create_empty_model(config, dtype = torch.float16, is_vision_model = False):
 
 @torch.inference_mode
 def set_additional_modules(new_model, quant_state_dict, config):
-    if hasattr(new_model, "language_model"):
+    # Locate the text submodule that holds `embed_tokens`. Three layouts are
+    # observed across transformers >= 5.0:
+    #   1. new_model.language_model.embed_tokens    (old VL: Gemma3, Mistral3,
+    #      Llama-VL via *ForConditionalGeneration with `language_model` exposed
+    #      directly on the outer class).
+    #   2. new_model.model.embed_tokens            (text-only models and
+    #      Qwen3-VL which folds embed_tokens onto the outer .model wrapper).
+    #   3. new_model.model.language_model.embed_tokens (Qwen2.5-VL: outer is
+    #      Qwen2_5_VLForConditionalGeneration, .model is Qwen2_5_VLModel which
+    #      stores the text model under .language_model).
+    # The prefix has to match the key path in quant_state_dict so the embed/
+    # norm/lm_head lookups below resolve correctly.
+    if hasattr(new_model, "language_model") and hasattr(new_model.language_model, "embed_tokens"):
         language_model = new_model.language_model
+        language_model_prefix = "model.language_model"
+    elif hasattr(new_model, "model") and hasattr(new_model.model, "language_model") \
+         and hasattr(new_model.model.language_model, "embed_tokens"):
+        language_model = new_model.model.language_model
         language_model_prefix = "model.language_model"
     else:
         language_model_prefix = "model"
--- a/unsloth_zoo/llama_cpp.py
+++ b/unsloth_zoo/llama_cpp.py
@@ -277,8 +277,25 @@ def install_package(package, sudo = False, print_output = False, print_outputs =
         install_cmd = f"{'sudo ' if sudo else ''}apt-get install {package} -y"
 
     print(f"Unsloth: Installing packages: {package}")
-    if not (IS_COLAB_ENVIRONMENT or IS_KAGGLE_ENVIRONMENT):
-        acceptance = input(f"Missing system packages. We need to execute `{install_cmd}` - do you accept? Press ENTER. Type NO if not.")
+    # Non-interactive contexts (CI, docker containers, headless notebook
+    # runners) close stdin so input() raises EOFError, which surfaces to
+    # the caller as "GGUF conversion failed: EOF when reading a line".
+    # Treat closed/non-TTY stdin as auto-accept so save_pretrained_gguf
+    # can install build deps on its own. UNSLOTH_AUTO_ACCEPT_INSTALL=1
+    # also forces auto-accept even when stdin is a TTY (for scripts that
+    # want to skip the prompt).
+    auto_accept = (
+        os.environ.get("UNSLOTH_AUTO_ACCEPT_INSTALL", "0") == "1"
+        or not (sys.stdin and sys.stdin.isatty())
+    )
+    if not (IS_COLAB_ENVIRONMENT or IS_KAGGLE_ENVIRONMENT) and not auto_accept:
+        try:
+            acceptance = input(f"Missing system packages. We need to execute `{install_cmd}` - do you accept? Press ENTER. Type NO if not.")
+        except EOFError:
+            # Stdin closed mid-call (eg. docker run -i without -t after a
+            # subprocess took stdin). Treat the same as auto-accept rather
+            # than abort the GGUF export.
+            acceptance = ""
         if "no" in str(acceptance).lower():
             raise RuntimeError(
                 f"Unsloth: Execution of `{install_cmd}` was cancelled!\n"\
--- a/unsloth_zoo/numpy_compat.py
+++ b/unsloth_zoo/numpy_compat.py
@@ -1,0 +1,75 @@
+# Unsloth Zoo - Utilities for Unsloth
+# Copyright 2023-present Daniel Han-Chen, Michael Han-Chen & the Unsloth team. All rights reserved.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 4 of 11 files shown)


danielhanchen:
**Codex review follow-up: closed out four open findings from the May 25 pass**

Pushed four small follow-up commits to this branch (78e152a, 65985ff, 9d602a8 stacked on the latest head):

**P1, unsloth_zoo/temporary_patches/notebook_deps.py – requires_backends retry** – transformers 5.x decorates each `is_<backend>_available` predicate with `@lru_cache`, so the cached `False` from the unpatched `_orig()` call gets reused on the retry even after the pip install succeeds. Now calls `.cache_clear()` on the predicate (if present) after flipping the `_<backend>_available` flag, so the retry actually re-evaluates the import.

**P1, unsloth_zoo/vllm_utils.py – _poison_flashinfer_if_unusable fallback** – the nvcc/ninja-missing branch already None-marks `flashinfer` in `sys.modules` and sets `UNSLOTH_VLLM_NO_FLASHINFER=1`, but it left a user-set `VLLM_ATTENTION_BACKEND=FLASHINFER` in place. That keeps vLLM pinned to FLASHINFER, defeating its own `try: import flashinfer except ImportError` fallback to FLASH_ATTN, and `vllm.LLM()` still crashes. Pop the env override on the way out so the import failure actually translates into the documented FLASH_ATTN fallback.

**P1, unsloth_zoo/vllm_utils.py – _decompose_size_nodes kwarg path** – when `v` is the size node directly, the existing `_replace_in_args([v], node, dims)[0]` would pick only the first dim under the depth-aware approach, so patterns like `torch.reshape(..., shape=x.size())` silently turn into `torch.reshape(..., shape=<first-dim-scalar>)`. Detect that case and pass `tuple(dims)` instead, preserving torch.Size semantics. List/tuple kwargs already had the correct depth-aware behaviour and are unchanged.

**P2, unsloth_zoo/temporary_patches/notebook_deps.py – VS Code notebook detection** – `VSCODE_PID` is set by every VS Code subprocess (terminals, debug runners, Python LSP), not just Jupyter kernels, so treating it as a notebook signal was firing `_ensure_notebook_chain()` on plain terminal sessions. Dropped from the env-var probe; real VS Code Jupyter kernels import `ipykernel` before `unsloth_zoo`, so the `sys.modules` check still catches them without the false-firing on terminals.

CI re-running now. Will re-poll until green.

danielhanchen:
Merged main into this branch and resolved the empty_model.py conflict: kept main's _unwrap_tensor helper and this PR's three-layout embed_tokens detection. Mergeable again. This pairs with unslothai/unsloth#5748 for the Docker images; #694 should land alongside it as well.

danielhanchen:
Validation update from the Docker image work (unsloth#5748): ran the GRPO notebook set inside the image with this branch installed, one variable changed at a time against main zoo.

- The `_decompose_size_nodes` port is confirmed to fix the `RuntimeError: Tried to erase Node size_N but it still had M users` failures. Advanced Llama3.2 GRPO LoRA and Qwen3 4B GRPO go from failing within 2 min on main zoo to passing (3 and 7 min) on this branch, unmodified.
- The FP8 pair (Llama 8B FP8, Qwen3 8B FP8) needs one more token on top of this branch: `convert_vllm_to_huggingface` in vllm_utils.py still passes `dtype=` to transformers 5.11's `FP8Linear`, whose signature dropped it (the weight dtype is forced to the FP8 dtype internally). Removing `dtype=dtype` from the has_bias branch (line ~1434) makes both notebooks pass (5 and 8 min). Worth folding into this PR.
- Known limit: the vision GRPO notebooks (Qwen2.5-VL 7B, Qwen3 VL 8B) remain broken even with this branch, but for unrelated reasons. Qwen2.5-VL trains 22 min then hits vLLM 0.19's multimodal processor cache desync (`Expected a cached item for mm_hash=...`); disabling the cache surfaces a CUDA illegal memory access at the next sleep-mode empty_cache sync. Qwen3 VL hangs in trainer.train() identically on main, #694 and this branch. Both look like vllm 0.19.x vision engine instability under standby cycling rather than anything this PR touches.

PR #703, opened 2026-05-28
fix: guard torch.distributed attrs missing on ROCm Windows

LeoBorcherding:
On Windows, the ROCm build of PyTorch ships without the distributed C extension (torch._C._distributed_c10d), so torch.distributed loads as a partial stub with several attributes missing entirely, including is_initialized, is_torchelastic_launched, and get_rank.

unsloth_zoo/utils.py grabbed all three at module import time with bare attribute access, causing an AttributeError the moment the module was imported, even in code paths like GGUF export that never use distributed features at all.

Fix: replace the three bare grabs with getattr(..., lambda: False/0) fallbacks so importing unsloth_zoo never crashes on platforms where torch.distributed is unavailable or stubbed.

Root cause tracked in ROCm/TheRock#3284 (libuv / torch.distributed missing on Windows ROCm builds). Companion fix for the export subprocess in unsloth/unsloth#5301.

Ref: https://github.com/ROCm/TheRock/issues/3284
Ref: https://github.com/unslothai/unsloth/pull/5301

### DIFF
--- a/unsloth_zoo/temporary_patches/utils.py
+++ b/unsloth_zoo/temporary_patches/utils.py
@@ -143,6 +143,120 @@ if _pil_mod is not None and hasattr(_pil_mod, "__version__"):
         pass
 del _sys, _pil_mod
 
+# ROCm on Windows ships PyTorch without the full torch.distributed C-extension
+# stack (torch._C._distributed_c10d, DeviceMesh, etc.).
+# torchao pulls the entire distributed chain in at module-import time
+# (torch.distributed._functional_collectives → _tensor → DeviceMesh …),
+# which cascades into ImportError even for code paths that never use
+# distributed features (e.g. plain LoRA training).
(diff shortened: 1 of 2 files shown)


LeoBorcherding:
Added two Windows ROCm import fixes in the latest commit.

The first guards bare attribute access on torch.distributed (is_initialized, get_rank, etc.) which the ROCm Windows build strips out entirely.

The second installs a meta path hook that intercepts all torchao imports and returns self-contained stubs. On Windows ROCm, torchao tries to pull in the full distributed stack at module load time, which cascades into a hard crash even for plain LoRA training that never touches distributed features. The stubs give each submodule a proper __spec__ so importlib.find_spec works, and return sentinel classes (not modules) for leaf attributes so isinstance() checks in peft pass correctly. The hook only activates when torchao actually fails to import, so Linux and CUDA are unaffected.

PR #705, opened 2026-05-31
Self-heal GGUF conversion when the converter environment is broken

danielhanchen:
### Problem

GGUF export (including from Unsloth Studio) can fail with an undiagnosable error:

```
Unsloth: GGUF conversion failed: ... unsloth_convert_hf_to_gguf.py ...
Command '[...]' returned non-zero exit status 1.
```

The real cause is hidden. `convert_to_gguf` runs the converter with
`capture_output=True` and, when `print_output=False` (the default unless
`UNSLOTH_ENABLE_LOGGING=1`), re-raises only the `CalledProcessError` string, so
the converter's actual Python traceback is discarded.

### Root cause

The converter's Python deps (`gguf`, `protobuf`, `sentencepiece`,
`mistral_common`) are installed only inside `install_llama_cpp()`, which returns
early when llama.cpp already exists ("llama.cpp found in the system. Skipping
installation."). The convert script is re-downloaded from llama.cpp master each
run, so a stale or broken `gguf` drifts out of sync with it and the subprocess
exits 1. Users had to manually uninstall and reinstall to recover.

### Fix

`convert_to_gguf` now:

- Surfaces the converter's stderr/stdout on failure instead of an opaque exit
  status, so any failure is diagnosable.
- On an import/module error, reinstalls the converter deps into the converter's
  own interpreter and retries the conversion once, so a stale package self-heals
  with no user action.
- Leaves genuine model errors untouched (no spurious reinstall).

`quantize_gguf` gets the same stderr surfacing.

### Tests

Adds `tests/test_convert_to_gguf_self_heal.py` covering three paths:

- stale-package error self-heals and the retry succeeds
- reinstall failure surfaces the converter stderr plus a clear note
- a genuine model error is surfaced as-is with no reinstall attempted

```
tests/test_convert_to_gguf_self_heal.py ... 3 passed
tests/test_convert_hf_to_gguf_patcher.py ... 22 passed
```

No behavior change on the happy path; the repair only runs after a failure that
looks like a broken environment.

### DIFF
--- a/tests/test_convert_to_gguf_self_heal.py
+++ b/tests/test_convert_to_gguf_self_heal.py
@@ -1,0 +1,119 @@
+"""Tests for convert_to_gguf's self-heal on a broken converter environment.
+
+A stale or missing converter package (usually `gguf`) makes the converter
+subprocess exit 1. convert_to_gguf should reinstall the deps into the
+converter's own interpreter and retry once, surface the real traceback if the
+failure persists, and never reinstall on a genuine model error.
+"""
+
+from __future__ import annotations
+
+import importlib.util
+import json
+import os
+import sys
+import textwrap
+import types
+from pathlib import Path
+
+import pytest
+
+
+def _load_llama_cpp_module():
+    repo_root = Path(__file__).resolve().parents[1]
+    module_path = repo_root / "unsloth_zoo" / "llama_cpp.py"
+    spec = importlib.util.spec_from_file_location("llama_cpp_under_test_self_heal", module_path)
+    module = importlib.util.module_from_spec(spec)
+    sys.modules[spec.name] = module
+    spec.loader.exec_module(module)
+    return module
+
+
+def _make_model_dir(tmp_path: Path) -> Path:
+    model_dir = tmp_path / "model"
+    model_dir.mkdir()
+    (model_dir / "config.json").write_text(json.dumps({"architectures": ["LlamaForCausalLM"]}))
+    return model_dir
+
+
+def _write_converter(tmp_path: Path, body: str) -> Path:
+    fake = tmp_path / "fake_convert.py"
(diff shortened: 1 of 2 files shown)


danielhanchen:
### Simulation and robustness pass

Drove `convert_to_gguf` and `quantize_gguf` with fake converter/quantizer scripts in isolated `uv` venvs across Python 3.9, 3.11, 3.12 and 3.13 (33 scenarios x 4 versions). Coverage:

- Happy paths (`print_output` on/off, `quantization_type="None"`, sharded output, VLM dual run) with zero reinstall calls, so the success path is unchanged.
- Self-heal across every dep-error marker, including Windows ("DLL load failed") and Linux/Mac ("undefined symbol") strings, and when the error is on stdout rather than stderr.
- Reinstall failure surfaces the converter output plus a clear note.
- Single-retry guard: a reinstall that does not actually fix the env stops after exactly one retry (no loop).
- Genuine model errors are surfaced as-is with no reinstall attempted.
- Exception type stays `RuntimeError`, so existing callers in `save.py` are unaffected.

Three follow-ups came out of the simulation and are in the second commit:

- Pinned `encoding="utf-8", errors="replace"` on the subprocess calls. `text=True` alone decodes with the locale encoding and strict errors, so non-UTF8 converter output (common on Windows) could raise `UnicodeDecodeError` outside the `except` block.
- Wrapped the reinstall in try/except so a failing pip invocation degrades to the original error.
- Bootstrap pip via `ensurepip` when the converter interpreter has none (eg uv venvs), then retry the install once.

Also verified end to end with a real install (no mocks): in a pip-less venv with `gguf` absent, `_reinstall_converter_deps` bootstrapped pip, installed the deps, and `import gguf, sentencepiece, google.protobuf` then succeeded.

This change is backend Python only (no frontend or hardware-path changes), so there is no browser or GPU surface to validate.

PR #706, opened 2026-05-31
Fix Gemma3 full finetuning on float32 weights (no-bf16 GPUs)

danielhanchen:
## Summary

Full finetuning of the Gemma3 family on a GPU without bfloat16 (V100, T4) crashed with `RuntimeError: expected mat1 and mat2 to have the same dtype, but got: c10::Half != float`.

The `UNSLOTH_FORCE_FLOAT32` patches for Gemma3 (`patch_Gemma3RMSNorm`, `patch_Gemma3MLP`, `patch_Gemma3Attention`) run the heavy reductions in float32 for fp16 overflow safety, then narrow back to float16 at every Linear boundary. That hard-coded float16 assumes the projection weights are float16, which holds for LoRA / QLoRA (base weights stay fp16). Full finetuning upcasts the trainable weights to float32, so a float16 activation then meets a float32 weight and the matmul fails.

## Root cause

In `unsloth_zoo/temporary_patches/gemma.py`, the forced patches always did `.to(torch.float16)` (plus the fp16 overflow clamp) at the RMSNorm output, the MLP `down_proj` input, and the attention `o_proj` input. The boundary dtype was fixed at float16 regardless of the actual weight dtype.

## Changes

All in `unsloth_zoo/temporary_patches/gemma.py`. The boundary dtype is now derived from the module's own weights instead of being hard-coded:

- Added `_module_compute_dtype(module, *attrs)`: returns the dtype of the first floating weight found (e.g. `down_proj.weight`), so the narrow target follows the real weights.
- Added `_narrow_to_boundary(x_fp32, dtype)`: float16 keeps the original clamp + downcast (bit-identical to before), float32 is a no-op (stays float32), bfloat16 clamps to its own range then casts.
- `patch_Gemma3RMSNorm`, `patch_Gemma3MLP`, `patch_Gemma3Attention`: the final `.to(torch.float16)` becomes `_narrow_to_boundary(..., _module_compute_dtype(self, ...))`.

The non-forced `*_generic` patches are untouched. All heavy reductions still run in float32 in both regimes, so the fp16 overflow safety is preserved.

## Behavior

- LoRA / QLoRA (fp16 weights): bit-identical to before. The fp16 clamp and downcast are unchanged, so there is no numerical change on the common path.
- Full finetuning (float32 weights): the whole Gemma3 block runs in float32, so there is no Half vs float mismatch. This is the correct pure float32 behavior for a model that cannot use float16.
- bfloat16 GPUs, AMD, Intel and MLX are unaffected: those use the untouched generic patches, and the fix uses only `tensor.to(dtype)` and `weight.dtype`, with no device specific calls.

## Validation

The crash reproduces on a Tesla T4 (sm_75, no bf16) and on a B200 (forcing `dtype=float16` to enter the `UNSLOTH_FORCE_FLOAT32` path with full finetuning), and the fix resolves it. Using the real transformers Gemma3 RMSNorm, MLP and attention modules with the patches applied:

- RMSNorm fp16 output is bit-identical to before (it always clamped). The MLP down_proj and attention o_proj boundaries now clamp to the fp16 range too (matching RMSNorm) rather than a bare downcast: identical to the old output for in-range activations, and clamped instead of overflowing to inf otherwise, which is strictly safer. LoRA and QLoRA losses match the old path to about four significant figures.
- float32 path matches a clean unpatched transformers float32 reference (max abs diff 0), with finite gradients.
- torch.compile enabled smoke test passes for both regimes (the boundary dtype is a per instance constant, so no graph break).
- Validated on a real Tesla T4 (sm_75, no bf16) together with unslothai/unsloth#5880: gemma3 full finetuning that crashed before now trains to a finite loss with 40 step stability (loss trends down), and LoRA / QLoRA still train with unchanged precision flags.

Only Gemma3 / Gemma3 text carry this hard-coded float16 assumption. gemma3n, gpt_oss and qwen3_5 already derive their boundary dtype from the running tensors, so they do not crash and need no change.

## Related

This is the downstream half of the V100 / no-bf16 full finetuning fix in unslothai/unsloth#5880, which corrects the trainer precision selection so the model reaches a clean float32 forward. With both, the Gemma3 family can be full finetuned on no-bf16 GPUs. The non-forced Gemma3 MLP `compute_dtype` handling in #462 is complementary and independent of this forced-path fix.

### DIFF
--- a/unsloth_zoo/temporary_patches/gemma.py
+++ b/unsloth_zoo/temporary_patches/gemma.py
@@ -449,6 +449,45 @@ pass
 TEMPORARY_PATCHES.append(patch_Gemma3MLP)
 
 
+def patch_Gemma3MultiModalProjector():
+    if os.environ.get("UNSLOTH_FORCE_FLOAT32", "0") == "0": return
+    try:
+        import transformers.models.gemma3.modeling_gemma3
+        transformers.models.gemma3.modeling_gemma3.Gemma3MultiModalProjector
+    except Exception as e:
+        return raise_error("Gemma3MultiModalProjector.forward", e)
+
+    def forward(self, vision_outputs: torch.Tensor):
+        batch_size, _, seq_length = vision_outputs.shape
+
+        reshaped_vision_outputs = vision_outputs.transpose(1, 2)
+        reshaped_vision_outputs = reshaped_vision_outputs.reshape(
+            batch_size, seq_length, self.patches_per_image, self.patches_per_image
+        )
+        reshaped_vision_outputs = reshaped_vision_outputs.contiguous()
+
+        pooled_vision_outputs = self.avg_pool(reshaped_vision_outputs)
+        pooled_vision_outputs = pooled_vision_outputs.flatten(2)
+        pooled_vision_outputs = pooled_vision_outputs.transpose(1, 2)
+
+        # mm_soft_emb_norm is a Gemma3RMSNorm, so under FORCE_FLOAT32 it can emit a float32
+        # activation (when its weight is upcast for stability) while mm_input_projection_weight
+        # stays a low-precision dtype -> "mat1 and mat2 must have the same dtype: float != Half"
+        # in the matmul below. Narrow the normed output to the projection weight's boundary
+        # dtype, exactly like the MLP/attention Linear boundaries: fp16 weights -> fp16 (with
+        # the overflow clamp, bit-identical to stock), fp32 weights (full finetuning) -> fp32.
+        normed_vision_outputs = self.mm_soft_emb_norm(pooled_vision_outputs)
+        boundary_dtype = _module_compute_dtype(self, "mm_input_projection_weight")
+        normed_vision_outputs = _narrow_to_boundary(normed_vision_outputs.to(torch.float32), boundary_dtype)
+
+        projected_vision_outputs = torch.matmul(normed_vision_outputs, self.mm_input_projection_weight)
+        return projected_vision_outputs.type_as(vision_outputs)
+    pass
+    patch_function(transformers.models.gemma3.modeling_gemma3.Gemma3MultiModalProjector, "forward", forward, fullgraph = False)
+pass
+TEMPORARY_PATCHES.append(patch_Gemma3MultiModalProjector)
+
+
 def patch_Gemma3Attention():
     if os.environ.get("UNSLOTH_FORCE_FLOAT32", "0") == "0": return
     try:
--- a/unsloth_zoo/temporary_patches/gemma.py
+++ b/unsloth_zoo/temporary_patches/gemma.py
@@ -74,10 +74,11 @@ def _module_compute_dtype(module, *attr_names):
 def _narrow_to_boundary(x_fp32, dtype):
     """Cast a float32 tensor to ``dtype`` for a Linear boundary.
 
-    For float16 we preserve the original overflow-safety clamp + downcast so the
-    fp16 path stays numerically identical to the unfixed forced patches. For
-    float32 this is a no-op (no narrowing happens). bfloat16 is clamped to its
-    own (very wide) range, which never actually clips, then downcast.
+    For float16 we clamp to the fp16 range then downcast, matching what the
+    RMSNorm patch already did. For in-range activations this equals the original
+    bare ``.to(float16)`` at the MLP/attention boundaries; for activations above
+    the fp16 max it clamps instead of overflowing to inf (strictly safer). For
+    float32 this is a no-op. bfloat16 clamps to its own (very wide) range first.
     """
     if dtype == torch.float32:
         return x_fp32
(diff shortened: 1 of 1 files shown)


danielhanchen:
Validated on a real Tesla T4 (sm_75, no native bf16, torch 2.11.0+cu128, transformers 4.57.6). Tested together with unsloth #5880 so the trainer selects pure float32 for the gemma3 family. BEFORE = unsloth_zoo main, AFTER = this branch.

Results
- gemma3 full finetuning (gemma-3-270m-it, dtype=float16, full_finetuning=True): on zoo main this crashes with `RuntimeError: expected mat1 and mat2 to have the same dtype, but got: c10::Half != float`. With this branch it selects pure float32 (`fp16=False, bf16=False, ACCELERATE_MIXED_PRECISION=no`) and trains to a finite loss.
- 40-step stability (this branch): all 40 losses finite, no NaN or inf, loss trends down (mean of first 10 = 2.02, mean of last 10 = 1.36).
- LoRA and QLoRA: precision flags unchanged (`fp16=False, bf16=False, AMP=no`), both train fine.

One thing to tweak: the change is not strictly bit-identical to main for fp16 LoRA/QLoRA, even though the description says so. The added `_narrow_to_boundary` calls put an fp16-range clamp at the MLP down_proj and attention o_proj boundaries, where main did a bare `.to(torch.float16)`. RMSNorm is genuinely identical, it always clamped. For LoRA/QLoRA the per-step losses match main to about 4 significant figures but diverge by roughly 1e-4 and compound over steps. I confirmed this is deterministic (an AFTER re-run reproduced the same losses exactly), so it is the clamp, not GPU noise. It only matters when an activation exceeds fp16 max, where main overflowed and this branch clamps, so it is safer, not a regression. Suggest softening the "bit-identical to before" wording for those two boundaries.

Context on how this was tested: I first tried to validate via #462, but that branch was 344 commits behind zoo main and would not import with unsloth #5880 (`ImportError: cannot import name '_unsloth_get_mm_token_id'`). This branch off current main is clean (ahead 1, behind 0) and gives a clean main vs main-plus-fix comparison. All results above use this branch.

Not testable on a 16 GB T4: gemma3n full FT (smallest is E2B at about 5B params) and gpt_oss full FT. gpt_oss QLoRA does run in float32 and is unchanged vs main.

danielhanchen:
Gemma notebook check on a real Tesla T4 (sm_75, no native bf16, torch 2.11.0+cu128, transformers 4.57.6, trl 0.25.1), stock vs this PR.

- STOCK = unsloth_zoo main `a69d9e6` + unsloth main `87ee06a`
- PR = this PR (#706) stacked on #707 + #709, with unsloth #5880 + #5898
- Each notebook run twice, seed 3407, max_steps 5, inference + local save kept.

Good news: **Gemma3N_(4B)-Vision passes**. On the T4 it correctly selects pure float32 (`UNSLOTH_FORCE_FLOAT32=1`, fp16=False bf16=False, never bf16) on both stock and this PR, trains all 5 steps, and the per-step loss and grad-norm are bit-identical stock vs PR (max abs diff 0.0), peak memory 14571 MB both. So this PR does not regress the gemma3n forced-float32 path.

Heads up on one path this PR does not fully cover: **Gemma3_(4B)-Vision** (gemma-3-4b, 4bit, vision) still crashes during `training_step` on both stock and this PR, with a different error on each:
- stock `a69d9e6`: `RuntimeError: expected scalar type BFloat16 but found Float`
- with this PR: `RuntimeError: expected mat1 and mat2 to have the same dtype, but got: float != c10::Half`

So #706 does change the forced-float32 behaviour (the error moves), but the Gemma3 vision tower still has a dtype boundary that ends up float vs half under forced float32 on the T4. The text/gemma3n paths are fine; the gemma-3-4b vision projection boundary looks like it needs the same weight-derived dtype treatment. Gemma3_(4B)-Vision-GRPO is a separate issue (CUDA OOM on 15GB during rollout, also on stock).

Happy to dig into the gemma-3-4b vision boundary if useful. The core forced-float32 fix here is correct and regression-free on the gemma3n vision notebook.

danielhanchen:
Follow-up: extend this PR to the gemma3 vision tower.

T4 (sm_75, no bf16) notebook validation surfaced that this PR shifted but did not fully resolve the gemma3 forced-float32 boundary for the vision model. Gemma3 (4B) Vision still failed with a Float vs Half dtype mismatch.

Root cause: the existing patches force the gemma3 text tower to float32, and `patch_Gemma3RMSNorm` makes every `Gemma3RMSNorm` emit float32 when its weight is upcast. `Gemma3MultiModalProjector.mm_soft_emb_norm` is a `Gemma3RMSNorm`, so under `UNSLOTH_FORCE_FLOAT32` its output became float32 while the sibling `mm_input_projection_weight` stayed low precision, so:

```
torch.matmul(normed_vision_outputs, self.mm_input_projection_weight)
RuntimeError: expected mat1 and mat2 to have the same dtype, but got: float != c10::Half
```

Fix: patch `Gemma3MultiModalProjector.forward` to narrow the normed vision features to the projection weight's boundary dtype, the same `_module_compute_dtype` + `_narrow_to_boundary` pattern used at the MLP and attention Linear boundaries (fp16 weights stay fp16 with the overflow clamp, full finetuning fp32 weights stay fp32), then run the matmul. The merge in `Gemma3Model.forward` already casts image features to `inputs_embeds.dtype`, so no merge-site change is needed. The patch is guarded on `UNSLOTH_FORCE_FLOAT32`, so bf16 and other normal paths are untouched.

Reproduction and validation (B200, spoofing no-bf16 by loading `unsloth/gemma-3-4b-pt` in float16, which forces float32 for the gemma3 family):

- Before: forward raises `float != c10::Half` inside `Gemma3MultiModalProjector_forward`.
- After: log shows `Patched Gemma3MultiModalProjector.forward`, forward and backward both run (loss 16.84).
- bf16 load (`UNSLOTH_FORCE_FLOAT32 = 0`): the projector patch is not applied and the path is unchanged (loss 16.83).

This addresses the Gemma3 (4B) Vision case from the T4 run. The other two gemma3-family T4 failures are separate and unrelated to this boundary: Gemma3 (4B) Vision GRPO is a 15 GB CUDA OOM, and Gemma3N (4B) Audio is a torch.compile/Dynamo `requires_grad_()` issue.

PR #707, opened 2026-05-31
Fix GRPO full finetuning KV cache dtype mismatch on no-bf16 GPUs

danielhanchen:
## Summary

GRPO full finetuning crashes during the generation rollout with `RuntimeError: index_copy_(): self and source expected to have the same dtype, but got (self) Float and (source) BFloat16` (or `Half` on a no-bf16 GPU). This blocks GRPO full finetuning on V100 / T4 and on any setup that runs the rollout under a low precision autocast.

## Root cause

Full finetuning upcasts the trainable weights to float32, so `model.dtype` is float32. The GRPO rollout runs `generate()` under the trainer's bf16/fp16 autocast. Inside attention, `apply_rotary_pos_emb` rotates the keys through the float32 `cos/sin` buffers, so the key reaching the cache is float32, while the value has no RoPE and stays the compute dtype (bf16/fp16).

`transformers/cache_utils.py` sizes the cache buffer lazily from the first key written (float32). The following value write then mismatches:

```
cache_utils.py  self.values.index_copy_(2, cache_position, value_states)
RuntimeError: index_copy_(): self and source expected to have the same dtype,
              but got (self) Float and (source) BFloat16
```

This is full finetuning specific: LoRA and QLoRA keep fp16/bf16 weights, so the keys and values agree and never diverge. It is also independent of `config.torch_dtype`: the buffer dtype comes from the runtime key dtype, not from the config, so changing `config.torch_dtype` does not address it.

## Fix

A new temporary patch (`patch_kv_cache_compute_dtype` in `unsloth_zoo/temporary_patches/misc.py`) aligns the key and value dtype at the cache write for every `CacheLayerMixin` subclass:

- First write: cast the key to the value (compute) dtype, so the rollout cache is allocated in fp16/bf16 (fast, half the KV memory).
- Later writes: cast both to the existing cache dtype, so they can never disagree.
- No-op when the key and value already match.

The value carries the true compute dtype because it is the un-rotated projection. This covers both the static cache (used by the vision fast generate path) and the dynamic cache (used by the llama fast generate path).

## Behavior

- Full finetuning rollout: the `index_copy_` crash is gone and the cache is allocated in the compute dtype (bf16/fp16) instead of float32, which is faster and uses half the KV memory.
- LoRA / QLoRA: no change. The keys and values already match, so the patch is a no-op and training is unaffected.
- Plain `model.generate()` for a full finetuned model (the float32 autocast inference path): the keys and values are both float32, so the patch is a no-op and the existing behavior is preserved.

## Validation

On `unsloth/Qwen2.5-0.5B-Instruct`, eager:

- Full finetuning GRPO, before: `index_copy_(): self Float vs source BFloat16`.
- Full finetuning GRPO, after: rollout first write `k_in=float32 v_in=bfloat16 -> cache buffer=bfloat16` for both the static and dynamic cache layers; the `index_copy_` crash is gone and the rollout completes.
- LoRA GRPO: trains to a finite loss, identical with and without the patch (the patch never fires).
- Plain `model.generate()` full finetuned: keys and values both float32, patch is a no-op, generation succeeds.

## Related

- Complements unslothai/unsloth#5880 (trainer precision selection for no-bf16 GPUs), which lets the no-bf16 full finetuning path reach the rollout in the first place.
- Orthogonal to the `config.torch_dtype` handling in PR #424; this fix does not require changing it.

## Note on a separate issue

After this fix, full finetuning GRPO reaches a distinct, pre-existing failure in the chunked selective log-softmax path (`chunk_hidden_states.to(lm_head.dtype) @ lm_head.t()`), where already projected logits are fed back through `lm_head` (a shape mismatch under eager, a Dynamo "same reduction dim" error under compile). That is unrelated to the KV cache dtype and is tracked separately.

### DIFF
--- a/unsloth_zoo/temporary_patches/misc.py
+++ b/unsloth_zoo/temporary_patches/misc.py
@@ -18,6 +18,7 @@ import torch
 import torch.nn as nn
 import inspect
 import importlib
+import functools
 from collections.abc import Mapping
 from typing import Any, List, Optional, Tuple, Union, Dict, Set, Callable
 from .common import TEMPORARY_PATCHES, torch_compile, _torch_compile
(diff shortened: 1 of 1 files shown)


danielhanchen:
Validated on a real Tesla T4 (sm_75, no native bf16, torch 2.11.0+cu128, transformers 4.57.6, trl 0.25.1). Tested together with unsloth #5880 so full finetuning selects fp16 autocast. BEFORE = unsloth_zoo main (a69d9e6), AFTER = this branch (620af9a), so the only difference is this commit. Model unsloth/Qwen2.5-0.5B-Instruct, dtype=float16, tiny GRPO (use_vllm=False, num_generations=2, max_steps=2, length reward), UNSLOTH_COMPILE_DISABLE=1.

All four runs select `fp16=True, bf16=False, ACCELERATE_MIXED_PRECISION=fp16`.

Full finetuning
- BEFORE (main): crash in the rollout at the first cache write. The RoPE-rotated key reaches the cache as float32 while the value stays fp16. Probe: `key=float32 value=float16 cache=<write failed>`.
  `RuntimeError: index_copy_(): self and source expected to have the same dtype, but got (self) Float and (source) Half`, at `cache_utils.py  self.values.index_copy_(2, cache_position, value_states)`.
- AFTER (this branch): index_copy_ is gone, the cache allocates in fp16. Probe: `key=float32 value=float16 cache=torch.float16`. The rollout completes, then the run hits the separate, already known #708 log-prob bug, exactly as expected.

LoRA (control, the patch should never fire)
- BEFORE and AFTER both train and are identical. The probe recorded no standard cache write for LoRA (the LoRA fast generation path does not go through CacheLayerMixin), consistent with the patch never firing.

Verdict: the rollout KV cache dtype mismatch is fixed and GRPO full finetuning is unblocked up to #708. LoRA is unaffected. LGTM.

The full-FT AFTER traceback for #708 is posted on that issue.

danielhanchen:
Stack confirmed end-to-end on a real Tesla T4. With this PR plus unsloth #5880 + #5898 and unsloth_zoo #709, GRPO full finetuning trains 10/10 steps on a no-bf16 T4 in both eager and compiled modes (DECISION fp16=True bf16=False AMP=fp16, finite train_loss, rewards and KL computed). The rollout cache fix here removes the index_copy_ Float vs Half crash, and the run no longer hits the #708 logits-shape error. LoRA is unaffected. Full end-to-end results are on unsloth #5898 and #709.

danielhanchen:
Confirmed unsloth_zoo/temporary_patches/misc.py has no KV cache dtype alignment today, so the RoPE-rotated float32 key still meets a half precision value at the first cache write during a full finetuning rollout. Will get this reviewed.

danielhanchen:
Confirmed the cache write path in unsloth_zoo/temporary_patches/misc.py still lets the RoPE-rotated float32 key size the buffer against a half precision value, and the T4 runs you posted line up. Will get this reviewed.

danielhanchen:
Confirmed the rollout still writes a float32 RoPE key against a half value in unsloth_zoo/temporary_patches/misc.py on main, and the T4 before/after runs here line up with that. Will get this reviewed.

Issue #708, opened 2026-05-31
GRPO full finetuning: chunked log-softmax receives logits instead of hidden states


danielhanchen:
## Summary

After the KV cache dtype fix in #707, GRPO full finetuning reaches a second, independent failure in the chunked selective log-softmax path. `chunked_hidden_states_selective_log_softmax` is called with already projected logits (last dim is the vocab size) instead of hidden states (last dim is the hidden size), so the `@ lm_head.t()` matmul has mismatched shapes.

This is full finetuning specific. LoRA and QLoRA GRPO are unaffected.

## Where

`unsloth_zoo/rl_replacements.py`, `chunked_hidden_states_selective_log_softmax` (line 74), at:

```python
chunk_logits = chunk_hidden_states.to(lm_head.dtype) @ lm_head.t()   # line 94
```

The function expects `hidden_states` of shape `(batch, seq, hidden)` and `lm_head` of shape `(vocab, hidden)`, so `chunk_hidden_states (n, hidden) @ lm_head.t() (hidden, vocab)`. In full finetuning it instead receives `(n, vocab)`, so the reduction dims do not match.

## Errors

Eager (with `TORCHDYNAMO_DISABLE=1`):

```
RuntimeError: mat1 and mat2 shapes cannot be multiplied (4x151936 and 896x151936)
```

`4x151936` is `(n, vocab)`: the input is logits, not hidden states (`hidden = 896`, `vocab = 151936` for Qwen2.5-0.5B).

Compiled (the function carries a hard `@torch.compile(dynamic = True, fullgraph = True)`, so this fires even with `UNSLOTH_COMPILE_DISABLE=1`):

```
torch._dynamo.exc.TorchRuntimeError: Dynamo failed to run FX node with fake tensors:
call_function matmul(FakeTensor(size=(((s47*s87 + 7)//8), s33)),
                     FakeTensor(size=(896, 151936))):
a and b must have same reduction dim, but got [((s47*s87 + 7)//8), s33] X [896, 151936].
```

## Likely cause

The chunked path depends on `UNSLOTH_RETURN_HIDDEN_STATES` so the model returns hidden states and the trainer projects them in chunks. In full finetuning the model appears to return logits instead, so the already projected logits get passed back through `lm_head`. This needs the full finetuning lm_head / `UNSLOTH_RETURN_HIDDEN_STATES` handling to be checked for the GRPO path.

## Reproduce

[...]

PR #709, opened 2026-05-31
Guard chunked log-softmax against receiving logits in GRPO

danielhanchen:
## Summary

Defensive guard for unslothai/unsloth-zoo#708. If a forward returns logits instead of hidden states (for example full finetuning, where `UNSLOTH_RETURN_HIDDEN_STATES` could not be honored), `chunked_hidden_states_selective_log_softmax` projected already projected logits through `lm_head` again and crashed:

```
RuntimeError: mat1 and mat2 shapes cannot be multiplied (Nx151936 and 896x151936)
```

## Change (`unsloth_zoo/rl_replacements.py`)

In `chunked_hidden_states_selective_log_softmax`, detect when the incoming tensor's last dim already equals the vocab size (`lm_head.shape[0]`), i.e. it is already logits, and skip the `lm_head` projection. Logits are exactly `hidden_states @ lm_head.t()`, so using them directly is numerically equivalent. It is a no-op for genuine hidden states (last dim is the hidden size), so LoRA / QLoRA are unchanged.

## Behavior

- A forward that returns logits no longer crashes the log-prob step; the logits are used directly.
- Genuine hidden states take the normal projection path, so there is no change for LoRA / QLoRA or any model that honors `UNSLOTH_RETURN_HIDDEN_STATES`.
- The added branch is on a static shape, so it does not break `torch.compile` (validated: full finetuning and LoRA both compile with zero recompiles).

## Validation

On `unsloth/Qwen2.5-0.5B-Instruct` (with unsloth-zoo#707 applied so the rollout reaches the log-prob step):

| Mode | Compile | Before | After |
|------|---------|--------|-------|
| full finetuning | eager | crash `(4x151936 and 896x151936)` | trains, finite loss |
| full finetuning | compiled | crash | trains, 0 recompiles |
| LoRA | eager | trains | trains, identical (no-op) |
| QLoRA | eager | trains | trains |

## Related

This is the consumer-side guard. The upstream cause (full finetuning forward returning logits) is fixed in unslothai/unsloth, which makes the forward return hidden states again so the chunked path stays memory efficient. This guard is complementary insurance against any model whose forward returns logits, and is independent of that fix.

### DIFF
--- a/unsloth_zoo/rl_replacements.py
+++ b/unsloth_zoo/rl_replacements.py
@@ -88,10 +88,19 @@ def chunked_hidden_states_selective_log_softmax(
     chunked_hidden_states = torch.chunk(flat_hidden_states, chunks=chunks, dim=0)
     chunked_index = torch.chunk(flat_index, chunks=chunks, dim=0)
 
+    # If UNSLOTH_RETURN_HIDDEN_STATES could not be honored (e.g. a full finetuning forward that
+    # returns logits), the incoming tensor is already logits, not hidden states. Detect that
+    # (last dim == vocab) and skip the lm_head projection; logits are exactly
+    # hidden_states @ lm_head.t(), so using them directly is equivalent. No-op for hidden states.
+    already_logits = hidden_states.shape[-1] == lm_head.shape[0]
+
     all_per_token_logps = []
 
     for chunk_hidden_states, chunk_index in zip(chunked_hidden_states, chunked_index):
-        chunk_logits = chunk_hidden_states.to(lm_head.dtype) @ lm_head.t()
+        if already_logits:
+            chunk_logits = chunk_hidden_states
+        else:
+            chunk_logits = chunk_hidden_states.to(lm_head.dtype) @ lm_head.t()
 
         if logit_scale_multiply != 0.0:
             chunk_logits = chunk_logits * logit_scale_multiply


danielhanchen:
Validated end-to-end on a real Tesla T4 (sm_75, no native bf16, torch 2.11.0+cu128, transformers 4.57.6, trl 0.25.1). Together with unsloth #5898 (the root fix that returns hidden states), this guard closes #708. I stacked all four GRPO full finetuning fixes into one test branch per repo (unsloth #5880 + #5898, unsloth_zoo #707 + #709) and ran tiny GRPO on unsloth/Qwen2.5-0.5B (dtype=float16, num_generations=2, max_steps=10, length reward).

Versions: unsloth 2026.5.8 (940fe0b), unsloth_zoo 2026.5.4 (14b0e30).

Results, all with `DECISION fp16=True bf16=False AMP=fp16`:
- full finetuning, eager (UNSLOTH_COMPILE_DISABLE=1): TRAINED OK, 10/10 steps, train_loss 0.009076, finite.
- full finetuning, compiled: TRAINED OK, 10/10 steps, train_loss 0.010111, finite.
- LoRA: TRAINED OK, 10/10 steps, train_loss 0.010101, finite.

No `index_copy_(): self Float vs source Half` and no `mat1 and mat2 shapes cannot be multiplied (8x151936 and 896x151936)`, in both the eager and compiled paths. KL grows from 0 to about 0.065 over 10 steps and the full finetuning completion length moves from 32 to 19.5 by step 10, so the rollout is computing real signal. LGTM.

PR #710, opened 2026-06-01
fix: replace exec/eval with importlib in patch_compiling_bitsandbytes

### LINKED ISSUE #311: 'peft' is not defined in patching_utils.py

Rahul007007:
Fixes #311

### Problem

`patch_compiling_bitsandbytes()` iterated over the dotted module strings `["bitsandbytes.nn.modules", "peft.tuners.lora.bnb"]` and used:

```python
exec(f"import {x}", globals(), locals())
layers = dir(eval(x))
```

to pull the module out and walk its attributes. Two problems:

1. `exec("import a.b.c", globals, locals)` does **not** reliably bind the top-level name `a` into the supplied namespaces; the subsequent `eval("a.b.c")` then raises `NameError: name 'a' is not defined`. This is the failure reported in #311 ("peft not defined") and is consistent with CPython's documented behaviour around `exec` with explicit `globals`/`locals` mappings.
2. If `peft` isn't installed, the user just sees a bare `ImportError` from inside an Unsloth internal patch with no hint that they need to `pip install peft`.

### Fix

Replace the `exec` / `eval` dance with `importlib.import_module(x)` + `getattr(module, fx)` — the standard, well-defined way to import a module by string name and walk its attributes. Wrap it in `try/except ImportError` so:

- if `bitsandbytes.nn.modules` somehow isn't importable, we just `continue`;
- if `peft.tuners.lora.bnb` is missing, we re-raise with a clear `ImportError("Unsloth: Please install peft via 'pip install peft'")`.

No behaviour change in the happy path (still calls `torch._disable_dynamo` on every `forward` that isn't already wrapped), but the unhappy paths no longer crash mysteriously.

### Notes

- Only touches the `< 0.46.0` branch; the `>= 0.46.0` branch is unchanged.
- Added `import importlib` at the top of the file.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/unsloth_zoo/patching_utils.py
+++ b/unsloth_zoo/patching_utils.py
@@ -48,16 +48,18 @@ def patch_compiling_bitsandbytes():
         for x in ["bitsandbytes.nn.modules", "peft.tuners.lora.bnb",]:
             try:
                 module = importlib.import_module(x)
-            except ImportError:
-                # peft is required for LoRA training; give a helpful message
+            except ImportError as e:
+                # peft is required for LoRA training
                 if x.startswith("peft"):
                     raise ImportError(
                         "Unsloth: Please install peft via `pip install peft`"
-                    )
+                    ) from e
+                if os.environ.get("UNSLOTH_ENABLE_LOGGING", "0") == "1":
+                    print(f"Unsloth: Skipping {x} - import failed: {e}")
                 continue
             for fx in dir(module):
                 try: layer = getattr(module, fx)
-                except: continue
+                except Exception: continue
                 if not hasattr(layer, "forward"): continue
                 if hasattr(layer.forward, "__wrapped__"): continue
                 layer.forward = torch._disable_dynamo(layer.forward)
--- a/unsloth_zoo/patching_utils.py
+++ b/unsloth_zoo/patching_utils.py
@@ -18,6 +18,7 @@ import torch
 import os
 import re
 import ast
+import importlib
 
 __all__ = [
     "patch_compiling_bitsandbytes",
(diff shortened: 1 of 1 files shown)


danielhanchen:
Thanks for the fix! Verified this resolves #311.

Root cause confirmed: Python 3.13 (PEP 667) changed `locals()` semantics, so `exec(f"import {x}", globals(), locals())` no longer persists the imported name and the following `eval` raises `NameError`. On Python 3.12 and below the old code worked by accident, which is why this only started showing up recently. With bitsandbytes < 0.46.0 every `FastModel.from_pretrained` call hits this.

Validation on Python 3.13.12:

- On main, loading any model with bitsandbytes < 0.46.0 crashes with `NameError: name 'peft' is not defined`. With this PR it loads and trains fine.
- The new importlib path wraps the exact same 15 forward attributes across `bitsandbytes.nn.modules` and `peft.tuners.lora.bnb` as the old code did on 3.12.
- Two seeded 61 step QLoRA runs (Qwen2.5 0.5B bnb 4bit) produce bitwise identical losses and grad norms versus a reference fix, with identical peak memory.
- The bitsandbytes >= 0.46.0 branch is untouched and stays a no-op, checked at 0.46.0 and 0.49.2.
- Missing peft raises the new helpful install message. Repeated calls do not double wrap.

Pushed a small commit on top (8a17d74):

- `raise ... from e` so an installed but broken peft keeps its original traceback instead of being mislabeled as missing peft.
- Log skipped bitsandbytes submodule import failures under `UNSLOTH_ENABLE_LOGGING=1` instead of continuing silently.
- Tightened the bare `except:` to `except Exception:` in the attribute loop.

All checks re-verified after the commit.

danielhanchen:
One correction to my validation comment above: the `raise ... from e` change chains the original cause into the traceback, but the surface message still says "Please install peft" when peft is installed and one of its transitive dependencies fails during import. #743 fixes that properly by classifying the failure via `ModuleNotFoundError.name`, so only a genuinely missing package gets the install hint and real errors surface unchanged.

PR #711, opened 2026-06-01
fix: normalize tuple/list patch_size to int in UnslothVisionDataCollator

### LINKED ISSUE #316: UnslothVisionDataCollator cannot handle a tuple as patch size

Rahul007007:
Fixes #316

### Problem

Some vision configs specify `vision_config.patch_size` as a `(height, width)` tuple/list rather than a single int. The reported example is [OpenGVLab/InternVL3-2B-hf](https://huggingface.co/OpenGVLab/InternVL3-2B-hf/blob/main/config.json) where `patch_size = [14, 14]`.

`UnslothVisionDataCollator` then uses `self.patch_size * 2` as a scalar `size_factor` for image resizing (in `_extract_images_videos_for_example`, `_extract_images_for_pc`, and `_resize_images_inplace`). When `patch_size` is a list, `* 2` is **list duplication** (`[14, 14] * 2 == [14, 14, 14, 14]`), not multiplication, so the downstream arithmetic in `smart_resize` / `quantize_to_factor` crashes with:

```
TypeError: unsupported operand type(s) for /: 'int' and 'list'
```

Reproduction (on the actual model from the issue) before this PR:

```
collator.patch_size = [14, 14] (type list)
File ".../vision_utils.py", line 996, in _resize_images_inplace
    new_w, new_h = quantize_to_factor(new_w), quantize_to_factor(new_h)
  File ".../vision_utils.py", line 971, in quantize_to_factor
    math.ceil(x/factor) if x >= image_size - 1e-6 else math.floor(x/factor+0.5)
TypeError: unsupported operand type(s) for /: 'int' and 'list'
```

### Fix

Normalize `self.patch_size` to a Python `int` immediately after it is assigned in `UnslothVisionDataCollator.__init__`:

| input                          | result                              |
|--------------------------------|-------------------------------------|
| `int` / numpy scalar           | `int(x)` (no-op for plain `int`)    |
| `list`/`tuple`, all equal      | first element                       |
| `list`/`tuple`, mismatched     | `max(...)` + `logger.warning(...)`  |
| empty `list`/`tuple`           | `IMAGE_FACTOR // 2` fallback        |

The `max` rule for non-square patches is the conservative choice for the way `patch_size` is used here (`size_factor = patch_size * 2`): a divisor for ensuring the resized image dimensions are valid in both axes. In practice every shipped VLM I checked has square patches, so the mismatched branch is defensive.

### Validation

Tested with the actual model configs (no weight download needed — `AutoConfig.from_pretrained` only):

| Case | Before | After |
|------|--------|-------|
| `OpenGVLab/InternVL3-2B-hf` (`patch_size=[14,14]`) | crashes in `_resize_images_inplace` | `patch_size=14`, image (800x600) -> (532, 392), both divisible by 28 |
| `Qwen/Qwen2-VL-2B-Instruct` (`patch_size=14`) | works | works, value preserved |
| `unsloth/Qwen2.5-0.5B-Instruct` (no `vision_config.patch_size`) | falls back to `IMAGE_FACTOR // 2` | unchanged |
| Synthetic `(14, 16)` | would crash | `patch_size=16` + warning |
| Synthetic `[]` | would crash | falls back to `IMAGE_FACTOR // 2` |

The end-to-end resize on InternVL3 was confirmed to succeed: PIL `Image.new("RGB", (800, 600))` -> `(532, 392)` after `_resize_images_inplace`, both divisible by `patch_size * 2 = 28`.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/unsloth_zoo/vision_utils.py
+++ b/unsloth_zoo/vision_utils.py
@@ -697,6 +697,25 @@ class UnslothVisionDataCollator:
             else:
                 self.patch_size = IMAGE_FACTOR // 2
 
+        # Some configs (e.g. InternVL3) specify patch_size as a (height, width)
+        # tuple/list, but UnslothVisionDataCollator uses it as a scalar size
+        # factor (`self.patch_size * 2`) for image resizing. Coerce to a single
+        # int: take the unique value for square patches, else fall back to the
+        # max so resized images remain divisible in both dimensions.
+        if isinstance(self.patch_size, (tuple, list)):
+            if len(self.patch_size) == 0:
+                self.patch_size = IMAGE_FACTOR // 2
+            elif all(p == self.patch_size[0] for p in self.patch_size):
+                self.patch_size = int(self.patch_size[0])
+            else:
+                logger.warning(
+                    f"Unsloth: non-square vision patch_size {tuple(self.patch_size)} "
+                    f"detected; using max for image size factor."
+                )
+                self.patch_size = int(max(self.patch_size))
+        else:
+            self.patch_size = int(self.patch_size)
+
         # Auto resize images to save VRAM!
         if resize == "min":
             try:
--- a/unsloth_zoo/vision_utils.py
+++ b/unsloth_zoo/vision_utils.py
@@ -697,11 +697,8 @@ class UnslothVisionDataCollator:
             else:
                 self.patch_size = IMAGE_FACTOR // 2
 
-        # Some configs (e.g. InternVL3) specify patch_size as a (height, width)
-        # tuple/list, but UnslothVisionDataCollator uses it as a scalar size
-        # factor (`self.patch_size * 2`) for image resizing. Coerce to a single
-        # int: take the unique value for square patches, else fall back to the
-        # max so resized images remain divisible in both dimensions.
+        # Configs like InternVL3 store patch_size as (height, width); coerce to
+        # an int since it is used as a scalar size factor (patch_size * 2).
         if isinstance(self.patch_size, (tuple, list)):
             if len(self.patch_size) == 0:
                 self.patch_size = IMAGE_FACTOR // 2
@@ -710,16 +707,23 @@ class UnslothVisionDataCollator:
             else:
                 logger.warning(
                     f"Unsloth: non-square vision patch_size {tuple(self.patch_size)} "
-                    f"detected; using max for image size factor."
+                    f"detected; using the least common multiple as the size factor."
                 )
-                self.patch_size = int(max(self.patch_size))
+                self.patch_size = math.lcm(*(int(p) for p in self.patch_size))
+        elif self.patch_size is None:
+            self.patch_size = IMAGE_FACTOR // 2
         else:
             self.patch_size = int(self.patch_size)
 
         # Auto resize images to save VRAM!
         if resize == "min":
             try:
-                self.image_size = model.config.vision_config.image_size
+                image_size = model.config.vision_config.image_size
+                # Configs like InternVL3 store image_size as [height, width];
+                # a tuple makes _resize_images_inplace do a fixed-size resize.
+                if isinstance(image_size, (tuple, list)):
+                    image_size = tuple(int(x) for x in image_size) if len(image_size) else None
+                self.image_size = image_size
             except Exception:
                 print("Unsloth: Model does not have a default image size - using 512")
                 self.image_size = 512


unsloth_zoo/vision_utils.py
```diff
@@ -697,6 +697,25 @@ def __init__(
...
+                logger.warning(
+                    f"Unsloth: non-square vision patch_size {tuple(self.patch_size)} "
+                    f"detected; using max for image size factor."
+                )
+                self.patch_size = int(max(self.patch_size))
+        else:
+            self.patch_size = int(self.patch_size)
+
         # Auto resize images to save VRAM!
         if resize == "min":
```
danielhanchen, line 719: **[2/15 reviewers, confirmed empirically]** The linked model still crashes after this PR when using the default `resize="min"`. `OpenGVLab/InternVL3-2B-hf` also stores `vision_config.image_size` as a list (`[448, 448]`), which lands unmodified in `self.image_size`. `_resize_images_inplace` only treats a `tuple` as a fixed-size resize (`type(image_size) is tuple`), so a list falls through to `side > image_size` and raises `TypeError: '>' not supported between instances of 'int' and 'list'`.

Reproduced with the real config: `collator([example])` works on this branch only with an explicit int `resize` (for example `resize=448`); with the default `resize="min"` it still fails for InternVL3.

Suggested fix at the assignment a couple of lines below (line 722), mirroring the `patch_size` normalization:

```python
            try:
                image_size = model.config.vision_config.image_size
                if isinstance(image_size, (tuple, list)):
                    image_size = tuple(int(x) for x in image_size) if len(image_size) else None
                self.image_size = image_size
```

A tuple takes the existing fixed-size resize path in `_resize_images_inplace`, so `[448, 448]` resizes images to exactly 448x448, and `None` falls back to no resizing.

unsloth_zoo/vision_utils.py
```diff
@@ -697,6 +697,25 @@ def __init__(
...
+            if len(self.patch_size) == 0:
+                self.patch_size = IMAGE_FACTOR // 2
+            elif all(p == self.patch_size[0] for p in self.patch_size):
+                self.patch_size = int(self.patch_size[0])
+            else:
+                logger.warning(
+                    f"Unsloth: non-square vision patch_size {tuple(self.patch_size)} "
+                    f"detected; using max for image size factor."
+                )
+                self.patch_size = int(max(self.patch_size))
```
danielhanchen: **[13/15 reviewers]** For genuinely non-square patches the `max()` rule does not do what the comment says. With `(14, 16)` the factor becomes `max * 2 = 32`, and `32 % 28 != 0`, so the 14 axis is not divisible by its required factor; the claim "resized images remain divisible in both dimensions" only holds when the smaller value divides the larger one. The least common multiple is the smallest scalar that satisfies both axes (`lcm(14, 16) = 112`, factor 224). This branch is defensive only (every shipped VLM we checked has square patches), so this is non-blocking, but if the branch exists it should be correct:

```suggestion
        # Some configs (e.g. InternVL3) specify patch_size as a (height, width)
        # tuple/list, but UnslothVisionDataCollator uses it as a scalar size
        # factor (`self.patch_size * 2`) for image resizing. Coerce to a single
        # int: take the unique value for square patches, else use the least
        # common multiple so both axes stay divisible.
        if isinstance(self.patch_size, (tuple, list)):
            if len(self.patch_size) == 0:
                self.patch_size = IMAGE_FACTOR // 2
            elif all(p == self.patch_size[0] for p in self.patch_size):
                self.patch_size = int(self.patch_size[0])
            else:
                logger.warning(
                    f"Unsloth: non-square vision patch_size {tuple(self.patch_size)} "
                    f"detected; using the least common multiple as the size factor."
                )
                self.patch_size = math.lcm(*(int(p) for p in self.patch_size))
```

unsloth_zoo/vision_utils.py
```diff
@@ -697,6 +697,25 @@ def __init__(
...
+        else:
+            self.patch_size = int(self.patch_size)
```
danielhanchen, line 716: **[4/15 reviewers]** An explicit `"patch_size": null` in a config now raises `TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'` at collator construction. Before this PR, `None` survived `__init__` and never crashed at all when the dataset uses a top-level `images` column (which skips `process_vision_info`) with `snap_to_patch_size=False`, the default. No transformers config class defaults `patch_size` to `None` (we checked all 110 vision configs in 4.57.6), so this is a corner case, but the guard is free and keeps this strictly safer than the previous behavior:

```suggestion
        elif self.patch_size is None:
            self.patch_size = IMAGE_FACTOR // 2
        else:
            self.patch_size = int(self.patch_size)
```

danielhanchen (dismissed):
Thank you for the PR! The goal of this PR is to fix #316, where `UnslothVisionDataCollator` crashes when `vision_config.patch_size` is a tuple/list. As a summary, this PR normalizes tuple/list `patch_size` values to a scalar `int` in `__init__` so the `size_factor = patch_size * 2` arithmetic in `process_vision_info` / `smart_resize` / `quantize_to_factor` no longer fails for configs such as `OpenGVLab/InternVL3-2B-hf` (`patch_size=[14, 14]`).

We verified the fix end to end (unit battery on the real configs plus 61-step LoRA training runs, bs 2, grad accum 3, seed 3407, bf16 LoRA r16):

**unsloth/Qwen2-VL-2B-Instruct (int `patch_size=14`, control)**

| Method | Samples/s | Peak VRAM | Losses [1, 2, 3, ..., 60, 61] | Grad norms [1, 2, 3, ..., 60, 61] |
|---|---|---|---|---|
| main | 2.681 | 5.78 GB | [2.99, 3.0, 2.824, ..., 0.887, 1.156] | [4.746, 7.699, 4.128, ..., 1.305, 1.33] |
| PR #711 | 2.685 | 5.78 GB | [2.99, 3.0, 2.831, ..., 0.879, 1.154] | [4.753, 7.697, 4.349, ..., 1.217, 1.058] |

Collator outputs on real data are bitwise identical between main and this PR (`input_ids`, `attention_mask`, `pixel_values`, `image_grid_thw`, `labels` over 8 samples, `torch.equal`), so the change is a provable no-op for every model with an int `patch_size`. The tiny loss drift above is standard bf16 kernel nondeterminism.

**OpenGVLab/InternVL3-2B-hf (`patch_size=[14, 14]`, the issue #316 model), run with explicit `resize=448`**

| Method | Samples/s | Peak VRAM | Losses [1, 2, 3, ..., 60, 61] | Grad norms [1, 2, 3, ..., 60, 61] |
|---|---|---|---|---|
| main | crashes: `TypeError: unsupported operand type(s) for /: 'int' and 'list'` | - | - | - |
| main + manual `patch_size=14` | 2.621 | 9.21 GB | [22.331, 22.445, 21.543, ..., 7.475, 7.499] | [17.353, 16.633, 15.808, ..., 0.145, 0.154] |
| PR #711 | 2.564 | 9.21 GB | [22.331, 22.445, 21.549, ..., 7.474, 7.499] | [17.352, 16.632, 15.776, ..., 0.145, 0.157] |

So the `patch_size` normalization itself is correct and well scoped. However, our review battery (12 automated reviewers plus 3 independent deep dives, 15 total) found one gap that keeps the linked issue from being fully fixed, plus two smaller items:

|---|---|---|
| 2/15, confirmed empirically | High | InternVL3-2B-hf still crashes with default `resize="min"`: the same config also ships `image_size=[448, 448]` as a list, `self.image_size` keeps it, and `_resize_images_inplace` fails at `side > image_size` with `TypeError: '>' not supported between instances of 'int' and 'list'`. The PR description's end-to-end claim only reproduces when an int `resize` is passed explicitly. |
| 13/15 | Medium | For non-square `(h, w)`, `max()` does not preserve divisibility on both axes (`(14, 16)` gives factor 32, but the 14 axis needs 28), and the code comment claims it does. Use `math.lcm` or fail loudly. Defensive only, no shipped VLM has non-square patches. |
| 4/15 | Medium | Explicit `patch_size=None` in a config now raises `TypeError: int(None)` at construction time, where it previously trained fine when the dataset uses a top-level `images` column with `snap_to_patch_size=False`. A one-line guard restores that. |

For context on overlapping PRs: #399 and #405 carry a simpler `patch_size[0]` variant of this fix (#412 was closed as superseded). This PR is the most thorough standalone version, so with the items above addressed it would be the one to merge for #316.

danielhanchen:
Pushed 591b3caf on top of this branch to address the three review findings directly:

- Normalized the config derived `image_size` the same way as `patch_size`: a list/tuple becomes a tuple (which takes the fixed-size resize path in `_resize_images_inplace`), an empty one becomes `None`. `OpenGVLab/InternVL3-2B-hf` now works with the default `resize="min"`, which previously still failed with `TypeError: '>' not supported between instances of 'int' and 'list'` even after the `patch_size` fix.
- An explicit `patch_size=None` now falls back to `IMAGE_FACTOR // 2` instead of crashing at `int(None)` during construction.
- Non-square patch sizes use `math.lcm` so both axes stay divisible by their patch factors, and the comment plus warning were reworded to match. Comments were also shortened.

Re-validated after the commit:

| Check | Result |
|---|---|
| Unit battery: real InternVL3-2B-hf and Qwen2-VL-2B configs plus synthetic `(14, 16)`, `[]`, numpy scalar, float and `None` patch sizes | 37/37 pass |
| InternVL3-2B-hf LoRA smoke run with default collator settings | trains, finite losses, no collator errors |
| Qwen2-VL-2B collator outputs vs main (`torch.equal` on all batch tensors) | bitwise identical, int configs unaffected |

Thanks for the fix @Rahul007007, this resolves #316 fully now.

PR #713, opened 2026-06-02
Capture outputs fixes for transformers v5

Datta0:
Fixes: unslothai/unsloth#4812

### DIFF
--- a/tests/test_compiler_output_capture.py
+++ b/tests/test_compiler_output_capture.py
@@ -1,0 +1,114 @@
+# Tests for the transformers v5 output capture helpers in unsloth_zoo.compiler:
+# patch_output_capture_targets and calls_output_capture_target.
+
+import types
+
+import pytest
+
+from unsloth_zoo.compiler import (
+    calls_output_capture_target,
+    patch_output_capture_targets,
+)
+
+output_capturing = pytest.importorskip(
+    "transformers.utils.output_capturing",
+    reason="output capture targets exist on transformers >= 5.2 only",
+)
+OutputRecorder = output_capturing.OutputRecorder
+
+
+class FakeRouter:
+    pass
+
+
+class FakeAttention:
+    pass
+
+
+class FakeDecoderLayer:
+    pass
+
+
+class ReplacementRouter:
+    __name__ = "FakeRouter"
+
+
+class ReplacementAttention:
+    __name__ = "FakeAttention"
+
+
+def make_modeling_file():
+    mod = types.ModuleType("fake_modeling")
+
+    class FakeModel:
+        _can_record_outputs = {
+            "router_logits": OutputRecorder(FakeRouter, index=0),
+            "hidden_states": FakeDecoderLayer,
+            "attentions": [OutputRecorder(FakeAttention, index=1)],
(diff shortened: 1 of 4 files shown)


danielhanchen:
Pushed c671a06d on top of this branch. While testing the capture fixes end to end on a tiny Qwen3 MoE (full finetuning, `output_router_logits=True`), the #4812 repro still crashed under the default `UNSLOTH_FUSED_FORWARD=1` before the retargeted hooks could matter, so this PR alone would close the issue while its repro still fails. Two stacked problems in `fused_losses`, both reproducible on transformers 4.57.6 and 5.5.0:

1. `forward_install.py` seeded the exec namespace from `forward.__globals__`. HF forwards are decorator wrapped, so that resolves to `transformers.utils.generic`, while `inspect.getsource` unwraps to the modeling source which references modeling level helpers. MoE bodies call `load_balancing_loss_func` and raised `NameError`. Now seeded from `inspect.unwrap(forward).__globals__` with the wrapper globals overlaid, a strict superset of the old namespace.
2. With names resolving, MoE forwards then hit `loss += router_aux_loss_coef * aux_loss` on the fused loss, which is an autograd Function view output, raising the view inplace `RuntimeError`. The adapter now clones the loss before returning.

Also added a small guard that skips installation with a warning when a rewritten body references names that resolve in neither the namespace nor builtins, so future drift degrades gracefully instead of failing mid training.

Verification, all with fused forward at its default:

| Config | Before | After |
|---|---|---|
| transformers 5.5.0, tiny Qwen3 MoE, 61 steps | NameError | Trains, `router_logits` length 4 per forward, losses match the non fused run within 0.001 |
| transformers 5.5.0, dense Qwen3 | OK | Bitwise identical losses |
| transformers 4.57.6, MoE and dense | NameError on MoE | Both train, MoE matches the non fused baseline |
| `tests/test_fused_forward_install.py` | n/a | 29 passed on 4.57.6, 5.5.0 and 5.10.2 |
| trl 0.22.2 / 0.27.1 / 1.5.1 | n/a | Fused install succeeds, suite passes |

The compiler changes themselves check out: control runs on zoo main reproduce the exact #4812 `IndexError`, this branch fixes them, capture specs end up targeting the rewritten classes, and dense plus 4.57.6 behavior is unchanged.

danielhanchen:
Also pushed 5d3b71cf with two follow ups:

1. `calls_output_capture_target` only matched submodules assigned directly in `__init__`, so models building captured submodules indirectly (Zamba2 constructs layers via `get_layers`) kept `fullgraph=True` around a forward that still carries `@capture_outputs`, which hard errors under compile once hidden states or attentions are requested. The detection now also checks the forward source for the preserved decorator. Verified no behavior change on the common paths: the emitted compiled caches for Qwen3 MoE and dense Qwen3 are identical before and after, since their capture decorated classes never reach the standalone compile path.
2. Unit tests for the two new helpers in `tests/test_compiler_output_capture.py`: spec name collection, `OutputRecorder` and bare class retargeting with list and tuple specs preserved, direct call detection, uncalled targets, and the decorated forward case. They skip below transformers 5.2 where `output_capturing` does not exist, and pass on 5.5.0.

PR #714, opened 2026-06-03
Logging cleanup

danielhanchen:
Logging cleanup


danielhanchen:
Split into #715 and #716

PR #717, opened 2026-06-04
fix: set _unsloth_grouped_mm_format on stock transformers GptOssExperts so MoE LoRA grouped_mm works

### LINKED ISSUE #683: gpt-oss-20b MoE LoRA: missing _unsloth_grouped_mm_format flag on transformers' GptOssExperts causes _grouped_mm contraction dim mismatch
## Summary

`unsloth_zoo`'s MoE LoRA path on `gpt-oss-20b` (with `transformers==5.5.0`) crashes on the first forward pass with a `torch._grouped_mm` contraction dimension mismatch. The root cause is a per-instance flag, `_unsloth_grouped_mm_format`, which `_get_moe_lora_io_dims` (in `unsloth_zoo/temporary_patches/moe_utils.py`) reads to decide between the `gpt-oss` `(num_experts, in_dim, out_dim)` weight layout and the Qwen3-MoE `(num_experts, out_dim, in_dim)` layout. The flag is set in Unsloth's `GptOssExperts` constructor — but the runtime instances under transformers 5.x come from `transformers.models.gpt_oss.modeling_gpt_oss.GptOssExperts`, not Unsloth's class, so the flag never gets set, the Qwen3 branch fires, the reshape doubles the inner dimension, and `torch._grouped_mm` correctly rejects the result.

A one-line runtime patch (iterate `model.modules()`, match by type name, set the flag) resolves it. Proposed upstream fix is to apply that loop inside `unsloth_zoo`'s MoE patching path so end users don't need to know about the flag at all.

## Environment

| Component | Version |
|---|---|
| torch | 2.10.0+cu128 |
| transformers | 5.5.0 |
| trl | 0.24.0 |
| peft | 0.19.1 |
| unsloth | 2026.5.5 |
| unsloth_zoo | 2026.5.3 |
| bitsandbytes | 0.49.2 |
| accelerate | 1.13.0 |
| triton | 3.6.0 |
| CUDA | 12.8 |

Reproduced on H100 80GB SXM, H100 NVL, H200 SXM, and A100 SXM. Same crash signature across all four — not hardware-specific.

## Reproducer

```python
import torch

    model_name="unsloth/gpt-oss-20b-BF16",
    max_seq_length=2048,
    dtype=torch.bfloat16,
    load_in_4bit=False,
    attn_implementation="eager",
)

model = FastLanguageModel.get_peft_model(
    model,
    r=64,
    target_modules=["q_proj", "k_proj", "v_proj", "o_proj",
                    "gate_proj", "up_proj", "down_proj"],
    lora_alpha=8,
    lora_dropout=0,
    bias="none",
    use_gradient_checkpointing="unsloth",
    random_state=42,
    use_rslora=True,
)

# Any forward pass triggers the crash; trainer.train() reproduces it directly.
```

## Observed behavior

On the first forward pass:

```
RuntimeError: torch._grouped_mm: contraction dim mismatch:
  first_weight.shape = (32, 5760, 64), expected (32, 2880, 64)
```

`5760 = 2 × 2880`. The doubled inner dimension is the signature of the wrong-layout branch firing in `_get_moe_lora_io_dims`.

## Root cause

`_get_moe_lora_io_dims` reads a per-instance flag, `_unsloth_grouped_mm_format`, to interpret the 3D expert tensor shape:

- `True` → `(num_experts, in_dim, out_dim)`, the `gpt-oss` convention
- absent / `False` → `(num_experts, out_dim, in_dim)`, the Qwen3-MoE convention

Default is "absent." Unsloth's `GptOssExperts` replacement class sets the flag to `True` in `__init__` — but under transformers 5.5.0, the runtime `GptOssExperts` instances are not constructed by Unsloth's class. They are instances of `transformers.models.gpt_oss.modeling_gpt_oss.GptOssExperts`. Unsloth's function-level monkey-patches still apply, but the class itself is never reconstructed, the flag is never set, and the wrong layout branch is taken.

## Fix (currently as user-side runtime patch)

```python
patched = 0
for module in model.modules():
    if type(module).__name__ == 'GptOssExperts':
        module._unsloth_grouped_mm_format = True
        patched += 1
assert patched > 0, "No GptOssExperts modules patched"
```

Two notes on the shape of the fix:

- **Type-name comparison rather than `isinstance`.** `isinstance(module, GptOssExperts)` against the import path most users would write would match whichever `GptOssExperts` happened to be imported into user code. The runtime instances are from `transformers`, not from `unsloth_zoo`, so depending on which import "wins," `isinstance` checks become brittle. String-name matching sidesteps this entirely.
- **The `assert` is load-bearing.** Without it, a future regression in transformers' internal class naming or in Unsloth's patching order would produce zero matches silently, and training would crash on the same dim mismatch error this fix is meant to prevent — except attributed to the wrong cause.

**Suggested upstream placement:** inside `unsloth_zoo`'s existing MoE patching path, run an equivalent loop on `model.modules()` immediately after the current MoE patches are applied. The conditions are stable across versions and not specific to any user stack.

## Verification

With the fix applied, the trainable-parameter signature confirms full MoE coverage across all 24 layers:

| Metric | Value |
|---|---|
| Trainable parameters | 739,639,296 |
| Total parameters | 21,654,396,480 |
| Trainable ratio | 3.4157% |
| `GptOssExperts` modules patched | 24 (one per transformer layer) |

For reference: attention-only LoRA on this model lands near 0.4% trainable. Partial MoE coverage (some layers caught, others missed) lands in the 1–2% range. The 3.4% figure is the empirical fingerprint of full coverage. Training proceeds normally — smoke step times 1.6–2.0s on H100 80GB SXM.

## Related

TRL issue #5222 (autograd deadlock when using stock `transformers + PEFT` with `target_parameters` on the same model) is a separate bug in the vanilla path, not the subject of this report. This issue is specific to Unsloth's MoE LoRA path, which is meant to provide the working alternative.

mvanhorn:
## Summary

Sets `_unsloth_grouped_mm_format = True` on every gpt-oss expert module during MoE patching so `_get_moe_lora_io_dims` in `unsloth_zoo/temporary_patches/moe_utils.py` picks the gpt-oss `(num_experts, in_dim, out_dim)` weight layout. This fixes the `torch._grouped_mm` contraction-dimension crash on the first forward of a gpt-oss-20b MoE LoRA run under transformers 5.x.

## Why this matters

Issue #683 reports that gpt-oss-20b with `transformers==5.5.0` crashes on the first forward with a grouped_mm shape mismatch, reproduced on H100/H200/A100. The per-instance `_unsloth_grouped_mm_format` flag read at `moe_utils.py:395` was only set in Unsloth's own `GptOssExperts.__init__`, but under transformers 5.x the live modules are the stock `transformers.models.gpt_oss.modeling_gpt_oss.GptOssExperts`, so the flag stays unset, the Qwen3-MoE branch fires, and the reshape doubles the inner dim. The patch walks `model.modules()`, matches gpt-oss expert modules by type name regardless of whether they are the stock or Unsloth class, and sets the flag, leaving the existing shape heuristic as a fallback and gating the loop so Qwen3 modules are untouched.

## Testing

`tests/test_moe_grouped_mm_format.py` covers the happy path (stock `GptOssExperts` gets the flag and resolves to gpt-oss ordering), layout selection (Qwen3-MoE modules do not get the flag and stay on the Qwen3 branch), the already-subclassed regression case, and a model with no expert modules not raising. Full suite runs in CI.

Fixes #683

### DIFF
--- a/tests/test_forward_native_moe_loop_lora.py
+++ b/tests/test_forward_native_moe_loop_lora.py
@@ -272,3 +272,46 @@ def test_forward_native_moe_loop_lora_dtype_precast_no_loop_alloc(dtype):
     rtol = 1e-4 if dtype == torch.float32 else 1e-1
     torch.testing.assert_close(out, ref, atol=atol, rtol=rtol)
     assert out.dtype == dtype
+
+
+class GptOssExperts(nn.Module):
+    """Stock gpt-oss layout: (E, in, out) weights, biases, clamped swiglu."""
+
+    def __init__(self, num_experts=4, hidden=16, intermediate=12):
+        super().__init__()
+        self.num_experts = num_experts
+        self.alpha = 1.702
+        self.limit = 7.0
+        self.gate_up_proj = nn.Parameter(
+            torch.randn(num_experts, hidden, 2 * intermediate) * 0.1)
+        self.gate_up_proj_bias = nn.Parameter(
+            torch.randn(num_experts, 2 * intermediate) * 0.1)
+        self.down_proj = nn.Parameter(
+            torch.randn(num_experts, intermediate, hidden) * 0.1)
+        self.down_proj_bias = nn.Parameter(torch.randn(num_experts, hidden) * 0.1)
+        self._unsloth_grouped_mm_format = True
+
+
+def test_forward_native_moe_loop_gpt_oss_matches_naive():
+    torch.manual_seed(0)
+    num_experts, hidden, num_tokens, top_k = 4, 16, 8, 2
+    experts = GptOssExperts(num_experts, hidden)
+    hidden_states = torch.randn(num_tokens, hidden)
+    top_k_index = torch.stack(
+        [torch.randperm(num_experts)[:top_k] for _ in range(num_tokens)])
+    top_k_weights = torch.softmax(torch.randn(num_tokens, top_k), dim=-1)
+
+    out = forward_native_moe_loop(experts, hidden_states, top_k_index, top_k_weights)
+
+    ref = torch.zeros_like(hidden_states)
+    for t in range(num_tokens):
+        for k in range(top_k):
+            e = top_k_index[t, k].item()
+            h = hidden_states[t] @ experts.gate_up_proj[e] + experts.gate_up_proj_bias[e]
+            gate, up = h[::2], h[1::2]
+            gate = gate.clamp(max=experts.limit)
+            up = up.clamp(min=-experts.limit, max=experts.limit)
+            inter = (up + 1.0) * (gate * torch.sigmoid(gate * experts.alpha))
+            ref[t] += top_k_weights[t, k] * (
+                inter @ experts.down_proj[e] + experts.down_proj_bias[e])
+    torch.testing.assert_close(out, ref, atol=1e-5, rtol=1e-4)
--- a/tests/test_moe_grouped_mm_format.py
+++ b/tests/test_moe_grouped_mm_format.py
@@ -199,3 +199,17 @@ def test_model_without_experts_does_not_raise_or_set_flags():
     assert patch_gpt_oss_grouped_mm_format(model) == 0
     assert not hasattr(model, "_unsloth_grouped_mm_format")
     assert not hasattr(model.dense, "_unsloth_grouped_mm_format")
+
+
+def test_lazy_path_does_not_flag_qwen_layout_named_gpt_oss():
+    class GptOssExperts(nn.Module):
+        def __init__(self):
+            super().__init__()
+            self.num_experts = 4
+            self.gate_up_proj = nn.Parameter(torch.empty(4, 24, 16))
+            self.down_proj = nn.Parameter(torch.empty(4, 16, 12))
+
+    experts = GptOssExperts()
+    assert _get_moe_lora_io_dims(_Wrapper("gate_up_proj", experts)) == (16, 24)
+    assert _get_moe_lora_io_dims(_Wrapper("down_proj", experts)) == (12, 16)
+    assert not hasattr(experts, "_unsloth_grouped_mm_format")
--- a/tests/test_moe_grouped_mm_format.py
+++ b/tests/test_moe_grouped_mm_format.py
@@ -69,6 +69,14 @@ class _Wrapper:
         return self.base_layer
 
 
+class _PeftLikeWrapper(nn.Module):
+    def __init__(self, model):
+        super().__init__()
+        self.base_model = nn.Module()
+        self.base_model.model = model
+        self.config = _Config("lora")
+
+
 def _unwrap_base_layer(module):
     while hasattr(module, "base_layer"):
         module = module.base_layer
@@ -163,6 +171,15 @@ def test_gpt_oss_experts_name_is_gated_by_model_type():
 
     assert patch_gpt_oss_grouped_mm_format(model) == 0
     assert not hasattr(experts, "_unsloth_grouped_mm_format")
+
+
+def test_gpt_oss_detection_unwraps_peft_like_model():
+    experts = GptOssExperts()
+    model = _Model("gpt_oss", experts)
+    wrapped_model = _PeftLikeWrapper(model)
+
+    assert patch_gpt_oss_grouped_mm_format(wrapped_model) == 1
+    assert experts._unsloth_grouped_mm_format is True
 
 
 def test_existing_gpt_oss_grouped_mm_flag_is_preserved():
--- a/tests/test_moe_grouped_mm_format.py
+++ b/tests/test_moe_grouped_mm_format.py
@@ -1,8 +1,10 @@
 import torch
 import torch.nn as nn
+import pytest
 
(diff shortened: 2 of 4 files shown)


Datta0:
Review/comment follow-up:

- Addressed the Gemini review items in `52ad048ee`: `_is_gpt_oss_model` now unwraps PEFT-style `base_model` / `model` containers when checking configs, and `importlib` is imported explicitly instead of relying on `import importlib.util` side effects.
- Added a regression test for direct PEFT-like wrapper detection, plus the earlier integration-style PEFT `target_parameters` test and guard tests for Qwen3-VL grouped layout and MXFP4 GPT-OSS experts.
- Focused suite passes locally: `python -m pytest -q tests/test_moe_grouped_mm_format.py tests/test_forward_native_moe_loop_lora.py tests/test_peft_paramwrapper_layout_drift.py` -> 19 passed.

GPT-OSS Colab 4-bit validation with the notebook settings:

- Notebook source checked: `~/unsloth_nbs/python_scripts/gpt-oss-(20B)-Fine-tuning.py` uses `transformers==4.56.2`, `max_seq_length=1024`, `load_in_4bit=True`, LoRA `r=8`, batch size 1.
- Reproduced that path in a pinned env: `transformers==4.56.2`, `trl==0.22.2`, `torch==2.10.0+cu130`, `bitsandbytes==0.49.2`, one visible GPU.
- `unsloth/gpt-oss-20b` resolves to `unsloth/gpt-oss-20b-unsloth-bnb-4bit`; all 24 `GptOssExperts` had `_unsloth_grouped_mm_format=True` after LoRA attach.
- End-to-end 1024-token train step passed: forward/backward/AdamW8bit peak reserved memory was 12.617 GiB, under the 15 GiB Colab target.
- Important caveat: first load still peaked at 19.295 GiB reserved before settling to 11.715 GiB. Using the direct BNB model name did not reduce it, and `max_memory={0: "13GiB", "cpu": "48GiB"}` did not cap it.
- Native MXFP4 with `UNSLOTH_MXFP4_NO_DEQUANTIZE=1` and BNB patch disabled was not a practical Colab alternative in this local test: it spent several minutes in CPU-side load/conversion and reached about 16 GB process RSS before `after_load`, so I stopped it.

Conclusion: this PR fixes the grouped-mm layout/LoRA issue and the actual training step is under 15 GiB, but the current GPT-OSS 20B BNB loader path still has a >15 GiB first-load spike. If strict free-Colab fit includes initial model load, that needs a separate lower-peak loader change; simply simplifying the notebook model name is not enough.

danielhanchen:
Thanks for the PR and the detailed writeup, this is a great find. I validated the fix end to end and pushed b86e4367 to this branch with two follow ups.

**Validation of the original fix**

- Latest main still reproduces #683 on transformers 5.5.0 with `unsloth/gpt-oss-20b-BF16`: first forward crashes in `torch._grouped_mm` with the contraction dim mismatch, 24 expert modules present and 0 flagged.
- This branch merged with latest main trains cleanly on the same config: 61 steps, all 24 `GptOssExperts` flagged, trainable params 739,639,296 (3.4157 percent, matching the issue fingerprint), loss 4.59 to about 1.0.
- The grouped_mm split LoRA forward matches an fp32 naive per-expert reference at cosine 0.99999 with rsLoRA enabled. The explicit flag is genuinely required since gpt-oss-20b has hidden == intermediate == 2880, so shape heuristics are ambiguous.
- A Qwen3-30B-A3B regression run shows identical losses with and without this change, and 0 experts flagged there as expected.

**What I pushed**

1. Gated the lazy flag set inside `_get_moe_lora_io_dims` on the gpt-oss weight signature (gate_up out dim equals twice the down in dim, matching expert counts and inner dims). The class name check alone could mark a same-named module storing the F.linear layout. The signature also disambiguates the square hidden dim case, so real gpt-oss still resolves. This addresses the model-gating review comment without giving up the fail-safe.
2. Fixed the non grouped_mm fallback this PR exposes. With `UNSLOTH_MOE_BACKEND=native_torch`, or on any GPU without `torch._grouped_mm` where the backend silently falls back, `patch_gpt_oss_moe_for_lora` installed `torch_native_forward`, which expects the bnb4bit ModuleList layout (`gate_up_projs`) and a `(T, num_experts)` routing convention, so training crashed with `AttributeError: 'GptOssExperts' object has no attribute 'gate_up_projs'` even with the flag fix. The fallback now routes through `forward_native_moe_loop`, which gained the gpt-oss interleaved gate/up split, clamped swiglu and per-expert biases, keyed the same way `forward_native_grouped_mm` does. bnb4bit, Qwen3 and Qwen3-VL paths are unchanged.

**Verification of the follow ups**

- 60 MoE tests pass on this branch including two new regression tests (lazy path leaves a Qwen-layout same-named class unflagged; the gpt-oss native loop matches a naive reference with biases and swiglu).
- On a B200 with transformers 5.5.0, the native_torch run now completes 61 steps with per-step losses matching the grouped_mm run (max diff 0.05, mean 0.013), and a grouped_mm rerun after the change is step for step identical to before it.
- transformers 4.57.6 is unaffected since the whole path sits behind the v5 gate.

PR #718, opened 2026-06-04
fix: avoid NameError when load_harmony_encoding fails in encode_conversations_with_harmony

mvanhorn:
## Summary

Separates the `openai_harmony` import from the `load_harmony_encoding` call in `encode_conversations_with_harmony` (`unsloth_zoo/temporary_patches/gpt_oss.py`) and extends the in-function guard so that when the encoding cannot be loaded the user sees a named, actionable error instead of `NameError: name 'encoding' is not defined`.

## Why this matters

Issue #522 reports a `NameError` at the `encoding.render_conversation_for_completion(...)` call. The module-level block did the `from openai_harmony import (...)` and the `encoding = load_harmony_encoding(HarmonyEncodingName.HARMONY_GPT_OSS)` inside one `try/except: pass`, while the function guard only checked that `SystemContent` was importable. The reporter confirmed `openai_harmony==0.0.8` was installed, so the import succeeded but `load_harmony_encoding` failed, the bare except swallowed it, the guard passed, and `encoding` was never bound. The fix constructs `encoding` inside the function and raises "Unsloth: failed to load the gpt-oss harmony encoding: <orig error>" when it is unavailable, while keeping the import guard so a missing package still raises the existing ImportError with the install hint.

## Testing

`tests/test_encode_conversations_with_harmony.py` covers the happy path (encoding loadable returns the decoded text and input ids), the missing-import path (ImportError with install hint), the monkeypatched `load_harmony_encoding` failure producing the named error rather than a NameError, and the `add_generation_prompt` True/False regression. Full suite runs in CI.

Fixes #522

### DIFF
--- a/tests/test_encode_conversations_with_harmony.py
+++ b/tests/test_encode_conversations_with_harmony.py
@@ -1,0 +1,243 @@
+from __future__ import annotations
+
+import importlib
+import importlib.machinery
+import importlib.util
+import pathlib
+import sys
+import types
+
+import pytest
+
+
+_MODULE_NAME = "unsloth_zoo.temporary_patches.gpt_oss"
+_PACKAGE_ROOT = pathlib.Path(__file__).resolve().parents[1] / "unsloth_zoo"
+
+
+def _package_shell(name, path):
+    module = types.ModuleType(name)
+    module.__path__ = [str(path)]
+    module.__package__ = name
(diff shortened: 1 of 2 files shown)


danielhanchen:
Thanks for the PR

mvanhorn:
Thanks for this one as well, @danielhanchen. No more NameError masking the real harmony encoding failure.

PR #721, opened 2026-06-04
Fix Qwen3.5 attention wrapper to accept new mlx-vlm kwargs

BardiaKoopah:
mlx-vlm 0.6.1 added position_embeddings and target_verify kwargs to Qwen3_5Attention.__call__. The patched wrapper rejected them with a TypeError on training start.

Add **kwargs to the signature and forward to the original call. Using **kwargs instead of named params for durability as mlx-vlm frequently adds new kwargs to this interface.

Fixes #6002 (first crash)

### DIFF
--- a/unsloth_zoo/mlx/loader.py
+++ b/unsloth_zoo/mlx/loader.py
@@ -511,7 +511,7 @@ def _fix_qwen35_attention_cache(model):
 
     original_attn_call = attn_cls.__call__
 
-    def patched_attn_call(self, x, mask=None, cache=None, position_ids=None):
+    def patched_attn_call(self, x, mask=None, cache=None, position_ids=None, **kwargs):
         # When training (cache=None) and position_ids=None, compute them
         if cache is None and position_ids is None:
             import mlx.core as mx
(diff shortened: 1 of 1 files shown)


danielhanchen:
Validated this on mlx 0.31.2 + mlx-vlm 0.6.2 (CPU backend) with a tiny random-weight `qwen3_5` model:

- Reproduced the exact `TypeError` from unslothai/unsloth#6002 with the current wrapper, and confirmed the unpatched attention dereferences `cache.offset` when `cache=None` (the original bug this patch function exists for).
- With this patch, both the training forward and `value_and_grad` run, and the synthesized `position_ids` path is bit-identical to passing explicit `position_ids` through the unpatched attention (max abs diff 0).
- `Qwen3_5Attention.__call__` has the identical signature across mlx-vlm 0.6.1, 0.6.2 and current main, and the decoder layer always passes `position_embeddings` and `target_verify`, so `**kwargs` forwarding is the right durable fix here.
- One subtlety checked: mlx-vlm only supplies `position_embeddings` together with non-None `position_ids` (both are computed and passed down together at the model level), so the `cache=None` synthesis never fires in that case, and the attention body ignores `position_ids` for rotary when `position_embeddings` is present. No extra guard is needed.

Two notes:

1. Scope: this fixes only the first crash in unslothai/unsloth#6002. The remaining VJP crash is covered by #738, which contains the same signature change and can rebase once this lands.
2. The branch needs a sync with main: `ff12f344` reworded the comment line inside this function, so there is a trivial comment-only conflict to resolve.

PR #722, opened 2026-06-05
fix(mlx): gate gemma4 kv sharing shim

Lyxot:
This PR fixes a stale MLX Gemma4 training workaround around KV sharing.

Bug:
- `mlx-vlm < 0.5.0` needed Unsloth’s `_TrainingKVStore` shim because Gemma4 shared-KV layers still routed through cache objects during training.
- `mlx-vlm >= 0.5.0` added native Gemma4 `shared_kv` threading, so applying the old shim can interfere with the upstream path.

Root cause:
- Unsloth always applied `_fix_gemma4_kv_sharing()` for Gemma4 VLM loads and LoRA setup.
- The workaround was correct for older `mlx-vlm`, but stale once upstream introduced native shared-KV coordination in `mlx-vlm` https://github.com/Blaizzy/mlx-vlm/pull/1030 / `v0.5.0`.

Fix:
- Detect native Gemma4 shared-KV support from the loaded backbone.
- Skip the legacy `_TrainingKVStore` shim when native `shared_kv` support is present.
- Keep the shim active for older `mlx-vlm` versions such as `0.4.4`.

### DIFF
--- a/unsloth_zoo/mlx/loader.py
+++ b/unsloth_zoo/mlx/loader.py
@@ -437,17 +437,35 @@ class _TrainingKVStore:
         return keys, values
 
 
+def _gemma4_has_native_shared_kv(backbone):
+    """Return whether mlx-vlm already threads Gemma4 shared K/V for training."""
+    layers = getattr(backbone, "layers", None) or []
+    previous_kvs = getattr(backbone, "previous_kvs", None)
+    if (
+        isinstance(previous_kvs, (list, tuple))
+        and layers
+        and len(previous_kvs) == len(layers)
+    ):
+        return True
+
+    try:
+        call_params = inspect.signature(backbone.__class__.__call__).parameters
+    except (TypeError, ValueError):
+        return False
+
+    return "shared_kv_sink" in call_params
+
+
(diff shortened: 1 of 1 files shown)


danielhanchen:
Validated this before it merged; posting the record. The gate is a capability check rather than a version heuristic, which is the right design: `_gemma4_has_native_shared_kv` keys off structural signals (`backbone.previous_kvs` with length matching `layers`, plus `shared_kv_sink` in the backbone call signature), and tracing mlx-vlm 0.6.2's gemma4 confirms the native path threads shared K/V independently of `cache`, so the legacy `_TrainingKVStore` shim is redundant exactly when the gate says so. Neither failure direction reproduces: it does not skip the shim where still needed, and does not apply it where native handling exists. Absent-attribute tolerance is clean (older mlx-vlm, missing attrs, and signature introspection failures all early-return), and `_kv_sharing_patched` keeps it idempotent. 16/16 local gate tests pass against the real mlx-vlm 0.6.2 `Gemma4TextModel` plus stubs for the legacy branches, and the Apple Silicon install plus MLX import smoke passed on this branch.

PR #723, opened 2026-06-05
fix: add audio support to UnslothVisionDataCollator

LeoBorcherding:
## Summary

- `UnslothVisionDataCollator` silently dropped `{"type": "audio"}` content parts — `extract_vision_info` only matched `image`/`image_url`/`video`, so audio-only batches produced no `input_features` and the audio encoder never ran
- Add `AUDIO_TOKENS` list (`<|audio|>` for Gemma 4) and include them in `get_padding_tokens_ids` so audio placeholder tokens are correctly masked in labels
- Add `extract_audio_info` to pull audio arrays out of message content parts
- Add `_extract_audio_for_example` method on the collator (checks top-level `example["audio"]` column first, falls back to inline extraction)
- Wire audio collection into `__call__` and pass `audio=` to the processor so the audio encoder actually runs

Fixes: https://github.com/unslothai/unsloth/issues/6004

## Test plan

- [x] Run the standalone repro from issue #6004 — `input_features` should now be present in the batch
- [x] Verify `<|audio|>` tokens are masked (`-100`) in labels, not trained as text
- [ ] Smoke test Gemma 4 E2B/E4B SFT with an audio dataset under `SFTTrainer`

### DIFF
--- a/unsloth_zoo/vision_utils.py
+++ b/unsloth_zoo/vision_utils.py
@@ -586,7 +586,12 @@ def extract_audio_info(conversations: Union[List[Dict], List[List[Dict]]]) -> Li
             if isinstance(content, list):
                 for ele in content:
                     if isinstance(ele, dict) and ele.get("type") == "audio" and "audio" in ele:
-                        audio_inputs.append(ele["audio"])
+                        audio = ele["audio"]
+                        # HuggingFace Audio feature dict -> raw waveform
+                        if isinstance(audio, dict):
+                            audio = audio.get("array")
+                        if audio is not None:
+                            audio_inputs.append(audio)
     return audio_inputs
 
 
@@ -890,9 +895,7 @@ class UnslothVisionDataCollator:
         if audios and self.truncation and self.max_seq_length:
             seq_len = batch["input_ids"].shape[1]
             if seq_len > self.max_seq_length:
-                for key in list(batch.keys()):
-                    if isinstance(batch[key], torch.Tensor) and batch[key].shape[-1] == seq_len:
-                        batch[key] = batch[key][..., :self.max_seq_length]
+                batch = self._truncate_sequence_tensors(batch, seq_len)
 
         # Cannot remove due to bidirectional attention from Gemma 3!
         # batch.pop("token_type_ids", None)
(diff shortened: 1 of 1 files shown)


LeoBorcherding:
Tested locally on AMD RX 9060 XT (gfx1200, ROCm 7.13), Gemma 4 E2B 4bit, transformers 5.5.0, unsloth-zoo main vs fix branch.

**Before (stock unsloth-zoo):**
```
keys: ['input_ids', 'attention_mask', 'mm_token_type_ids', 'labels']
has input_features: False
<bos><|turn>system
You transcribe audio. <turn|>
<|turn>user
<|audio|>Please transcribe this audio.<turn|>
<|turn>model
hello<turn|>
```

**After (this PR):**
```
keys: ['input_ids', 'attention_mask', 'mm_token_type_ids', 'input_features', 'input_features_mask', 'labels']
has input_features: True
<bos><|turn>system
You transcribe audio. <turn|>
<|turn>user
<|audio><|audio|><|audio|><|audio|><|audio|><|audio|><|audio|><|audio|><|audio|><|audio|><|audio|><|audio|><|audio|><|audio|><|audio|><|audio|><|audio|><|audio|><|audio|><|audio|><|audio|><|audio|><|audio|><|audio|><|audio|><|audio|><audio|>Please transcribe this audio.<turn|>
<|turn>model
hello<turn|>
```

`input_features` and `input_features_mask` are now in the batch. The single `<|audio|>` placeholder expands to N tokens as expected and the audio encoder actually runs.

Also verified that `<|audio|>` tokens are masked to -100 in labels and not trained as text:
```
audio token id: 258881
audio token positions in input_ids: [15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27, 28, 29, 30, 31, 32, 33, 34, 35, 36, 37, 38, 39]
all <|audio|> tokens masked to -100 in labels: True
```

SFT forward pass confirmed: audio encoder ran during model forward (audio_seen_in_forward: True). Full backward pass requires a bitsandbytes ROCm build that isn't available on this machine but that is unrelated to this fix.

All bot review comments have been addressed across the 4 commits on this PR.

LeoBorcherding:
Completed the SFTTrainer smoke test on AMD RX 9060 XT (gfx1200, ROCm 7.13), Gemma 4 E2B 4bit, unsloth 2026.6.1.

Required setting `BNB_ROCM_VERSION=72` to point bitsandbytes at the correct DLL (`rocm72.dll` vs the default `rocm713.dll` lookup). This is a separate Windows ROCm setup issue filed at unslothai/unsloth#6032 and not related to this fix.

3 forward passes with loss computed successfully:

```
Running 3 forward passes...
  step 1: loss=9.8145, input_features in batch: True
  step 2: loss=9.6671, input_features in batch: True
  step 3: loss=9.8428, input_features in batch: True

=== Smoke test results ===
input_features seen in forward: True
losses: [9.8145, 9.6671, 9.8428]
PASS  audio encoder ran in all 3 forward passes, loss computed successfully
```

All 3 items from the test plan are now checked off.

LeoBorcherding:
Is ready for final review and then merge

danielhanchen:
Thanks for the PR @LeoBorcherding! The core fix is correct and I verified it end to end, so I pushed two commits on top: one hardening the new audio path and one merging latest main into the branch.

**What I verified**

- The issue #6004 repro now produces `input_features` and `input_features_mask`, the placeholder expands correctly, and all audio tokens are masked to -100 in labels.
- Text only, image only and prompt-completion batches are bit identical to main, so no behavior change for existing vision and text users.
- A 61 step Gemma 4 E2B LoRA ASR run: losses and grad norms match a hand-rolled reference collator (final losses 0.156/0.199 vs 0.161/0.211), and the audio tower ran in 176/176 batches. On current main the audio tower runs in 0/176 batches and the loss plateaus about 2.4x higher, so this PR also fixes our audio notebooks which rely on this collator.

**Fixes pushed**

1. The manual truncation block sliced any tensor whose last dim equals the sequence length. That clipped `input_features` whenever seq_len happened to equal the mel bin count (128) and `input_features_mask` whenever it equaled the frame count. It also ignored left padding (the Gemma 4 default), so short rows in a truncated batch kept only their `<pad>` tokens and lost all content. Truncation now slices only `input_ids`, `attention_mask`, `token_type_ids` and `mm_token_type_ids`, respects the padding side, and raises a clear error if it would cut into the expanded audio tokens, which would otherwise crash at model forward with "Audio features and audio tokens do not match".
2. A top-level `audio` column holding a flat `list[float]` waveform was iterated as one clip per sample, which crashes with `StopIteration` on batches of 2 or more rows. A flat list is now treated as a single clip, and lists of HF Audio dicts are unwrapped to their arrays.
3. Inline `{"type": "audio", "audio": {"array": ..., "sampling_rate": ...}}` content (a decoded HF Audio feature) crashed the feature extractor because the dict was passed through unchanged. It is now unwrapped to the array, matching the top-level path.
4. `example["audio"] = None` returned early and skipped inline audio in the messages, silently reproducing the original text-only training bug for nullable audio columns. It now falls back to inline extraction.
5. Audio inside `completion` for prompt-completion data was silently dropped while its placeholder stayed in the rendered text. It now raises a clear error asking the user to move audio parts into `prompt`.

**Note for Gemma 4 audio users**: transformers 5.5.0 has an upstream processor bug where the audio placeholder count is recomputed from the raw waveform and disagrees with the feature extractor by one token for many clip lengths, for example a 3.97s clip yields 99 placeholders vs 100 encoder outputs, which crashes training on real datasets regardless of this PR. This was fixed in transformers 5.10.0 by deriving the count from `input_features_mask`, so Gemma 4 audio finetuning should use transformers >= 5.10.0. I verified this PR against both 5.5.0 and 5.10.2.

PR #728, opened 2026-06-07
fix: skip vllm tests when vllm is not installed

marksverdhei:
Fixes test collection errors in CI when vllm is not installed by using pytest.importorskip.

### DIFF
--- a/tests/test_vllm_to_hf_conversion.py
+++ b/tests/test_vllm_to_hf_conversion.py
@@ -3,6 +3,7 @@ sys.path.insert(0, os.path.join(os.path.dirname(__file__), ".."))
 
 import types
 import pytest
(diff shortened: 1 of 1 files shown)


danielhanchen:
Thanks @marksverdhei for digging into this CI failure! The root cause turned out to be in patch_gemma4_vllm_lora_support itself: it imported vllm_lora_worker_manager (which hard-imports vllm.config) before its early return. That is fixed on main by #726 along with a hardened regression test, so the conversion tests now pass without vllm installed. A module-level importorskip would have skipped the entire conversion test file in environments without vllm, which is exactly where Core CI runs, so I reverted the line in-branch and am merging with a net-zero diff to keep the contribution in the history.

PR #731, opened 2026-06-08
fix: add Qwen3.5 SSM for GGUF

Datta0:


danielhanchen:
Thanks @Datta0! Good catch on the stale checkout problem. The converter script is always fetched from llama.cpp master but gguf-py comes from the local checkout, so a checkout predating Qwen3.5 cannot map the new tensors. I pushed d253cf8 which reworks the patch since the previous insertion did not actually add the entries that matter:

1. The old logic anchored on `# qwen3next` comment lines, so it only duplicated names those entries already cover (`linear_attn.conv1d`, `dt_proj`, `A_log`, `norm`, `out_proj` are byte identical strings, so re-adding them changes nothing). The Qwen3.5 split projections never got written because `SSM_BETA` and `SSM_ALPHA` have no qwen3next lines to anchor on.
2. The split GDN projections are now inserted in the blocks llama.cpp master actually uses: `in_proj_qkv` under `ATTN_QKV`, `in_proj_z` under `ATTN_GATE`, `in_proj_b` under `SSM_BETA`, `in_proj_a` under `SSM_ALPHA`.
3. Insertion now anchors on the block headers and is idempotent per entry, so partially patched files get completed and blocks missing from very old checkouts are skipped instead of guessed at.
4. `dt_bias` needs no mapping entry. The master converter renames `dt_bias` to `dt_proj.bias` before mapping, so the existing `dt_proj` alias is the correct target.
5. The patched file is validated with `ast.parse` before writing, and the patch now runs in `_download_convert_hf_to_gguf` outside the lru cache against the resolved llama.cpp dir, covering checkouts pulled after the first conversion and `UNSLOTH_LLAMA_CPP_SCRIPTS_DIR` overrides.
6. Moved the test under `tests/` since pytest only collects from there, and it now asserts each alias lands in its exact block. I also verified against the real master `tensor_mapping.py` with the qwen3.5 lines stripped: the patch restores all four aliases in the same blocks master uses, runs idempotently, and leaves a current master file byte identical.

PR #745, opened 2026-06-10
Resolve file:// URIs per RFC 8089 in media readers

danielhanchen:
## Problem

The media readers handle the file:// scheme by slicing off the first 7 characters: `_read_video_torchcodec` always, `_read_video_torchvision` only for torchvision below 0.19, and `fetch_image` for images. `_read_video_decord` does not handle the scheme at all. Naive slicing breaks two valid URI forms:

- `file://localhost/data/clip.mp4` becomes `localhost/data/clip.mp4`, which no decoder can open
- percent-encoded paths such as `file:///data/my%20clip.mp4` are never decoded, so the literal `%20` path does not exist

Both fail deep inside the decoder with unrelated errors, for example `ValueError: nframes should in interval [2, 0], but got 0`, instead of reading a file that exists.

unslothai/unsloth#5136 added pre-training validation that resolves these URI forms correctly per RFC 8089, so validation passes them while the decode side then fails. This PR makes the decode side agree with the validator.

## Fix

Adds `resolve_file_uri_to_path`: resolves `file://` URIs with an empty or `localhost` authority via `urlparse` + `url2pathname` (percent-decoding included), and returns everything else unchanged (plain paths, http/https, data:, non-local authorities, non-strings). Applied in `_read_video_torchvision`, `_read_video_decord`, `_read_video_torchcodec` and `fetch_image`.

| Input | Before | After |
|-------|--------|-------|
| `file:///data/clip.mp4` | works (torchcodec/old torchvision), fails (decord) | works |
| `file://localhost/data/clip.mp4` | cryptic decoder error | works |
| `file:///data/my%20clip.mp4` | cryptic decoder error | works |
| plain path, http(s), data: | unchanged | unchanged |

## Tests

`tests/test_file_uri_resolution.py`: 14 tests covering the resolver (empty and localhost authority, percent-encoding, non-local authority passthrough, degenerate URIs, non-strings) plus integration through `fetch_image` and `fetch_video` with a synthesized mp4 across all URI forms.

Verified locally: 14/14 new tests pass, the existing `tests/test_vision_collator_audio.py` suite still passes (27/27), and `file://localhost` plus percent-encoded URIs now decode end to end through `process_vision_info` where both previously failed.

### DIFF
--- a/tests/test_file_uri_resolution.py
+++ b/tests/test_file_uri_resolution.py
@@ -1,0 +1,99 @@
+"""Tests for resolve_file_uri_to_path and file:// handling in media readers.
+
+Naive `path[7:]` stripping broke file://localhost/... URIs (producing the
+unopenable "localhost/...") and never decoded percent-escapes, so valid URIs
+failed deep inside the video decoder with unrelated errors.
+"""
+import os
+
+import pytest
+
+from unsloth_zoo.vision_utils import resolve_file_uri_to_path
+
+
+def test_plain_path_unchanged():
+    assert resolve_file_uri_to_path("/tmp/clip.mp4") == "/tmp/clip.mp4"
+
+
+def test_http_url_unchanged():
+    url = "https://example.com/clip.mp4"
+    assert resolve_file_uri_to_path(url) == url
+
+
+def test_data_uri_unchanged():
+    uri = "data:video/mp4;base64,AAAA"
+    assert resolve_file_uri_to_path(uri) == uri
+
+
+def test_non_string_unchanged():
+    for value in (None, 42, ["frame.jpg"], {"url": "x"}):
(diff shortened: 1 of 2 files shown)


danielhanchen:
Addressed all review comments in ef5b3fdd:

- file://C:/x legacy Windows form: the drive letter parses as the URI authority, so the resolver now folds a two character drive netloc back into the path before the localhost check. The old slicing behavior accepted this form on Windows, so this also avoids a regression there. Added a test for it.
- Tests now build URIs with Path.as_uri instead of string concatenation and normalize separators in the resolver assertions, so they pass on Windows as well.
- Removed the unused os import.

42 tests pass locally (15 in the new file plus the existing vision collator suite).

PR #748, opened 2026-06-10
diffusion_studio: serve DiffusionGemma in Unsloth Studio with the optimized visual decoder

danielhanchen:
Run DiffusionGemma in Unsloth Studio over the GGUF path.

DiffusionGemma is a block-diffusion model, so llama.cpp serves it through a dedicated diffusion runner rather than the autoregressive server. `diffusion_studio` drives the optimized on-device visual decoder (`llama-diffusion-gemma-visual-server` from ggml-org/llama.cpp PR #24423) and wraps it in an OpenAI-compatible shim, so Studio can serve DiffusionGemma as an ordinary llama.cpp model. The shim streams the committed answer text and appends a self-contained ```html artifact that replays the per-step denoising canvas, which Studio auto-renders for DiffusionGemma so you watch the 256-token canvas resolve out of noise. Autoregressive flows are untouched.

## What's here
- `visual_engine.py` - keeps the visual server resident (model loaded once) and streams the per-step canvas (already decoded to text) plus the committed answer. Launches the server with PR_SET_PDEATHSIG so a parent exit never orphans a GPU process.
- `shim.py` - `python -m unsloth_zoo.diffusion_studio.shim --gguf ... --gpu ... --port ...`; OpenAI `/v1/chat/completions` (streaming + non-streaming) + `/health`; appends the `canvas_player.html` artifact.
- `canvas_player.html` - self-contained denoising-canvas player (animation, scrub, speed).

## Self-tokenizing
The visual server tokenizes, applies the GGUF chat template and detokenizes itself from the GGUF's own embedded tokenizer, so this package needs no tokenizer files. The shim sends chat messages as JSON; the server returns text frames + committed text.

## How Studio uses it
Studio detects a DiffusionGemma GGUF (`general.architecture = diffusion-gemma`) and launches this shim via `-m unsloth_zoo.diffusion_studio.shim` for any quant of `unsloth/diffusiongemma-26B-A4B-it-GGUF`, then auto-renders the canvas inline. No tokenizer directory and no manual connection are required.


ronnyli:
I'm new to Unsloth and llama.cpp but would this PR allow for using diffusion gemma within Claude Code? This is the first local model I tried that can actually respond to me in seconds instead of minutes on my 32GB M1 Mac.

PR #751, opened 2026-06-10
feat(mlx): implement resume_from_checkpoint for MLX training

BardiaKoopah:
## Summary

Studio's Resume button silently restarted MLX runs from step 0. The MLX worker called `trainer.train()` with no args, and the MLX trainer didn't accept `resume_from_checkpoint`. The frontend already exposes a Resume action (and `can_resume_run` doesn't gate by hardware), so this was a real user-facing bug.

This PR adds the trainer-side support. The Studio worker change (passing the kwarg through) is a [companion PR in `unslothai/unsloth`](https://github.com/unslothai/unsloth/pulls?q=is%3Apr+author%3ABardiaKoopah+mlx-studio-resume-wiring).

## What gets saved

The checkpoint loop already wrote `adapters.safetensors` + `adapter_config.json` via `save_trainable_adapters`. After a successful adapter save we now also write, best-effort:

- **`optimizer_state.safetensors`** — `mx.save_safetensors(tree_flatten(optimizer.state))`. For Adam/AdamW this is per-parameter `m,v` + scalar `step` and `learning_rate`. Round-trip is byte-equal in mlx 0.31.2.
- **`trainer_state.json`** — `{global_step, train_loss_history}`. JSON rather than safetensors because these are scalars/lists, not tensors.

If the new writes fail, we print a warning and keep going. The adapter save was already successful at that point; only resume is unavailable.

This matches the contract Studio's backend already expects: `studio/backend/core/training/resume.py:has_resume_state()` looks for `trainer_state.json` in either `output_dir/` or `output_dir/checkpoint-N/`.

## What gets restored

`train()` now accepts `resume_from_checkpoint`. `_train_inner` does three things, in order, after building the optimizer:

1. **`model.load_weights(adapters.safetensors, strict=False)`** — the model already has LoRA wrappers (Studio's pipeline does `get_peft_model` before training), so we only need to load the trained adapter tensors back into them.

2. **`load_optimizer_state`** — restores Adam moments and the optimizer step. Without this, Adam would silently restart its moment estimates and the first post-resume step would be effectively a learning-rate-scaled gradient with zero momentum.

3. **`load_trainer_state`** — restores `global_step` and `train_loss_history`. `global_step` is used to fast-forward the loop counter and `batch_idx`; `train_loss_history` is restored so the UI's loss curve stays continuous across the resume boundary.

If `trainer_state.json` or `optimizer_state.safetensors` is missing on resume, we raise rather than silently restart from step 0 with a fresh optimizer. Silent restart is the worse outcome — it gives the user false confidence.

## Loop fast-forward

On resume we change the loop start to begin at `_resume_step * grad_accum + 1`, set `batch_idx = _resume_step * grad_accum`, and (for streaming mode where there is no random-access batch list) consume `_resume_step * grad_accum` items from the generator before entering the loop. All four data paths (`create_batches` / `iterate_training_batches` / `create_vlm_batches` / `iterate_vlm_training_batches`) take `seed=args.seed` and produce deterministic ordering, so fast-forward is sufficient — we don't need to save iterator position to disk.

The LR scheduler is a pure function of step, so once it starts at the right value the schedule is automatically correct.

## Verification

Stop-resume harness on M2 16GB, Qwen3-0.6B + unsloth/LaTeX_OCR, max_steps=10, save_steps=5, grad_accum=4:

| Step | Fresh loss | Resumed loss |
|------|-----------|--------------|
| 6    | 2.1686279773712158 | 2.168627977371216 |
| 7    | 1.6476788520812988 | 1.6476788520812988 |
| 8    | 1.4659109115600586 | 1.4659109115600586 |
| 9    | 1.4719936847686768 | 1.4719936847686768 |
| 10   | 1.4772168397903442 | 1.4772168397903442 |

Every loss after the resume boundary matches the fresh run bit for bit. LR schedule matches (`2e-04, 1.6e-04, 1.2e-04, 8e-05, 4e-05`). Checkpoint dir contains the expected 4 files.

## Note on Studio integration

This PR alone isn't enough to unblock the Studio Resume button — Studio's worker must also pass `resume_from_checkpoint` through to `trainer.train()`. That's a 2-line change in a companion PR.

### DIFF
--- a/unsloth_zoo/mlx/trainer.py
+++ b/unsloth_zoo/mlx/trainer.py
@@ -546,10 +546,10 @@ class MLXTrainer:
         self._prior_metal_limits = {}
 
     def train(self, resume_from_checkpoint: str | None = None):
-        # Stash for _train_inner. None = fresh start, a path = resume.
-        self._resume_from_checkpoint = resume_from_checkpoint
         """Run MLX-native training loop following mlx-lm's compiled-step pattern
         with gradient accumulation. Returns a dict of training metrics."""
+        # Stash for _train_inner. None = fresh start, a path = resume.
+        self._resume_from_checkpoint = resume_from_checkpoint
         args = self.args
         model = self.model
         args.patch_mode = normalize_mlx_patch_mode(getattr(args, "patch_mode", "patched"))
@@ -1081,7 +1081,7 @@ class MLXTrainer:
                         f"Unsloth: streaming dataset exhausted while "
                         f"fast-forwarding to resume step {_resume_step}. "
                         f"Dataset may be shorter than the killed run consumed."
-                    )
+                    ) from None
 
         for it in range(_resume_step * grad_accum + 1, total_steps * grad_accum + 1):
             if self.stop_requested:
--- a/unsloth_zoo/mlx/trainer.py
+++ b/unsloth_zoo/mlx/trainer.py
@@ -66,6 +66,10 @@ from .utils import (
     collect_mlx_texts,
     save_lora_adapters,
     save_trainable_adapters,
+    save_optimizer_state,
+    load_optimizer_state,
+    save_trainer_state,
+    load_trainer_state,
     collect_mlx_lora_adapter_tensors,
     iter_mlx_lora_modules,
     apply_gradient_checkpointing,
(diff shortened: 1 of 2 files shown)


danielhanchen:
Reviewed and validated. The restore order (adapters, optimizer state, trainer state), the hard error instead of a silent restart when state files are missing, the LR being re-derived from the loop counter, and the fast-forward arithmetic across the four batch paths all check out. Verified the optimizer and trainer state round-trips and the missing-state errors on the test shim; the MLX test files pass.

Pushed one cleanup commit on top: moved the train() docstring back above the first statement (it had become a dead string literal after the resume stash) and chained the streaming fast-forward RuntimeError with from None so the bare StopIteration does not add noise. No behavior change.

PR #754, opened 2026-06-11
MLX: fix gated-delta (Qwen3.5) training and add fused backward kernels

Lyxot:
## Summary

Gated-delta (linear attention) models — qwen3_5, qwen3_5_moe, qwen3_next, kimi_linear — could not be trained properly on MLX: the memory-efficient custom VJP never reached any call site, and the training recurrence ran as an unrolled per-token Python loop that is both slow and hostile to `mx.compile`. This PR fixes the two bugs and replaces the recurrence backward with fused Metal kernels: Qwen3.5-0.8B LoRA training goes from ~140 tok/s with corrupt gradients (current main) to **~1,100 tok/s with correct gradients at 5x less memory**.

## What changed

- **Rebind stale `gated_delta_update` consumer imports** — `patch_gated_delta()` only rebound `mlx_lm.models.gated_delta.gated_delta_update`, but every consumer (mlx_lm qwen3_5 / qwen3_next / kimi_linear, mlx_vlm qwen3_5) binds the symbol via from-import at module load, so the custom VJP had been dead code since it was introduced. An identity-matched `sys.modules` sweep rebinds stale references; modules shipping their own implementation are never touched.

- **Install the VJP for qwen3_next and kimi_linear too** — the patch trigger was gated on `"qwen3_5" in model_type`; it now detects gated-delta layers structurally (class name + `A_log`/`dt_bias` pair, deliberately excluding Mamba/SSM mixers which share the parameters but not the recurrence).

- **Fused Metal backward kernels** — forward reuses mlx-lm's fused `gated_delta_kernel` per 64-step chunk (chunk-boundary states fall out as the chunk outputs); backward replays chunk states into scratch (K1), then reverse-scans with atomic gradient accumulation (K2). The per-layer graph drops from ~25k unrolled ops to ~100 kernel launches, which also makes whole-step `mx.compile` viable. Calls outside kernel support (mask present, head dim not a multiple of 32, no Metal) fall back to the ops VJP automatically. Kernel pointer arithmetic is bounds-clean (no out-of-range pointer formation at chunk edges), and the reverse scan pre-reduces threadgroup rows in shared memory before issuing global atomics for d_q/d_k, cutting atomic traffic 4x (dispatch falls back to the ops VJP for head dims that would leave a partial threadgroup).

- **Vectorized-gating support** — extends the kernels to kimi_linear's per-column gating (`g: [B, T, Hv, Dk]`) via the same template-flag pattern mlx-lm's forward kernel uses. Every gated-delta architecture in mlx-lm now trains on the kernel path.

- **Unify with #738's VLM patch** — `patch_gated_delta_vlm`'s mlx-vlm >= 0.6 branch now routes training through the fused-kernel dispatch (its previous ops-VJP routing under whole-step `mx.compile` would hit the `compile_fuse` blow-up this PR eliminates); its legacy-layout branch is dropped (the sweep already rebinds mlx-vlm 0.4-0.5 from-imports); the sweep recognizes the sibling patch instead of warning about a foreign implementation.

- **Tests** — consolidated into `tests/test_mlx_gated_delta_vjp.py`: binding-sweep and detection tests run on CI via the torch shim; gradient-parity tests (Metal-only) pin both the ops and kernel VJPs against plain `mx` autodiff at batch sizes 2-4, across GQA, multi-chunk, uneven tails, and both gating forms.

## Benchmark

Qwen3.5-0.8B, bf16 LoRA r=8, batch 2, seq <= 2048, gradient checkpointing on, 10 steps (~37.7k trained tokens) over a synthetic dataset of 16 long prompts (repeated reasoning filler in 5 length buckets up to the seq cap; identical rows for all four configurations), M-series 128 GB, mlx 0.31.2 / mlx-lm 0.31.3 / mlx-vlm 0.5.0:

| configuration | avg tok/s | wall time | peak GPU | 
|---|---|---|---|
| main | 141 | 272 s | 37.5 GB |
| main, `compile=False` | 136 | 277 s | 12.8 GB |
| upstream `mlx_vlm.lora` | 89 | ~425 s | 38.9 GB |
| **this PR** | **1,244** | **32 s** | **7.4 GB** |

(Upstream `mlx_vlm.lora` measured on 0.5.0; on 0.6.2 it crashes with `Primitive::vjp not implemented for CustomKernel` before the first step.)

## Validation

- **Gradient parity vs plain `mx` autodiff** (not vs the previous implementation — which is how the batch-stride bug evaded its original verification): <= 1e-7 fp32 / bf16-rounding bounded across GQA, multi-chunk, uneven tails, scalar + vectorized gating, batch 2-4, for both the ops and kernel VJPs.
- **Cross-framework parity vs CUDA**: the trainer-owned LoRA harness (fingerprint-verified identical token batches, trainer-owned AdamW + matched LR schedule, TF32 off) shows mean per-step loss difference < 0.005 against `UnslothTrainer` on an RTX 4070 SUPER over 30 steps of OpenMathReasoning-mini, for qwen3_5-0.8b through these kernels as well as qwen3-0.6b / gemma3-270m controls.
- **Real-model runs**: Qwen3.5-0.8B end-to-end (training, save, cold-process reload with loss parity); Kimi-Linear-48B-A3B 4-bit QLoRA (binding sweep connects, vectorized kernels under quantized projections, finite decreasing loss at 29.2 GB peak).
- Atomic accumulation makes low-order gradient bits nondeterministic across runs, matching Metal reduction-order behavior elsewhere in the backend.


Lyxot:
@codex review

danielhanchen:
@codex review

danielhanchen:
@codex review

PR #759, opened 2026-06-11
Install prebuilt llama.cpp release binaries before compiling from source

danielhanchen:
## What

`install_llama_cpp` currently clones llama.cpp and compiles it from source. This PR makes it install prebuilt binaries first, with the existing clone and compile flow kept unchanged as the fallback:

- CPU installs (the `save_pretrained_gguf` path, which forces `gpu_support=False` since `llama-quantize` is CPU-only) use the official CPU archives from ggml-org/llama.cpp GitHub releases
- GPU installs (`gpu_support=True`) use the same prebuilt CUDA/ROCm/Metal bundles Unsloth Studio publishes on unslothai/llama.cpp, selected via the release manifest and verified against its published sha256 list

## How

CPU path (upstream ggml-org releases):
- Resolve the latest release tag (or an `UNSLOTH_LLAMA_TAG` pin) through the GitHub API, using `GH_TOKEN`/`GITHUB_TOKEN` when set
- Select the CPU asset for the host: `ubuntu-x64`/`ubuntu-arm64` tar.gz on Linux, `macos-arm64`/`macos-x64` tar.gz on macOS, `win-cpu-x64`/`win-cpu-arm64` zip on Windows

GPU path (unslothai/llama.cpp bundles, same source as Studio):
- CUDA: read `llama-prebuilt-manifest.json` and pick the narrowest coverage bundle whose supported SMs include the host GPU (torch device capability), preferring the runtime line that matches torch's CUDA major; that line's portable build and the other line are queued as follow-up attempts
- ROCm: map `gcnArchName` to the per-gfx family bundles and copy the hipblaslt/rocblas Tensile kernel trees next to the libs
- macOS: the fork's Metal bundles, no torch detection needed
- Verify downloads against `llama-prebuilt-sha256.json` before extraction

Shared machinery:
- Download and extract into a staging dir (path-traversal guarded), place binaries and shared libs where `check_llama_cpp` and `quantize_gguf` already look (folder root on Linux/macOS for the `$ORIGIN` RPATH, `build/bin/Release` on Windows)
- Hydrate `convert_hf_to_gguf.py`, `conversion/` and `gguf-py/` from the same tag's source tarball so the converter machinery and tensor mappings stay consistent with the binaries
- Validate `llama-quantize --help` in staging before activating atomically, then write an `UNSLOTH_PREBUILT_INFO.json` marker recording repo, tag and asset
- Teach the folder integrity check to accept marker-bearing installs; prebuilt folders lack the `src`/`ggml`/`common` dirs of a checkout and were previously rmtree'd as corrupted on the next call
- Every failure (rate limit, missing asset, corrupt archive, checksum mismatch, bundle that cannot load on the host) logs the reason and advances to the next candidate and ultimately to the unchanged compile path; `UNSLOTH_LLAMA_FORCE_COMPILE=1` bypasses prebuilts entirely

## Verification

- 27 CPU tests in `tests/test_llama_cpp_prebuilt.py` (asset selection matrices for CPU and GPU, runtime-line preference, ROCm family mapping, macOS routing, tag pinning, gates, extraction safety, mocked full installs, sha256 mismatch and other fallback drills, reuse without deletion); 102 tests pass across the llama_cpp suites
- Live CPU on Linux x64: prebuilt install of `b9596` completes in about 4 seconds versus minutes of cloning plus compiling; second call reuses the install
- Live GPU on a B200 (SM 100, torch CUDA 12.8): selects `app-b9585-linux-x64-cuda12-newer.tar.gz`, installs in 12 seconds versus a much longer CUDA source build, `llama-quantize` validates, and `llama-server --list-devices` enumerates the GPU through the bundled CUDA backend
- End to end with the Llama 3.2 3B conversational notebook flow: trained 60 steps, exported `q4_k_m` twice, once with a compiled llama.cpp and once through the prebuilt path on a fresh dir. The prebuilt run shows zero clone or cmake/make activity and the two Q4_K_M GGUFs are byte-identical in size (2,019,378,176 bytes)
- Fallback drills: a bogus `UNSLOTH_LLAMA_TAG` falls back cleanly without creating the target folder, and `UNSLOTH_LLAMA_FORCE_COMPILE=1` skips the prebuilt path

### DIFF
--- a/tests/test_llama_cpp_prebuilt.py
+++ b/tests/test_llama_cpp_prebuilt.py
@@ -1,0 +1,273 @@
+# Tests for the prebuilt-first llama.cpp install path: release resolution,
+# asset selection, archive extraction/placement, fallback-to-compile contract,
+# and reuse of an existing prebuilt install without deletion.
+
+from __future__ import annotations
+
+import importlib.util
+import io
+import json
+import os
+import sys
+import tarfile
+from pathlib import Path
+
+import pytest
+
+
+def _load_llama_cpp_module():
+    repo_root = Path(__file__).resolve().parents[1]
+    module_path = repo_root / "unsloth_zoo" / "llama_cpp.py"
+    spec = importlib.util.spec_from_file_location("llama_cpp_under_test_prebuilt", module_path)
+    module = importlib.util.module_from_spec(spec)
+    sys.modules[spec.name] = module
+    spec.loader.exec_module(module)
+    return module
+
+
+llama_cpp = _load_llama_cpp_module()
+
+IS_POSIX = os.name == "posix"
+
+FAKE_QUANTIZE = "#!/bin/sh\necho 'usage: llama-quantize [--help] model-f32.gguf'\nexit 0\n"
+FAKE_QUANTIZE_BROKEN = "#!/bin/sh\necho 'segfault'\nexit 1\n"
+
+
+def _add_file(tar, name, data, mode = 0o644):
+    raw = data.encode() if isinstance(data, str) else data
+    info = tarfile.TarInfo(name)
+    info.size = len(raw)
+    info.mode = mode
+    tar.addfile(info, io.BytesIO(raw))
+
+
+def _make_binary_archive(path, tag, quantize_script = FAKE_QUANTIZE):
+    with tarfile.open(path, "w:gz") as tar:
+        _add_file(tar, f"llama-{tag}/llama-quantize", quantize_script, mode = 0o755)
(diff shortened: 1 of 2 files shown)


danielhanchen:
Added GPU prebuilt support in 991e680: gpu_support=True now installs the CUDA/ROCm/Metal bundles published on unslothai/llama.cpp (the same ones Studio uses), selected via the release manifest by host SM and torch CUDA runtime line, sha256 verified, with per-candidate validation and fallback to source compilation. CPU installs keep using upstream ggml-org archives. Verified live on a B200: the cuda12-newer bundle installs in 12 seconds and llama-server enumerates the GPU. PR description updated with the full selection logic.

PR #761, opened 2026-06-11
[staging] Metal regression tests verifying issue 6002 fixes on main

danielhanchen:
Staging validation only, not for merge. Adds the four Metal regression tests covering both unslothai/unsloth#6002 crashes (attention kwargs TypeError, gated-delta and fused MRoPE VJP) plus the mlx-ci step to run them, on top of current main. Will be closed once the macOS run confirms; the workflow_dispatch route kept getting cancelled in queue.


danielhanchen:
Staging validation superseded: the identical Metal test suite already ran fully green against the merged code in https://github.com/unslothai/unsloth-zoo/actions/runs/27288537041, and unslothai/unsloth#6002 is closed on that evidence. Queued runs on this branch kept getting cancelled, so no point keeping it open.

PR #764, opened 2026-06-11
fix: decode llama.cpp subprocess output as UTF-8 on Windows

dylanschroers:
Part of unslothai/unsloth#2660

## Problem

`unsloth_zoo/llama_cpp.py` clones, cmake-builds, quantizes and converts llama.cpp through text-mode subprocess calls (`text=True` / `universal_newlines=True`) that never set an `encoding`. On Windows the output is decoded as cp1252 (the locale default) instead of UTF-8. The build/convert pipeline prints non-ASCII along the way (localized cmake and compiler messages on non-English Windows, progress and box-drawing glyphs, model/tensor names, file paths), so when a byte cp1252 can't map shows up (0x9d is the usual one, it sits inside the UTF-8 encoding of common punctuation) the read raises `UnicodeDecodeError` and the build or convert aborts, often after a long compile. Valid-but-wrong bytes corrupt the captured text silently instead.

Reproduced on Windows (cp1252, `PYTHONUTF8` unset) using the `try_execute()` pattern:

```
UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d in position 24: character maps to <undefined>
```

This is the shared-engine half of the bug, the one both unsloth and Studio call under the hood. The `unsloth` save.py half is unslothai/unsloth#6218.

## Fix

- `unsloth_zoo/llama_cpp.py`
  - add `encoding="utf-8", errors="replace"` to all 12 text-mode subprocess calls, matching each call's existing spacing. Covers `try_execute` (streams the build/convert output), `_convert_to_gguf`, the cmake configure/build steps, and the winget/gcc/which/dpkg/rpm/pacman probes. `errors="replace"` means a stray byte can neither crash a build nor silently mangle the output.
- `tests/test_llama_cpp_subprocess_utf8_encoding.py` (new)
  - AST drift test: parses `llama_cpp.py` and fails if any text-mode subprocess call is missing `encoding="utf-8"`. No torch or built llama.cpp needed. Flagged all 12 call sites before the change, passes after.
  - behavioural test: spawns a real subprocess emitting a curly quote, asserts the raw bytes fail under cp1252 and read cleanly with the new kwargs.

Validation: `pytest tests/test_llama_cpp_subprocess_utf8_encoding.py` passes 3/3 (1 failed before the fix); `python -m compileall` is OK; `ruff check --select E9,F63,F7,F82` is clean.

### DIFF
--- a/tests/test_llama_cpp_subprocess_utf8_encoding.py
+++ b/tests/test_llama_cpp_subprocess_utf8_encoding.py
@@ -1,0 +1,125 @@
+# Regression tests for the Windows UnicodeDecodeError class in the llama.cpp
+# build/run path (unslothai/unsloth#2660).
+#
+# On Windows the default text encoding is the locale code page (e.g. cp1252),
+# not UTF-8. ``subprocess.run`` / ``subprocess.Popen`` opened in text mode
+# (``text=True`` / ``universal_newlines=True``) without an explicit
+# ``encoding`` therefore decode child output with cp1252. The llama.cpp
+# clone/cmake/quantize/convert pipeline emits non-ASCII content -- localized
+# compiler/cmake messages, progress / box-drawing glyphs, and model/tensor
+# names or file paths (e.g. ``C:\\Users\\Jose\\...``). When a byte undefined
+# in cp1252 appears (e.g. ``0x9d``, which sits inside the UTF-8 encoding of
+# common punctuation), the read raises ``UnicodeDecodeError`` and aborts the
+# build/quantize/convert.
+#
+# ``test_llama_cpp_subprocess_text_calls_declare_utf8_encoding`` is a
+# source-level drift detector: it parses ``unsloth_zoo/llama_cpp.py`` (no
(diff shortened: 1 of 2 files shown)


danielhanchen:
Confirmed the text-mode subprocess calls in unsloth_zoo/llama_cpp.py (try_execute at line 498, the cmake steps, and _convert_to_gguf) still have no encoding pinned, while other call sites already use encoding="utf-8", errors="replace", so this lines them up. Will get this reviewed.

danielhanchen:
Confirmed the text-mode subprocess calls in unsloth_zoo/llama_cpp.py (try_execute, the cmake configure/build steps, the package probes) still decode with the locale code page, and pinning utf-8 with errors="replace" matches what the rest of the file already does. Will get this reviewed.

danielhanchen:
Confirmed the text-mode subprocess calls in unsloth_zoo/llama_cpp.py (for example try_execute and _convert_to_gguf) still have no explicit encoding on main, so the cp1252 decode path is real. Will get this reviewed.

PR #766, opened 2026-06-12
Fix transformers 5.x drift: FP8Linear constructor signature and Gemma3 double-BOS strip under padding

danielhanchen:
Two transformers 5.x drift fixes, both backwards compatible with 4.57.6.

## FP8Linear constructor signature drift

transformers 4.57.6 builds `FP8Linear(in_features, out_features, bias, dtype, block_size, device, activation_scheme)`. transformers 5.x changed the signature to `(in_features, out_features, block_size, activation_scheme, scale_fmt, has_bias)`: `bias`, `dtype`, `device` are gone and `has_bias` arrived. `vllm_utils.py` passed the 4.x layout positionally, so FP8 models crashed on 5.x with a TypeError.

The fix builds a superset kwargs dict (`bias`, `has_bias`, `dtype`, `device`, `block_size`, `activation_scheme`) and filters it against `inspect.signature(FP8Linear.__init__)` before calling, with a VAR_KEYWORD guard so a future `**kwargs` signature receives everything. Works unchanged on 4.57.6, 5.x, and signatures in between.

## Gemma3 processor double-BOS under batched padding

The existing Gemma3 patch stripped a duplicated BOS by checking `input_ids[:2]` after tokenization with padding. Under left padding only the longest row starts with `[bos, bos]`, so the strip re-ragged the batch and desynced `attention_mask`.

The fix tokenizes unpadded, strips the duplicate BOS per row while keeping the attention mask in sync, then pads once via `tokenizer.pad` with the pad kwargs filtered against its signature (transformers 5 changed which kwargs `pad` accepts).

## Validation

Unit tests for both paths pass on transformers 4.57.6 and 5.11 (same scripts, both versions).

End to end inside the unsloth/unsloth Docker validation image (B200, torch 2.10 cu128, vllm 0.19.1, transformers 5.11), notebooks from unslothai/notebooks run with these fixes applied:

| Notebook | Result |
| --- | --- |
| Gemma3_(4B)-Vision-GRPO | pass, 13.6 min |
| Llama_FP8_GRPO | pass, 8.6 min |
| Qwen3_8B_FP8_GRPO | pass, 12.1 min |
| Qwen3_(4B)-GRPO (control, no FP8/Gemma path) | pass, 11.9 min |

Both fixes are runtime monkey patches scoped to the existing patch sites; no behavior change when the installed transformers matches the old signatures.

### DIFF
--- a/unsloth_zoo/temporary_patches/gemma.py
+++ b/unsloth_zoo/temporary_patches/gemma.py
@@ -189,11 +189,23 @@ def patch_Gemma3Processor():
         # text_inputs = self.tokenizer(text=text, **output_kwargs["text_kwargs"], return_tensors="np")
         return_mm_token_type_ids = output_kwargs["text_kwargs"].pop("return_mm_token_type_ids", True)
 
-        text_inputs = self.tokenizer(text=text, **output_kwargs["text_kwargs"])
-        # Fix double BOS tokens
+        # Tokenize WITHOUT padding: stripping a double BOS after padding only
+        # shortens rows that still start with [bos, bos] (under left padding,
+        # just the longest row), re-ragging the batch and desyncing
+        # attention_mask. Strip first, then pad once.
+        text_kwargs = dict(output_kwargs["text_kwargs"])
+        pad_kwargs = {k: text_kwargs.pop(k) for k in ("padding", "max_length", "pad_to_multiple_of", "padding_side") if k in text_kwargs}
+        text_inputs = self.tokenizer(text=text, **text_kwargs)
+        # Fix double BOS tokens, keeping attention_mask in sync
         double_bos_token_id = [self.tokenizer.bos_token_id]*2
         input_ids = text_inputs["input_ids"]
-        text_inputs["input_ids"] = [x[1:] if x[:2] == double_bos_token_id else x for x in input_ids]
+        stripped = [x[1:] if x[:2] == double_bos_token_id else x for x in input_ids]
+        if "attention_mask" in text_inputs:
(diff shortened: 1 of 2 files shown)


danielhanchen:
Re-validated on current main via an integration branch and a sim battery in isolated uv venvs (transformers 4.57.6 and 5.x).

FP8Linear signature filter (the unique part of this PR): 16/16 simulation cases pass. The before-repro confirms the bug: the pre-fix positional/keyword construction raises `TypeError: unexpected keyword argument 'dtype'` against the real transformers 5.x FP8Linear signature; the filter constructs correctly against the 4.57.6 layout, the 5.x layout, a **kwargs signature, a keyword-only signature, and a future-extra-arg signature, with has_bias True and False. End to end, the FP8 GRPO notebooks (Llama_FP8_GRPO, Qwen3_8B_FP8_GRPO) train to completion on a B200 image carrying this fix.

Note on the Gemma3 double-BOS hunk: the maintainer's open zoo#695 independently fixes the same bug with a different approach (tokenize-with-padding then re-pad trimmed rows, vs this PR's tokenize-unpadded-strip-pad-once). Both are correct at the token-id level (verified across padding x side x batch combinations, 56/56 sim cases). Since 695 is likely to land first, the Gemma3 hunk here is redundant with it; happy to drop it and keep this PR to just the FP8Linear fix if that is cleaner for review.

danielhanchen:
Closing as superseded.

The FP8Linear constructor half is now handled by signature-robust construction on current main, and the Gemma3 double-BOS handling was reworked through the `_fix_double_bos_and_pad` + `_resolve_truncation` path, which also preserves truncation for over-length rows (this PR's version dropped `max_length` before padding and lost that).

If the `inspect.signature`-based FP8Linear construction is still worth keeping purely as future-proofing, I can cherry-pick just that piece onto current main in a small follow-up.

PR #768, opened 2026-06-12
Fix vision GRPO under vLLM standby, bundle converter resolution, DeepseekV2MoE rename

danielhanchen:
Three independent runtime fixes, found while validating the unslothai/unsloth Docker images notebook by notebook (B200, torch 2.10 cu128, vllm 0.19.1, transformers 5.11).

## 1. Vision GRPO crashes under vLLM standby

With `UNSLOTH_VLLM_STANDBY=1` and a multimodal model, the CuMemAllocator sleep/wake cycle corrupts CUDA VMM: `cudaErrorIllegalAddress` at `empty_cache()` on the first sleep, or `Expected a cached item for mm_hash=...` desyncs. Text GRPO under the same standby setting is unaffected.

Two layers:
- `load_vllm` disables standby for vision models unless `UNSLOTH_VLLM_STANDBY_VISION=1`. This is also the faster configuration: `import unsloth` strips `expandable_segments` from the allocator config at import time when standby is requested, which cannot be undone later in the process, so a gated vision run avoids both the crash and the slow allocator.
- `patch_vllm_reset_caches_on_sleep` wraps `vllm.LLM.sleep` to clear the multimodal (P0 sender / P1 receiver, in lockstep) and encoder caches while their CUDA memory is still mapped, then delegates. Defense in depth for forced standby; every reset is guarded so any vllm version without these methods degrades to current behavior. Registered only in the standby path; no-op for text models.

Validation: Qwen2_5_7B_VL_GRPO previously crashed at about 22 minutes with both signatures; with the gate it runs 20/20 cells, 0 errors (TrainOutput step 4 in 238 s), and a 31 minute forced-standby run with the sleep wrapper showed zero crash signatures. Text control Qwen3_(4B)-GRPO passes unchanged.

## 2. GGUF conversion picks a mismatched converter for prebuilt bundles

When only `UNSLOTH_LLAMA_CPP_PATH` is set (the Docker images bake a prebuilt bundle there), the converter resolution downloaded the latest GitHub `convert_hf_to_gguf.py` and ran it against the bundle's older `conversion/ModelBase`, crashing with an unexpected `target_model_dir` kwarg. New `_resolve_bundle_convert_script()` prefers the bundle's own co-versioned converter when, and only when, the bundle ships a paired `conversion/` package. Source builds, monolith installs and hosts without a bundle are unchanged.

Validation: Llama3_(8B)-Ollama went from fail to pass (7.4 min) inside the image.

## 3. DeepseekV2MoE renamed in transformers 5

transformers 5 renamed `DeepseekV2MoE` to `DeepseekV2Moe`; trust_remote_code models (DeepSeek-OCR) still import the old name. A temporary patch aliases it back when absent; no-op on 4.x. Note DeepSeek-OCR then hits a separate transformers 5 strict dataclass validation of its checkpoint config (`kv_lora_rank=None`), which is a checkpoint level incompatibility and deliberately not patched here.

All changes are pure python guards over existing parameters and env vars, no version specific imports: compatible with transformers 4.57.6 and 5.x, TRL 0.22.2 through 1.x, vLLM 0.11.2 onward, and inert on non NVIDIA platforms.

### DIFF
--- a/tests/test_deepseek_v2_moe_alias.py
+++ b/tests/test_deepseek_v2_moe_alias.py
@@ -1,4 +1,4 @@
-"""PR #768: patch_deepseek_v2_moe_alias.
+"""Tests for the DeepseekV2MoE compatibility alias (patch_deepseek_v2_moe_alias).
 
 transformers 5.x renamed DeepseekV2MoE -> DeepseekV2Moe; trust_remote_code models
 (e.g. DeepSeek-OCR) still import the old name. The patch aliases it back when only
--- a/tests/test_llama_cpp_bundle_converter.py
+++ b/tests/test_llama_cpp_bundle_converter.py
@@ -1,15 +1,14 @@
-"""PR #768: _resolve_bundle_convert_script (llama.cpp bundle converter fallback).
+"""Tests for the llama.cpp bundle converter resolver (_resolve_bundle_convert_script).
 
 A prebuilt llama.cpp bundle ships convert_hf_to_gguf.py co-versioned with its own
 conversion/ package. When UNSLOTH_LLAMA_CPP_SCRIPTS_DIR is unset, the resolver
-prefers the bundle's converter *only when* the paired conversion/__init__.py is
-present; otherwise it returns None (monolith installs / hosts without a bundle
-fall through to the network).
+prefers the bundle's converter only when the paired conversion/ package (both
+__init__.py and base.py) is present; otherwise it returns None (monolith installs
+/ hosts without a paired package fall through to the network).
 
 Loads llama_cpp.py in isolation (spec_from_file_location), matching the idiom in
 tests/test_convert_hf_to_gguf_patcher.py, and monkeypatches LLAMA_CPP_DEFAULT_DIR
-to a tmp tree. No network, no GPU. Closes a coverage gap: no existing test names
-this function.
+to a tmp tree. No network, no GPU.
 """
 
 from __future__ import annotations
@@ -25,7 +24,7 @@ import pytest
 def _load_llama_cpp_module():
     repo_root = Path(__file__).resolve().parents[1]
     module_path = repo_root / "unsloth_zoo" / "llama_cpp.py"
-    spec = importlib.util.spec_from_file_location("llama_cpp_under_test_pr768", module_path)
+    spec = importlib.util.spec_from_file_location("llama_cpp_bundle_under_test", module_path)
     module = importlib.util.module_from_spec(spec)
     sys.modules[spec.name] = module
     spec.loader.exec_module(module)
@@ -122,7 +121,7 @@ def test_underscore_wins_when_both_converters_present(mod, monkeypatch, tmp_path
 
 
 def test_source_or_new_layout_checkout_also_resolves(mod, monkeypatch, tmp_path):
-    # F6: the resolver keys purely on (conversion/__init__.py + converter), so a
+    # The resolver keys purely on (conversion/{__init__,base}.py + converter), so a
     # new-layout source checkout under LLAMA_CPP_DEFAULT_DIR also resolves locally
     # rather than hitting the network. This is intended (its converter is
     # genuinely co-versioned with its own conversion/) and documented here.
--- a/tests/test_pr768_bundle_converter.py
+++ b/tests/test_pr768_bundle_converter.py
@@ -1,0 +1,147 @@
+"""PR #768: _resolve_bundle_convert_script (llama.cpp bundle converter fallback).
+
+A prebuilt llama.cpp bundle ships convert_hf_to_gguf.py co-versioned with its own
+conversion/ package. When UNSLOTH_LLAMA_CPP_SCRIPTS_DIR is unset, the resolver
+prefers the bundle's converter *only when* the paired conversion/__init__.py is
+present; otherwise it returns None (monolith installs / hosts without a bundle
+fall through to the network).
+
+Loads llama_cpp.py in isolation (spec_from_file_location), matching the idiom in
+tests/test_convert_hf_to_gguf_patcher.py, and monkeypatches LLAMA_CPP_DEFAULT_DIR
+to a tmp tree. No network, no GPU. Closes a coverage gap: no existing test names
+this function.
+"""
+
+from __future__ import annotations
+
+import importlib.util
+import os
+import sys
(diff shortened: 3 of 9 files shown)


danielhanchen:
Validation results for this branch, run inside the unslothai/unsloth#5748 Docker image (B200, torch 2.10 cu128, vllm 0.19.1, transformers 5.11), zoo installed from this branch.

| Notebook | Result | Notes |
| --- | --- | --- |
| Llama3_(8B)-Ollama | pass, 5.6 min | was failing with the mismatched converter (unexpected `target_model_dir` kwarg); the bundle converter resolution fixes it |
| Qwen3_VL_(8B)-Vision-GRPO | pass, 13.4 min | trains and completes; this notebook never used vLLM (`fast_inference = False`), confirming the gate does not touch non vLLM vision runs |
| Qwen2_5_7B_VL_GRPO | blocked on current main by a separate pre-existing bug | details below |

On the Qwen2_5_7B_VL_GRPO row: current main fails inside `load_vllm` with vLLM's `RuntimeError: Tried to erase Node size_3 but it still had 1 users in the graph` before any change from this branch can matter. Forcing `UNSLOTH_VLLM_STANDBY_VISION=1` (which restores the exact pre-branch standby path at load time) reproduces the same error, so the failure is independent of this branch. #695 carries `patch_vllm_decompose_size_nodes` which targets exactly this error, so the two PRs compose.

The standby fixes themselves were validated on a #695-vintage build where load succeeds:
- gate active (standby auto disabled for the vision model): 20/20 cells, 0 errors, TrainOutput step 4 in 238 s, where the unpatched run crashed at about 22 min with `cudaErrorIllegalAddress` / `Expected a cached item for mm_hash` 
- standby forced on with the sleep wrapper registered: 31 minutes of training with zero crash signatures

Text GRPO control (Qwen3 4B) passes unchanged.

danielhanchen:
Re-validated on current main via an integration branch (this PR + maintainer zoo#695) and a sim battery.

Unique parts of this PR, all validated:
- Vision standby gate (disable sleep mode for multimodal unless UNSLOTH_VLLM_STANDBY_VISION=1): gate truth table 8/8 cases; only (standby on, vision, no override) flips to off.
- Pre-sleep multimodal/encoder cache reset wrapper: verified against fake vllm.LLM variants (with/without reset_mm_cache, reset_encoder_cache, llm_engine; raising resets; double-patch idempotency; resets fire before the original sleep; no-op for text). Together these eliminate the vision GRPO standby crash class (cudaErrorIllegalAddress / mm_hash desync); Qwen2_5_7B_VL_GRPO ran clean at 238 s with the gate, where the unpatched run crashed at ~22 min.
- Bundle converter resolution: tmp-layout sim over conversion-package present/absent x converter-file present/absent x dir-missing x SCRIPTS_DIR set/unset. End to end, Llama3_(8B)-Ollama flips fail -> pass on the image (the bug was the latest GitHub convert_hf_to_gguf.py run against the bundle's older ModelBase).

Note on the DeepseekV2MoE alias hunk: the maintainer's zoo#695/main already carries a more comprehensive DeepSeek-V2 fix (the alias plus a DeepSeek-OCR generation patch). This PR's small alias is redundant with it and can be dropped; the vision standby and bundle converter parts are the unique contributions.

26/26 sim cases pass; no old pathway regressed.

danielhanchen:
Pushed a follow-up (15866d2) that hardens the three fixes after a closer look at the standby and converter paths.

**vLLM standby**
- `patch_vllm` now gates the standby patches on `UNSLOTH_VLLM_STANDBY != "0"` to match `load_vllm`. A truthy value other than `1` (e.g. `2`) was previously enabling standby in `load_vllm` while skipping `patch_vllm_reset_caches_on_sleep`, leaving standby active but unpatched.
- When the vision gate disables standby, `load_vllm` now also sets `UNSLOTH_VLLM_STANDBY=0`. The gate only flipped a local variable before, so the env-keyed consumers (the earlier `patch_vllm` call via `fast_inference_setup`, and the GRPO sleep-mode arg in `unsloth/models/rl.py`) still saw `1` and could drive sleep mode even though the engine was built with `enable_sleep_mode=False`.
- `patch_vllm_reset_caches_on_sleep` now wraps `sleep` on `LLM`, `LLMEngine` and `AsyncLLMEngine` (per class, idempotent), so the `use_async` / `use_engine` paths are no longer bypassed.

**GGUF bundle converter**
- `_resolve_bundle_convert_script` now requires `conversion/base.py` as well as `conversion/__init__.py`, the same signal `_detect_converter_layout` uses. A partial bundle with only `__init__.py` could previously be selected and then mis-classified as monolith, failing on `conversion.base` instead of falling through to the network.

**DeepseekV2MoE alias**
- Skips on transformers < 5 before importing, so we no longer eagerly import the `deepseek_v2` module on every startup where the rename does not exist.

Added unit tests for all three areas under `tests/test_pr768_*`.

Validation: gate truth table 8/8; text and vision GRPO under standby run clean (Qwen2.5-VL-7B, forced vision standby, 10 sleep/wake cycles) with no `cudaErrorIllegalAddress` or `mm_hash` signatures on torch 2.10 + vLLM 0.19.1, composed with the decompose-size-nodes fix from #695.

danielhanchen:
Follow-ups pushed (8a0ad3b):

- `load_vllm`: the vision standby gate now also falls back to `hasattr(config, "vision_config")`, so it still fires if a caller does not pass `is_vision_model=True`.
- `patch_vllm_reset_caches_on_sleep`: a cache reset that exists but raises now warns once instead of staying silent.

Companion change in the main repo: unslothai/unsloth#6420 makes the GRPO `vllm_enable_sleep_mode` arg read the engine's actual `enable_sleep_mode` rather than the raw env var, so the TRL config can no longer disagree with the gated engine.

Re-validated end to end on B200 + vLLM 0.19.1 with both changes installed:

| Run | Result |
| --- | --- |
| Text GRPO, `STANDBY=1` | standby stays on, sleep/wake each step, no crash |
| Vision GRPO, default gate | standby fully disabled, zero sleep/wake during training, 6/6 steps clean |
| Vision GRPO, `STANDBY_VISION=1` (forced) | 8 sleep/wake cycles, 8/8 steps, no `cudaErrorIllegalAddress` or `mm_hash` signatures |

Before these changes the gated vision run still drove a sleep/wake every step despite the warning; it is now zero.

PR #769, opened 2026-06-12
diffusion_studio: split shim thought channels into reasoning_content

raydeStar:
The shim streamed DiffusionGemma's committed text verbatim, so thought markers rendered as raw text in chat UIs. The model emits two thinking
dialects: the chat template's native <|channel>thought ... <channel|>
and DeepSeek-style <think> ... </think> when thinking is not explicitly templated. Split both out of committed text into OpenAI-style reasoning_content (matching llama-server's reasoning extraction) for both streaming and non-streaming responses, so chat UIs fold the thinking into a collapsible section instead of showing raw markers.

Streaming withholds a trailing partial marker until the next block commit, so a marker split across the 256-token canvas block boundary never leaks fragments into either channel. Handles unterminated thoughts (length-truncated generations) and multiple thought blocks.

Before:

image

After:

image

### DIFF
--- a/tests/test_diffusion_shim_thought_split.py
+++ b/tests/test_diffusion_shim_thought_split.py
@@ -1,0 +1,166 @@
+# Unsloth Zoo - Utilities for Unsloth
+# Copyright 2023-present Daniel Han-Chen, Michael Han-Chen & the Unsloth team. All rights reserved.
+#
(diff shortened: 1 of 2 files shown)


danielhanchen:
Thanks!

PR #770, opened 2026-06-12
fix(diffusion_studio): add torch/lib to PATH on Windows so visual-server loads CUDA backend

Anai-Guo:
## Summary

On Windows, the DiffusionGemma visual-server (`llama-diffusion-gemma-visual-server`) silently runs CPU-only: the model loads into system RAM instead of VRAM even though Studio passes `NGL=99` / `CUDA_VISIBLE_DEVICES=0`. Probing `ggml-cuda.dll` returns **error 126 (MOD_NOT_FOUND)** — the DLL is present but its CUDA runtime deps (`cudart64_13.dll`, `cublas64_13.dll`, …) are not on the child's DLL search order.

Those DLLs **are** installed, in `torch/lib` (bundled by the `torch+cuXXX` wheel). The problem is that `VisualServer.__init__` only sets `LD_LIBRARY_PATH`, which is a no-op on Windows — so `torch/lib` never makes it onto `PATH` for the spawned server.

This is the same root cause as the llama-server bug fixed in unslothai/unsloth#5491; the Windows `PATH` bootstrap was never carried over to the diffusion runner.

## Fix

When `os.name == "nt"`, prepend torch's bundled `torch/lib` to `env["PATH"]` before spawning the server, guarded on the dir actually shipping a `cudart64_*.dll`. The version glob keeps it CUDA-major-agnostic (works for `cudart64_12`, `cudart64_13`, …). Linux behaviour (the existing `LD_LIBRARY_PATH` line) is unchanged; non-Windows platforms skip the new branch entirely. If torch isn't importable or `torch/lib` has no cudart, the helper returns `None` and `PATH` is left untouched.

## Verification (from unslothai/unsloth#6273)

Prepending `torch/lib` to `PATH` makes the CUDA backend load:
```
ggml_cuda_init: found 1 CUDA devices (Total VRAM: 24563 MiB):
  Device 0: NVIDIA GeForce RTX 4090, compute capability 8.9, VMM: yes
load_backend: loaded CUDA backend from ...\ggml-cuda.dll
```

Fixes unslothai/unsloth#6273.

---

### DIFF
--- a/tests/test_diffusion_visual_engine_env.py
+++ b/tests/test_diffusion_visual_engine_env.py
@@ -1,0 +1,77 @@
+# Unsloth Zoo - Utilities for Unsloth
+# Copyright 2023-present Daniel Han-Chen, Michael Han-Chen & the Unsloth team. All rights reserved.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+# Coverage for the DiffusionGemma visual-server subprocess env builder
+# (visual_engine._build_subprocess_env): the bundled CUDA runtime must reach the
+# child's library path so it loads the CUDA backend instead of falling back to CPU
+# (Windows unsloth#6273, Linux/WSL #6303), and torch/lib must not shadow the C++
+# binary's system libs on Linux.
+
+from __future__ import annotations
+
+import os
+
+from unsloth_zoo.diffusion_studio import visual_engine as V
+
+
+def test_linux_prepends_nvidia_wheels_and_excludes_torch_lib(monkeypatch):
+    monkeypatch.setattr(V, "_bundled_cuda_lib_dirs", lambda: ["/wheel/nvidia/cu13/lib"])
+    monkeypatch.setattr(V, "_torch_cuda_lib_dir", lambda: "/py/torch/lib")
+    env = V._build_subprocess_env(
+        "/opt/dg/llama-diffusion-gemma-visual-server",
+        gpu = "3", maxtok = 4096,
+        base_env = {"PATH": "/usr/bin", "LD_LIBRARY_PATH": "/sys/cuda"},
+        os_name = "posix",
+    )
+    ld = env["LD_LIBRARY_PATH"].split(os.pathsep)
+    assert ld[0] == "/opt/dg"                      # binary dir first
+    assert "/wheel/nvidia/cu13/lib" in ld          # bundled CUDA runtime present
+    assert ld[-1] == "/sys/cuda"                    # inherited path kept, last
+    assert all("torch" not in p for p in ld)       # torch/lib excluded on Linux
+    assert env["PATH"] == "/usr/bin"               # PATH untouched on Linux
+    assert env["CUDA_VISIBLE_DEVICES"] == "3"
+    assert env["NGL"] == "99"
+    assert env["MAXTOK"] == "4096"
+
+
+# Note: paths here are colon-free so the assertions hold regardless of the host's
+# os.pathsep (on a Linux CI host os.pathsep is ":", which would split a "C:\..." path).
+def test_windows_prepends_torch_lib_to_path_only(monkeypatch):
+    monkeypatch.setattr(V, "_bundled_cuda_lib_dirs", lambda: ["/wheel/nvidia/cu13/lib"])
+    monkeypatch.setattr(V, "_torch_cuda_lib_dir", lambda: "/py/torch/lib")
+    env = V._build_subprocess_env(
+        "/dg/llama-diffusion-gemma-visual-server.exe",
+        gpu = "0", maxtok = 0,
+        base_env = {"PATH": "/winpath"},
+        os_name = "nt",
+    )
+    assert env["PATH"].startswith("/py/torch/lib" + os.pathsep)         # torch/lib first on PATH
+    assert "/wheel/nvidia/cu13/lib" not in env.get("LD_LIBRARY_PATH", "")  # not forced on Windows
+    assert env["MAXTOK"] == "0"
+
+
+def test_windows_without_torch_cudart_leaves_path_untouched(monkeypatch):
+    monkeypatch.setattr(V, "_torch_cuda_lib_dir", lambda: None)
+    env = V._build_subprocess_env(
+        "/dg/srv.exe", base_env = {"PATH": "/winpath"}, os_name = "nt",
+    )
+    assert env["PATH"] == "/winpath"
+
+
+def test_bundled_cuda_lib_dirs_finds_nvidia_next_to_install(tmp_path, monkeypatch):
+    fake_file = tmp_path / "unsloth_zoo" / "diffusion_studio" / "visual_engine.py"
+    fake_file.parent.mkdir(parents = True)
+    fake_file.write_text("")
+    for pkg in ("cu13", "cudnn"):
+        (tmp_path / "nvidia" / pkg / "lib").mkdir(parents = True)
+    monkeypatch.setattr(V, "__file__", str(fake_file))
+    dirs = V._bundled_cuda_lib_dirs()
+    assert str(tmp_path / "nvidia" / "cu13" / "lib") in dirs
+    assert str(tmp_path / "nvidia" / "cudnn" / "lib") in dirs
+    assert all("torch" not in d for d in dirs)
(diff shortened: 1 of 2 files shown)


rrfaria:
Same issue here:

Fixes an issue where `llama-diffusion-gemma-visual-server` falls back to CPU on Linux/WSL when the host's system CUDA version differs from the python environment's CUDA packages version (e.g., system has CUDA 12.4 installed in `/usr/local/`, but python virtualenv uses `nvidia-cuda-runtime-cu13` with `libcudart.so.13`).

### The Problem

For regular text models, `LlamaCppBackend` in `studio/backend/core/inference/llama_cpp.py` dynamically resolves and appends the virtualenv's pip-installed `nvidia/cu*/lib` paths to `LD_LIBRARY_PATH` so the dynamically-loaded CUDA backend `libggml-cuda.so` can resolve dependencies like `libcudart.so.13`.

However, for visual/diffusion models (Gemma 3D/DiffusionGemma), the subprocess is spawned by `VisualServer` in `unsloth_zoo/diffusion_studio/visual_engine.py`. This script only appends `self.server_bin` directory to `LD_LIBRARY_PATH`, leaving out the python package folders. 

Because `libcudart.so.13` and `libcublas.so.13` cannot be found by the system dynamic linker inside `visual_engine.py`'s environment, loading `libggml-cuda.so` fails, and the runner silently falls back to CPU execution (`libggml-cpu.so`).

### The Solution

Mirror the `LD_LIBRARY_PATH` resolution block from `llama_cpp.py` inside `visual_engine.py` to append the venv's pip-installed CUDA package directories (`nvidia/cu*/lib`, `nvidia/cudnn/lib`, `nvidia/nvjitlink/lib`) before spawning the visual server process.

Here is the proposed change:

```diff
--- a/unsloth_zoo/diffusion_studio/visual_engine.py
+++ b/unsloth_zoo/diffusion_studio/visual_engine.py
@@ -95,3 +95,19 @@ class VisualServer:
         env = dict(os.environ, CUDA_VISIBLE_DEVICES=str(gpu), NGL="99", MAXTOK=str(maxtok))
-        env["LD_LIBRARY_PATH"] = os.path.dirname(self.server_bin) + os.pathsep + env.get("LD_LIBRARY_PATH", "")
+        
+        # Add pip-installed CUDA runtime packages from the virtualenv to LD_LIBRARY_PATH
+        import glob as _glob
+        import sys
+        lib_dirs = [os.path.dirname(self.server_bin)]
+        for _nv_pattern in [
+            os.path.join(sys.prefix, "lib", "python*", "site-packages", "nvidia", "cu*", "lib"),
+            os.path.join(sys.prefix, "lib", "python*", "site-packages", "nvidia", "cudnn", "lib"),
+            os.path.join(sys.prefix, "lib", "python*", "site-packages", "nvidia", "nvjitlink", "lib"),
+        ]:
+            for _nv_dir in _glob.glob(_nv_pattern):
+                if os.path.isdir(_nv_dir):
+                    lib_dirs.append(_nv_dir)
+
+        existing_ld = env.get("LD_LIBRARY_PATH", "")
+        new_ld = os.pathsep.join(lib_dirs)
+        env["LD_LIBRARY_PATH"] = f"{new_ld}{os.pathsep}{existing_ld}" if existing_ld else new_ld
         self.env = env
```

Extra tip:

in my case I have 2 gpu and 20GB VRAM, to avoid out of memory I used:
[diffusiongemma-26B-A4B-it-Q4_K_M.gguf](https://huggingface.co/unsloth/diffusiongemma-26B-A4B-it-GGUF/blob/main/diffusiongemma-26B-A4B-it-Q4_K_M.gguf)
```
CUDA_VISIBLE_DEVICES=0,1 NGL=22 unsloth studio -p 8888
```

with 2024 context size. and using old gpus:

image


NGL controls how many layers will be offloaded to the GPU.


and better, if you want to automatically calculate how many layers 

``` python
def _get_free_vram_mb(cuda_devices_str):
    """Query free VRAM using nvidia-smi for the selected visible devices."""
    try:
        res = subprocess.run(
            ["nvidia-smi", "--query-gpu=memory.free", "--format=csv,noheader,nounits"],
            stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True, check=True
        )
        all_gpus_free = [int(x) for x in res.stdout.strip().split("\n") if x.strip()]
        if cuda_devices_str:
            visible_idx = [int(i.strip()) for i in cuda_devices_str.split(",") if i.strip()]
            free_vram = [all_gpus_free[idx] for idx in visible_idx if idx < len(all_gpus_free)]
        else:
            free_vram = all_gpus_free
        return sum(free_vram)
    except Exception:
        # Fallback to 12GB if nvidia-smi fails
        return 12000
def _get_gguf_metadata(model_path):
    """Retrieve model file size and try to read the block count metadata."""
    try:
        file_size_mb = os.path.getsize(model_path) / (1024 * 1024)
    except Exception:
        file_size_mb = 16000  # ~15.6 GiB fallback
    
    total_layers = 30  # DiffusionGemma default
    try:
        from gguf import GGUFReader
        reader = GGUFReader(model_path)
        for key in reader.fields.keys():
            if key.endswith(".block_count"):
                total_layers = int(reader.fields[key].parts[-1][0])
                break
    except Exception:
        pass
    return file_size_mb, total_layers
def _calculate_auto_ngl(model_path, cuda_devices_str, reserve_vram_mb=5500):
    """Calculate the optimal NGL based on available VRAM and model layers."""
    free_vram = _get_free_vram_mb(cuda_devices_str)
    model_size_mb, total_layers = _get_gguf_metadata(model_path)
    vram_for_weights = free_vram - reserve_vram_mb
    if vram_for_weights <= 0:
        return 0
    mb_per_layer = model_size_mb / total_layers
    optimal_ngl = int(vram_for_weights // mb_per_layer)
    return min(total_layers, max(0, optimal_ngl))
```

and replace line unsloth_zoo/diffusion_studio/visual_engine.py:
```diff
- ngl = os.environ.get("NGL", "99")
+ ngl = os.environ.get("NGL")
+      if ngl is None:
+          try:
+              ngl = str(_calculate_auto_ngl(self.gguf, cuda_devices))
+          except Exception:
+              ngl = "99"
```

danielhanchen:
Thanks for this, and for flagging the overlap with #6303. I pushed a follow-up commit on top of your Windows fix that adds the Linux/WSL sibling so both land together.

What it does:
- `_bundled_cuda_lib_dirs()` collects the venv pip `nvidia-*` CUDA wheel lib dirs and `_build_subprocess_env` prepends them to `LD_LIBRARY_PATH` ahead of any inherited system CUDA on non-Windows hosts. This is the Linux analogue of your `torch/lib` on `PATH` fix and resolves the system-`libcudart` WSL crash in #6303 (cudaGetDeviceCount in the PTX JIT) as well as the silent CPU fallback when no system CUDA is present.
- `torch/lib` is intentionally left off `LD_LIBRARY_PATH` on Linux: it also bundles `libgomp`/`libstdc++`, which there could shadow the system libs the C++ visual-server links against, and the runtime is already covered by the `nvidia-*` wheels. This matches the llama-server resolver in studio `llama_cpp.py`. (Windows is unchanged, your `torch/lib` on `PATH` is what ships `cudart64_*.dll` there.)
- I extracted the env construction into a pure `_build_subprocess_env` helper and verified the {Windows, Linux, WSL, macOS} x {NVIDIA, AMD, CPU} matrix, including that no-GPU hosts behave exactly as before.

The MAXTOK / non-causal-canvas OOM clamp from #6303 is a separate concern and is not included here.

Anai-Guo:
Thanks @danielhanchen — the Linux/WSL analogue looks right. Prepending the `nvidia-*` wheel lib dirs ahead of inherited system CUDA on `LD_LIBRARY_PATH`, and deliberately leaving `torch/lib` off on Linux (to avoid shadowing system `libgomp`/`libstdc++` that the C++ visual-server links against) both make sense, and it mirrors the llama-server resolver. LGTM. 🤖 Generated with [Claude Code](https://claude.com/claude-code)

danielhanchen:
Pushed a small refinement plus tests. Made `_bundled_cuda_lib_dirs` resolve the nvidia-* wheel lib dirs relative to the installed unsloth_zoo (and the env purelib) instead of the `sys.prefix` glob, so it also works under editable and non-standard-prefix installs. Kept torch/lib excluded on Linux and on PATH for Windows.

I confirmed the torch/lib choice empirically on a CUDA box via /proc maps: with this builder the visual server resolves libcudart/libcublas from `nvidia/cu13/lib` and reaches `ggml_cuda_init`, whereas adding torch/lib on Linux only shadows the binary's libgomp with torch's copy (no CUDA benefit), so excluding it is correct. Added env-builder unit tests.

I am treating this as the canonical CUDA-runtime fix for the visual server, covering both Windows (#6273) and Linux/WSL (#6303). Credit to ThrownLemon for the Linux/WSL diagnosis in #771, which I have reduced to just its MAXTOK canvas clamp on top of this.

danielhanchen:
Thanks!

PR #771, opened 2026-06-13
fix(diffusion): load bundled CUDA runtime + auto-size canvas for DiffusionGemma on Linux/WSL2

ThrownLemon:
## What

Fixes DiffusionGemma (`llama-diffusion-gemma-visual-server`) on Linux / WSL2, where it currently crashes at startup with `munmap_chunk(): invalid pointer` (SIGABRT) and never loads. Linux analogue of #6273; full diagnosis in #6303.

Two changes in `unsloth_zoo/diffusion_studio/visual_engine.py`:

1. **Prepend the bundled CUDA runtime to `LD_LIBRARY_PATH`.** The runner previously set `LD_LIBRARY_PATH` to only the binary directory, so a system CUDA toolkit on the ldconfig path (e.g. `/usr/local/cuda`, CUDA 13.3) got loaded instead of the venv's bundled `nvidia-cu13` (13.0). The system runtime then does a bad `free()` inside `libnvidia-ptxjitcompiler.so.1` / the WSL `libcuda.so.1.1` during `cudaGetDeviceCount()`. New helper `_bundled_cuda_lib_dirs()` adds `site-packages/nvidia/*/lib` (cu13 first) and `torch/lib`.

2. **Clamp the non-causal canvas (`MAXTOK`).** The canvas is non-causal, so the compute buffer grows ~quadratically with size; the model's full context (32768) reserves a ~77 GB buffer → `cudaMalloc failed` on 24 GB. Values above 8192 fall back to the server's `MAXTOK=0` auto-size path.

## Backtrace (root cause)

```
#6  malloc_printerr ("munmap_chunk(): invalid pointer")
#8  __libc_free ()
#9  __cuda_CallJitEntryPoint ()   from libnvidia-ptxjitcompiler.so.1
#10 ...                           from /usr/lib/wsl/drivers/.../libcuda.so.1.1
#17 cudaGetDeviceCount ()         from /usr/local/cuda/.../libcudart.so.13
#18 ggml_cuda_init () -> #23 llama_backend_init () -> #24 main ()
```

## Test plan

- Env: WSL2 Ubuntu 24.04, RTX 4090 24 GB (driver 610.47), system CUDA 13.3, `torch` 2.10.0+cu130, `unsloth_zoo` 2026.6.5, llama.cpp `b9616`.
- **Before:** `unsloth run --model unsloth/diffusiongemma-26B-A4B-it-GGUF` → `HTTP 500: Failed to load GGUF model`; binary crashes even on `--help`.
- **After:** loads cleanly (`ggml_cuda_init: found 1 CUDA devices: RTX 4090`), `canvas=256, MAXTOK=4096`, ~23.7 GB VRAM, **~138 tok/s** via the OpenAI shim.

## Notes

The `MAXTOK <= 8192 else auto-size` clamp is a pragmatic guard for single 24 GB GPUs; maintainers may prefer to drive the threshold from detected free VRAM. The `LD_LIBRARY_PATH` change is the actual crash fix and is independent of the clamp.

Fixes #6303


danielhanchen:
Thanks, the canvas clamp is a good catch. I reduced this PR to just that change. The Linux/WSL CUDA runtime libpath fix overlapped with #770 (same `_bundled_cuda_lib_dirs` helper, which would have conflicted), so I moved that to #770 as the single env builder and stripped it here, leaving the MAXTOK clamp as the focused change. I extracted it into a small `_canvas_maxtok` helper with unit tests (oversized defers to auto-size 0, in-range honored).

This should merge after #770 and will rebase cleanly onto its `_build_subprocess_env`. Verified on hardware that an oversized 32768 request now reaches the runner as MAXTOK=0 (auto-size) and 4096 passes through unchanged.

danielhanchen:
Thanks!

PR #772, opened 2026-06-14
Fix harmony analysis channel using thinking instead of content

### LINKED ISSUE #246: encode_conversation_with_harmony try to refer "content" when "thinking" in message.

mvanhorn:
## Summary
In `temporary_patches/gpt_oss.py`, `encode_conversations_with_harmony` built the
`analysis`-channel `Message` from `message["content"]` inside the `"thinking"`
branch, so the reasoning text in `message["thinking"]` was silently dropped and
the visible answer was emitted on the wrong channel. This now encodes the
`analysis` channel from `message["thinking"]`, and when the same assistant turn
also has non-empty `content`, additionally emits a `final`-channel message so the
answer is preserved instead of being lost. The `tool_calls` and plain final-only
branches are unchanged.

## Why this matters
Assistant turns carrying both reasoning and an answer previously lost their
reasoning and mis-routed their content, corrupting the encoded Harmony
conversation used for gpt-oss training/inference. See
https://github.com/unslothai/unsloth-zoo/issues/246

## Testing
`python3 -m pytest tests/test_encode_conversations_with_harmony.py -q` -> 15 passed.
Four new hermetic CPU-only cases cover thinking+content (analysis carries thinking,
final carries content), thinking-only (single analysis message), plain content
(unchanged final-only), and a mixed multi-turn conversation confirming channel
ordering and that the tool branches stay untouched.

Fixes #246

### DIFF
--- a/tests/test_encode_conversations_with_harmony.py
+++ b/tests/test_encode_conversations_with_harmony.py
@@ -517,6 +517,93 @@ def test_mixed_multi_turn_preserves_channels_and_branches(
     assert tool_message.content == "tool result"
 
 
+def test_assistant_thinking_with_tool_calls_keeps_both(
+    monkeypatch,
+    gpt_oss_module,
+):
+    # An assistant turn carrying BOTH reasoning and a tool call must emit the analysis
+    # message AND the commentary tool call. The tool call must not be dropped.
+    messages = [
+        {
+            "role": "assistant",
+            "thinking": "decide to call the tool",
+            "tool_calls": [{"name": "lookup", "arguments": "{\"city\": \"Paris\"}"}],
+        },
+    ]
+    assistant_messages = _encode_and_capture_messages(
+        monkeypatch, gpt_oss_module, messages
+    )
+
+    channels = [m.channel for m in assistant_messages]
+    assert channels == ["analysis", "commentary"]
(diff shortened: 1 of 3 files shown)


danielhanchen:
@codex review

mvanhorn:
Glad this is in, thanks @danielhanchen. Using the right channel for harmony analysis instead of thinking fixes the gpt_oss patch.

mvanhorn:
Appreciate this fix landing, @danielhanchen. Pulling the analysis channel from content instead of thinking should keep the harmony reasoning intact.

PR #774, opened 2026-06-15
fix: strip MTP config keys before GGUF export to prevent block_count inflation

ugifractal:
Strip `mtp_num_hidden_layers` and `unsloth_fixed_mtp` from `config.json` (both top-level and `text_config`) before the llama.cpp converter reads it.

**Three changes:**

1. **`unsloth_zoo/mlx/utils.py`** — `save_pretrained_gguf`: strip MTP keys from temp config after saving the merged model. Also restores `architectures` key from original HF config (`mlx-vlm.save_config` strips it).

2. **`unsloth_zoo/llama_cpp.py`** — `convert_to_gguf`: strip MTP keys from config.json on disk before running `convert_hf_to_gguf.py`. This is the central entry point that covers **all** callers (Studio UI, direct API, etc.).

3. **`unsloth_zoo/llama_cpp.py`** — `convert_to_gguf`: make architecture/vision checks graceful when `architectures` key is missing (MLX-style configs).

Fixes GGUF export for Qwen3.5-2B and other MTP-enabled models on MLX.

### DIFF
--- a/unsloth_zoo/llama_cpp.py
+++ b/unsloth_zoo/llama_cpp.py
@@ -2126,8 +2126,7 @@ def convert_to_gguf(
 
     # Check if arch is supported
     supported_types = (supported_vision_archs or set()) | (supported_text_archs or set())
-    if supported_types:
-        assert("architectures" in config_file)
+    if supported_types and "architectures" in config_file:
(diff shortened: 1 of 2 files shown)


danielhanchen:
Thanks!

PR #777, opened 2026-06-15
Optimize LoRA merge to 16bit save and add end-to-end merge correctness suite

danielhanchen:
## Summary

Optimizes the LoRA-merge to 16bit safetensors save path (`merge_and_overwrite_lora` in `unsloth_zoo/saving_utils.py`) and adds the first end-to-end correctness coverage for the whole merge pipeline. Output is tensor-identical to the previous behavior (byte-identical for the safetensors-written shards real models ship); only memory and timing change.

Rebased onto current `main`, so it now includes the Gemma-4/Qwen3.5 narrow-expert fused MoE fix (#779). Without that fix the fused merge on this branch dropped expert deltas; the new tests caught it, and the rebase resolves it cleanly.

## Optimizations

- Drop the per-tensor `device_empty_cache()` / `device_synchronize()` churn in the merge hot loop (standard and mxfp4 paths). Cleanup now runs only after a real GPU merge on a large tensor (>= 256MB) plus once at shard end. The post-merge `.cpu()` is synchronous, so removing the per-key sync introduces no async-copy hazard.
- Stream the vocab-resized shard rewrite to a temp file with an atomic `os.replace` (lower peak RAM, and crash-safe where `main` overwrote the shard in place). When free disk cannot fit the transient shard, fail closed and leave the original shard intact rather than risk a non-atomic in-place overwrite; set `UNSLOTH_ALLOW_NON_ATOMIC_RESIZED_REWRITE=1` to opt into the in-place rewrite for tight-disk jobs.
- Zero-copy mmap write: replace the `ctypes` / `bytes()` buffer copy in `_write_tensor_direct_torch` with a zero-copy `reshape(-1).view(torch.uint8)` + `memoryview` (the existing `.contiguous()` stays, guaranteeing the view shares storage). Halves peak RAM on large tensors and fixes a latent crash on 0-dim scalar tensors.
- Hoist per-shard-invariant lookups out of the merge loop; transfer the fused-MoE LoRA matrices to device once per layer instead of per expert.

## End-to-end correctness test suite

New `tests/test_merge_e2e_*.py` (plus `tests/_merge_e2e_helpers.py`) drive the real `merge_and_overwrite_lora` path on tiny but architecturally-real configs and verify every saved tensor against an independent reference: adapted modules equal `base + scale * (B @ A)`; every untargeted tensor (norms, router, biases, untargeted experts, vision tower) stays byte-identical to the base. This is the first coverage of the merge orchestration itself (key remap, sharding, pass-through, resized rewrite, MoE dispatch); existing tests only unit-test the merge primitives.

Coverage:
- Dense (Llama / Qwen3 / Mistral / Gemma2): full, attention-only, MLP-only, per-module alpha+rank patterns, bf16 output, forced multi-shard.
- Per-expert MoE (Qwen3-MoE, GLM4-MoE), including the no-expert-LoRA negative case proving experts stay byte-identical when untargeted.
- Fused-3D MoE (gpt-oss on 4.57.6 and 5.x; Qwen3.5 / Gemma4 / LFM2 gated to 5.x): grouped expert LoRA via `target_parameters`, per-expert reference with auto-detected orientation, distinct dims so a transpose bug cannot pass by luck.
- Vocab-grow / resized rewrite: embed/lm_head grow with old rows preserved.
- Vision (Gemma3): language-only LoRA merges exactly while the vision tower stays byte-identical.

Green on both CI transformers lanes (4.57.6 and 5.x). Families whose tiny-config key layout drifts across transformers versions skip with a clear reason rather than fail.

## Notes

- Verified output-neutral: running the same seeded merge under this branch vs `main` produces bit-identical merged tensors across dense, per-expert MoE, fused MoE, and resized scenarios.
- Touches the same file as #773 (VLM key-format remap). If #773 lands first this will rebase on top of it; the vision test already exercises the composite-model key remap path.

### DIFF
--- a/tests/_merge_e2e_helpers.py
+++ b/tests/_merge_e2e_helpers.py
@@ -1,0 +1,552 @@
+# Unsloth Zoo - Utilities for Unsloth
+# Copyright 2023-present Daniel Han-Chen, Michael Han-Chen & the Unsloth team. All rights reserved.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+"""Shared helpers for the end-to-end LoRA merge-to-16bit correctness suite.
+
+These helpers drive the real public merge path (`merge_and_overwrite_lora`) on
+tiny but architecturally-real models and check every saved tensor against an
+INDEPENDENT reference:
+
+    adapted module : merged == (base.float() + scale * (B @ A)).to(save_dtype)
+    pass-through   : merged is byte-identical to the base tensor
+
+The reference is built directly from the LoRA A/B/scaling read off the live PEFT
+model (not from the merge implementation), so it cannot mask a merge bug. Tiny
+configs use *distinct* hidden / intermediate sizes so fused-expert orientation is
+unambiguous and transpose bugs are visible.
+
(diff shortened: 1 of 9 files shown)


danielhanchen:
@codex review

PR #784, opened 2026-06-17
fix(mlx): align response-mask batching with CUDA

Lyxot:
## Summary

This improves MLX response-mask dataset parity with CUDA for both text and VLM training.

- Filter text samples whose labels become entirely `-100` after `train_on_responses_only`, avoiding batches with no trainable tokens.
- Apply the same fully-masked filtering to VLM response-mask datasets before batching, including unsized MLX VLM streaming batches.
- Preserve CUDA behavior for VLM prompt/completion rows by handling `completion_only_loss` separately instead of applying the response-mask closure again.
- Expose `completion_only_loss` on `MLXTrainingConfig` so direct MLX users can opt out of prompt masking for prompt/completion datasets.
- Keep response masking compatible with wrapped MLX/HF tokenizers, including processor/tokenizer wrappers.
- Forward `last_response_only` into the CUDA masking helper.
- Avoid double BOS insertion when text is already rendered with BOS, while preserving tokenizer-added BOS for raw text rows.
- Support dict eval datasets and record per-split eval metrics.
- Align VLM prompt/completion collation closer to CUDA, including prompt/completion split masking, padding-side handling, token type ids, truncation side, and CUDA-matching image extraction behavior.
- Match CUDA's VLM prompt/completion image extraction behavior: message-list rows use embedded images only, while plain prompt/completion rows follow CUDA's top-level `images` processing path and suppress extraction errors.

The loss / grad curve tracks the CUDA reference more closely on Gemma3 + Thytu/ChessInstruct over 30 steps.

图片

## Validation

- `python -m pytest tests/test_mlx_trainer_internals.py tests/test_mlx_vlm_label_masks.py -q -k 'not cce_backward_via_torch_autograd'`
  - `90 passed, 1 deselected`
- `python -m pytest tests/test_mlx_trainer_internals.py tests/test_mlx_vlm_label_masks.py tests/test_pr684_review_fixes_a.py -q -k 'not thread4 and not compiler_review_guards_are_present and not cce_backward_via_torch_autograd'`
  - `102 passed, 6 deselected`

### DIFF
--- a/tests/test_mlx_trainer_internals.py
+++ b/tests/test_mlx_trainer_internals.py
@@ -350,13 +350,13 @@ def test_evaluate_dict_eval_datasets_records_split_metrics():
     trainer.model = Model()
     trainer.stop_requested = False
 
-    def loss_fn(_model, name):
+    def loss_fn(_model, name, _lengths, _labels):
         if name == "small":
             return mx.array(1.0), mx.array(2)
         return mx.array(3.0), mx.array(6)
 
     loss, ppl = trainer._evaluate(
-        {"small": [("small",)], "large": [("large",)]},
+        {"small": [("small", None, None)], "large": [("large", None, None)]},
         loss_fn,
         is_vlm=False,
     )
--- a/tests/test_mlx_trainer_internals.py
+++ b/tests/test_mlx_trainer_internals.py
@@ -331,6 +331,31 @@ def test_train_on_responses_only_forwards_last_response_only(monkeypatch):
     assert received["last_response_only"] is True
 
 
+def test_response_mask_tokenizer_rejects_encode_only_tokenizer():
+    from unsloth_zoo.mlx.trainer import _resolve_response_mask_tokenizer
+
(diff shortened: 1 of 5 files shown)


Lyxot:
@codex review

danielhanchen:
Pushed a commit addressing two of the review points:

- Trainable-label filters (text and VLM) now check labels[1:] to match the causal shift the loss applies, so a row whose only unmasked token is the first position is filtered instead of slipping through as a zero-token batch.
- A raw string prompt paired with a chat-style completion is now coerced to a user message before combining, so mixed raw/chat prompt-completion rows render and split through one template instead of raising on None + list.

The remaining point about sequence-aligned prompt extras (for example Mllama cross_attention_mask or processor position_ids) only affects processors that emit per-token side tensors and is not exercised by the Qwen3-VL or Gemma3 paths here. I left that for a follow-up since dropping or recomputing those fields needs model-specific handling.

PR #785, opened 2026-06-17
fix(mlx): cast Gemma3 language MLP activation to fp32 to prevent backward NaN

BardiaKoopah:
## Summary

Mirrors `_fix_gemma3_vision_mlp_fp32_activation` for the language tower. Adds a sibling helper `_fix_gemma3_language_mlp_fp32_activation` that wraps `mlx_vlm.models.gemma3.language.MLP.__call__` to compute the GEGLU activation in fp32, then casts back to the source dtype before the multiply with `up_proj`.

## Bug

`gemma-3-4b-it` under multimodal LoRA training NaN's at every language layer 0-11 on the first backward pass, propagating to all upstream parameters and poisoning the run from step 2 onward.

## Why

`gelu_approx`'s tanh form is `0.5 * x * (1 + tanh(sqrt(2/pi) * (x + 0.044715 * x^3)))`. The `0.044715 * x^3` term's forward intermediates overflow in bf16 for the activation magnitudes that arise on the multimodal embed-then-decode path. The backward through `gelu_approx` then produces NaN gradients which propagate up the residual chain.

The existing vision-side patch covers the SigLIP MLP but doesn't address the language tower's separate GEGLU activation site.

## Fix

Cast `gate_proj`'s output to fp32 before `gelu_approx`, cast back to source dtype before the multiply with `up_proj`. All matmuls stay in native dtype — only the activation function runs in fp32.

## Verification

A/B on `gemma-3-4b-it` + `unsloth/LaTeX_OCR`, bs=1, seq=1024, language LoRA:

| | step 1 Grad | step 1 Loss | Avg loss (5 steps) | Peak mem |
|---|---|---|---|---|
| Vanilla | **nan** | 6.4850 | nan | 11.4 GB |
| Patched | **39.11** | 6.4751 | 4.91 | 11.7 GB |

Patched run loss trajectory: 6.48 → 5.02 → 4.80 → 3.94 → 4.32. Bit-identical across patched runs, confirming determinism.

Additional configs (seq=512, 4-bit):
- Vision + language LoRA combo: step-1 Grad=39.41, Avg=4.06, peak 11.87 GB ✓
- Vision-only LoRA: step-1 Grad=11.36, Avg=5.54, peak 11.51 GB ✓

### DIFF
--- a/unsloth_zoo/mlx/loader.py
+++ b/unsloth_zoo/mlx/loader.py
@@ -961,6 +961,50 @@ def _fix_gemma3_vision_mlp_fp32_activation(model=None):
     return True
 
 
+def _fix_gemma3_language_mlp_fp32_activation(model=None):
+    """Compute the language-tower GEGLU activation in fp32, then cast back.
+
+    Mirrors _fix_gemma3_vision_mlp_fp32_activation for the language tower.
+    Without this, the gelu_approx forward intermediates (x^3 term in the
+    tanh-form) overflow in bf16 / fp16 for some prompts and produce NaN
+    gradients in the backward pass — observed at language layers 0-11 of
+    gemma-3-4b-it during multimodal LoRA finetuning. The vision-side fix
+    above is not sufficient because the language MLP has its own activation
+    site that does not share that code path.
+    """
+
+    try:
+        import mlx.core as mx
+        import mlx.nn as nn
+        language_module = importlib.import_module("mlx_vlm.models.gemma3.language")
+    except Exception:
+        return False
+
+    mlp_cls = getattr(language_module, "MLP", None)
+    if mlp_cls is None:
+        return False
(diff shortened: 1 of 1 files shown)


danielhanchen:
Pushed a small commit to keep this scoped to the documented language MLP fix.

The vision encoder helper (`_fix_gemma3_vision_encoder_fp32_layernorm`) was also forcing the SigLIP self-attention to run fully in fp32, tagged `[BISECT-B]` and left over from bisection. That is unrelated to the language tower NaN and conflicts with the stated approach here of keeping matmuls in native dtype and only running the activation in fp32, so I reverted that hunk. The diff is now just the language GEGLU fp32 cast.

If the vision attention fp32 turns out to be needed, please send it as its own PR with before/after numbers so it can be reviewed on its own merits.

Verified on an Apple Silicon runner that the language MLP patch applies cleanly against mlx_vlm and the MLX training smoke stays green.

PR #786, opened 2026-06-17
test(mlx): verify VLM resume_from_checkpoint matches fresh run step-for-step

BardiaKoopah:
## Summary

PR #684's full-validation test suite has a text-only resume test (`test_resume_from_checkpoint_matches_fresh_run`) but no equivalent for VLMs, even though #751 implemented the resume path for any model type going through `MLXTrainer`. This adds the missing VLM test alongside the existing text one.

## Coverage

The new test mirrors the text version exactly:

1. Train Qwen2-VL-2B for 6 steps with `save_steps=3`
2. Verify `trainer_state.json`, `optimizer_state.safetensors`, and `adapters.safetensors` all wrote at `checkpoint-3` with `global_step=3`
3. Construct a fresh `MLXTrainer` instance, resume from `checkpoint-3`, train to step 6
4. Assert per-step losses match the fresh run within 1e-5 relative tolerance

Exercises the resume path through the multimodal collator and image processor in addition to the text-only path, catching VLM-specific regressions the existing text test would miss.

## Refactor

Two small helpers extracted to avoid duplication between the existing `test_vlm_lora_training_e2e` and the new resume test:

- `_color_square_dataset(colors)` — synthetic VLM dataset of solid-color squares
- `_make_vlm_trainer(tmp_path, dataset, **overrides)` — VLM trainer factory mirroring the existing `_make_text_trainer`

The existing VLM e2e test was left as-is to keep the diff scoped to a test addition rather than a refactor. The new helpers are used by the new test only.

## Verification

Ran locally on M2 16GB:

    ============================== 1 passed, 5 warnings in 34.95s ===============================

Fresh run losses: 1.9341, 1.7966, 1.5733 (steps 4-6)
Resumed run losses: 1.9341, 1.7966, 1.5733 (steps 4-6)

Byte-identical to display precision; passed the 1e-5 relative-tolerance assertion.

## Why this matters

The resume mechanics in #751 were verified on text-only models because the trainer is model-agnostic at the resume layer. But the VLM data path is genuinely separate (image processor state, multimodal collator iterator, `create_vlm_batches` / `iterate_vlm_training_batches`). A future change that breaks any of those without breaking text resume would currently pass CI; this test closes that gap.

### DIFF
--- a/tests/test_mlx_pr684_full_validation_metal.py
+++ b/tests/test_mlx_pr684_full_validation_metal.py
@@ -228,3 +228,105 @@ def test_vlm_lora_training_e2e(tmp_path):
     _assert_finite(hist)
     saved = glob.glob(os.path.join(str(tmp_path), "**", "*.safetensors"), recursive=True)
     assert saved, "no adapter safetensors saved at end of VLM training"
+
+
+def _color_square_dataset(colors):
+    """Synthetic VLM dataset: one solid-color 64x64 square per message."""
+    from PIL import Image
+    return [
+        {
+            "messages": [
+                {
+                    "role": "user",
+                    "content": [
+                        {"type": "image"},
+                        {"type": "text", "text": "What color is this square?"},
+                    ],
+                },
+                {
+                    "role": "assistant",
+                    "content": [{"type": "text", "text": f"This square is {color}."}],
+                },
+            ],
+            "images": [Image.new("RGB", (64, 64), color)],
+        }
+        for color in colors
+    ]
+
+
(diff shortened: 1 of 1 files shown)


danielhanchen:
Applied both review suggestions to the new VLM resume test: switched the checkpoint path assertions to pathlib, and added del trainer plus gc.collect() before loading the resumed model so memory constrained runners (for example 16GB Macs) do not hold two 2B models at once. Left the existing text resume test unchanged to keep the diff scoped to the new test.

PR #789, opened 2026-06-18
fix(mlx): detect bitsandbytes models early on Apple Silicon

BardiaKoopah:
## Summary

bitsandbytes-quantized models cannot load on Apple Silicon — bnb's NF4 sidecar tensors (`.absmax`, `.quant_state.bitsandbytes__nf4`, etc.) need the bitsandbytes runtime for dequantization, and bnb has no Metal backend.

Before this patch, users hit one of two opaque errors:
- `load_in_4bit=True` → "per-module MLX quantization metadata ... but requested affine" (talks about quant configs, never mentions bnb)
- `load_in_4bit=False` → "Received N parameters not in model: ...absmax, ...nested_absmax, ...quant_state.bitsandbytes__nf4..." (cryptic, surfaces internal NF4 sidecar names)

Both errors fired AFTER downloading multi-gigabyte weights.

## Fix

Reads `config.json`'s `quant_method` field early in `from_pretrained` and raises a clear error suggesting the non-bnb variant. Detection happens before any weight download, so the failure is near-instant (~0.1s).

## Behavioral note

This guard also fires when `force_requantize=True` is passed on a bnb model. The previous `load_in_4bit=True` error pointed users at `force_requantize` as an escape hatch, but that path never actually succeeded on Mac (bnb NF4 weights cannot be dequantized without the bitsandbytes runtime), so surfacing the underlying reason is strictly clearer.

## Related

Complements the existing frontend filter that greys out bnb models in the Studio picker. Defense in depth: frontend prevents most users from selecting these models; loader catches direct API callers and anything that bypasses the picker.

## Verification

12-case harness on M2 across `{non-quant, bnb-4bit, already-MLX-4bit}` × `{text, VLM}` × `{load_in_4bit True, False}`:
- All 4 bnb cases now hit the clear error in ~0.1s
- The 8 non-bnb cases are unchanged

Edge-case checks:
- non-bnb models don't have `quant_method == "bitsandbytes"` (verified across 4 sampled models from `unsloth/` and `mlx-community/`)
- Suggested variant names resolve to real HF repos (verified for `Qwen3-VL-4B-Thinking-unsloth-bnb-4bit` → `Qwen3-VL-4B-Thinking` and similar)
- The `-bnb-Xbit` suffix regex doesn't false-positive on names where the pattern appears mid-string (e.g. `model-bnb-4bit-fp16`)

### DIFF
--- a/unsloth_zoo/mlx/loader.py
+++ b/unsloth_zoo/mlx/loader.py
@@ -25,6 +25,7 @@ import importlib
 import inspect
 import math
 import os
+import re
 import tempfile
 import types
 import warnings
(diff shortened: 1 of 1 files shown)


danielhanchen:
Thanks for this. The same early bitsandbytes detection (same quant_method check, the -bnb-4bit suffix strip, and the Apple Silicon error message) has since landed on main via #815, which goes a step further and dequantizes the bnb weights into a working MLX 4-bit model rather than only rejecting, and also handles LoRA adapter directories and VLM repos as separate cases. Since main already covers this path and then some, I am closing this as superseded. Appreciate the contribution.

PR #795, opened 2026-06-19
training_utils: respect PEFT bias training in prepare_model_for_training (#2343)

danielhanchen:
### Summary

Fixes unslothai/unsloth#2343: bias training is silently disabled. `prepare_model_for_training`
froze every parameter that was not a LoRA adapter weight:

```python
if ".lora_A." in name or ".lora_B." in name or ".lora_magnitude_vector" in name:
    requires_grad = True
else:
    requires_grad = False   # also froze biases PEFT had marked trainable
```

So `LoraConfig(bias="all")` / `bias="lora_only"` had no effect: PEFT enables the biases, then this
loop re-freezes them. The reporter saw the trainable-parameter count stay identical regardless of the
`bias` setting, which I reproduced on a tiny model:

```
bias='none'  total_bias=14  trainable_after_PEFT= 0  trainable_after_prepare= 0
bias='all'   total_bias=14  trainable_after_PEFT=14  trainable_after_prepare= 0   <- bug
```

### Change

Keep PEFT's decision for parameters whose name ends in `.bias`: preserve their incoming
`requires_grad` and upcast the trainable ones to float32 like the LoRA adapters. `bias="none"` (the
default) leaves biases frozen, so the common LoRA / QLoRA path is byte-identical. Full finetuning is
untouched (the change is inside the `if not full_finetuning:` branch).

After the fix:

```
bias='all'   total_bias=14  trainable_after_PEFT=14  trainable_after_prepare=14   <- fixed
bias='none'  total_bias=14  trainable_after_PEFT= 0  trainable_after_prepare= 0   <- unchanged
```

### Tests

`tests/test_prepare_model_for_training_bias.py` (pure CPU, tiny random Llama):

- marked biases stay trainable, default biases stay frozen, the non-bias base weights stay frozen;
- a trainable bias is upcast to fp32 while a frozen one keeps its loaded dtype;
- a real forward / backward populates the trainable bias `.grad` and leaves a frozen weight with none;
- a PEFT end-to-end check for `bias="all"` / `bias="none"` (gated on peft being importable, since peft
  is excluded on darwin/arm64).

Existing `prepare_model_for_training` suites still pass locally (the fp32-norms, use-cache, and
upstream source-pattern / signature tests), so there is no regression to the LoRA, full-FT, or norm
paths.

### DIFF
--- a/tests/test_prepare_model_for_training_bias.py
+++ b/tests/test_prepare_model_for_training_bias.py
@@ -1,0 +1,200 @@
+"""Regression tests for unsloth#2343: bias training was silently disabled.
+
+Pre-fix bug: `prepare_model_for_training`'s LoRA branch froze every parameter
+that was not a LoRA adapter weight:
+
+    if ".lora_A." in name or ".lora_B." in name or ".lora_magnitude_vector" in name:
+        requires_grad = True
+    else:
+        requires_grad = False   # <- also froze biases PEFT had marked trainable
+
+So `LoraConfig(bias="all")` / `bias="lora_only"` had no effect: PEFT enabled the
+biases, then this loop re-froze them. The fix keeps the PEFT decision for params
+whose name ends in `.bias` (preserve their incoming requires_grad and upcast the
+trainable ones to fp32 like LoRA adapters). `bias="none"` (the default) leaves
+biases frozen, so the common path is byte-identical.
+
+Pure CPU, tiny random Llama. The PEFT end-to-end test is gated on peft being
+importable (peft is excluded on darwin/arm64).
+"""
+import os
+
+import pytest
+import torch
+from transformers import LlamaConfig, LlamaForCausalLM
+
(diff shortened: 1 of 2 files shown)


danielhanchen:
@codex review

oobabooga:
Ran the suite against a real tiny Llama and peft, all 6 pass: bias="all" stays trainable, bias="none" stays frozen, non-PEFT biases stay frozen, the trainable bias keeps its loaded dtype, grad flows, and modules_to_save is handled consistently (both weight and bias via the patch_modules_to_save block). Minor: the PR description still says biases are upcast to fp32 "like the LoRA adapters", but the code (correctly) keeps the loaded dtype to avoid the bf16-weight/fp32-bias matmul mismatch. Worth updating the text. LGTM ✅

danielhanchen:
@codex review

PR #796, opened 2026-06-19
Fix MLX GGUF export on macOS (import, apt-get fallback, optimized build)

danielhanchen:
## Summary

GGUF export from an MLX-trained model (`save_pretrained_gguf` in `unsloth_zoo/mlx/utils.py`) was broken on a fresh macOS machine where llama.cpp is not already present. Three issues stacked up on the install path:

1. **`NameError: name 'install_llama_cpp' is not defined`.** When `check_llama_cpp` raises (no llama.cpp yet), the function falls back to `install_llama_cpp`, but that name was never added to the `from ..llama_cpp import (...)` block, so the fallback crashed immediately.

2. **macOS hit the Linux build-dependency path.** After fixing (1), `install_llama_cpp` resolves missing system build-deps through `apt-get`, which does not exist on macOS, so it aborted with `RuntimeError: [FAIL] Unsloth: apt-get does not exist? Is this NOT a Linux / Mac based computer?`.

3. **The macOS source build was unoptimized.** `_install_llama_cpp_macos` configured cmake without `-DCMAKE_BUILD_TYPE=Release`. macOS uses a single-config generator, so `--config Release` on the build step is ignored and the binaries are built with no optimization, making `llama-cli` inference extremely slow.

LoRA and merged-16bit exports were never affected (they do not touch llama.cpp).

## Fix

1. Add `install_llama_cpp` to the existing `..llama_cpp` import block (it is already part of `llama_cpp.__all__` next to `check_llama_cpp`).

2. On macOS, route the install fallback to `_install_llama_cpp_macos` (a cmake + Metal source builder already defined in `mlx/utils.py` but never called) instead of the apt-get-based generic installer, then re-probe with `check_llama_cpp`.

3. Set `-DCMAKE_BUILD_TYPE=Release` in `_install_llama_cpp_macos` so the built binaries are optimized.

## Validation

Verified on a real Apple Silicon runner (macOS 14, M1) by training `unsloth/gemma-3-270m-it` with MLX LoRA and exporting all three formats end to end:

- LoRA and merged-16bit export -> reload in a fresh process -> generation matches the in-memory model (post-train loss 0.0003).
- GGUF: `save_gguf` builds llama.cpp via cmake on first run and writes a real GGUF (`gemma-3-270m-it.BF16.gguf`, 517 MB, 236 tensors); the file parses back as a valid GGUF.

Before this change, the macOS GGUF phase logged `GGUF SKIPPED: NameError ...`, then (after the import-only fix) `apt-get does not exist ...`; `gguf_supported` was `false`.

### DIFF
--- a/tests/test_llama_cpp_prebuilt.py
+++ b/tests/test_llama_cpp_prebuilt.py
@@ -380,6 +380,39 @@ def test_extract_and_place_linux(tmp_path):
     assert (install / "libggml-base.so").is_file()
 
 
+def _make_binary_zip(path, tag):
+    import zipfile
+    with zipfile.ZipFile(path, "w") as z:
+        z.writestr(f"llama-{tag}/llama-quantize.exe", b"MZ fake exe")
+        z.writestr(f"llama-{tag}/llama-cli.exe", b"MZ fake exe")
+        z.writestr(f"llama-{tag}/ggml.dll", b"fake dll")
+        z.writestr(f"llama-{tag}/llama.dll", b"fake dll")
+        z.writestr(f"llama-{tag}/LICENSE", b"MIT-ish")
+
+
+def test_extract_and_place_windows_zip(monkeypatch, tmp_path):
+    # The fork/ggml Windows bundle is a .zip; _place_prebuilt_binaries must land
+    # the executables + DLLs under build/bin/Release (where check_llama_cpp looks
+    # on Windows). Exercise the Windows placement on any host by patching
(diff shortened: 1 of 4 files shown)


danielhanchen:
@codex review

danielhanchen:
@codex review

danielhanchen:
@codex review

danielhanchen:
@codex review

danielhanchen:
@codex review

danielhanchen:
@codex review

danielhanchen:
@codex review

PR #799, opened 2026-06-20
vision_utils: actionable error when the processor has no chat template

danielhanchen:
### Summary

When a base (non instruction tuned) vision checkpoint has no chat template (for example `unsloth/Llama-3.2-11B-Vision`, the non-Instruct base), `UnslothVisionDataCollator` raised a bare error from `apply_chat_template`:

```
RuntimeError: Cannot use apply_chat_template ...
```

This message does not name the model or tell the user how to fix it.

### Change

In `unsloth_zoo/vision_utils.py`:

- Add `_resolve_processor_model_name` and `_raise_chat_template_error` helpers before the `UnslothVisionDataCollator` class.
- Replace the two bare `raise RuntimeError(e) from e` handlers in `UnslothVisionDataCollator.__init__` with `_raise_chat_template_error(e, processor, model)`.

The new message names the model and tells the user to use an `-Instruct` / `-it` checkpoint or set `tokenizer.chat_template`, while preserving the original error text.

### Behavior

- Only the chat-template error is special-cased; the message includes the resolved model name and the original error.
- All other errors are re-raised unchanged via `raise RuntimeError(error) from error`.

### DIFF
--- a/unsloth_zoo/vision_utils.py
+++ b/unsloth_zoo/vision_utils.py
@@ -682,6 +682,42 @@ import PIL.Image
 LANCZOS = PIL.Image.Resampling.LANCZOS
 from .dataset_utils import train_on_responses_only as _train_on_responses_only
 
+
+def _resolve_processor_model_name(processor, model = None):
+    """Best-effort human readable name for the processor / model in errors."""
+    for obj in (processor, getattr(model, "config", None), model):
+        if obj is None:
+            continue
+        for attr in ("name_or_path", "_name_or_path"):
+            name = getattr(obj, attr, None)
+            if isinstance(name, str) and len(name) > 0:
+                return name
+    if processor is not None:
+        return processor.__class__.__name__
+    return "this model"
+
+
+def _raise_chat_template_error(error, processor, model = None):
+    """Turn an apply_chat_template failure into an actionable Unsloth error.
+
+    The most common cause is a base (non instruction tuned) checkpoint whose
+    processor has no chat template, so vision conversations cannot be formatted.
+    """
+    name = _resolve_processor_model_name(processor, model)
+    msg = str(error)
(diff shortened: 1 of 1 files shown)


danielhanchen:
@codex review

danielhanchen:
Follow-up commit: the recovery hint now points users at `processor.chat_template`, since the collator calls `processor.apply_chat_template` and setting `tokenizer.chat_template` on a separate tokenizer would not fix the failure.

danielhanchen:
@codex review

danielhanchen:
Checked this against the actual transformers error and it is a false positive for the path this branch handles. _raise_chat_template_error only wraps processor.apply_chat_template (the vision collator path), and that raises:

  "Cannot use apply_chat_template because this processor does not have a chat template."

which contains the spaced phrase "chat template", so `"chat template" in msg.lower()` matches. The underscore wording you quoted ("...because tokenizer.chat_template is not set") is the tokenizer-only message, which is not what reaches this branch. I confirmed the processor wording is stable across transformers 4.52 -> 5.12.1 (every release I checked still contains "chat template"). Leaving the predicate as-is.

danielhanchen:
@codex review

danielhanchen:
Addressed in b22d4d7. The current message did already contain the space form (`...does not have a chat template.` / `Cannot use chat template functions because tokenizer.chat_template is not set...`), but older Transformers raised `Cannot use apply_chat_template() because tokenizer.chat_template is not set...`, which only has the underscore form. The match now also catches `chat_template` (which covers `apply_chat_template` too), so those older paths get the actionable guidance instead of the bare upstream error. The helper is only reached from inside an `apply_chat_template` failure handler, so broadening the substring cannot misfire on unrelated errors.

PR #801, opened 2026-06-20
gpt-oss: fix bnb-4bit<->16bit compiled-cache mode switch breaking merged_16bit reload

danielhanchen:
## Problem

Fine-tuning `unsloth/gpt-oss-20b-unsloth-bnb-4bit` in 4-bit, saving with `save_pretrained_merged(save_method="merged_16bit")`, then reloading the merged checkpoint failed with:

```
Unsloth: Critical error since some weights are not initialized.
... Some weights of the model checkpoint were not used when initializing GptOssForCausalLM:
['model.layers.0.mlp.router.bias', 'model.layers.0.mlp.router.weight', ...]
```

A bnb-4bit gpt-oss load could also fail at load time with `AttributeError: GptOssExperts has no attribute 'down_projs'`.

## Root cause

The compiled `unsloth_compiled_module_gpt_oss.py` is a single per-model-type file that hardcodes **either** the BnB layout (`router.linear.weight` + `ModuleList` experts) **or** the stock layout (`router.weight` + fused 3D experts), depending on which classes were active when it was generated.

A bnb-4bit load swaps `transformers.models.gpt_oss.modeling_gpt_oss.GptOssTopKRouter` / `GptOssExperts` to the BnB variants and writes a BnB-flavoured compiled module. When a stock 16-bit checkpoint (e.g. a `merged_16bit` export of a 4-bit-trained model) is loaded afterwards, the stale BnB-flavoured compiled module re-installs the nested-linear router, so the checkpoint's flat `router.weight`/`router.bias` are "not used" and transformers raises the critical error.

## Fix

`patch_gpt_oss_bnb4bit_auto` now:
- restores the stock classes and invalidates the cached compiled module when the current load is **not** bnb-4bit (so a stale BnB swap from an earlier 4-bit load does not break a later 16-bit reload), and
- invalidates a stale stock-built compiled module the first time it switches **into** bnb mode.

`_invalidate_gpt_oss_compiled_module()` drops both the in-memory `sys.modules` entry and the on-disk cached `.py` so the next compile regenerates against the currently-active classes.

This pairs with unslothai/unsloth (loader: build `UNSLOTH_MODEL_NAME` fresh per load) which stops a stale `_load_in_4bit_` flag from leaking across save->reload subprocesses.

## Verification

End-to-end on gpt-oss-20b (B200, transformers 4.57.6), train -> save_lora + save_merged_16bit -> reload in a fresh process, all reproducing the fine-tuned phrase:

| precision | load | train | reload_lora | reload_merged_16bit |
|---|---|---|---|---|
| 4-bit (`-unsloth-bnb-4bit`) | OK | loss -> 0.0001 | reproduces phrase | reproduces phrase |
| 16-bit (`-BF16`) | OK | loss -> 0.0002 | reproduces phrase | reproduces phrase (MoE 48/48 experts merged) |

The exact merged checkpoint that previously raised "weights are not initialized" now reloads cleanly and generates the trained answer. The 16-bit path is unchanged (no regression).

## Tests

Adds two regressions to `tests/test_zoo_history_regressions.py`:
- `test_invalidate_gpt_oss_compiled_module_drops_sys_modules_and_file` (functional, no GPU).
- `test_gpt_oss_bnb4bit_auto_restores_and_invalidates_on_non_4bit_load` (AST contract). Both fail on the pre-fix function and pass on the fix.

## Known limitation

A single process that loads 4-bit **and then** reloads a 16-bit merged checkpoint without a fresh interpreter still mismatches, because the already-patched transformers `GptOssForCausalLM`/`GptOssMLP` class objects persist. The realistic notebook flow (reload in a new session/process) is fully covered.

### DIFF
--- a/tests/test_zoo_history_regressions.py
+++ b/tests/test_zoo_history_regressions.py
@@ -477,26 +477,16 @@ def test_patch_compiling_bitsandbytes_missing_peft_helpful_error(monkeypatch):
 
 # ---------------------------------------------------------------------------
 # Regression: gpt-oss bnb-4bit <-> 16bit compiled-cache mode switch.
-#
-# The compiled `unsloth_compiled_module_gpt_oss.py` is a single per-model-type
-# file that hardcodes EITHER the BnB layout (router.linear.weight + ModuleList
-# experts) OR the stock layout (router.weight + fused 3D experts), depending on
-# which classes were active when it was generated. A bnb-4bit load swaps the
-# transformers GptOssTopKRouter/GptOssExperts globals and writes a BnB-flavoured
-# compiled module. Reusing that stale module when later loading a stock 16bit
-# checkpoint (e.g. a merged_16bit export of a 4bit-trained model) re-installs the
-# BnB router, so the checkpoint's router.weight/.bias are "not used" and
-# transformers raises "Unsloth: Critical error since some weights are not
-# initialized". `patch_gpt_oss_bnb4bit_auto` must (a) restore the stock classes
-# and invalidate the compiled module when the current load is NOT bnb-4bit, and
-# (b) invalidate a stale stock-built module the first time it switches INTO bnb.
+# The compiled gpt_oss module hardcodes the BnB or stock router/experts layout. Reusing a
+# stale BnB-built module for a later 16bit load re-installs the BnB router -> "weights not
+# initialized". patch_gpt_oss_bnb4bit_auto must restore stock classes + invalidate the module
+# when the load is not bnb-4bit, and drop a stale stock module when switching into bnb.
 # ---------------------------------------------------------------------------
 
 
 def test_invalidate_gpt_oss_compiled_module_drops_sys_modules_and_file(tmp_path, monkeypatch):
-    """`_invalidate_gpt_oss_compiled_module` must remove both the in-memory
-    `unsloth_compiled_module_gpt_oss` entry and the on-disk cached .py so the
-    next compile regenerates against the currently-active classes."""
+    """_invalidate_gpt_oss_compiled_module removes the sys.modules entry and the on-disk .py so
+    the next compile regenerates against the active classes."""
     import sys
     import types
     from unsloth_zoo.temporary_patches import gpt_oss as _M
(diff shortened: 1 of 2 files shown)


danielhanchen:
Coupled with unslothai/unsloth#6504. Both changes are required together to fix the gpt-oss bnb-4bit -> merged_16bit reload: this PR restores the stock router/experts and invalidates the compiled gpt_oss module on the 4bit<->16bit switch, and unsloth#6504 stops UNSLOTH_MODEL_NAME from carrying a stale _load_in_4bit_ flag across loads. Please merge as a pair.

danielhanchen:
Follow-up commit addressing the cross-process gap raised in review: the on-disk compiled gpt_oss module is now invalidated via a flavor marker independent of the in-process UNSLOTH_GPT_OSS_BNB4BIT_PATCHED flag (which is unset in a fresh process), so a fresh 16bit load after a 4bit load drops the stale BnB-built module instead of importing the wrong router/experts layout. Also removes the .pyc via cache_from_source + importlib.invalidate_caches() and covers the temp-fallback cache dir. Added a regression test.

danielhanchen:
@codex review

danielhanchen:
Follow-up commit addressing the codex review of the flavor-marker change:
- Gated the flavor sync to GPT-OSS loads, so loading an unrelated model no longer syncs the cache to 'stock' and cannot delete a valid bnb4bit GPT-OSS compiled module.
- Removed the unconditional invalidate on the first bnb load of a process (the in-process flag is process-local, so it deleted a matching cache every fresh session). _sync_gpt_oss_compiled_flavor is now the sole invalidator and only drops the module on a real flavor mismatch, preserving a matching cache.
Added a regression test that an unrelated model load leaves the GPT-OSS cache intact.

danielhanchen:
@codex review

danielhanchen:
This is handled in combination with unslothai/unsloth#6504, which is the companion change that builds UNSLOTH_MODEL_NAME from the model TYPE tokens (get_transformers_model_type) plus load flags rather than the raw path. So for a merged gpt-oss reloaded from a user-named dir like ./merged_16bit, the name still contains "gpt_oss" (from config.model_type), and this gate runs the flavor sync correctly regardless of the directory name. With #6504 also now encoding the effective bnb state, the bnb4bit/stock decision matches the actual checkpoint. The two PRs are meant to land together.

danielhanchen:
@codex review

PR #805, opened 2026-06-20
Fix GRPO eval crash: keep gradient checkpointing buffers out of inference mode

danielhanchen:
## Bug

GRPO training with `use_gradient_checkpointing = "unsloth"` crashes when evaluation is enabled (and intermittently during training), with:

```
RuntimeError: Inplace update to inference tensor outside InferenceMode is not allowed.
You can make a clone to get a normal tensor before doing inplace update.
```

The traceback ends at the Unsloth smart gradient checkpointing offload copy:

```
unsloth_zoo/gradient_checkpointing.py, in forward
    x.copy_(arg, non_blocking = True)
RuntimeError: Inplace update to inference tensor outside InferenceMode is not allowed.
```

The reporter noted it does not happen on every evaluation, just randomly, and that it only happens when evaluation is on. See unslothai/unsloth#3828.

## Root cause

GRPO computes the reference / old per-token log-probs under `torch.inference_mode` (`unsloth/models/rl_replacements.py` `_get_per_token_logps`, via `_get_inference_mode_context_manager`). The model is still in training mode at that point (GC enabled), so a smart-GC forward can run while that inference context is active.

The smart-GC offload uses persistent global CPU and GPU buffers. If one of those buffers is allocated (`torch.empty(...)`) or copied into (`x.copy_(arg)`) while the `inference_mode` context is active, the buffer permanently becomes an inference tensor. A later, normal training / eval offload pass then runs `x.copy_(arg, non_blocking=True)` on that inference-tensor buffer outside inference mode and raises.

This explains the intermittency: it only triggers when a buffer allocation or copy happens to land inside the `inference_mode` window, and only on the offload path (hidden state larger than the 2MB minimum).

Minimal isolated reproduction:

```python
import torch
with torch.inference_mode():
    x = torch.empty(1024, device="cpu", pin_memory=True)   # becomes an inference tensor
arg = torch.randn(1024, device="cuda")
x.copy_(arg)   # RuntimeError: Inplace update to inference tensor ...
```

## Fix

Add a small `_no_inference_mode()` context manager (`torch.inference_mode(False)` plus `torch.no_grad()`) and wrap every persistent buffer allocation (`initialize_unsloth_gradient_checkpointing`, the in-forward buffer extension, and `reset_unsloth_gradient_checkpointing_buffers`) and every inplace buffer `copy_` (forward CPU offload and backward GPU restore) with it.

- Buffers are now always normal tensors regardless of the caller's context.
- The offload copies never inplace-update an inference tensor.
- `no_grad` is kept because the offload copies are pure data movement that must not enter grad mode (otherwise it hits "a view created in `no_grad` is modified in grad mode").
- Older torch without a boolean arg to `inference_mode` falls back to plain `no_grad`.

## Backwards compatibility

- Does not touch the GRPO training-mode GC handling from #5269 (`for_training` / `unsloth_unwrap_model_for_generation` / `unsloth_fast_generate`). That fix governs which GC mode is restored after generation; this fix is purely about not tainting the offload buffers with inference state.
- Gradients are bit-identical to the non-checkpointed reference (max diff `0.0`), so training dynamics are unchanged.
- The OOM / double-buffer `except` handling around buffer allocation is preserved.
- All GC modes (`"unsloth"`, `True`, `False`), `eval_strategy` on and off, and vLLM on and off were exercised.

## Before / after

Deterministic unit test that allocates the GC buffers inside `inference_mode` (mirroring the GRPO log-prob pass) and then runs a normal offload forward:

| State    | CPU buffers `is_inference` | Offload forward `x.copy_(arg)` |
|----------|----------------------------|--------------------------------|
| Before   | `[True, True, True, True]`  | crash: Inplace update to inference tensor |
| After    | `[False, False, False, False]` | OK |

End-to-end GRPO (Qwen2.5-1.5B, LoRA, `use_gradient_checkpointing="unsloth"`, `beta=0.04`, `eval_strategy="steps"`, ~500 token completions so the offload path is active): training and eval both complete, the "Will smartly offload gradients to save VRAM" path is active throughout, and final `evaluate()` returns metrics with no crash.

## Test matrix

| GC mode      | eval | vLLM | Result |
|--------------|------|------|--------|
| `"unsloth"`  | on   | off  | pass   |
| `"unsloth"`  | on   | on   | pass   |
| `"unsloth"`  | off  | off  | pass (memory savings preserved) |
| `True`       | on   | off  | pass   |
| `False`      | on   | off  | pass   |
| `"unsloth"` (large model, offload path active) | on | off | pass |
| GC fwd+bwd vs reference grads | - | - | max diff 0.0 |

Fixes unslothai/unsloth#3828

### DIFF
--- a/unsloth_zoo/gradient_checkpointing.py
+++ b/unsloth_zoo/gradient_checkpointing.py
@@ -67,7 +67,7 @@ def _no_inference_mode():
     # no_grad avoids the "view created in no_grad modified in grad mode" error.
     try:
         leave_inference = torch.inference_mode(False)
-    except TypeError:
+    except (TypeError, AttributeError):
         # Older torch without inference_mode(bool); no_grad alone is enough.
         leave_inference = nullcontext()
     with leave_inference, torch.no_grad():
--- a/unsloth_zoo/gradient_checkpointing.py
+++ b/unsloth_zoo/gradient_checkpointing.py
@@ -681,7 +681,7 @@ class UnslothCheckpointFunction(torch.autograd.Function):
 
             device_autocast_ctx = torch.amp.autocast(
                 device_type=ctx.device_type, **ctx.device_autocast_kwargs
-            ) if torch.amp.is_autocast_available(ctx.device_type) else contextlib.nullcontext()
+            ) if torch.amp.is_autocast_available(ctx.device_type) else nullcontext()
 
             # detached_inputs = detach_variable(tuple(inputs))
             detached_inputs = []
--- a/unsloth_zoo/gradient_checkpointing.py
+++ b/unsloth_zoo/gradient_checkpointing.py
@@ -59,8 +59,8 @@ DOUBLE_BUFFER_HEADROOM = 512 * 1024 * 1024 # min free CUDA memory to enable doub
 @contextmanager
 def _no_inference_mode():
     # GC buffers must not be inference tensors, else a later offload copy_ raises
-    # "Inplace update to inference tensor" (GRPO log-probs run under inference_mode,
-    # unsloth#3828). Leave inference_mode but stay in no_grad for the copies.
+    # "Inplace update to inference tensor" under GRPO's inference_mode (unsloth#3828).
+    # Leave inference_mode but stay in no_grad.
     try:
         leave_inference = torch.inference_mode(False)
     except (TypeError, AttributeError):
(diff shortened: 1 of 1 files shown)


danielhanchen:
@codex review

oobabooga:
Live CUDA repro: a GC buffer allocated under inference_mode becomes an inference tensor and the later copy_ raises "Inplace update to inference tensor outside InferenceMode", exactly as in #3828. With _no_inference_mode() the buffers come back as normal tensors (is_inference()=False) and copy_ succeeds (CPU and GPU). Also checked the real forward path (buffer view created inside inference_mode then wrapped copy_) works, and swept the file: all buffer allocs (5) and inplace copy_ (2) are wrapped. LGTM ✅

danielhanchen:
@codex review

PR #806, opened 2026-06-20
Fix LoRA scaling count mismatch on merge for Qwen2.5-VL exports (#2966)

danielhanchen:
## Summary

Fixes the empty `AssertionError` when saving / exporting a finetuned Qwen2.5-VL model via `save_pretrained_merged`, reported in unslothai/unsloth#2966.

The traceback ends at:

```
File ".../unsloth_zoo/saving_utils.py", line 313, in create_lora_statistics
    assert(module_count == lora_A_count == lora_B_count == scaling_count)
AssertionError:
```

## Root cause

`create_lora_statistics` counts each LoRA layer four ways while walking the model:

- `lora_A_count`, `lora_B_count`, `module_count` are incremented by name suffix (`.lora_A.default`, `.lora_B.default`, `.base_layer`), so they fire for any LoRA-wrapped module.
- `scaling_count` was only incremented when the module is an instance of one of the peft `Linear*` classes discovered by `get_lora_layer_modules()` (it scans `peft.tuners.lora.*` for classes named `Linear*` / `*Linear`).

When a peft build, quant backend (HQQ / EETQ expose no `Linear*`-named LoRA class), or a vision tower wraps a layer in a class that scan does not surface, or the wrapper only exposes the singular `active_adapter` instead of `active_adapters`, `scaling_count` falls short of the other three counters. The four counts disagree, so:

- on older builds the bare `assert` raised the empty `AssertionError` in the issue, and
- after the assert was softened to a warning, the merge proceeds but those layers keep the default `alpha = 0`, so their LoRA delta is silently dropped from the merged weights (`W.addmm_(lora_B, lora_A, alpha=0)`).

## Fix

- Add a small `_get_lora_scaling(module)` helper that resolves the active adapter from either `active_adapters` (plural, current peft) or `active_adapter` (singular, older / custom wrappers) and returns the scaling, falling back to `0.0` only if it genuinely cannot be resolved.
- Use it from the strict `Linear_LoRA_Layers` branch and broaden the fallback branch to match on `scaling` plus either active-adapter attribute, so every real LoRA layer is counted and merged with its true alpha regardless of its wrapper class or peft version.

This keeps `scaling_count` aligned with the suffix-based counters and removes the silent `alpha = 0` drop. Non-LoRA modules are untouched (they carry no `scaling` dict), so dense / MoE / text / other vision models are unaffected.

## Testing

Reproduced on a B200 with transformers 5.x, peft 0.18.1, bnb-4bit Qwen2.5-VL:

- Forced the class-discovery gap (bnb LoRA class hidden) on a real Qwen2.5-VL adapter: before the fix `create_lora_statistics` reports `modules=412, lora_A=412, lora_B=412, scaling=0` and every captured `alpha` is `0`; after the fix all 412 entries align and carry their true alpha.
- A minimal layout mirroring the issue (LoRA wrapper not in `Linear_LoRA_Layers`, singular `active_adapter`): before -> empty `AssertionError` / `alpha = 0`; after -> counts align, `alpha` correct.
- End-to-end `save_pretrained_merged` on Qwen2.5-VL-3B succeeds.
- Text model + `use_rslora=True` (the dynamic-scaling case raised in the thread): 168 LoRA entries, all alphas captured at `16/sqrt(8)`.

Added CPU-only unit tests in `tests/test_saving_utils_lora_remap_count.py` covering `_get_lora_scaling` (plural / singular / unresolved) and count alignment for a class-mismatched adapter. Full file: 38 passed.

Fixes unslothai/unsloth#2966

### DIFF
--- a/tests/test_saving_utils_lora_remap_count.py
+++ b/tests/test_saving_utils_lora_remap_count.py
@@ -36,7 +36,12 @@ from unsloth_zoo.saving_utils import (  # noqa: E402
     LoraStats,
     _infer_prefix_and_remap,
     _count_backed_lora_modules,
+    _get_lora_scaling,
+    create_lora_statistics,
 )
+
+import torch  # noqa: E402
+import torch.nn as nn  # noqa: E402
 
 
 def _lw(keys):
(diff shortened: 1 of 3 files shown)


danielhanchen:
@codex review

oobabooga:
All 38 bundled tests pass, including the exact bug condition: a wrapper exposing only the singular active_adapter is now counted via _get_lora_scaling, so scaling_count aligns with the other counters and the alpha=0 silent-drop is gone. MoE/vision-tower/mxfp4 counting cases also covered. LGTM ✅

danielhanchen:
Addressed the remaining Codex finding (the CPU-only CI crash) in 38c2d0ad.

Root cause: the new `create_lora_statistics` test reaches `get_lora_layer_modules()`, which imports `peft.tuners.lora.bnb`. On a CPU-only runner bitsandbytes is genuinely absent, and the test installed a bare `types.ModuleType("bitsandbytes")` whose `__spec__` is `None`. The `find_spec("bitsandbytes")` probe run by transformers `_is_package_available` and peft `is_bnb_available` then raised `ValueError: bitsandbytes.__spec__ is None`, crashing collection before the regression ran.

Two changes:

1. `tests/test_saving_utils_lora_remap_count.py` now installs unsloth_zoo's permissive bitsandbytes stub via `inject_into_sys_modules()` instead of a bare module. The stub carries a real `__spec__`, so the availability probes resolve without raising. peft then defines its bnb LoRA classes normally (they subclass `torch.nn.Module`, not `bnb.nn.X`, so the import succeeds), and `get_lora_layer_modules()` returns its full set.

2. `unsloth_zoo/peft_utils.py`: `get_lora_layer_modules()` now skips a single un-importable optional backend (bnb/eetq/awq/aqlm/hqq/gptq) instead of crashing the whole discovery. This makes `create_lora_statistics` resilient on genuinely CPU-only or partial-backend installs as well. There is no behavior change when the backends are present (the returned set is identical).

Validation (peft 0.19.1, transformers 4.57.6):
- Test passes 38/38 with bitsandbytes present and 38/38 under a faithful absent-bitsandbytes simulation (find_spec returns None, import raises ModuleNotFoundError).
- `get_lora_layer_modules()` returns 13 classes with bnb present (unchanged) and 11 with bnb absent (only the bnb classes dropped, no crash).
- Broader saving/MoE regression suite stays green (113 passed, 3 skipped).
- The `#2966` regression assertion is intact: it still merges via `create_lora_statistics` and asserts alpha = 2.0 across the three layers, so it would fail if the scaling fix regressed.

danielhanchen:
@codex review

danielhanchen:
@codex review

danielhanchen:
@codex review

PR #809, opened 2026-06-20
Add lfm2_moe/ernie4_5_moe 4bit experts support and w1/w3/w2 expert merge

danielhanchen:
## Problem

LFM2.5-8B-A1B (lfm2_moe) and ERNIE-4.5-21B-A3B (ernie4_5_moe / ernie4_5_vl_moe) fail two ways that other MoE families (qwen3_moe, glm4_moe, deepseek_v3_moe, gemma4_moe) do not:

1. **4bit QLoRA training raises on the first backward.** On transformers >= 5 these models load the fused experts class (Lfm2MoeExperts / Ernie4_5_MoeExperts / Ernie4_5_VLMoeMoeExperts) whose native grouped-mm `forward` cannot consume a packed 2D bitsandbytes `Params4bit` `gate_up_proj`. It raises `IndexError: Dimension out of range (expected in [-2, 1], got 2)`. unsloth-zoo already ships the cure (`get_forward_moe_backend()`, dequantize + grouped_mm) and wires it for qwen3/glm4/deepseek/gemma4, but never for lfm2/ernie.

2. **merged_16bit merge maps 0 experts.** `_merge_moe_experts_file` recognizes the fused `gate_up_proj`/`down_proj` and the per-expert `gate_proj/up_proj/down_proj` disk layouts. LFM2 instead stores its experts per-expert as `w1`(gate)/`w3`(up)/`w2`(down) (`Lfm2MoeMLP.forward = w2(silu(w1(x)) * w3(x))`) while the runtime module is the fused `Lfm2MoeExperts`. Neither path matched, so the experts LoRA was silently not merged and the Step-7 sanity check raised `Saving LoRA finetune failed since # of LoRAs = 62 does not match # of saved modules = 18`.

## Fix

1. New `temporary_patches/lfm2_moe.py` and `ernie4_5_moe.py` that register the experts-forward backend patch (and the grouped-mm LoRA extractor) for the fused experts classes, mirroring `qwen3_moe.py` exactly (module-level `patch_function`, not a closure, so the compiled cache regenerates correctly). Registered in `temporary_patches/__init__.py`.

2. `_merge_moe_experts_file` now detects the per-expert disk-tensor naming **from the shard header** via `_detect_moe_perexpert_scheme` (`gate_proj/up_proj/down_proj` or `w1/w3/w2`), so the existing per-expert merge primitives handle both layouts. Detection is by on-disk tensor names, not model name, so any future arch with the same layout is covered and every other layout falls through to the unchanged default.

## Before / after

- Before: LFM2/ERNIE 4bit train crashes on the first backward; LFM2 16bit `save_pretrained_merged(merged_16bit)` raises the count mismatch.
- After: LFM2.5-8B-A1B 4bit + 16bit and ERNIE-4.5-21B-A3B-PT 4bit train, merge, and reload in a fresh process and reproduce the fine-tuned output.

## Does it preserve old pathways

Yes. The two new patch files are strict no-ops on transformers < 5 (the experts classes do not exist) and never touch any non-lfm2/ernie model. The merge change only adds the `w1/w3/w2` scheme as an alternative; when a shard uses the standard names the detected scheme is identical to today's hardcoded `gate_proj/up_proj/down_proj`, so qwen3/deepseek/glm4/gemma4/gpt-oss merges are byte-identical.

## Validation

- LFM2.5-8B-A1B 16bit: merge applies 2112/2112 expert tensors, fresh-process reload reproduces the phrase at CE ~1e-5.
- ERNIE-4.5-21B-A3B-PT: 18432/18432 expert tensors merged, CE ~1e-5.
- LFM2/ERNIE 4bit: train completes (no IndexError), loss drops, reload reproduces the phrase.
- Regression: Qwen3-30B-A3B (existing fused MoE) train -> merge -> reload unchanged (the new files and the header-detected scheme are no-ops for it), and a dense model is unaffected.
- The changed Python is transformers-version-agnostic and TRL-independent, and platform-agnostic (pure Python, only `os.path` string handling).

## Known limitations (documented, not active for these models)

The header detector keys off expert 0's tensors in a shard; a future MoE whose experts are split across multiple safetensors so that a shard lacks expert 0 would fall back to the default scheme for that shard. LFM2/ERNIE do not shard experts that way (verified). Mixed fused+per-expert layouts in one checkpoint are also out of scope.


danielhanchen:
Follow-up commit addressing the merge review notes:
- Per-expert scheme detection now probes any expert index (not just expert 0), so a multi-shard checkpoint whose shard holds only later experts (e.g. prefix.7.w3.weight) is still detected and merged.
- moe_num_experts is seeded from the shard header for schemes the pre-scan does not count (w1/w3/w2), so a down_proj-only adapter merges every expert instead of deriving 1.
Both are no-ops for standard gate/up/down layouts and fused 3D archs. On the native_torch backend note: forward_native_moe_loop applies the gate-up delta exactly once and 4bit experts route through the bnb4bit backend before reaching that loop, so there is no double application.

danielhanchen:
@codex review

danielhanchen:
Follow-up commit: registered lfm2_moe and ernie4_5_moe in TEMPORARY_PATCHES_SUBMODULES so test_temporary_patches_submodule_list_is_complete passes (it globs every *.py and asserts each is listed).

danielhanchen:
@codex review

danielhanchen:
@codex review

danielhanchen:
Good catch on the precedence. Fixed in `96a04f7`.

The concern is real: for a per-expert (w1/w3/w2) checkpoint whose experts are split across shards, a shard holding only a low-index subset (e.g. experts 0-15 of 64) reports 16 from its header. The old code seeded that shard-local count into `moe_num_experts` before resolving, so it short-circuited `_resolve_moe_num_experts` and overrode the authoritative count. Since that count drives the per-expert LoRA slicing stride, a wrong value corrupts every merged expert.

Worth noting where it was already safe vs. where it was not:
- Standard merge flow (model loaded, `lora_stats.module` live): the inner per-expert slicing already re-resolves via `_resolve_num_experts_from_lora_stats`, and the fused-3D path reads `num_experts` straight off the weight shape. So the actual slicing in that path was not corrupted.
- The gap was the outer iteration bound and the no-live-module path, which trusted the shard-local seed.

The fix resolves the authoritative count first (live module `num_experts`, else the fused-LoRA shape) and lets the shard header only RAISE a missing or too-low count, never lower an authoritative one. This keeps the original behavior that the header seed was added for (a `down_proj`-only adapter that would otherwise derive 1 still merges all experts), while making the header a last-resort bound rather than the slicing count.

I extracted the decision into `_resolve_moe_num_experts_with_header(...)` and added `tests/test_moe_num_experts_header_precedence.py` pinning the precedence (module/fused-LoRA count wins; header only raises). The existing MoE merge e2e suite still passes.

danielhanchen:
@codex review

danielhanchen:
@codex review

danielhanchen:
Looked at this again carefully. The shard-local `_count_moe_experts_in_header` value can only become the resolved count when the authoritative per-module resolution returns nothing, and that does not happen for a real MoE expert merge:

- For every shard file, before `_merge_moe_experts_file` runs, the #5410 block (saving_utils.py ~597-604) calls `_resolve_num_experts_from_lora_stats` on the live LoRA module and writes the result into `moe_num_experts[prefix]` when it is > 1. That walks `module.base_layer` for `num_experts` / `num_local_experts` / `num_routed_experts` / `num_moe_experts` / `num_experts_per_group`.
- The fused experts module for both targeted families exposes it directly: `Lfm2MoeExperts.num_experts = config.num_experts` and `Ernie4_5_MoeExperts.num_experts = config.moe_num_experts`. So the true full count (e.g. 64) is seeded shard-independently.
- `_resolve_moe_num_experts_with_header` then hits the cache branch and returns 64; the header only raises (`hdr_ne > num_experts`), and a middle shard's 32 never exceeds 64. The re-seed runs per file, so a low-index shard processed first is still corrected.

The header (max index + 1) path is reached only when no live module exposes a count at all, i.e. there is no expert module to read a stride from in the first place; in that last-resort case it is still strictly better than the single-projection derivation of 1. Since any fused-expert module stores `num_experts` by construction (the gate_up_proj/down_proj Parameters are shaped with it), the undercount is not reachable for the in-scope models, so I have left the resolution as-is rather than risk regressing the working merge. Happy to add an explicit config-num_experts fallback if you want belt-and-suspenders for a hypothetical module that omits the attribute.

PR #812, opened 2026-06-20
Fix Llama/Mixtral MoE LoRA merge dropping unfused expert deltas (#5403)

danielhanchen:
## Summary

Fixes the wrong-dimension / dropped-delta MoE merge for Llama-family and Mixtral MoE models trained with LoRA on the unfused expert weights (`w1`, `w2`, `w3`).

In transformers v5 the Mixtral architecture fuses experts in memory (`w1 + w3 -> mlp.experts.gate_up_proj`, `w2 -> mlp.experts.down_proj`), so LoRA attaches to the fused `mlp.experts` parameters. However `save_pretrained` writes the experts back to disk in the original unfused layout:

```
model.layers.0.block_sparse_moe.experts.0.w1.weight
model.layers.0.block_sparse_moe.experts.0.w2.weight
model.layers.0.block_sparse_moe.experts.0.w3.weight
```

The merge in `_merge_and_overwrite_lora` only recognised the new per-expert names (`mlp.experts.N.gate_proj/up_proj/down_proj`) and the fused 3D layout (`mlp.experts.gate_up_proj`). The legacy `block_sparse_moe.experts.N.w1/w2/w3` keys matched none of them, so the expert LoRA was silently dropped: the merged experts came out byte-identical to the base while attention layers merged normally. That is what the user in unslothai/unsloth#5403 hit (expert tensors look untouched / mismatched on reload).

## Root cause

`_merge_and_overwrite_lora` iterates the on-disk safetensor keys and routes each to a per-expert merge helper by regex. There was no branch for the legacy Mixtral disk naming, and `_lora_key_has_backing` (used by the save-count check) did not know the `mlp.experts` LoRA is backed by `block_sparse_moe.experts` shards.

## Fix

- Map legacy `block_sparse_moe.experts.N.w1/w2/w3.weight` disk keys back to the renamed `mlp.experts` LoRA (`w1 -> gate`, `w3 -> up`, `w2 -> down`) and merge them with the existing per-expert helpers.
- Count the experts under the `mlp.experts` prefix when pre-computing `num_experts` so out-of-range guards work.
- Teach `_lora_key_has_backing` the `mlp.experts -> block_sparse_moe.experts` alias so the save-count check matches what the merge writes.

All three changes are gated on the `block_sparse_moe.experts` disk naming, so dense models and other MoE layouts (Qwen MoE, GPT-OSS, Gemma4) are byte-identical to before. The merged checkpoint stays unfused on disk, so it remains a drop-in replacement for the original Mixtral-style model.

## Validation

- New CPU regression `tests/test_moe_merge_legacy_mixtral_cpu.py`: builds a legacy `w1/w2/w3` shard, merges, and asserts every expert tensor changed, no fallback was recorded, the count check passes, and the per-expert deltas match the analytic LoRA delta. It fails on `main` (`0 == 24`) and passes with the fix.
- Existing merge/MoE suite stays green: `tests/test_merge_e2e_moe_per_expert.py`, `test_merge_e2e_moe_fused.py`, `test_merge_e2e_dense.py`, `test_moe_merge_e2e_cpu.py`, `test_moe_fused_narrow_expert_merge.py`, `test_unsloth_zoo_lora_merge.py`, `test_moe_lora_extractor_coverage.py` and others (67 passed, 2 skipped).
- End-to-end on a B200 with a tiny `MixtralForCausalLM` (transformers 5.12.1, peft 0.19.1) trained with `target_modules=["q_proj","k_proj","v_proj","o_proj","w1","w2","w3"]`:
  - before: 0 / 24 expert tensors merged (deltas dropped, merged == base).
  - after: 24 / 24 expert tensors merged, 0 fallbacks, shapes preserved, disk layout still unfused, reloaded model runs a forward; the merged weights match what `peft.merge_and_unload` produces (the residual logit gap is the pre-existing grouped_mm-forward vs weight-merge difference shared by every fused MoE, not introduced here).

Fixes unslothai/unsloth#5403


danielhanchen:
@codex review

oobabooga:
The bundled CPU suite passes (3/3), and it's a real merge, not just structural: it builds a legacy block_sparse_moe.experts.N.w1/w2/w3 shard, merges, and asserts every expert tensor changed (no fallback) with the per-expert deltas matching the analytic LoRA delta (max_err < 1e-4), across both the .base_layer and .gate_up_proj paths. Gated on the legacy naming, so dense/other-MoE layouts are untouched. LGTM ✅

danielhanchen:
@codex review

danielhanchen:
@codex review

danielhanchen:
@codex review

danielhanchen:
On the P2 about `num_experts` being shard-local in the legacy w1/w2/w3 path: the count passed at the call site is only a fallback. Both merge helpers re-resolve it from the wrapped module before slicing:

- `_merge_moe_gate_or_up_expert` -> `num_experts = _resolve_num_experts_from_lora_stats(lora_stats, num_experts)`
- `_merge_moe_down_proj_expert` -> same

`_resolve_num_experts_from_lora_stats` walks `lora_stats.module(.base_layer ...)` for `num_local_experts` / `num_experts` / etc., so on a real Mixtral the true expert count is used regardless of what the shard key scan produced (which for legacy w1/w2/w3 keys is `None`, since those keys don't match the `gate_proj|up_proj|down_proj` scan that seeds `moe_num_experts`). The shard-local value only survives if the wrapped module exposes no expert-count attribute at all.

So this is already handled. I added a comment at the call site to make the re-resolution explicit, and a regression test (`test_legacy_mixtral_both_fused_keys_resolve_num_experts`) for the one combination that had no coverage: gate_up keyed on `.gate_up_proj` and down on `.down_proj` with no `.base_layer` / bare `.experts` companion, so the `#5410` override never fires. The merge still produces the correct per-expert deltas.

Also merged `origin/main` into the branch so it carries the latest `saving_utils.py` and friends rather than going stale.

danielhanchen:
@codex review

PR #813, opened 2026-06-22
Gemma-4 E-series: make cross-layer KV sharing survive use_cache=False and gradient checkpointing

danielhanchen:
## Problem

Gemma-4 E2B / E4B (`num_kv_shared_layers > 0`) implement cross-layer KV sharing THROUGH the cache: a store layer writes its full-length K/V into `past_key_values.shared_layers`, and each shared layer reads them back. Both are gated on `past_key_values is not None`. So when `past_key_values` is None the sharing silently breaks and every shared layer recomputes its own (wrong) K/V, producing garbage logits (huggingface/transformers#45242).

`past_key_values` is None in exactly the two cases that matter for training:
- `use_cache=False` (every QLoRA tutorial sets `config.use_cache=False`), and
- gradient checkpointing, where `GradientCheckpointingLayer` nulls the `past_key_values` kwarg before the attention runs (`transformers/modeling_layers.py`).

The previous wrapper injected a `DynamicCache` at the model forward and forced `use_cache=True`. That fixed plain `use_cache=False`, but was **defeated by gradient checkpointing** (the layer nulls the cache before the attention sees it), so QLoRA training stayed broken.

## Before / after (transformers 5.5.0, google/gemma-4-E2B-it, num_kv_shared_layers=20)

Plain forward, no Unsloth:
| path | top-1 | teacher-forced loss |
|---|---|---|
| use_cache=True | `Here` | 6.84 (correct) |
| use_cache=False | `**` | 16.17 (garbage) |

Under gradient checkpointing the shared layers fall back for both paths. Instrumenting the attention: pre-fix `shared retrieve=0 fallback=20`; post-fix `retrieve=20 fallback=0`, loss back to the correct ~6.8.

Real Unsloth QLoRA (`FastModel` + `use_gradient_checkpointing="unsloth"`) on a single repeated row: loss `4.86 -> 0.000` (memorises), generation coherent.

## Fix

Decouple the within-forward shared KV from `past_key_values`. The model forward attaches a fresh tiny carrier (a `shared_layers` dict + no-op `update()`) to every `Gemma4TextAttention`, and the attention uses it whenever the real `past_key_values` is None. Store/retrieve then works under `use_cache=False` AND gradient checkpointing. This mirrors transformers' own later `shared_kv_states` fix.

**Forwards-compatible:** capability-gated to a strict no-op on builds whose attention forward already takes a `shared_kv_states` argument (the upstream fix), so it only applies where the cache-dependent mechanism is present.

## No-break

- No-op for `num_kv_shared_layers == 0` (31B / 26B-A4B) and non-Gemma-4 models (the attention list is cached empty).
- Generation (`use_cache=True`) is unchanged: the real cache is non-None, so the carrier is never substituted.
- No-op on transformers builds that pass `shared_kv_states` (capability gate).

## Tests

CPU regression tests (no GPU/model): patches stay registered and actually wire the carrier (guards against silently disabling the fix), carrier semantics, the capability gate (True without `shared_kv_states`, False with), and the attention carrier-substitution rule (substitute only when `past_key_values` is None).

### DIFF
--- a/tests/test_zoo_history_regressions.py
+++ b/tests/test_zoo_history_regressions.py
@@ -571,6 +571,41 @@ def test_gemma4_attention_carrier_substitution():
     assert seen["pkv"] is None, "without a carrier, pass through unchanged"
 
 
+def test_gemma4_clear_shared_kv_carrier_releases_kv():
+    """_gemma4_clear_shared_kv_carrier drops the producer K/V the carrier pinned and
+    removes the module attribute, so a no_grad/eval forward does not hold the carrier
+    (and its K/V) alive until the next forward. No-op when there is no carrier."""
+    from unsloth_zoo.temporary_patches.gemma4 import (
+        _Gemma4SharedKVCarrier, _gemma4_clear_shared_kv_carrier,
+    )
+
+    class _Attn:
+        pass
+
+    a0, a1 = _Attn(), _Attn()
+    carrier = _Gemma4SharedKVCarrier()
+    carrier.shared_layers[0] = ("k", "v")
+    a0._unsloth_shared_kv_carrier = carrier
+    a1._unsloth_shared_kv_carrier = carrier
+
+    class _Model:
+        pass
+
+    model = _Model()
+    model._unsloth_gemma4_attns = [a0, a1]
+
+    _gemma4_clear_shared_kv_carrier(model)
+    assert carrier.shared_layers == {}, "producer K/V must be released"
+    assert not hasattr(a0, "_unsloth_shared_kv_carrier"), "carrier attr must be removed"
+    assert not hasattr(a1, "_unsloth_shared_kv_carrier")
+
+    # Idempotent / no-op when nothing is attached.
+    model2 = _Model()
+    _gemma4_clear_shared_kv_carrier(model2)  # no _unsloth_gemma4_attns -> no error
+    model3 = _Model(); model3._unsloth_gemma4_attns = []
+    _gemma4_clear_shared_kv_carrier(model3)  # empty list -> no error
+
+
 def test_gemma4_force_nonreentrant_checkpointing(monkeypatch):
     """The GC fix overrides _gradient_checkpointing_func on gemma-4 text decoder layers
     that are being checkpointed with a NON-reentrant torch checkpoint, resolving the
--- a/tests/test_zoo_history_regressions.py
+++ b/tests/test_zoo_history_regressions.py
@@ -713,47 +713,73 @@ def test_gemma4_pristine_checkpoint_recovered_via_set_once_capture(monkeypatch):
         del _ckpt._unsloth_pristine_checkpoint
 
 
-def test_gemma4_carrier_rejects_two_forwards_before_one_backward():
+def _make_gemma4_fake_model(torch, *, kv_shared, checkpointed):
+    """A stand-in Gemma-4 model: one decoder layer whose self_attn class is named
+    Gemma4TextAttention, with configurable KV sharing and gradient checkpointing flags."""
+    Gemma4TextAttention = type("Gemma4TextAttention", (), {})
+
+    class _Layer:
+        def __init__(self):
+            self.self_attn = Gemma4TextAttention()
+            self.self_attn.is_kv_shared_layer = kv_shared
+            self.gradient_checkpointing = checkpointed
+
+    layer = _Layer()
+
+    class _Model:
+        def modules(self):
+            return [self, layer, layer.self_attn]
+
+    return _Model()
+
+
+def test_gemma4_carrier_rejects_two_forwards_only_when_unsafe():
     """The module-scoped 5.5.0/5.5.1 carrier cannot serve two live checkpointed graphs, so a
-    second grad-enabled forward before the first's backward (DPO/contrastive) must raise rather
-    than silently corrupt gradients. Single-forward + backward, then re-forward, stays allowed."""
+    second grad-enabled forward before the first's backward must raise -- but ONLY when KV sharing
+    AND gradient checkpointing are both active (else no carrier is re-read in backward). Models
+    without sharing (31B/26B-A4B) or with checkpointing off must keep working with DPO/contrastive."""
     import torch
     import pytest
     from unsloth_zoo.temporary_patches.gemma4 import _make_kv_shared_use_cache_false_safe_forward
 
     w = torch.nn.Parameter(torch.randn(4, 4))
 
-    class _Model:
-        # no Gemma4TextAttention -> force/attach helpers are strict no-ops; only the
-        # outstanding-forward marker logic is under test here.
-        def modules(self):
-            return []
-
     def _orig(self, x):
         return x @ w  # stands in for last_hidden_state; requires grad
 
     wrapped = _make_kv_shared_use_cache_false_safe_forward(_orig, attach_carrier=True)
-    m = _Model()
     x = torch.randn(2, 4, requires_grad=True)
 
+    # Unsafe: KV sharing + checkpointing -> arm + reject the overlapping forward.
+    m = _make_gemma4_fake_model(torch, kv_shared=True, checkpointed=True)
     out1 = wrapped(m, x)
     assert getattr(m, "_unsloth_gemma4_carrier_outstanding", False) is True
-
-    # second forward before any backward -> reject
     with pytest.raises(RuntimeError, match="two forward passes before a single backward"):
         wrapped(m, x)
-
-    # after the first graph's backward, the marker clears and a new forward is allowed
+    # after the first graph's backward the marker clears and a new forward is allowed
     out1.sum().backward()
     assert getattr(m, "_unsloth_gemma4_carrier_outstanding", False) is False
     wrapped(m, x)
     assert getattr(m, "_unsloth_gemma4_carrier_outstanding", False) is True
 
+    # Safe (no KV sharing, e.g. 31B/26B-A4B): two forwards before one backward must NOT raise.
+    m_noshare = _make_gemma4_fake_model(torch, kv_shared=False, checkpointed=True)
+    wrapped(m_noshare, x)
+    wrapped(m_noshare, x)  # would raise if the marker were armed
+    assert getattr(m_noshare, "_unsloth_gemma4_carrier_outstanding", False) is False
+
+    # Safe (KV sharing but checkpointing off): backward uses saved activations, never re-reads
+    # the carrier, so overlapping forwards must NOT raise.
+    m_nockpt = _make_gemma4_fake_model(torch, kv_shared=True, checkpointed=False)
+    wrapped(m_nockpt, x)
+    wrapped(m_nockpt, x)
+    assert getattr(m_nockpt, "_unsloth_gemma4_carrier_outstanding", False) is False
+
     # eval / no_grad never arms the marker (no backward to guard)
-    delattr(m, "_unsloth_gemma4_carrier_outstanding")
+    m_eval = _make_gemma4_fake_model(torch, kv_shared=True, checkpointed=True)
     with torch.no_grad():
-        wrapped(m, x)
-    assert getattr(m, "_unsloth_gemma4_carrier_outstanding", False) is False
+        wrapped(m_eval, x)
+    assert getattr(m_eval, "_unsloth_gemma4_carrier_outstanding", False) is False
 
 
 # ---------------------------------------------------------------------------
--- a/tests/test_zoo_history_regressions.py
+++ b/tests/test_zoo_history_regressions.py
@@ -580,28 +580,36 @@ def test_gemma4_force_nonreentrant_checkpointing(monkeypatch):
     cross-layer KV sharing under reentrant/offloaded gradient checkpointing."""
     import functools
     import torch.utils.checkpoint as _ckpt
+    from unsloth_zoo.temporary_patches import gemma4 as g4
     from unsloth_zoo.temporary_patches.gemma4 import _gemma4_force_nonreentrant_checkpointing
 
     def _pristine(function, *args, use_reentrant=None, **kw):
         return ("pristine", use_reentrant)
     def unsloth_checkpoint(function, *args, **kw):
         return "shim"
-    unsloth_checkpoint.__name__ = "unsloth_checkpoint"  # the reentrant-forcing shim
-    # Simulate Unsloth smart GC having globally patched torch's checkpoint.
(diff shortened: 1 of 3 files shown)


danielhanchen:
Verified forwards and backwards compatibility across the full gemma-4 lifespan in transformers (every 5.x release; pulled the real modeling_gemma4.py from each wheel and ran the live gate + patch install in actual venvs).

| transformers | gemma-4 | shared_kv_states | past_key_values arg | gate | fix behavior |
|---|---|---|---|---|---|
| <= 5.4.0 (all 4.x + early 5.x) | absent | n/a | n/a | True (except) | ImportError caught, no patch, no crash |
| 5.5.0, 5.5.1 | present | absent | args[3] | True | fix installed, args[3] is past_key_values (correct) |
| 5.5.2 - 5.12.1 (latest) | present | present | args[4] | False | strict no-op, patch never installed |

Key point: transformers backported the native shared_kv_states fix mid-patch-release at 5.5.2 (not 5.6.0), which shifts past_key_values from args[3] to args[4]. A version-range check would have applied the args[3] logic on 5.5.2-5.5.4 and grabbed shared_kv_states instead of the cache. Because the gate inspects the attention signature rather than the version string, it flips to no-op exactly when the native argument appears, wherever that lands.

Net: the patch is active only on 5.5.0-5.5.1 (the precise window where gemma-4 exists but the cache-free path is still broken), and is a verified no-op everywhere else, including the latest 5.12.1. Confirmed on 5.5.2+ that neither the attention carrier nor the model-forward wrapper is installed, so vanilla transformers runs untouched.

Latest commit documents this version window in the code comments.

danielhanchen:
@codex review

danielhanchen:
@codex review

danielhanchen:
Confirmed the P1 is real and fixed it. The carrier made the forward correct under use_cache=False + gradient checkpointing, but the producer (KV-source) layer's k_proj/v_proj got wrong gradients under REENTRANT checkpointing, which is the default Unsloth path (use_gradient_checkpointing="unsloth" forces use_reentrant=True for its offloaded checkpointer). Each layer is recomputed independently in reverse order during backward, so the ~20 shared layers recompute before their producer and read the detached no-grad K/V from the first forward; the gradient from the shared attention never reaches the producer's k/v projections.

Confirmed empirically on gemma-4-E2B-it (fp32, grads vs the use_cache=True reference):

| checkpointing | producer k/v_proj grad cosine, before -> after |
|---|---|
| torch reentrant | 0.14-0.55 -> 1.0000 |
| torch non-reentrant | 1.0 -> 1.0 (already correct) |
| Unsloth offloaded smart GC | 0.14-0.55 -> 1.0000 |

The loss/logits were correct in every case (15.13375), so this was a silent gradient bug that the earlier loss-goes-to-zero memorization test did not catch.

Fix: while the carrier is active, force the gemma-4 text decoder layers to use non-reentrant checkpointing, whose connected recomputation graph keeps the shared K/V grad-connected. It resolves the pristine torch checkpoint because Unsloth's smart GC globally swaps in a shim that hard-forces use_reentrant=True. This trades Unsloth's CPU activation offload for correct gradients on these ~1-2B E-series models, and is scoped to the transformers 5.5.0-5.5.1 window where the carrier runs (5.5.2+ native shared_kv_states makes the whole patch a no-op).

Validation: real Unsloth QLoRA e2e (FastModel + use_gradient_checkpointing="unsloth", smart GC active) trains loss 4.88 -> 0.000, no crash/OOM; 21/21 CPU regression tests pass including a new one for the override + pristine-checkpoint resolution.

danielhanchen:
@codex review

danielhanchen:
Two follow-ups from this review round:

P1 (skipping the GC fix on native shared_kv_states) -- confirmed real, fixed. I tested native transformers 5.12.1 directly and it has the exact same producer-gradient bug under reentrant GC (producer k/v_proj grads diverge from the use_cache=True reference, cosine 0.14), because native shared_kv_states is a dict passed as a kwarg, which GradientCheckpointingLayer bakes into its partial rather than passing as a checkpoint input, so the reverse-order recomputation drops the producer gradient just like the carrier did. The previous commit gated the non-reentrant forcing behind the carrier (needs_cache), so it never ran on 5.5.2+.

Fixed in 18643b5 by decoupling the two fixes: the non-reentrant checkpointing forcing (gradient correctness) now runs for any model with E-series KV sharing on every transformers version, while the carrier (cache-free forward) stays gated to 5.5.0-5.5.1. Verified producer grads match the reference to cosine 1.0 under reentrant and Unsloth offloaded GC on both transformers 5.5.0 (carrier) and 5.12.1 (native); strict no-op for non-E-series gemma-4.

P2 (clear the carrier after non-checkpointed forwards) -- I did not change this, because clearing the carrier after the forward would break the gradient-checkpointing backward: under GC the producer writes the shared K/V during the forward and the consumers read it during the backward recomputation, so the carrier must survive until backward completes. A fresh carrier is created and re-attached at the start of every forward, so there is no cross-step contamination, and with the decoupling above the carrier is now only attached on 5.5.0-5.5.1 (on 5.5.2+ the attribute is never set). The only residual is one forward's shared K/V held until the next forward overwrites it, which is negligible.

danielhanchen:
@codex review

danielhanchen:
Thanks for the thorough pass. Addressed across `18643b5`, `4bbfb68`, and `6fd484f`. Status per finding:

**P1 (line 517) - keep the non-reentrant GC fix on native `shared_kv_states`:** Resolved by the restructure in `18643b5`. The early return that skipped the wrapper on 5.5.2+ builds is gone. The model-forward wrapper is now installed on every version, and it calls `_gemma4_force_nonreentrant_checkpointing(self)` unconditionally (the carrier attach is the only part gated on the 5.5.0-5.5.1 window). Confirmed empirically on native transformers 5.12.1 + gemma-4-E2B: producer k/v_proj grads diverge under reentrant GC (cosine ~0.14) before the patch and match the `use_cache=True` reference (cosine 1.0) after.

**P2 (line 450) - unwrap stacked Unsloth checkpoint shims:** Fixed in `4bbfb68`. The pristine torch checkpoint is now captured at import (before any Unsloth GC patch swaps `torch.utils.checkpoint.checkpoint`), and the `_old_checkpoint` fallback unwraps in a loop while the candidate name is one of the known Unsloth shim names, so a stacked shim can no longer be selected as the base.

**P3 (line 473) - preserve existing checkpoint kwargs when forcing non-reentrant:** Fixed in `4bbfb68`. The override now copies the existing `_gradient_checkpointing_func` partial's keywords (minus `use_reentrant`) into the new partial, so `preserve_rng_state`, `context_fn`, `determinism_check`, etc. are kept and only `use_reentrant` is overridden.

**P2 (line 500) - clear the per-forward carrier after non-checkpointed forwards:** Fixed in `6fd484f`. When no autograd graph is built (`not torch.is_grad_enabled()`, i.e. eval/inference/no_grad), the carrier and its pinned producer K/V are released in a `finally` as soon as the forward returns. The training/GC-backward path is untouched (it must keep the carrier alive until backward), so the validated gradient fix is byte-identical there.

**P2 (line 500) - do not store the per-forward carrier on modules (multi-forward):** Documented as a limitation specific to the 5.5.0-5.5.1 carrier bridge in `6fd484f`. The carrier must live on the module because that is the only channel that survives into gradient-checkpoint backward recompute: recompute re-runs the decoder layers, not the model-forward wrapper, so a forward-scoped contextvar/stack would already be gone by backward. Keeping two checkpointed forward graphs alive before a single backward (summing microbatch/contrastive losses) therefore lets the second forward overwrite the carrier. The clean fix is exactly the function-scoped `shared_kv_states` that transformers added in 5.5.2, which also makes the carrier unnecessary, so the supported path for that pattern is transformers >= 5.5.2. Standard single-forward and gradient-accumulation training (one backward per forward) is unaffected.

All `tests/test_zoo_history_regressions.py` gemma-4 tests pass, including new coverage for the kwarg preservation, stacked-shim unwrap, and carrier-clear behaviors.

danielhanchen:
@codex review

danielhanchen:
This is a real limitation, but it is specific to the transformers 5.5.0-5.5.1 carrier bridge and cannot be fixed cleanly within it; the structural reason is worth spelling out.

The carrier has to live on the attention module because that is the only channel that survives into gradient-checkpoint backward recompute. Under checkpointing, recompute re-runs each decoder layer's forward (reading `self._unsloth_shared_kv_carrier`), but it does NOT re-run the model-level forward wrapper that creates the carrier. So anything forward-scoped (a contextvar, a stack, a local) is already gone by the time recompute reads it. The only per-forward channel that would survive is a checkpoint INPUT (non-reentrant checkpoint saves and restores the inputs to each checkpointed call), but the shared K/V reaches the attention through `past_key_values`, which `GradientCheckpointingLayer` nulls before the attention runs (that nulling is the whole reason the carrier exists). We do not control the decoder layer's call signature, so there is no non-invasive way to make the carrier a saved checkpoint input.

The consequence is exactly what you describe: with gradient checkpointing and more than one forward graph alive before backward (DPO/contrastive, or summing microbatch losses), the second forward overwrites the single module carrier and the first graph's recompute reads the later batch's K/V. Standard single-forward training and gradient accumulation (one backward per forward) are unaffected, since the carrier is consumed by that forward's backward before the next forward replaces it.

The clean fix is transformers' own function-scoped `shared_kv_states` argument, which is passed per forward and is therefore immune to this. That landed in 5.5.2, and this patch already detects it by signature and no-ops the carrier there (only the version-independent non-reentrant GC fix remains). So the carrier is a bridge for exactly two patch releases (5.5.0, 5.5.1), and on those the supported path for multi-forward-graph training is `transformers >= 5.5.2`. I have documented this scope in the code rather than contort the carrier, since the only correct per-forward channel is the upstream mechanism the carrier exists to stand in for.

danielhanchen:
@codex review

danielhanchen:
Triaged all three in eb19927.

**Avoid returning a checkpoint shim as pristine (gemma4.py:397).** Real bug, fixed. The unwrap loop only read the fixed module-level `_ckpt._old_checkpoint`, so with stacked shims it advanced one level and could return a shim; `_gemma4_force_nonreentrant_checkpointing` then wrapped that shim with `use_reentrant=False`, silently bypassing the gradient fix. It now follows each shim's own `_old_checkpoint` first (which survives stacking, where a single module attr is overwritten) and returns `None` when no genuine checkpoint is reachable. The caller already early-returns on `base is None`, so the worst case is now a safe no-op (fix skipped) rather than wrapping the offloaded checkpointer. The common import-before-patch path is unchanged (it returns the import-captured ref).

**Clear the carrier after backward (gemma4.py:539).** Looked hard at this. The carrier's producer K/V is checkpointed away in the original forward, so it has no grad_fn and cannot carry a backward hook; and it must stay live through the entire backward because GC recompute reads it in reverse order (consumers recompute before the producer). The only reliable 'end of backward' signal is `Variable._execution_engine.queue_callback`, a private API not used anywhere else in unsloth_zoo, and it would have to live in the per-layer recompute hot path. For a single producer layer's K/V (tens of MB at E2B/E4B context) held across one `optimizer.step()` and freed at the next forward, on the narrow 5.5.0-5.5.1 window only, that machinery is not worth the risk. I documented the lifetime in the code instead. 5.5.2+ already frees it sooner via the function-scoped `shared_kv_states`.

**Keep per-forward carriers isolated (gemma4.py:485).** This is the documented limitation already called out in the wrapper docstring: the carrier is module-scoped precisely because that is the only channel that survives GC backward recompute (recompute re-runs the layers, not the forward wrapper), so two checkpointed graphs alive before one backward (DPO chosen/rejected, contrastive, summed microbatch) can overwrite it. Function-scoped isolation is exactly what 5.5.2+ `shared_kv_states` provides; on 5.5.0-5.5.1 there is no per-graph channel to scope it to. Single-forward and gradient-accumulation training are unaffected. No code change.

danielhanchen:
@codex review

danielhanchen:
Both addressed in d93922c.

**P1 - Preserve the carrier for each checkpointed forward.** You're right that the module-scoped carrier can't serve two live checkpointed graphs, and on 5.5.0/5.5.1 there's no per-graph channel to scope it to (that's exactly what 5.5.2+ function-scoped `shared_kv_states` provides). Rather than silently corrupt gradients, the wrapper now detects the unsafe pattern and fails loudly. It marks a carrier outstanding when a grad-enabled forward attaches one, and clears the marker via a grad hook on the forward output (which fires at the start of that graph's backward). A second grad-enabled forward seen while a prior one is still outstanding (DPO chosen/rejected, contrastive, summed microbatch losses) raises a clear error pointing to transformers >= 5.5.2. Single-forward and gradient-accumulation are unaffected, because the prior graph's backward clears the marker before the next forward. Eval/no_grad never arms it.

**P2 - Fall back past stacked Unsloth checkpoint shims.** Correct - my earlier 'return None' avoided wrapping a shim but then left E-series layers on reentrant GC in exactly the stacked-shim case. The real problem is that all three patchers store `_old_checkpoint` at module level, so stacking overwrites the pristine reference. Fixed at the source: `gradient_checkpointing.py` now stashes the genuine `torch.utils.checkpoint.checkpoint` set-once (`_unsloth_pristine_checkpoint`) on the first patch, before any shim stacks, and `_resolve_pristine_checkpoint` prefers it. So the non-reentrant fix now applies regardless of import order or how many shims are stacked. Added regression tests for both (multi-forward rejection + set-once pristine recovery through stacked shims); the full gemma4 suite passes.

danielhanchen:
@codex review

danielhanchen:
Good catch on both - the multi-forward guard I added was over-eager. Fixed in aee75ef.

Both findings are the same root issue: the marker armed whenever the output required grad, ignoring whether the module-scoped carrier is actually re-read in backward. It isn't, in two cases:
- **No KV sharing (31B / 26B-A4B):** `_gemma4_attach_shared_kv_carrier` caches an empty attn list and attaches nothing, so there is no carrier to clobber.
- **KV sharing but gradient checkpointing off:** backward uses the saved activations and never re-enters attention to read the carrier.

Both of those wrongly raised on DPO/contrastive/summed-loss objectives. The marker and the raise are now gated on `_gemma4_carrier_overlap_unsafe(self)`, which is true only when the model has KV sharing AND at least one Gemma-4 layer currently has `gradient_checkpointing` enabled. Only then does a second forward overwrite a carrier the first graph's recompute still needs. The error message also now mentions disabling gradient checkpointing as an alternative to upgrading. Regression test extended to cover the unsafe case (raises) plus both safe configs (no-share and checkpointing-off must not raise); full gemma4 suite passes.

danielhanchen:
@codex review

PR #814, opened 2026-06-22
peft_utils: add finetune_audio_layers for Gemma 4 / Gemma 3N audio + embedder LoRA

danielhanchen:
## Summary

`get_peft_regex` builds an ordered matcher: `[vision|language model-part] .*? [self_attn|mlp component] .*? [linear leaf]`. Gemma 4 (`Gemma4ForConditionalGeneration`) and Gemma 3N (`Gemma3nForConditionalGeneration`) keep the visual/audio path in flat embedder Linears and a conformer `audio_tower` whose leaves carry no attention/mlp token, so the matcher can never select them:

- `embed_vision.embedding_projection`, `embed_audio.embedding_projection` (multimodal projectors)
- `vision_embedder.patch_dense`, `vision_tower.patch_embedder.input_proj`
- `audio_tower` conformer: `ffw_layer_1/2`, `lconv1d.linear_start/linear_end`, `attention.post`, `q/k/v_proj`, `relative_k_proj`, `pos_proj`, `input_proj_linear`, `output_proj`

So `finetune_vision_layers=True` never adapts the projector, and no flag at all reaches the audio encoder. This also silently broke the Gemma audio notebooks once explicit `target_modules` lists began routing through the layer-scope filter (their audio leaf names have no model-part tag, so they match nothing).

## Change

- Add `finetune_audio_layers` (default `False`) to `get_peft_regex`.
- Add dedicated name-anchored branches for `audio_tower` (literal leaf list) and the vision/audio embedder projectors. Each branch is appended only when it actually matches a Linear in the model, and ends at the leaf (PEFT uses `re.fullmatch`).
- Fix the no-layers guard so audio-only selection is allowed without the base matcher degenerating into every attention/mlp leaf.

## Backwards compatibility

Because each branch is appended only when it matches a module in the given model, the produced regex is byte-identical for every architecture that lacks these Gemma module-name segments. The audio branch reaches `audio_tower` only when `finetune_audio_layers=True` (default off), so Qwen2-Audio / Qwen2.5-Omni and similar are unchanged unless explicitly opted in.

## Tests

`tests/test_peft_regex_audio.py`:
- byte-identical regex across existing architectures (Llama, Qwen, Qwen2-VL, Llava, Gemma3, Qwen2-Audio, Whisper) with the flag off
- 0 -> N attachment for Gemma 4 E2B/12B and Gemma 3N E4B with the flag on (audio_tower + `embed_audio.embedding_projection`; the vision flag attaches `embed_vision.embedding_projection`)
- conv / norm modules never matched; audio-only does not leak into the language stack

Verified against the real Gemma 3N E4B module tree (meta device): audio off matches 245 language Linears and 0 audio; audio on adds 133 `audio_tower` Linears plus the `embed_audio.embedding_projection` projector, language set unchanged, no conv/norm matched.

## Notes

- Pairs with the unsloth PR threading `finetune_audio_layers` through `FastBaseModel.get_peft_model`.
- Composes with #798 (`mixer` tag) and #807 (drop trailing `.*?`).

### DIFF
--- a/tests/test_peft_regex_audio.py
+++ b/tests/test_peft_regex_audio.py
@@ -1,0 +1,315 @@
+# Unsloth Zoo - Utilities for Unsloth
+# Copyright 2023-present Daniel Han-Chen, Michael Han-Chen & the Unsloth team. All rights reserved.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+"""
+Regression tests for get_peft_regex's audio/vision-embedder LoRA targeting.
+
+Goals (mirrors the #798 byte-identical methodology):
+  * finetune_audio_layers defaults False and is strictly additive: for every
+    existing architecture the produced regex never contains the audio branches
+    when the flag is off, and is byte-identical with the flag on when the model
+    has no audio_tower / Gemma embedder modules.
+  * With the flag on, Gemma 4 / Gemma 3N attach their audio_tower Linears and the
+    embed_audio.embedding_projection projector (0 -> N); the vision flag attaches
+    embed_vision.embedding_projection. conv / *norm modules are never matched
+    (they are not nn.Linear, so not even candidates).
+
+Module names are taken verbatim from the real checkpoint tensor names
+(`get_safetensors_metadata` on unsloth/gemma-4-E2B-it, unsloth/gemma-4-12B-it,
+unsloth/gemma-3n-E4B-it), so the synthetic trees are faithful. We use synthetic
+trees for determinism / no downloads; get_peft_regex only inspects named_modules()
++ isinstance(_, nn.Linear).
+"""
+import re
+import pytest
+import torch
+import torch.nn as nn
+
+from unsloth_zoo.peft_utils import get_peft_regex
+
+
+# ----------------------------------------------------------------------------- helpers
+def _add(root: nn.Module, path: str, leaf_module: nn.Module):
+    *mids, leaf = path.split(".")
+    cur = root
+    for m in mids:
+        if not hasattr(cur, m):
+            setattr(cur, m, nn.Module())
+        cur = getattr(cur, m)
+    setattr(cur, leaf, leaf_module)
+
+
+class FakeModel(nn.Module):
+    """Builds a module tree from (path -> module-kind) specs."""
+    def __init__(self, linear_paths, other_paths=(), name="fake"):
+        super().__init__()
+        self.model = nn.Module()
+        for p in linear_paths:
+            _add(self.model, p, nn.Linear(8, 8, bias=False))
+        for p, kind in other_paths:
+            _add(self.model, p, kind())
+        class _Cfg:
+            _name_or_path = name
+            model_type = "fake"
+        self.config = _Cfg()
+
+
+def linear_names(model):
+    return [n for n, m in model.named_modules() if isinstance(m, nn.Linear)]
+
+
+def matched(regex, names):
+    # PEFT uses re.fullmatch; assert on that (no DOTALL needed - names have no newlines).
+    return {n for n in names if re.fullmatch(regex, n)}
+
+
+def leaf(name):
+    # Strip a trailing Gemma4ClippableLinear ".linear" suffix only (do NOT touch a
+    # leaf literally named e.g. "linear_start"), then take the last path segment.
+    if name.endswith(".linear"):
+        name = name[: -len(".linear")]
+    return name.rsplit(".", 1)[-1]
+
+
+# ----------------------------------------------------------------------------- fixtures
+def _text_stack(prefix, n=2, heads=("q_proj", "k_proj", "v_proj", "o_proj"),
+                mlp=("gate_proj", "up_proj", "down_proj")):
+    out = []
+    for i in range(n):
+        out += [f"{prefix}.{i}.self_attn.{h}" for h in heads]
+        out += [f"{prefix}.{i}.mlp.{m}" for m in mlp]
+    return out
+
+
+# Existing architectures (no Gemma embedder / no audio LoRA target today).
+LLAMA      = _text_stack("language_model.layers") + ["lm_head"]
+QWEN3      = _text_stack("language_model.layers") + ["lm_head"]
+QWEN2_VL   = (_text_stack("language_model.layers")
+              + [f"visual.blocks.{i}.attn.{p}" for i in range(2) for p in ("qkv", "proj")]
+              + [f"visual.blocks.{i}.mlp.{p}" for i in range(2) for p in ("fc1", "fc2")]
+              + ["visual.merger.mlp.0", "visual.merger.mlp.2"])
+LLAVA      = (_text_stack("language_model.layers")
+              + [f"vision_tower.vision_model.encoder.layers.{i}.self_attn.{p}"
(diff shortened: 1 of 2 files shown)


danielhanchen:
@codex review

danielhanchen:
Thanks for the reviews. Addressed in 715fff77:

- **Positional API** (Codex): moved `finetune_audio_layers` after `target_modules`, so a positional call like `get_peft_regex(model, True, True, True, True, ["q_proj","v_proj"])` still binds the list to `target_modules`.
- **`re.search` vs `re.fullmatch`** (Gemini): the candidate-branch pre-check now uses `re.fullmatch` (PEFT's own matcher), avoiding substring false positives like `embedding_projection` matching `embedding_projection_classifier`.
- **Fallback ordering** (Codex): the component-only fallback now runs *after* the new branches and never for audio-only, so vision-only / language-off on Gemma 3N no longer broadens into `language_model.*`.
- **Respect explicit `target_modules`** (Codex): the audio/embedder branches are intersected with an explicit list when one is given.
- **Respect attention/mlp flags** (Codex): the conformer attention leaves (`q/k/v/post/...`) and MLP leaves (`ffw_layer_*`) are gated by `finetune_attention_modules` / `finetune_mlp_modules`.
- **Cleanup** (code-quality): dropped the unused `torch` import in the test.

New tests lock each behavior (vision-only no-language-leak, explicit-target intersect, attn/mlp scoping, positional `target_modules`).

Not adopting: adding `out_proj`/`fc1`/`fc2` to generalize to Qwen2-Audio / Qwen2.5-Omni / Voxtral. Those have no Unsloth notebook, and we want `finetune_audio_layers` scoped to the audio models we ship notebooks for (Gemma 4 / Gemma 3N). The Qwen2-Audio fixture remains purely a no-leak / byte-identical guard.

danielhanchen:
@codex review

danielhanchen:
Second round addressed in 47add503:

- **Segment boundary** (Codex P2): the audio branch now requires a `.` before the leaf (`audio_tower\.(?:.*\.)?<leaf>`), so `target_modules=["k_proj"]` no longer also matches `...relative_k_proj`. New test locks this.
- **Respect attention/mlp flags** (Codex P2 x2): every audio leaf is now gated -- attention (`q/k/v/relative_k/pos/post`) behind `finetune_attention_modules`, and the feed-forward / lightweight-conv / subsample / output projections behind `finetune_mlp_modules` -- instead of attaching the conv/projection leaves unconditionally.

Not changing:

- **Gemma 4 wrapper match** (Codex P1): matching `Gemma4ClippableLinear` wrappers (e.g. `...q_proj` alongside `...q_proj.linear`) is already handled -- `unsloth/models/vision.py` patches `LoraModel._create_and_replace` to redirect any `Gemma4ClippableLinear` target to its `.linear` child, and `unsloth_zoo` does the same for reload (`patch_Gemma4ClippableLinear_peft_reload`). This is exactly how the standard matcher already targets the `vision_tower` ClippableLinear wrappers, so no regex change is needed.
- **Qwen `out_proj`/`fc1`/`fc2`** (Codex P2): intentionally not added -- `finetune_audio_layers` is scoped to the audio models we ship notebooks for (Gemma 4 / Gemma 3N); Qwen2-Audio / Qwen2.5-Omni / Voxtral have none. A test now asserts the flag does not target those Qwen-specific leaves.

Verified on the real Gemma 3N E4B tree: audio (attn+mlp) still attaches 245 language + 133 audio_tower + 1 embed_audio; mlp-only correctly drops the conformer attention leaves.

danielhanchen:
@codex review

danielhanchen:
Third review round addressed in `6735c559`. 78/78 tests pass; re-verified on the real Gemma 3N E4B tree (`init_empty_weights`): audio OFF = 245 language / 0 audio, audio ON = 245 + 133 audio_tower + 1 projector, attn-only correctly drops the projector (0), mlp-only keeps it.

| Feedback | Resolution |
|---|---|
| `finetune_audio_layers` shifts positional `vision_tags`/`language_tags`/... callers (Codex P2, :61) | **Fixed.** Moved the flag to the very end of the signature, after all the tag-override params, so every existing positional slot (including `target_modules` and the four tag lists) is preserved. New `test_positional_tag_overrides_not_shadowed` locks it. |
| Projector branches ignore the attn/mlp flags (Codex P2, :185) | **Fixed.** The `embed_audio` / `embed_vision` projectors are feed-forward projections, so they are now gated under `finetune_mlp_modules`; attention-only runs no longer pick them up. New `test_projector_branches_gated_under_mlp_flag`. |
| Embedder branches are not actually Gemma-only; Mistral3 lists the same names in `empty_model.py` (Codex P2, :190) | **Fixed.** Added a model-type gate (`model_type` / `architectures` contains `gemma4` or `gemma3n`) around the audio + embedder branches, so a non-Gemma checkpoint that happens to carry `embed_vision.embedding_projection` / `vision_tower.patch_embedder.input_proj` keeps its byte-identical target set. New `test_embedder_branches_gated_to_gemma_model_type`. |
| Pre-check should use `re.fullmatch` not `re.search` (Gemini, :196) | **Already in place.** The `extra_branches` pre-check at line 195 already uses `re.fullmatch(branch, name, flags=re.DOTALL)`, matching PEFT's matcher exactly. No change needed. |
| Unnecessary `pass` (code-quality bot, :143) | **Declined.** `pass` as a block terminator is the existing idiom throughout this file (and the wider codebase); removing one would be inconsistent with the surrounding style. |

danielhanchen:
Re-checked against the latest commit `6735c559`. The bots re-fired the previous round's comments on the new commit; all are already addressed there:

- **Limit embedder branches to Gemma (Codex P2, :204)** - resolved in `6735c559`. The audio + embedder branches are now gated on `_is_gemma_mm` (a `model_type` / `architectures` check for `gemma4` / `gemma3n`), so the `embed_vision.embedding_projection` / `vision_tower.patch_embedder.input_proj` names shared with Mistral3 in `empty_model.py` are never appended on a non-Gemma checkpoint. `test_embedder_branches_gated_to_gemma_model_type` builds a Mistral3-typed model carrying those exact Linears and asserts they are not targeted, while the same names DO attach on a Gemma model. Verified on the real Gemma 3N E4B tree (245 language / 0 audio off, 245 + 133 + 1 on).
- **`re.fullmatch` pre-check (Gemini, :210)** - already in place. The `extra_branches` comprehension (line 209) uses `re.fullmatch(branch, name, flags=re.DOTALL)`; the suggestion matches the current code.
- **Unnecessary `pass` (code-quality bot, :143/:213)** - declined; `pass` block-terminators are the existing idiom throughout this file.

CI is green on this PR (14 checks passing).

danielhanchen:
@codex review

danielhanchen:
Good catch - fixed in `5b8274fc` (P1, :193).

On Gemma 4 each audio projection is a `Gemma4ClippableLinear` wrapper that appears in `named_modules()` as both the wrapper (`...q_proj`, not an `nn.Linear`) and its inner Linear (`...q_proj.linear`). The blanket optional `(?:\.linear)?` full-matched both, and since PEFT runs the regex over every module it would try to adapt the unsupported wrapper / double-process the inner Linear.

The fix classifies each leaf/path against the real Linear names (`linear_modules`, which already holds the appended `.linear` children) and emits a branch only for the form that genuinely exists - the bare module and/or the `.linear` child:

```python
def _linear_aware_branches(cores):
    out = []
    for core in cores:
        if any(re.fullmatch(r".*?" + core + r"\.linear", n, flags=re.DOTALL) for n in linear_modules):
            out.append(r"(?:.*?" + core + r"\.linear)")
        if any(re.fullmatch(r".*?" + core, n, flags=re.DOTALL) for n in linear_modules):
            out.append(r"(?:.*?" + core + r")")
    return out
```

So Gemma 3N's bare leaves and Gemma 4's `.linear` children each match exactly once, and the wrapper container is never matched. New `test_gemma4_audio_matches_inner_linear_not_wrapper` checks against the full module list (not just Linears) and reproduces/locks the regression: before the fix the regex matched 22 wrapper containers under `audio_tower`; after, 0. The bare audio Linears (subsample / output projections) and the inner `.linear` attention/ffw Linears all still attach. 80 tests pass; real Gemma 3N E4B unchanged (133 `audio_tower` + 1 projector).

(The `vision_tower.encoder` ClippableLinear wrappers are matched by the pre-existing main regex and handled by the `_create_and_replace` monkeypatch - that path is unchanged by this PR.)

danielhanchen:
@codex review

PR #815, opened 2026-06-23
feat(mlx): runtime bnb dequant -> MLX affine 4-bit on Apple Silicon

BardiaKoopah:
Replaces the early-bnb error introduced in #789 with a runtime dequantization path. When the loader detects a bitsandbytes-quantized repo (quant_method == 'bitsandbytes'), it uses bnb itself to dequantize the NF4 weights to fp16 and saves a clean copy to a temp dir, then recursively calls FastMLXModel.from_pretrained on the temp dir so MLX's existing affine 4-bit path re-quantizes the model.

Per Daniel Han's instruction: 'we don't need bitsandbytes - we can use MLX 4-bit - just wire it to those.' The user gets an MLX-quantized model; they never see bnb in the result.

Implementation details:

- _dequantize_bnb_to_tempdir lifts unsloth_zoo's Apple Silicon bnb stub (sys.meta_path finder) to reach the real wheel, calls AutoModelForCausalLM.from_pretrained(...).dequantize(), saves to a fresh tempdir with a clean config (no quantization_config), then restores the stub. Bnb registers torch custom ops at import that cannot be re-registered, so the real modules are cached after first import and reused on subsequent calls.

- Recursive load: the temp dir is passed back through from_pretrained with load_in_4bit kept enabled, so MLX's affine 4-bit applies as if the user had loaded a non-quantized base.

- Cleanup: the temp dir is rmtree'd in a try/finally after the recursive load completes. _hf_repo / _src_path are restored to the original repo name on the returned model so downstream code doesn't see implementation details.

- Fallback: ImportError on bitsandbytes -> the same clear, actionable error from #789 (suggesting the non-bnb variant). Other exceptions in the dequant path wrap the underlying error with chained context.

- bnb is not a hard dependency. Import happens inside the bnb branch only.

Verified locally on M2 with a 13-case harness:

| case                                | result                |
|-------------------------------------|-----------------------|
| TEXT non_quant + load_in_4bit=True  | success (no change)   |
| TEXT non_quant + load_in_4bit=False | success (no change)   |
| TEXT bnb + load_in_4bit=True        | success (~60s, new)   |
| TEXT bnb + load_in_4bit=False       | success (~48s, new)   |
| TEXT mlx_4bit + load_in_4bit=True   | success (no change)   |
| TEXT mlx_4bit + load_in_4bit=False  | success (no change)   |
| VLM non_quant + load_in_4bit=True   | success (no change)   |
| VLM non_quant + load_in_4bit=False  | success (no change)   |
| VLM bnb + load_in_4bit=True/False   | clean error (scope)   |
| VLM mlx_4bit + True/False           | success (no change)   |
| VERIFY: TEXT bnb -> MLX 4bit + gen  | success (~52s)        |

Final inference on the resulting MLX-quantized model produces sane output (e.g. 'The capital of France is Paris...').

Known limitations:

- VLM bnb is out of scope: AutoModelForCausalLM can't load Qwen3VLConfig. The dequant path raises a clean wrapped error pointing the user at the non-bnb variant. Wiring up AutoModelForImageTextToText + the mlx-vlm load path would be a follow-up.

- Memory: ~8 GB scratch copy of the fp16 model is materialized during dequant before MLX requantizes. Single-load is fine on a 16 GB Mac; sequential loads in one process need torch.mps.empty_cache between.

- Speed: ~45-60s per 4B model load (dominated by transformers load + fp16 save + MLX requant). One-time cost per load.

- Requires the nightly bitsandbytes wheel that ships Apple Silicon support, plus 'kernels<0.14' (newer kernels API breaks the bnb call site). Both are environment setup details, not loader concerns.

- unsloth_zoo currently stubs bnb out on Apple Silicon via a meta_path finder. The patch lifts the stub temporarily to reach the real wheel. Long-term, removing or conditioning that stub would simplify this path.

### DIFF
--- a/unsloth_zoo/mlx/loader.py
+++ b/unsloth_zoo/mlx/loader.py
@@ -2836,6 +2836,14 @@ def _dequantize_bnb_to_tempdir(source, *, token, trust_remote_code):
                 _strip_quant_meta(getattr(cfg, _sub, None))
         _strip_quant_meta(model.config)
         tmpdir = tempfile.mkdtemp(prefix="unsloth_bnb_dequant_")
+        # transformers 5.x: the dequantized model still carries weight-name
+        # conversions with no reverse op, so save_pretrained's
+        # revert_weight_conversion raises NotImplementedError. The dequantized
+        # weights need no conversion; clear it. No-op on 4.x (attr absent).
+        try:
+            model._weight_conversions = []
+        except Exception:
+            pass
         model.save_pretrained(tmpdir, safe_serialization=True)
         # VLMs need the full processor (image preprocessor + tokenizer); text
         # models only need the tokenizer. Use the right class so downstream
@@ -3769,6 +3777,19 @@ class FastMLXModel:
                     "Unsloth can dequantize and re-quantize via MLX"
                 )
             except Exception as _bnb_exc:
+                # bnb-4bit loading in transformers requires `accelerate`, which
+                # is excluded from the darwin-arm64 deps — exactly where this
+                # path runs. Detect that specific failure and point the user at
+                # the real fix instead of misreporting it as a bnb problem.
+                if "accelerate" in str(_bnb_exc).lower():
+                    raise ValueError(
+                        f"Unsloth: loading the bitsandbytes model '{model_name}' "
+                        "requires the `accelerate` package, which isn't installed "
+                        "on this machine.\n"
+                        "  - Install it with: pip install accelerate\n"
+                        "  - Then re-run; Unsloth will dequantize and re-quantize "
+                        "via MLX."
+                    ) from _bnb_exc
                 raise ValueError(
                     f"Unsloth: failed to dequantize the bitsandbytes model "
                     f"'{model_name}' "
--- a/unsloth_zoo/mlx/loader.py
+++ b/unsloth_zoo/mlx/loader.py
@@ -20,11 +20,15 @@ No GPU deps: uses mlx-lm (text) and mlx-vlm (VLM) instead of unsloth.models
 (which pulls in CUDA kernels).
 """
 
+import gc
 import json
 import importlib
 import inspect
 import math
 import os
+import re
+import shutil
+import sys
 import tempfile
 import types
 import warnings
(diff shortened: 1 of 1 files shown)


oobabooga:
I ran this end to end on a real M1 (macOS 14.5) to check it. Good news first: the approach is sound. PyPI `bitsandbytes==0.49.2` runs on Apple Silicon now (I confirmed an NF4 dequant on MPS), so the #789 block is no longer needed, and on transformers 4.57.1, with accelerate present, your path works fully: bnb dequant, MLX 4-bit requant, coherent generation, and the bnb stub restored cleanly.

Two things stop it working on a clean Apple Silicon install. Both are small, and I verified the fixes on the Mac:

1. transformers 5.5.0 (the newest version we support). `save_pretrained` on the dequantized model crashes with `NotImplementedError` inside transformers' `revert_weight_conversion`: the dequantized model still carries weight-name conversions that have no reverse op. Fix is to clear them right before the `save_pretrained` call in `_dequantize_bnb_to_tempdir`:

   ```python
   try:
       model._weight_conversions = []
   except Exception:
       pass
   ```

   I patched exactly this and re-ran the full path green on 5.5. It is a no-op on 4.x, so it is safe across the whole supported range.

2. accelerate. bnb-4bit loading in transformers requires accelerate, but our pyproject excludes it on darwin-arm64, which is exactly where this path runs. So on a clean Mac install the load fails, and your handler reports it as "failed to dequantize, try the non-bnb variant", which points the user the wrong way. Cleanest fix is to add accelerate to the darwin-arm64 dependency set so it works out of the box. If we would rather keep the Mac install lean, then instead detect the accelerate error and tell the user to `pip install accelerate`, just don't report it as a bitsandbytes problem.

With those two in it is mergeable. Push when ready and I will re-run the full path on the Mac to confirm.

BardiaKoopah:
For the transformers 5.x save crash, I clear model._weight_conversions right before save_pretrained. And for the accelerate issue, the handler now detects a missing accelerate failure and tells the user to pip install accelerate instead of misreporting it as a bnb problem. I kept the Mac deps lean rather than adding accelerate to darwin-arm64, but I'm happy to bundle it instead if you'd rather it work out of the box, just lmk

oobabooga:
Both fixes check out on the Mac: 5.5.0 loads and generates cleanly, and without accelerate you get the correct pip install message. Keeping accelerate as a manual install on Mac (not a hard dep) is the right call. Merging.

oobabooga:
@codex review

PR #816, opened 2026-06-23
Generalize pad_token repair into a shared module

danielhanchen:
## What

Generalizes the `pad_token` repair in `patch_tokenizer` into a shared single source of truth, `unsloth_zoo/pad_token.py` (`fix_pad_token`), so the same logic can back the equivalent fix in unsloth.

## Why

The previous heal was an inline regex scan over a delimiter-joined string of `added_tokens_decoder`. It worked for the common cases but had a few sharp edges:

- The closed-bracket grade (`endswith(">", "|>", "]", ")")`) ran on the matched prefix (e.g. `<SPECIAL_`) rather than the resolved full token (e.g. `<SPECIAL_20>`), so prefix families were mis-graded.
- `check_pad_token[0]` was indexed even after the code decided the candidate was not a single token, which can `IndexError` on a zero-token candidate.
- The model name used in the log was computed twice and reset to `"Model"` when only one of `model` / `model_config` was present.

## What changed

- New `unsloth_zoo/pad_token.py` (AGPL-3.0) with `fix_pad_token(tokenizer, model=None, model_config=None, *, is_vision_model=None, allow_add=True)`. It scans `added_tokens_decoder` token by token, keeps the vision-token denylist for text-only models (issue #4104), prefers a closed reserved token that encodes to exactly one in-vocab id, and only adds a brand new `<｜PAD▁TOKEN｜>` (then raises `RuntimeError`) when nothing suitable exists.
- `patch_tokenizer` now delegates to `fix_pad_token` and keeps the existing `model.config` / `generation_config` `pad_token_id` stamping and the `max_length` update. `POSSIBLE_RESERVED_TOKENS` / `VISION_RESERVED_TOKENS` are re-exported from the new module for backwards compatibility.
- The reserved-token list is expanded (Llama-4, Kimi `[PAD]`, Cohere `<PAD>`, GLM `[MASK]`, Granite `<fim_pad>`, DeepSeek `<｜▁pad▁｜>` / `<|EOT|>`, Phi-4 `<|dummy_`, Qwen `<|endoftext|>`, etc.). `<|endofprompt|>` is left commented out since it clashes with GPT-OSS.

## Behavior

- No-op when the tokenizer already has a valid pad token distinct from eos (verified against `unsloth/Mistral-Medium-3.5-128B`, which ships a valid `<pad>`).
- Heals a text-only Qwen3 / Qwen2.5 tokenizer that ships a vision pad token to `<|endoftext|>` (never a vision token).
- Idempotent on a second call.

## Tests

`tests/test_pad_token.py` (offline, no network, no torch): heal, no-op, `pad == eos`, missing pad, `allow_add=False` defer, manual fallback raises, the closed-bracket and zero-token regressions, and the vision-token-never-selected invariant.

Companion unsloth change (swapping the narrow `_fix_vision_pad_token` for this shared call, behind a guarded import) follows in unsloth#6524.

### DIFF
--- a/tests/test_pad_token.py
+++ b/tests/test_pad_token.py
@@ -126,3 +126,13 @@ def test_vision_token_never_selected_for_text_model():
     tok = _qwen3_text()
     fix_pad_token(tok)
     assert tok.pad_token not in VISION_RESERVED_TOKENS
+
+
+def test_vision_model_keeps_its_vision_pad():
+    # A real multimodal model (model_type without "vl", e.g. llava) must be
+    # detected as vision so its vision pad_token is left untouched.
+    tok = _qwen3_text()  # pad_token = "<|vision_pad|>"
+    cfg = type("Cfg", (), {"model_type": "llava"})()
+    res = fix_pad_token(tok, model_config = cfg)
+    assert res["changed"] is False
+    assert tok.pad_token == "<|vision_pad|>"
--- a/tests/test_pad_token.py
+++ b/tests/test_pad_token.py
@@ -2,9 +2,23 @@
 # Copyright 2023-present the Unsloth team. All rights reserved.
 """Offline tests for unsloth_zoo.pad_token.fix_pad_token (no network, no torch)."""
 
+import importlib.util
+import os
+
 import pytest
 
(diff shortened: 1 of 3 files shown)


danielhanchen:
@codex review

danielhanchen:
@codex review

PR #821, opened 2026-06-23
Feat(mlx): Wire report_to wandb/tensorboard for standalone MLXTrainer

BardiaKoopah:
## Summary

Wires `report_to="wandb"` / `"tensorboard"` for the standalone
`MLXTrainer`, so logging works directly from a notebook — not only
through Studio. The `report_to` field already existed on
`MLXTrainingConfig` but was never read anywhere in `trainer.py`, so a
user setting `report_to="wandb"` got silently nothing and had to
hand-register `add_step_callback`. CUDA users get this for free from
`TrainingArguments`; this closes that gap for the MLX path.

## Implementation

- New `_setup_report_to_callbacks()` method, called from `train()`. It
  parses `report_to` (accepts a string or a list), and for each backend
  requested, initializes a W&B run and/or a TensorBoard `SummaryWriter`,
  then registers internal step/eval callbacks via the existing
  `add_step_callback` / `add_eval_callback` hooks.
- Log keys mirror Studio's `worker.py` exactly so notebook and Studio
  runs chart identically:
  - step: `train/loss`, `train/learning_rate`, `train/tokens_per_sec`,
    `train/peak_gb`, `train/num_tokens`, and `train/grad_norm`
    (conditional).
  - eval: `eval/loss`, `eval/perplexity`.
- Handles are stored on `self._report_to_handles` and closed in the
  existing `finally` block in `train()` (`SummaryWriter.close()` /
  `wandb.finish()`), so they flush even if training raises.

## Deliberate choices

- **W&B auth via environment** (`WANDB_API_KEY` / `WANDB_PROJECT`,
  default project `"unsloth-mlx"`). No token field added to
  `MLXTrainingConfig` — matches how a CUDA `TrainingArguments` user
  already authenticates, and keeps the config surface unchanged.
- **Every logger call is wrapped** so a logging failure never interrupts
  training. A failed `wandb.init` (e.g. wandb not installed, or no key)
  degrades to a no-op with a printed warning rather than raising.
- **TB omits `train/num_tokens`; W&B includes it.** This asymmetry is
  intentional — it matches `worker.py`'s existing TB and W&B blocks
  exactly.
- **Eval logging is loss/perplexity only**, matching `worker.py`.
  Forwarding the newly added `eval_mean_token_accuracy` would require a
  wider eval-callback signature (currently `fn(step, eval_loss,
  perplexity)`); left as a follow-up.

## Testing

Verified on short gemma-3-4b-it LoRA runs:

- `report_to="none"` (default): trains normally, no writer created —
  confirmed no regression for the common path (`_report_to_handles`
  absent).
- `report_to="tensorboard"` + `eval_steps`: training completes and an
  `events.out.tfevents` file is written to `output_dir/runs` with the
  expected tags — `train/loss`, `train/learning_rate`,
  `train/tokens_per_sec`, `train/peak_gb`, `train/grad_norm`,
  `eval/loss`, `eval/perplexity`.
- `report_to="wandb"` (offline mode): run opens, all six `train/*` keys
  log with real values (verified via the run summary), and the run
  finishes cleanly.
- `report_to="wandb"` with wandb not installed: degrades to a no-op with
  a printed warning, training continues.
- List form (`["wandb", "tensorboard"]`) and `finally`-block cleanup
  verified by inspection.


oobabooga:
I tested it on a real Apple Silicon box (macOS 14.5, mlx-lm with wandb and tensorboard) on short SmolLM-135M LoRA runs, and the core feature works well:

- TensorBoard logs `train/loss`, `train/learning_rate`, `train/tokens_per_sec`, `train/peak_gb`, `train/grad_norm`, `eval/loss`, `eval/perplexity`, with `train/num_tokens` left out on purpose.
- wandb logs all of those plus `train/num_tokens`.
- `report_to="none"` stays completely inert (no writer, no callback), and the list form (`["wandb","tensorboard"]`) opens both.

So the wiring and the key names (which match `worker.py`) are all correct.

One thing to fix before this is mergeable. `_setup_report_to_callbacks()` adds the step/eval callbacks every time `train()` is called, but nothing ever removes them. So if someone calls `train()` twice on the same trainer (easy to do in a notebook), the callbacks pile up and the second run double-logs. I reproduced it: the callback list grows from 1 to 2, and two 4-step runs write 12 loss points total instead of the expected 8 (the second run logs everything twice). These callbacks are added internally, so a user has no way to clean them up themselves.

The fix is small: remember the callbacks when you add them, and remove them in the `finally` block that already closes the handles. Store the pair right after registering:

```python
self.add_step_callback(_on_step)
self.add_eval_callback(_on_eval)
self._report_to_handles = (wandb_run, tb_writer)
self._report_to_callbacks = (_on_step, _on_eval)
```

Then unwind them where you close the writers:

```python
for _cb in getattr(self, "_report_to_callbacks", ()):
    if _cb in self._step_callbacks: self._step_callbacks.remove(_cb)
    if _cb in self._eval_callbacks: self._eval_callbacks.remove(_cb)
self._report_to_handles = (None, None)
self._report_to_callbacks = ()
```

I tested this exact change on the same box: after it, calling `train()` twice logs cleanly with no duplicates, and both backends still work. That is the only blocker I found. Add this and I think it is good to merge.

BardiaKoopah:
@oobabooga good catch. Fixed as you suggested and storing the callbacks when registered and remove them in the same finally that closes the writers. Confirmed with your repro.

oobabooga:
Confirmed your fix on the same box: a repeated `train()` no longer piles up callbacks or double-logs, and both backends still log cleanly. Since then #824 (NEFTune) landed and conflicted with this, so I merged `main` into your branch and resolved it, your `report_to` code untouched and the two features coexisting. CI is green. Thanks @BardiaKoopah. Merging.

oobabooga:
@codex review

PR #822, opened 2026-06-23
Feat(mlx): Add early stopping + load_best_model_at_end to MLXTrainer

BardiaKoopah:
## Summary

Adds early stopping and best-model tracking to `MLXTrainer`, matching HF
Trainer's `EarlyStoppingCallback` + `load_best_model_at_end`. Previously a
run that overfit kept only the final (often worse) adapters, with no way
to automatically retain the best checkpoint or stop once eval stopped
improving.

## New config fields (`MLXTrainingConfig`)

- `load_best_model_at_end: bool = False`
- `metric_for_best_model: str = "eval_loss"`
- `greater_is_better: bool = False`
- `early_stopping_patience: int = 0`  (0 = disabled)

## Behavior

- After each eval, reads `metric_for_best_model` from the eval metrics
  dict. On improvement, saves adapters to `output_dir/best/` and resets
  the patience counter; on no-improvement, increments it and, once
  `early_stopping_patience` is reached, sets `stop_requested` to halt.
- If `load_best_model_at_end`, the best adapters are reloaded in-place
  (`model.load_weights`, the same mechanism `resume_from_checkpoint`
  uses) before the final save.
- Both features require `eval_steps > 0`. With no eval, tracking is a
  no-op: `_best_step` stays `None` and the reload is skipped — confirmed
  not to create a `best/` dir or alter behavior.
- Unknown `metric_for_best_model` raises a clear `ValueError` listing the
  available eval keys, rather than a mid-run `KeyError`.
- NaN eval loss counts as non-improvement (`nan < best` is `False`), so a
  diverged run early-stops cleanly and never overwrites `best/` with NaN
  weights.

## Notes

- Defaults to `eval_loss` / `greater_is_better=False`. Once #820
  (`eval_mean_token_accuracy`) merges,
  `metric_for_best_model="eval_mean_token_accuracy"` with
  `greater_is_better=True` works automatically — the metric is read by
  name from the eval dict.
- Training loop, gradient path, and existing checkpoint logic are
  untouched; best-model saving writes to a separate `best/` dir and does
  not interact with `save_steps` checkpoints.

## Testing

Short gemma-3-4b-it LoRA runs:

- **Feature off (defaults):** trains normally, no `best/` dir, no early
  stop — confirmed no regression for existing users.
- **`load_best_model_at_end=True`:** `best/adapters.safetensors` written,
  "Restored best model from step N" prints, model restored before final
  save.
- **`eval_steps=0` + `load_best_model_at_end=True`:** no-op, no crash.
- **`early_stopping_patience` with a divergent LR:** eval loss worsens,
  trainer prints the early-stop message and halts before `max_steps`.
- **`greater_is_better=True`:** best tracks the highest eval value (the
  opposite comparison branch), confirming the path the accuracy metric
  will use.
- **Divergence/NaN:** `best/` verified to contain only finite weights
  (476 tensors, 0 NaN/Inf) — best-tracking refused to save a diverged
  model.
- Patience comparison and the unknown-metric `ValueError` additionally
  verified by direct logic checks.


oobabooga:
Nice work on this, the feature is solid and the description made it easy to follow. I ran it on a real Apple Silicon box (M1, MLX 0.31.2 and mlx_lm 0.31.3) with short SmolLM-135M LoRA runs and confirmed the main paths: `load_best_model_at_end` saves `best/` and restores it before the final save, `early_stopping_patience` stops the run early under a diverging LR, and with both features off nothing changes (no `best/`, runs to `max_steps`) even when eval is active. So the core is working.

There's one bug to fix before this is mergeable, in the `_improved` check (the "is this eval the new best?" decision that drives both saving `best/` and the patience counter).

The problem is the first eval. That check is:

```python
_improved = (
    self._best_metric is None
    or (_cur > self._best_metric if _greater else _cur < self._best_metric)
)
```

On the first eval `self._best_metric is None`, so `_improved` is `True` no matter what the value is. If that first eval is NaN (easy to hit with a high LR), the NaN gets recorded as the best (`_best_step=1`, `_best_metric=nan`) and `best/` is written for it. From then on nothing can ever beat it, because any real value compared against NaN is `False` (`3.0 < nan` is `False`). So a later genuinely-good eval never replaces it, and `load_best_model_at_end` restores that diverged first checkpoint. Your description says NaN counts as non-improvement, and that's true once a good best exists, but not for the very first eval.

I reproduced it directly: with a hot LR the first eval is NaN and the run ends with `_best_step=1`, `_best_metric=nan`.

The fix is to treat NaN as "not an improvement" by rejecting it up front (`_cur == _cur` is `False` only for NaN):

```python
_improved = (
    _cur == _cur  # reject NaN: a diverged eval must never become "best"
    and (
        self._best_metric is None
        or (_cur > self._best_metric if _greater else _cur < self._best_metric)
    )
)
```

With that, a first-eval NaN behaves like the mid-run divergence you already handle: nothing is written to `best/`, `_best_step` stays `None`, the restore is skipped, and early stopping still fires. I verified the fix on the same box, and the happy-path and early-stop runs are unchanged.

That single change makes it mergeable. One optional thing you could clean up while you're in there: when only `early_stopping_patience` is set (no `load_best_model_at_end`), you still write `best/` on every improvement, but early stopping only needs the metric comparison, not the saved adapters. Gating the `save_trainable_adapters` call on `load_best_model_at_end` avoids those extra writes. Not a blocker, just tidier.

BardiaKoopah:
@oobabooga Thanks for the clean repro. I fixed it with the _cur == _cur NaN rejection you suggested, and also with a hot LR the first eval is a real value, later evals go NaN, and the NaN evals are now correctly rejected as non-improvements instead of overwriting best. Early stopping still fires btw and the good checkpoint is restored.

oobabooga:
Confirmed the NaN guard on a real M1: first-eval NaN no longer poisons best, and a finite-then-NaN run keeps and restores the good checkpoint. Thanks for the quick fix. Merging.

oobabooga:
@codex review

PR #823, opened 2026-06-23
tokenizer_utils: match IGNORED_TOKENIZER_NAMES case-insensitively in fix_untrained_tokens

oobabooga:
Follow-up to unslothai/unsloth#6620.

`fix_untrained_tokens` resets a model's untrained token embeddings, but it is supposed to skip the models listed in `IGNORED_TOKENIZER_NAMES` (the Qwen2.5-Coder tokenizers, which should be left alone). That skip never actually fired.

The list is built fully lowercased (it is defined in the `unsloth` package), but the check compared it against the raw model id, which is mixed-case:

```python
# IGNORED_TOKENIZER_NAMES contains "unsloth/qwen2.5-coder-7b-instruct"
# model.config._name_or_path is   "unsloth/Qwen2.5-Coder-7B-Instruct"
if model.config._name_or_path in IGNORED_TOKENIZER_NAMES:   # never True
    return
```

The mixed-case id was never found in the lowercased set, so the early return never ran and `fix_untrained_tokens` modified the embeddings of the very models the list exists to protect.

### Fix

Lowercase the id before the check so it matches the lowercased list:

```python
if (model.config._name_or_path or "").lower() in IGNORED_TOKENIZER_NAMES:
    return
```

The `or ""` keeps a `None` id from crashing `.lower()` (the old `in` check silently tolerated `None`). This is the zoo half of the same bug; the `unsloth` side is unslothai/unsloth#6620, and both are needed to fully skip these models.

### Test

`tests/test_ignored_tokenizer_casing.py` (CPU-only, stubbed model, no download): a mixed-case ignored id now returns early, and a non-ignored id still runs. Fails before the change, passes after.


danielhanchen:
@codex review

oobabooga:
@codex review

oobabooga:
@codex review

oobabooga:
@codex review

oobabooga:
@codex review

PR #824, opened 2026-06-23
Feat(mlx): Add NEFTune (neftune_noise_alpha) for text models to MLXTrainer

BardiaKoopah:
# Add NEFTune (`neftune_noise_alpha`) for text models to MLXTrainer
 
Adds NEFTune (Noisy Embedding Fine-Tuning) to the MLX trainer, matching the CUDA path. Unsloth's CUDA code wires NEFTune via TRL (`tokenizer_utils.py` imports `neftune_post_forward_hook` and registers it with `embed_tokens.register_forward_hook`), and it's exposed through HF `TrainingArguments` / TRL `SFTConfig` as `neftune_noise_alpha`. MLX had no equivalent. This closes that gap for **text models**.
 
NEFTune adds uniform noise `±alpha/sqrt(seq_len·dim)` to the input embeddings during training (no-op in eval), which regularizes instruction fine-tuning on small datasets.
 
## Config
 
`neftune_noise_alpha: float = 0.0` (0 = disabled; the paper uses 5–15).
 
## Implementation note — why `__class__` reassignment
 
The PyTorch approach (forward hook) has no MLX equivalent — MLX `nn.Module` has no `register_forward_hook`. Two obvious MLX substitutes both fail:
 
- Overriding `emb.__call__` on the *instance* silently no-ops — MLX/Python resolve `__call__` on the type, not the instance (verified).
- Swapping `embed_tokens` for a wrapper module *does* intercept, but breaks tied-embedding models (Qwen/Llama/Gemma): the CCE path reads `embed_tokens.weight` as the LM head (`_get_lm_head_layer` returns `embed_tokens` for tied models), and a wrapper hides `.weight` (`AttributeError` on Qwen2.5).
Instead, `_install_neftune` makes `embed_tokens` an instance of a dynamically-created subclass of its own type (`emb.__class__ = _NEFTuneEmbed`), overriding only `__call__`. Because it's a real subtype, the object is otherwise unchanged: `.weight` / `.scales` / `.bits` inherit natively (tied LM-head works), `__call__` resolves on the subtype (interception fires), and teardown is a one-line `emb.__class__ = base` in the `finally`. Verified on the real quantized tied Qwen2.5-0.5B that `trainable_parameters()`, full-model forward, adapter saving, and `isinstance` all behave correctly while active.
 
## Scope
 
Text models only. VLMs print a clear "not yet supported" notice and train normally (noise skipped) — the merged vision+text embedding path needs separate handling, left as a follow-up. (Note: even the community mlx-tune project, which has DPO/ORPO/GRPO/CPT on MLX, does not implement NEFTune.)
 
## Testing (Qwen2.5-0.5B, alpaca-cleaned)
 
- alpha=0 vs alpha=5: "NEFTune enabled" prints for alpha=5, training losses differ (noise lands), embedding class restored to base after `train()`.
- Structural safety on the real quantized tied model: param collection, forward, adapter save, isinstance, and class restore all verified while the subclass is active.
- VLM run (gemma-3-4b): prints the "not supported" notice, trains without crashing.

### DIFF
--- a/unsloth_zoo/mlx/trainer.py
+++ b/unsloth_zoo/mlx/trainer.py
@@ -78,6 +78,7 @@ from .utils import (
     apply_gradient_checkpointing,
     remove_gradient_checkpointing,
     _is_vlm_model,
+    _get_text_model,
 )
 from .compile import (
     build_compile_policy,
@@ -502,6 +503,7 @@ class MLXTrainingConfig:
 
     # Eval
     eval_steps: int = 0  # 0 = disabled
+    neftune_noise_alpha: float = 0.0  # 0 = disabled (text models only)
 
     # SFT-specific (from SFTConfig, for API compat)
     dataset_text_field: str = "text"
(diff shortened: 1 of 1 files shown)


oobabooga:
Verified your NEFTune path on real Apple Silicon (MLX 0.31.2), not just by reading it. I ran short LoRA fits on a tied, quantized model (SmolLM-135M-4bit): alpha=0 is a clean no-op, alpha=5 prints the notice and shifts the loss, `.weight` stays readable so the tied LM head still works, and the embedding class is restored after training. The subclass-over-wrapper choice holds up.

I also checked the one thing that could silently break this: `mx.random` inside `mx.compile`. Because the trainer threads `mx.random.state` through the compiled step, the noise genuinely resamples every step instead of freezing at trace time. I confirmed distinct draws step to step both in eager mode and under real `mx.compile`, and that eval stays a clean no-op. No regression against main.

Thanks @BardiaKoopah. Merging.

oobabooga:
@codex review

PR #833, opened 2026-06-26
dataset_utils: add mask_thinking_tokens to mask the </think> token

Sushankthatipally:
Adds `mask_thinking_tokens`, a convenience function in the same spirit as
`train_on_responses_only`, that masks the thinking closing token (`</think>`
by default) to `-100` in the labels so it isn't trained on.

Requested in unslothai/unsloth#6695. Nemotron Ultra masks out `</think>`
during training so the model isn't trained on the reasoning terminator and
doesn't memorise a fixed reasoning length. This turns that into a one-liner,
composable after `train_on_responses_only`:

    trainer = train_on_responses_only(trainer, instruction_part=..., response_part=...)
    trainer = mask_thinking_tokens(trainer)

The function mirrors `train_on_responses_only`'s interface (`tokenizer`,
`return_function`, `num_proc`, `torch.Tensor`/list labels, `IterableDataset`)
and handles `</think>` whether it tokenises to one id or several. It's purely
subtractive: only positions matching `think_token` are set to `-100`; existing
`-100`s and every other label are left untouched. Tokenised datasets that don't
yet carry labels fall back to deriving them from `input_ids`.

Added offline CPU-only unit tests in `tests/test_mask_thinking_tokens.py`
covering single/multi-token `</think>`, multiple occurrences, label
preservation, tensor and list inputs, and the no-labels fallback.

Happy to follow up with a companion re-export from `unsloth.chat_templates`
(like `train_on_responses_only`) once this lands, and to rename the function
if you'd prefer something else.

### DIFF
--- a/tests/test_mask_thinking_tokens.py
+++ b/tests/test_mask_thinking_tokens.py
@@ -1,0 +1,134 @@
+# SPDX-License-Identifier: AGPL-3.0-only
+# Copyright 2023-present the Unsloth team. All rights reserved.
+"""Offline tests for unsloth_zoo.dataset_utils.mask_thinking_tokens (no network)."""
+
+import importlib.util
+import os
+
+import pytest
+import torch
+
+# Load dataset_utils.py directly from its file so the test is self-contained and
+# does not run unsloth_zoo/__init__.py (which pulls the full stack).
+_DATASET_UTILS_PATH = os.path.join(
+    os.path.dirname(os.path.dirname(os.path.abspath(__file__))),
+    "unsloth_zoo", "dataset_utils.py",
+)
+_spec = importlib.util.spec_from_file_location("_unsloth_dataset_utils_offline", _DATASET_UTILS_PATH)
+_dataset_utils = importlib.util.module_from_spec(_spec)
+_spec.loader.exec_module(_dataset_utils)
+mask_thinking_tokens = _dataset_utils.mask_thinking_tokens
+
+
+class FakeTokenizer:
+    """Minimal stand-in: encodes the think token to a fixed id sequence."""
+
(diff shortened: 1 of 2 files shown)


Sushankthatipally:
Thanks for the reviews! Pushed an update addressing most of the points:

- Trainer-vs-Dataset guard: added a TypeError if a Dataset is passed as the
  first arg (it would otherwise silently no-op), pointing to return_function=True.
- Tokenizer resolution: switched to getattr(..., None) or getattr(..., None) so a
  None processing_class falls back to tokenizer, plus a clear ValueError when no
  tokenizer is available.
- Tensor preservation: use_tensors is now set when labels is a tensor too.
- Collator: switch to DataCollatorForSeq2Seq for non-packing trainers (same as
  train_on_responses_only) so the masked -100s aren't rebuilt from input_ids
  during collation.
- Tests: added coverage for the TypeError, the ValueError, and tensor-label
  preservation (13 tests, CPU-only/offline).

Two I held off on, with reasoning:

- Context-merged </think> encodings: the reasoning models this targets (Qwen3,
  DeepSeek-R1, Nemotron) tokenize </think> as an atomic special token that doesn't
  merge with adjacent whitespace, so the bare-id match is correct. Routing through
  _find_common_token_ids also pulls in its [0] fallback (which can resolve to
  token-id 0 / pad) for tokens that don't encode cleanly. Happy to add
  context-probing if you'd like to support non-special-token templates.
- Filtering fully-masked rows: train_on_responses_only filters because it masks the
  entire instruction, which can zero out a row. This masks only the single
  terminator token, so a row with any other trainable label can't become fully
  -100. Glad to add the filter for symmetry if you'd prefer it.

PR #834, opened 2026-06-27
Auto-detect instruction/response parts in train_on_responses_only

danielhanchen:
### Summary

`train_on_responses_only` currently requires the caller to hand write `instruction_part` and `response_part` for every model, for example:

```python
trainer = train_on_responses_only(
    trainer,
    instruction_part = "<|start_header_id|>user<|end_header_id|>\n\n",
    response_part    = "<|start_header_id|>assistant<|end_header_id|>\n\n",
)
```

This PR lets it auto-detect both from the tokenizer chat template when neither is passed, so the same call works for any supported model:

```python
trainer = train_on_responses_only(trainer)
```

If a tokenizer already has `_unsloth_input_part` / `_unsloth_output_part` set (the Ollama path), those are still used. Passing both parts manually behaves exactly as before. Passing only one now raises a clearer error.

### How detection works

A new helper `get_chat_template_parts(tokenizer)` renders a short multi turn probe conversation with sentinel content and reads the markers off the rendered string, model agnostic:

- `response_part` is the assistant generation prompt, obtained by diffing `apply_chat_template(..., add_generation_prompt=True)` against the plain render and stripping the shared turn terminator.
- `instruction_part` is the text that precedes user content (the user header), taken from the gap between assistant content and the next user content, with the leading terminator removed.
- A few practical fixes are baked in: take the most common gap across turns (Qwen3 injects an empty think block into the last assistant turn), strip trailing spaces that would merge into the first content token (Mistral `[/INST]`), and a headerless fallback for Mistral style `[INST]` / `[/INST]`.

Finally it validates: the detected markers must appear as token cores in a tokenized probe, otherwise masking would silently train on nothing (some older templates use role tags that are not atomic tokens and whose ids shift with context). In that case it raises and asks for explicit parts instead of failing quietly.

### Models verified

Detection plus an end to end mask check (decode the unmasked tokens, confirm they equal the assistant turns and contain no user text) passes on:

| Family | instruction_part / response_part |
| --- | --- |
| Llama 3.1 / 3.2 / 3.3 | `<|start_header_id|>user<|end_header_id|>\n\n` / `...assistant...` |
| Llama 4 | `<|header_start|>user<|header_end|>\n\n` / `...assistant...` |
| Gemma 2 / 3 / 3n | `<start_of_turn>user\n` / `<start_of_turn>model\n` |
| Qwen 2.5 / 2.5-Coder / 3 / 3-MoE | `<|im_start|>user\n` / `<|im_start|>assistant\n` |
| Phi 3 / 3.5 | `<|user|>\n` / `<|assistant|>\n` |
| Phi 4 | `<|im_start|>user<|im_sep|>` / `<|im_start|>assistant<|im_sep|>` |
| Mistral v0.3 / Small / Nemo | `[INST]` / `[/INST]` |
| DeepSeek R1 distill (Qwen, Llama) | `<｜User｜>` / `<｜Assistant｜>` |
| SmolLM2 / SmolLM3 | `user\n` / `assistant\n`, `<|im_start|>...` |
| Qwen2-VL, Llama 3.2 Vision | unwraps the processor, same as text |

For each of these the auto detected strings match the values used in the conversational notebooks. gpt-oss masks the final channel correctly but keeps its existing bespoke handling for reasoning channels. Tokenizers whose role tags are not atomic (older TinyLlama / Zephyr style) now raise a clear message asking for explicit parts rather than masking nothing.

### Notes

- Backwards compatible: explicit `instruction_part` / `response_part` and the `_unsloth_input_part` path are unchanged.
- No new dependencies. `re` and `collections.Counter` are imported locally inside the helper.
- `get_chat_template_parts` is exported for direct use.


danielhanchen:
Expanded the validation to 54 tokenizers across about 30 architecture families. For each one detection runs, then an end to end mask check decodes the unmasked tokens and confirms they equal the assistant turns with no user text leaking. All 54 pass:

- Llama 2, 3, 3.1, 3.2, 3.3, 4, Nemotron Nano
- Gemma 2, 3, 3n (1b, 4b, 27b, 270m)
- Qwen 2.5, 2.5-Coder, 2-VL, 3, 3-VL, 3-MoE (30B, 235B), QwQ
- Phi 3, 3.5, 3-medium, 4, 4-mini
- Mistral v0.3, Small, Nemo, Mixtral, Devstral, Magistral, Ministral
- DeepSeek R1 distill (Qwen, Llama)
- SmolLM2, SmolLM3, OLMo 2, Falcon3, InternLM 2.5, EXAONE 3.5, GLM 4, IBM Granite 3.1, Cohere Command R, Yi 1.5, StableLM 2 Zephyr, OpenHermes
- gpt-oss 20b and 120b (final channel)

Llama 2 and Mixtral style `[INST]` / `[/INST]` templates auto-detect to `[INST]` and `[/INST]`. Tokenizers whose role tags are not atomic tokens (older TinyLlama and Zephyr style) raise the clear "pass instruction_part and response_part" message instead of silently masking nothing. I confirmed the existing matcher also masks nothing for those when given the markers manually, so the raise is consistent with current behaviour rather than a regression.

The second commit strips leading separators (space, bos, eos) from the detected markers and validates each against a tokenized probe, trying a leading space variant for SentencePiece tokenizers.

danielhanchen:
Validated against every model referenced in unslothai/notebooks. I scanned all 433 notebooks plus the READMEs and python scripts, collected 240 model repo ids, then normalized away the quant and dtype suffixes (`-bnb-4bit`, `-unsloth-bnb-4bit`, `-FP8`, `-BF16`) and dropped non conversational repos (TTS, OCR, embeddings, BERT, whisper, datasets). That leaves 158 unique repos. For each I download the tokenizer config only, run detection, then a full mask check that decodes the unmasked tokens and confirms they equal the assistant turns with no user text leaking.

Results over the 158:

- 89 PASS (conversational models with a chat template)
- 1 FAIL: `Erland/DeepSeek-R1-0528-Qwen3-8B`, a third party repo whose chat template has a typo. It emits lowercase `<｜assistant｜>`, which is not a token in the vocab (5 pieces), instead of the real `<｜Assistant｜>` (1 token). The official `unsloth/DeepSeek-R1-0528-Qwen3-8B` passes. Detection faithfully reflects whatever the template emits.
- 1 SKIP, safe raise: `zephyr-sft`, whose role tags are not atomic tokens. The existing matcher also masks nothing here when given the markers manually, so raising is consistent with current behaviour.
- 39 base or pretrain models with no chat template (Llama 3.1 8B base, Qwen2.5 7B base, gemma 2/7b, mistral 7b, etc). Auto detect raises a clear message there, which is correct since you would not mask responses on a base model.
- 6 repo ids that 404 as written (Llama 3.1 405B, codellama 34b, and similar).
- 22 unreleased or future tokenizers that do not load in the current transformers (gemma 4, Ministral 3, ERNIE 4.5 VL, GLM 4.7). For gemma 4 I registered its special tokens and verified the bundled template detects to `<|turn>user\n` and `<|turn>model\n` and masks correctly, so the logic is ready once the tokenizer ships.

Net: every loadable conversational tokenizer in the notebooks auto detects correctly, the single FAIL is a broken third party template whose official counterpart passes, and base models raise the clear "pass instruction_part and response_part" message rather than masking nothing.

danielhanchen:
Two more rounds of validation: a token exact mask audit, and a retry of the unreleased tokenizers on transformers 5.x.

### Token exact mask audit

The docs section on training on completions only masks out inputs and trains on assistant responses. I confirmed this exactly, not just by substring, using char to token offset mapping over a 3 turn conversation:

- every token of every assistant message CONTENT is unmasked (trained on)
- no token of any user or system message CONTENT is ever unmasked

Result: 24 of 24 models pass, including Llama 3 and 4, Gemma 2/3/3n, Qwen 2.5/3/2-VL/3-VL, Phi 3.5/4/4-mini, Mistral and Mixtral and Mistral-Small, Llama 2, DeepSeek R1, SmolLM2, gpt-oss, Granite, Cohere Command R, and the Llama and Qwen vision models.

I also checked equivalence to the exact markers in the docs: for the Llama 3 family and the Gemma family the auto detected markers produce byte identical labels to the hand written `instruction_part` and `response_part` from the guide. Llama 4 renames the header tokens to `<|header_start|>` and is detected correctly for that checkpoint.

The masks differ from a naive incremental render only on scaffolding tokens (an empty Qwen3 think block, the gpt-oss channel header, the Llama 2 `<s>` between turns). Those are not content and match what the existing manual markers already produce, so there is no input leakage and no change in behaviour.

### transformers 5.x retry of the unreleased tokenizers

The earlier LOAD-ERR rows were unreleased tokenizer classes that do not exist in transformers 4.57. On transformers 5.12.1 they load, and all pass the same token exact check:

| Model | instruction_part / response_part |
| --- | --- |
| gemma 4 (12b, E2B, 31B, 26B-A4B) | `<|turn>user\n` / `<|turn>model\n` |
| Ministral 3 (3B, 8B, 14B, Reasoning) | `[INST]` / `[/INST]` |
| ERNIE 4.5 21B | `<\|end_of_sentence\|>User:` / `Assistant:` |
| GLM 4.7 Flash | `<\|user\|>` / `<\|assistant\|></think>` |
| Nemotron 3 Nano | `<\|im_start\|>user\n` / `<\|im_start\|>assistant\n<think></think>` |
| Falcon H1, LFM2.5, Granite 4.0 | im_start and start_of_role markers |

Only ERNIE 4.5 VL still fails to load, because its vision processor imports PIL and decord, which is unrelated to the chat template. The text ERNIE 4.5 passes.

Net: the auto detection masks inputs and trains on completions exactly as documented across every loadable conversational tokenizer, on both transformers 4.57 and 5.12.

danielhanchen:
@codex review

danielhanchen:

- Marker anchoring (Codex): for tokenizers where the bos token doubles as the turn opener (for example SmolLM2 has `bos == <|im_start|>`), the old code stripped the bos and detected a bare `user\n` / `assistant\n`, which could then match inside user text. Now bos/eos are stripped only in the headerless branch (Mistral `[INST]`), and header markers have only whitespace stripped, so SmolLM2 detects the anchored `<|im_start|>user\n` / `<|im_start|>assistant\n`. Verified with an adversarial conversation whose user content literally contains "assistant\n": no user tokens leak into the labels.
- Empty/None special tokens (Gemini): `specials` is now filtered to non-empty strings, so an empty string can no longer cause an infinite loop in `strip_shared` and `None` can no longer raise a TypeError when sorting by length.
- Processor chat template (Codex): when the chat template lives only on the processor, detection now renders with the processor and validates token ids with the inner tokenizer. Mainstream VLM processors (Qwen2-VL, Llama 3.2 Vision) keep the template on both, so their behaviour is unchanged.
- Tail slice guard (Gemini): the user-probe tail lambda returns "" when the sentinel is missing.
- Validation error (Gemini): the error now names the failing part and shows the detected value.

On the empty-core IndexError note: the validator already guards this with `if core and any(...)`, so an empty core fails validation rather than slipping through. The trailing `pass` after the return matches the house style used after every function in this file.

Re-ran the full sweep over the cached tokenizers: 118 pass, 42 safe raises (base models with no chat template, plus the legacy non-atomic TinyLlama/Zephyr), and the single fail is the third party `Erland/DeepSeek-R1-0528-Qwen3-8B` whose template has a lowercase `<｜assistant｜>` typo (the official `unsloth/DeepSeek-R1-0528-Qwen3-8B` passes).

danielhanchen:
Addressed both points from the latest review (commit a2700c2). Both came from the same root cause in the headerless branch: it stripped `eos + "\n"` as a unit, so the newline got glued to the eos and dropped.

- Empty eos (Codex): when `tok.eos_token` is unset, `eos == ""`, so `eos + "\n"` was a bare `"\n"`. That stripped a real anchor off both markers. Now only `eos` is stripped (and `strip_lead` skips empty prefixes), so a `"\n### Human:"` / `"\n### Assistant:"` template with no eos keeps both anchors. Verified with eos forced to `None`.
- Plain-text newline anchor (Codex): for `"\n### Human:"` / `"\n### Assistant:"` templates the instruction marker lost its leading `"\n"` (consumed by `eos + "\n"`) while the response kept it, leaving a bare `### Human:` that could match inside message content. Detection is now symmetric and anchored: `vicuna_old` returns `instruction_part = "\n### Human:"`, `response_part = "\n### Assistant:"`.

Verification after the change:
- 24/24 token-exact on the content-exact suite (every assistant-content token trained, zero user-content tokens trained), unchanged.
- The four `unsloth.chat_templates` plain-text templates (`vicuna`, `vicuna_old`, `unsloth`, `alpaca`) stay token-exact, now with anchored markers.
- Full cached-tokenizer sweep unchanged: 118 pass, 42 safe-raise, and the single fail is still the third party `Erland/DeepSeek-R1-0528-Qwen3-8B` typo template (the official `unsloth/DeepSeek-R1-0528-Qwen3-8B` passes).

danielhanchen:
@codex review

danielhanchen:
Ran a fresh full review pass. Verdict was approve with nits, and the headline sweep reproduced independently (118 pass, 42 safe-raises, 0 user-content leaks across the cached conversational tokenizers; adversarial user content containing literal marker substrings stays leak-free for SmolLM2, Qwen3 and Llama-3).

Fixed the one item worth addressing before merge (commit 32a6256):

- Processor-only chat template (Medium): train_on_responses_only unwrapped the processor to the inner tokenizer before auto-detecting, so for VLMs whose chat template lives only on the processor the detection hit the "No chat_template" raise and the processor fallback inside get_chat_template_parts never ran. It now passes the original processor into detection while the token matcher still uses the inner tokenizer. Verified: a processor whose template lives only on the processor now auto-detects and masks correctly through the integrated path; mainstream VLMs (Qwen2-VL, Llama 3.2 Vision) carry the template on both objects so they are unchanged. 24/24 content-exact and the 118/42/1 sweep are unchanged.

The remaining notes were optional hardening and left as is: validate() always checks the force_match=True core (the default the matcher uses, and auto-detect produces the same strings as the docs, so this is pre-existing and only relevant if someone combines auto-detect with force_match=False), gap_mode's equal-frequency tie-break relies on insertion order (the 3-turn probe makes ties effectively impossible), and the zero-width-space sentinels degrade to a clean raise if a template mangles them.

danielhanchen:
Checked chat-template auto-detection on six requested models (4 Nemotron text models, Nemotron Nano Omni, Qwen2.5-Omni). All six auto-detect and mask correctly. One real fix came out of it (commit 46ab605).

Detection results (instruction / response markers), each verified token-exact (assistant content trained, user content masked):
- NVIDIA-Nemotron-Nano-9B-v2: `\n<SPECIAL_12>\n<SPECIAL_11>User\n` / `\n<SPECIAL_11>Assistant\n`
- Nemotron-3-Nano-30B-A3B, NVIDIA-Nemotron-3-Super-120B-A12B, NVIDIA-Nemotron-3-Nano-Omni: `\n<|im_start|>user\n` / `\n<|im_start|>assistant\n`
- NVIDIA-Nemotron-3-Nano-4B: `\n<|im_start|>user\n` / `\n<|im_start|>assistant\n` (its tokenizer needs a newer transformers to load via AutoTokenizer, this is a version thing, not a detection issue)
- Qwen2.5-Omni-7B: `<|im_start|>user\n` / `<|im_start|>assistant\n`

The fix: the three Nemotron reasoning templates emit `<|im_start|>assistant\n<think></think>` as the generation prompt and truncate history-turn thinking to an empty `<think></think>`, while keeping the real reasoning on the final (trained) turn. The detected response marker was carrying that empty tag, so it matched only the history turns and left the final turn (reasoning plus answer) fully masked. Detection now drops a trailing empty paired tag, so the marker is just the assistant header and matches every turn. Verified on all three reasoning models: with reasoning data the final turn's reasoning and answer are now trained and the user turns stay masked. 24/24 content-exact and the 118/42/1 cached sweep are unchanged.

danielhanchen:
Added a committed regression test that sweeps every model in unslothai/notebooks (commit bbceb95).

tests/test_notebook_chat_templates.py drives the 163 notebook repos (mostly unsloth/*, pinned in tests/notebook_models.json) and for each asserts get_chat_template_parts auto-detects and masks content-exact on two conversations: a plain multi-turn one and a reasoning one whose final assistant turn carries a think block. The reasoning probe is the part that catches a response marker carrying an injected empty think tag (it would mask the trained turn) - exactly the Nemotron reasoning class fixed here.

It is a network/integration test, off by default so normal CI is unaffected. Enable with UNSLOTH_TEST_NOTEBOOK_MODELS=1; set UNSLOTH_NOTEBOOK_TOK_CACHE to reuse a tokenizer cache. Models that need a newer transformers, are gated, or have no atomic role markers are skipped, not failed.

Local run: 95 pass, 0 fail, 70 skip. The 70 skips are 39 base models with no chat template, 23 models whose tokenizer class needs transformers 5.x to load, 6 gated/renamed repos, 1 known non-atomic-tag template, plus the base-model placeholders.

danielhanchen:
Two follow-ups on auto-detect for the notebook + vision models.

1. Notebook sweep is now dynamic (commit af5e7ae). The test scans a notebooks checkout (git clone, or UNSLOTH_NOTEBOOKS_DIR) and extracts every model named via from_pretrained / model_name, dropping quant suffixes and non-conversational repos, then unions the pinned list as a fallback. A newly added notebook is now covered automatically without editing the test. Local run over the current notebooks: 95 pass, 0 fail (the rest skip: base models, models needing a newer transformers, gated repos).

2. Structured-content VLM templates (commit b1fcca8). Some vision processors (Pixtral) require content as a list of blocks and raised an UndefinedError when the probe used a plain string. Detection now probes string content first, then structured content, and raises a clear message only if neither renders. Verified auto-detect across the vision processors: Qwen2-VL / Qwen2.5-VL -> ChatML, Gemma 3 / 3n -> <start_of_turn>, Gemma 4 -> the new <|turn>user/<|turn>model, Llama 3.2 Vision -> header, Pixtral -> [INST]/[/INST], Qwen3.5 -> ChatML. Each masks content-exact. 24/24 content-exact and the cached sweep are unchanged.

danielhanchen:
Addressed both MLX points in 6ace7e1.

The MLX `train_on_responses_only` wrapper resolved `trainer.tokenizer` down to the inner HF tokenizer before calling the shared auto-detect helper, which dropped the context detection needs. Added `_resolve_autodetect_template_source` and routed the auto-detect path through it (the resolved callable tokenizer is still used for token matching and text batch encoding, so explicit-marker calls are unchanged):

- Use the processor for VLM auto-detection. When markers are omitted, detection now uses the VLM processor (preferring `trainer.processor`, then a processor passed as the tokenizer, then `model._processor`) so a processor-only chat template is visible. The HF helper still unwraps to the inner tokenizer for token matching. Verified on Pixtral-12B (template on the processor, none on the inner tokenizer): the old path raised "No chat_template" while the new path auto-detects `[INST]` / `[/INST]`.

- Normalize the MLX template before auto-detecting. The configured override (`MLXTrainingConfig.chat_template` / `vlm_chat_template`) is now applied via `normalize_mlx_chat_template` / `normalize_vlm_processor_chat_template` before detection, so markers come from the same template the batches are rendered with. Verified that with an override set, detection follows the override's markers rather than the tokenizer's original template.

Added `tests/test_mlx_autodetect_template_source.py` (runs under the mlx-on-torch simulation shim, hermetic, skips offline) covering the processor-only, override, and explicit-marker cases.

danielhanchen:
Fixed in 5b227a8. The VLM branch of `_resolve_autodetect_template_source` now prefers `source` when it looks like a processor before falling back to `trainer.processor` / `model._processor`. Since the wrapper already resolves the explicit `tokenizer=` override into `source` ahead of `trainer.tokenizer`, an explicit processor override now drives auto-detection as well as token matching, so the detected markers and the returned mask function agree. Added a regression test (`test_vlm_explicit_processor_override_wins_over_trainer_processor`) alongside the existing processor-only and override cases.

danielhanchen:
Leaving the guard as-is here after digging in. The suggestion is to relax _processor_ready_for_detect so it accepts a processor whose processor.tokenizer is a wrapper (not a PreTrainedTokenizerBase), since _resolve_response_mask_tokenizer can unwrap it. Two reasons it should stay strict:

1. Not reachable in the mlx-vlm path. The VLM processor comes from mlx_vlm.load (or is rebuilt via AutoTokenizer.from_pretrained in _repair_degraded_vlm_processor), so processor.tokenizer is always a transformers PreTrainedTokenizerBase. The mlx-lm TokenizerWrapper only wraps text tokenizers from mlx_lm.load and never nests inside a VLM processor.

2. Relaxing the guard would crash rather than fix. get_chat_template_parts renders fine through a wrapper because attribute access is delegated, but the next step, _find_common_token_ids, calls the tokenizer directly: `tokenizer(x, add_special_tokens = False)`. TokenizerWrapper delegates via __getattr__, which does not make the object callable (obj(...) looks up __call__ on the type, bypassing __getattr__), so handing the processor straight to the helper raises "TokenizerWrapper object is not callable". I verified with a non-callable wrapper stand-in: relaxing the check turns todays clear, actionable "No chat_template ... pass instruction_part and response_part" error into a TypeError deep in token matching.

So the strict guard plus the graceful fallback is the safer behavior. If a real VLM processor with a wrapped inner tokenizer ever shows up, the correct fix is a small proxy that pairs the processor template (for rendering) with the resolved callable inner (for matching), not relaxing the guard. Happy to add that proxy if such a processor appears.

danielhanchen:
@codex review

danielhanchen:
All four addressed in 2305073 (markers verified identical across every cached tokenizer/processor before vs after).

- Prefer the processor template (get_chat_template_parts): the probe now renders with the processor's chat_template when it has one (that is what VLM batching renders through and it can differ from the inner tokenizer's text template); token ids are still validated with the inner tokenizer.

- Probe mixed VLM content shapes: added a per-role probe (user parts + assistant string) so processors that require the assistant collapsed to a plain string (handled elsewhere by _collapse_vlm_assistant_content) auto-detect instead of raising.

- Headerless templates reach the fallback: when add_generation_prompt is a no-op (identical renders, e.g. Mistral [INST]) the assistant header is now forced empty, so instruction_part is `[INST]` rather than `<eos>[INST]`. Verified on a Mistral-shaped template: before it detected `<eos>[INST]` (masking the turn-terminating eos), now `[INST]`.

- Ignore stale cached markers after overrides (mlx/trainer.py): a chat_template / vlm_chat_template override now clears any _unsloth_input_part / _unsloth_output_part cached from a prior Unsloth template, so the HF helper re-detects from the new template instead of masking with the old markers.

Added regression tests for the headerless, stale-override, and processor-template-preference cases.

danielhanchen:
@codex review

danielhanchen:
Addressed in bcbc978. Three were real and fixed; two I am leaving with reasons below. Markers verified identical across all cached tokenizers/processors before vs after.

Fixed:

- Preserve template-provided marker attrs: the clear now runs BEFORE normalize, not after. A raw Jinja override supplies no markers (so the HF helper re-detects), while an Unsloth template name/tuple sets fresh correct _unsloth_input_part / _unsloth_output_part via get_chat_template, which are now preserved instead of deleted.

- Keep VLM override aligned with batching: detection now uses the processor that actually renders the masked batches. With return_function=False the trainer renders via _resolve_vlm_processor (trainer.processor / trainer.tokenizer / model._processor), so a tokenizer= override is not applied to those batches and no longer drives detection; with return_function=True the caller applies the returned mask, so the explicit override is preferred. This reconciles the earlier "honor the override" request with the batching path.

- Strip turn terminators from user marker: in the branch where the generation prompt equals the assistant header but only assistant turns carry eos (<|user|>...<|assistant|>...</s>), the instruction marker now has the eos stripped, so a non-final assistant turn keeps its eos trainable. Verified: before it detected <eos><|user|>, now <|user|>.

Not changing:

- Accept wrapped processor tokenizers for detection: same as the prior review on this thread. In the mlx-vlm path processor.tokenizer is always a PreTrainedTokenizerBase, so this is not reachable; and relaxing the check would move the failure into _find_common_token_ids, which calls the tokenizer directly (tokenizer(x, add_special_tokens=False)) - an mlx-lm TokenizerWrapper is not callable, so it would raise TypeError instead of the current clear "pass instruction_part and response_part" error. The strict guard plus graceful fallback is the safer behavior.

- Probe all VLM render shapes supported by batching: the additional shapes _render_vlm_messages falls back to (_flatten_vlm_messages_to_content_parts and _flatten_vlm_content_for_text_template) drop the role structure into a flat content-parts list, which has no user/assistant boundaries to read instruction/response markers from. The role-message probes already cover both shapes _render_vlm_messages uses for role-structured rendering (assistant-list and assistant-collapsed). A flat-parts-only processor has no role markers to detect, so raising there is correct rather than emitting garbage markers.

danielhanchen:
@codex review

PR #836, opened 2026-06-28
train_on_responses_only: faster filtering, safer collator and num_proc, clearer errors

danielhanchen:
### Summary

Four focused, backward compatible improvements to `train_on_responses_only` in `unsloth_zoo/dataset_utils.py`, motivated by profiling and by the most common issues users file about this function. The masking output is unchanged (verified token exact across many models); these change the surrounding plumbing to be faster, safer, and clearer.

### 1. Faster fully masked filtering (about 37 percent faster end to end)

`train_on_responses_only` masks labels in one `dataset.map`, then drops rows whose labels are all -100 (to avoid NaN loss) with a separate `dataset.filter`. Profiling 5000 multi turn samples (1.77M tokens):

| step | time |
| --- | --- |
| inner masking | 0.13s |
| `dataset.map` | 0.83s |
| `dataset.filter` (removes 0 rows in the common case) | 1.28s |

`filter` rewrites the whole Arrow table even when it drops nothing. This PR scans the `labels` column cheaply (0.42s) and only calls `select` when a row is actually fully masked, so the common case skips the rewrite. End to end drops from about 2.1s to about 1.3s. When rows do need dropping, the surviving set and the "Removed N out of M" message are identical to before (verified over randomized datasets).

### 2. Smarter num_proc

Spawning workers for this fast op on a small dataset costs more than it saves (`num_proc=4` measured slower than `num_proc=1`), and the CPU count heuristic caused Windows infinite spawn loops (#3211, #3397). This PR single processes datasets below a threshold while preserving the existing non fork guard and any explicit user provided `num_proc`.

### 3. Do not clobber vision collators

A trainer using `UnslothVisionDataCollator` had it silently replaced with `DataCollatorForSeq2Seq` at the end of `train_on_responses_only`, which drops image handling and breaks VLM training (`UnslothVisionDataCollator` is not a `DataCollatorForSeq2Seq` subclass, so the old guard did not protect it). It already supports `train_on_responses_only=True`, `instruction_part`, `response_part` on the collator itself. This PR detects vision/processor collators and keeps them, printing guidance to enable masking on the collator. Text collators (`DataCollatorForSeq2Seq`, `DataCollatorForLanguageModeling`) behave exactly as before.

### 4. Actionable truncation error

The most common report about this function is the opaque "All labels are -100" error, whose real cause is almost always truncation: `max_length` (default 1024) cut off the assistant response marker before masking could find it (#2364, #3198, #2771, #2900). When the dataset comes back overwhelmingly masked and the response marker is absent or the rows sit at `max_length`, this PR raises a clear message naming truncation and suggesting `SFTConfig(max_length = max_seq_length)`, instead of failing later with a zero loss. A few bad rows are still just dropped as before; the error only fires when the whole dataset is unusable.

### Tests and verification

- Masking output token exact unchanged: every assistant content token trained, zero user content tokens trained, across Llama 3/4, Gemma 2/3, Qwen 2.5/3/2-VL, Phi 4, Mistral, DeepSeek R1, Granite and more.
- Filter equivalence: new scan plus `select` drops the same rows as the old `filter` across randomized datasets, with the same message.
- num_proc: small dataset auto -> single process; explicit value honored; non fork guard preserved.
- Vision collator: real `UnslothVisionDataCollator` is not replaced and guidance is printed; text collators replaced exactly as before; no misfire.
- Truncation: tiny `max_length` raises with an actionable message; healthy data does not; partial truncation only drops bad rows.
- Existing suite: `tests/test_zoo_history_regressions_deep.py` passes (34 passed), including the `sft_prepare_dataset` column removal regression.

### Notes

- No new dependencies. `select_columns` and `Dataset.iter` are available in the pinned `datasets` floor.
- The `return_function=True` path and packing path are unaffected (both return or branch before the new code).
- Backward compatible: explicit `instruction_part`/`response_part`, the `_unsloth_input_part` path, IterableDataset, and dict eval datasets all behave as before.

### DIFF
--- a/tests/test_vlm_collator_masking.py
+++ b/tests/test_vlm_collator_masking.py
@@ -20,10 +20,8 @@ import os, sys, types, tempfile
 
 HERE = os.path.dirname(os.path.abspath(__file__))
 sys.path.insert(0, os.path.dirname(HERE))
-# NOTE: unsloth_zoo is imported lazily inside _check (not at module top). Importing it
-# runs unsloth_zoo/__init__.py, which raises ImportError when the separate `unsloth`
-# package is absent; at module scope that would break pytest collection before the
-# "OFF by default" skip gate in test_vlm_collator_masking() can run.
+# unsloth_zoo is imported lazily inside _check: at module scope its __init__ raises
+# ImportError without the `unsloth` package, breaking collection before the skip gate runs.
 
 ENABLED = os.environ.get("UNSLOTH_TEST_VLM_COLLATOR", "") not in ("", "0", "false")
 CACHE = os.environ.get("UNSLOTH_VLM_PROC_CACHE", os.path.join(tempfile.gettempdir(), "unsloth_vlm_proc"))
@@ -49,9 +47,8 @@ def _fetch(repo):
     path = os.path.join(CACHE, repo.replace("/", "__"))
     cached = os.path.isdir(path) and any(f.startswith("tokenizer") for f in os.listdir(path))
     if not cached:
-        # Downloads are opt-in. Without UNSLOTH_TEST_VLM_COLLATOR an uncached repo is
-        # skipped (raise -> _check returns SKIP), never fetched, even when CACHE already
-        # exists from a prior partial run - so the sweep stays truly off by default.
+        # Downloads are opt-in: without the flag an uncached repo raises (-> _check SKIP)
+        # instead of fetching, even if CACHE exists from a prior partial run.
         if not ENABLED:
             raise RuntimeError(f"{repo} not cached; set UNSLOTH_TEST_VLM_COLLATOR=1 to download")
         from huggingface_hub import snapshot_download
--- a/tests/test_vlm_collator_masking.py
+++ b/tests/test_vlm_collator_masking.py
@@ -86,10 +86,13 @@ def _content_exact(collator, proc):
 
 
 def _check(repo, ins, res):
-    from transformers import AutoProcessor, AutoConfig
-    from unsloth_zoo.vision_utils import UnslothVisionDataCollator
-    from unsloth_zoo.dataset_utils import train_on_responses_only
+    # Import inside the try so a zoo-only CI without the `unsloth` package (whose absence makes
+    # unsloth_zoo/__init__ raise ImportError) turns into a SKIP, not a hard error, when run() is
+    # reached via an existing cache dir with the env flag unset.
     try:
+        from transformers import AutoProcessor, AutoConfig
+        from unsloth_zoo.vision_utils import UnslothVisionDataCollator
+        from unsloth_zoo.dataset_utils import train_on_responses_only
         path = _fetch(repo)
         proc = AutoProcessor.from_pretrained(path)
         model = _mock_model(AutoConfig.from_pretrained(path))
--- a/tests/test_vlm_collator_masking.py
+++ b/tests/test_vlm_collator_masking.py
@@ -20,8 +20,10 @@ import os, sys, types, tempfile
 
 HERE = os.path.dirname(os.path.abspath(__file__))
 sys.path.insert(0, os.path.dirname(HERE))
-from unsloth_zoo.vision_utils import UnslothVisionDataCollator
-from unsloth_zoo.dataset_utils import train_on_responses_only
+# NOTE: unsloth_zoo is imported lazily inside _check (not at module top). Importing it
+# runs unsloth_zoo/__init__.py, which raises ImportError when the separate `unsloth`
+# package is absent; at module scope that would break pytest collection before the
+# "OFF by default" skip gate in test_vlm_collator_masking() can run.
 
 ENABLED = os.environ.get("UNSLOTH_TEST_VLM_COLLATOR", "") not in ("", "0", "false")
 CACHE = os.environ.get("UNSLOTH_VLM_PROC_CACHE", os.path.join(tempfile.gettempdir(), "unsloth_vlm_proc"))
@@ -82,6 +84,8 @@ def _content_exact(collator, proc):
 
 def _check(repo, ins, res):
     from transformers import AutoProcessor, AutoConfig
+    from unsloth_zoo.vision_utils import UnslothVisionDataCollator
+    from unsloth_zoo.dataset_utils import train_on_responses_only
     try:
         path = _fetch(repo)
         proc = AutoProcessor.from_pretrained(path)
--- a/tests/test_vlm_collator_masking.py
+++ b/tests/test_vlm_collator_masking.py
@@ -1,0 +1,138 @@
+"""End-to-end: train_on_responses_only must actually mask for vision models.
+
+For one representative model per marker style this builds a REAL
+UnslothVisionDataCollator and runs a real image+text example through __call__,
+two ways, asserting the labels are content-exact (assistant response trained, user
+text masked, image/pad tokens masked):
+  (A) constructor:  UnslothVisionDataCollator(..., train_on_responses_only=True, parts)
+  (B) in-place:     bare UnslothVisionDataCollator(model, processor), then
+                    train_on_responses_only(trainer) enables masking on the collator.
+Path (B) is the case where a user keeps the default vision collator and calls
+train_on_responses_only(trainer) separately (most vision notebooks build the bare
+collator).
+
+Network/integration test, OFF by default. Enable with
+    UNSLOTH_TEST_VLM_COLLATOR=1 pytest tests/test_vlm_collator_masking.py
+Processors are fetched config-only; set UNSLOTH_VLM_PROC_CACHE to reuse a cache dir.
+Models that need a newer transformers to load are skipped, not failed.
+"""
+import os, sys, types, tempfile
+
+HERE = os.path.dirname(os.path.abspath(__file__))
+sys.path.insert(0, os.path.dirname(HERE))
+from unsloth_zoo.vision_utils import UnslothVisionDataCollator
+from unsloth_zoo.dataset_utils import train_on_responses_only
+
+ENABLED = os.environ.get("UNSLOTH_TEST_VLM_COLLATOR", "") not in ("", "0", "false")
+CACHE = os.environ.get("UNSLOTH_VLM_PROC_CACHE", os.path.join(tempfile.gettempdir(), "unsloth_vlm_proc"))
+ALLOW = ["*.json", "*.jinja", "*.model", "*.txt", "tokenizer*", "merges*", "vocab*"]
+
+# one model per marker style: ChatML, Gemma-3-turn, Gemma-4-turn, Llama-header,
+# Mistral-[INST]. The Gemma-4 / Qwen3.5 entries need a newer transformers and skip
+# (not fail) where it is too old to load them.
+MODELS = [
+    ("unsloth/Qwen2-VL-2B-Instruct", "<|im_start|>user\n", "<|im_start|>assistant\n"),
+    ("unsloth/Qwen3.5-2B", "<|im_start|>user\n", "<|im_start|>assistant\n"),
+    ("unsloth/gemma-3-4b-it", "<start_of_turn>user\n", "<start_of_turn>model\n"),
+    ("unsloth/gemma-3n-E4B-it", "<start_of_turn>user\n", "<start_of_turn>model\n"),
+    ("unsloth/gemma-4-E2B-it", "<|turn>user\n", "<|turn>model\n"),
+    ("unsloth/Llama-3.2-11B-Vision-Instruct",
+     "<|start_header_id|>user<|end_header_id|>\n\n", "<|start_header_id|>assistant<|end_header_id|>\n\n"),
+    ("unsloth/Pixtral-12B-2409", "[INST]", "[/INST]"),
+]
+USER_TXT, ASST_TXT = "ZEBRAQUESTION", "PENGUINANSWER"
+
+
+def _fetch(repo):
+    from huggingface_hub import snapshot_download
+    path = os.path.join(CACHE, repo.replace("/", "__"))
+    if not (os.path.isdir(path) and any(f.startswith("tokenizer") for f in os.listdir(path))):
+        snapshot_download(repo, local_dir=path, allow_patterns=ALLOW, token=os.environ.get("HF_TOKEN"))
+    return path
+
+
+def _mock_model(config):
+    import torch
+    vc = getattr(config, "vision_config", None)
+    ns = types.SimpleNamespace
+    vcfg = ns(patch_size=getattr(vc, "patch_size", 14) if vc is not None else 14,
+              image_size=getattr(vc, "image_size", 336) if vc is not None else 336,
+              model_type=getattr(vc, "model_type", "") if vc is not None else "")
+    return ns(config=ns(torch_dtype="float16", vision_config=vcfg), max_seq_length=2048,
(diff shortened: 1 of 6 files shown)


danielhanchen:

- Vision collator masking (Codex): the previous version kept the `UnslothVisionDataCollator` but only printed guidance, so masking was silently not applied (the collator rebuilds labels from the processor at collate time and ignores the dataset labels). Now, when a vision collator without masking is detected, response masking is enabled on the collator in place and the text dataset path is skipped, so `train_on_responses_only(trainer)` actually masks for VLMs while keeping image handling. Detection uses the MRO so subclasses are covered (Gemini). The three vision notebooks (Gemma 4 E2B, Qwen3.5 2B, Gemma 4 31B) call `UnslothVisionDataCollator` directly and never call `train_on_responses_only(trainer)`, so they are unaffected either way.
- Truncation on a bad prefix (Codex): the diagnostic no longer samples only the first 100 rows. It runs off the global filter scan and fires only when the whole dataset is overwhelmingly masked, so a sorted or sharded dataset with a fully masked prefix but valid rows afterwards is no longer aborted; the bad rows are just dropped. Verified with a dataset of 150 bad rows followed by 50 valid rows.
- NumPy/tensor rows (Gemini, Codex): the label scan and the truncation marker check now convert via `tolist` (duck typed), so torch and numpy formatted datasets do not iterate element by element or raise on an ambiguous array truth value.
- Custom transforms (Codex): if `select_columns(["labels"])` raises because a `set_transform` needs other columns, the filter falls back to `dataset.filter`.
- Threshold: the single process cutoff is now 5000 rows.

Also added in vision_utils: Gemma 4 masks `<|image|>` and `<|audio|>` through the tokenizer image_token/audio_token attributes already, but the begin/end image and audio delimiters (`<|image>`, `<image|>`, `<|audio>`, `<audio|>`) and `<|video|>` were not masked. These are now in IMAGE_TOKENS / AUDIO_TOKENS, matching how Gemma 3 image delimiters are handled, so they are not trained on. Confirmed all seven Gemma 4 multimodal token ids are covered.

Tests: masking output stays token exact across models, the new filter drops the same rows as the old `dataset.filter` over randomized data, end to end stays about 1.3s on the 5000 sample case, and the existing `test_zoo_history_regressions_deep.py` suite passes (34 passed).

danielhanchen:
Addressed both points from the latest review (commit ada2699):

- Do not abort when valid rows remain: the truncation diagnostic is now fatal only when no rows survive the filter (`len(keep) == 0`). When 90 percent or more of rows are masked away but some valid rows remain, it prints a warning naming the likely truncation cause and then trains/evaluates on the surviving rows, matching the old `dataset.filter` behaviour. Verified: 95 bad rows plus 5 valid rows warns and keeps the 5; an all-masked split still raises.
- Vision collator with tokenizer-default parts: when the parts came from the tokenizer's `_unsloth_input_part`/`_unsloth_output_part` (the no-explicit-parts path), the nested call to enable masking on the collator no longer passes those inferred strings back. It detects that the collator's tokenizer already carries the parts and lets the nested call read them, so it configures the collator instead of hitting the "already set" guard. Verified end to end with a tokenizer carrying `_unsloth_*` parts.

Re-ran the suite: 18 of 18 unit checks pass, masking stays token exact, the filter still drops the same rows as the old path, end to end stays about 1.3s, and `test_zoo_history_regressions_deep.py` passes (34 passed).

danielhanchen:
@codex review

danielhanchen:
Addressed all four points from the latest review (commit 28aa730).

- Unsupported vision collator no longer silently skipped (P2): the early return now only happens once masking is actually active. If the collator already has a callable masking function it is left alone; an `UnslothVisionDataCollator` without masking (including `train_on_responses_only = False`, not just `None`) gets a masking function installed in place; and a processor-style collator we cannot reliably configure now raises with guidance to build `UnslothVisionDataCollator(..., train_on_responses_only = True, ...)` rather than returning as if masking were applied.
- Truncation diagnosis no longer blames short mismatches (P3): a fully masked row only counts as truncation evidence when it sits at the length cap (`len(input_ids) >= max_length`). Short fully masked rows, or any dataset with no known `max_length`, fall back to the generic error so a wrong `response_part` / chat template is not misreported as a truncation that needs a bigger `max_length`. Rows at the cap still raise the truncation message.
- Filter fast path no longer retains every kept row (P2): the scan now tracks only the fully masked row indices. The common healthy case collects an empty list and returns without building a per-row keep list, so a large clean corpus does not pay one Python int per surviving row. The keep set is only materialized when something is actually dropped.
- MLX mask list mirrors the new Gemma 4 tokens (P2): `_IMAGE_TOKEN_STRINGS` in `mlx/utils.py` now includes the Gemma 4 image/video/audio delimiters (`<|image>`, `<image|>`, `<|video|>`, `<|audio|>`, `<|audio>`, `<audio|>`), matching the additions to `IMAGE_TOKENS` / `AUDIO_TOKENS`, so the MLX VLM loss masks them even when the tokenizer does not expose them as `image_token` / `video_token` attributes.

Verification:
- New regression tests for the four cases pass (foreign vision collator raises, `False` flag still installs masking, short wrong-template row is not blamed on truncation while an at-cap row still is, filter drops only fully masked rows).
- Existing suites green: the 18 faster/safer unit checks, `test_zoo_history_regressions_deep.py` (34 passed) and `test_mlx_save_export_regressions.py` (30 passed).
- Confirmed all six Gemma 4 delimiters resolve to real token ids on `unsloth/gemma-4-E2B-it` and are present in the MLX mirror.

danielhanchen:
@codex review

danielhanchen:
Follow-up (commit 3e7dbe1): standardized the VLM token lists so the MLX and CUDA paths share one source of truth.

Previously `vision_utils.py` (CUDA collator) and `mlx/utils.py` (MLX trainer) each kept their own copy of the image/audio placeholder tokens, which is exactly how they drifted apart in the first place. Moved the canonical lists into a new lightweight `unsloth_zoo/vlm_tokens.py` (no torch/mlx imports) exposing `IMAGE_TOKENS`, `AUDIO_TOKENS`, and a combined `VLM_PLACEHOLDER_TOKENS`. Both backends now import from it:

- `vision_utils.py`: `from .vlm_tokens import IMAGE_TOKENS, AUDIO_TOKENS` (re-exported, so existing imports keep working).
- `mlx/utils.py`: `_IMAGE_TOKEN_STRINGS = tuple(VLM_PLACEHOLDER_TOKENS)`.

While reconciling, the CUDA list picked up the InternVL `</image>` token that only the MLX list had. MLX token coverage is unchanged (verified by set equality against the previous list). Adding `</image>` on the CUDA side is safe: tokenizers without it return None (filtered) or the unk id that is already present from the other non-vocab tokens.

Verification: 64 repo regression tests pass (`test_zoo_history_regressions_deep.py`, `test_mlx_save_export_regressions.py`), the 18 faster/safer unit checks and the 8 review-fix checks pass, `get_padding_tokens_ids` still resolves on a real vision tokenizer, and all six Gemma 4 delimiters remain covered.

danielhanchen:
Ran a fresh full review pass. Verdict was approve with nits, with the four functional changes verified correct (filter fast path provably drops the same rows as the old dataset.filter across list/numpy/torch labels, vision-collator branch matrix sound, Gemma 4 tokens confirmed real and distinct in the actual tokenizer, MLX loses no coverage). Addressed the actionable items (commit 43a5744):

- All-masked-by-wrong-markers now fails clearly (Low): when every row is masked and it is not truncation (the instruction_part/response_part do not match the template), it previously returned an empty dataset behind a "removed ... after truncation" message. It now raises a clear error naming the response marker and pointing at the template. The truncation path (rows at the length cap) still raises its own max_length guidance, and partial masking still just drops the bad rows.
- Filter fallback keeps num_proc (Low): the custom-transform fallback to dataset.filter now passes num_proc like the main path.
- Do not mask the unk token (Low): get_padding_tokens_ids built padding ids via convert_tokens_to_ids, so tokens a tokenizer lacks collapsed onto the unk id and unk got masked from loss (e.g. Gemma unk id 3). It now drops the unk id, matching what the MLX path already does, so the two backends behave the same. Verified Gemma 3 no longer masks id 3 while keeping its real image tokens, and Qwen2-VL is unchanged.

Remaining nits (comment wording on the select() fast path, and that MLX coverage expanded rather than stayed equal) were left as documentation-only.

danielhanchen:
Checked Gemma 3n audio coverage (commit d36fcf4).

Gemma 3n's image tokens (`<image_soft_token>`, `<start_of_image>`, `<end_of_image>`) are the same as Gemma 3, so they were already masked. Its audio tokens use a different style from Gemma 4 though: `<audio_soft_token>`, `<start_of_audio>`, `<end_of_audio>`. Only the soft token is reachable via the `audio_token` attribute; the begin/end delimiters were in neither backend's list and were masked by neither (the attribute lists only carry `boi`/`eoi`, not `boa`/`eoa`), so they would have been trained on.

Added the Gemma 3n audio triplet to `AUDIO_TOKENS`. Verified on `unsloth/gemma-3n-E2B-it`: all six Gemma 3n media token ids (image + audio) are now masked by the CUDA `get_padding_tokens_ids` and present in the shared list the MLX path consumes. Gemma 4 coverage is unchanged (all seven still masked), and the repo regression suites stay green.

danielhanchen:
Checked image/vision/audio token coverage on six requested models (commit 03b0356). The four text-only Nemotrons (Nano-4B, Nano-9B-v2, Nano-30B-A3B, Super-120B-A12B) have no media tokens, so nothing to add. The two Omni models had gaps:

- NVIDIA-Nemotron-3-Nano-Omni (InternVL-style): the `<image>` context token was already masked, but the `` / `</img>` wrappers, `<video>`, and the sound tokens `<so_embedding>` / `<so_start>` / `<so_end>` were not. Added.
- Qwen2.5-Omni: `<|vision_pad|>` was already masked, but `<|IMAGE|>`, `<|VIDEO|>`, `<|AUDIO|>`, `<|vision_bos|>` / `<|vision_eos|>` and `<|audio_bos|>` / `<|audio_eos|>` were not (this model uses uppercase placeholders and bos/eos delimiters, distinct from the Qwen2-VL lowercase pad style already covered). Added.

Verified on the actual tokenizers that every real media token id for both models is now masked by get_padding_tokens_ids, and since both backends source from vlm_tokens the MLX path picks them up too. Repo regression suites stay green.

danielhanchen:
Added a committed VLM token coverage test (commit 3b676de).

tests/test_vlm_token_coverage.py pins the image/video/audio placeholder tokens for every vision and omni family shipped in unslothai/notebooks (Llama Vision, Qwen2-VL/2.5-VL, Qwen2.5-Omni, PaliGemma/Llava/InternVL, Nemotron Nano Omni, Pixtral, Gemma 3/3n/4, Cohere) and asserts each is present in vlm_tokens, so neither backend silently stops masking one. It also checks the lists have no duplicates and that the MLX mirror derives from the shared list instead of hardcoding its own copy (which is how the two drifted before). Offline and fast (40 tokens across 12 families), so it runs in normal CI. The live per-model id verification stays in the network sweep.

danielhanchen:
Verified train_on_responses_only end to end for the vision collators (commit af0eba7), since this was never actually checked.

Across the vision notebooks, 60 of 84 build the collator and only the Paddle OCR ones pass train_on_responses_only into it; the rest use a bare UnslothVisionDataCollator(model, processor) and none call train_on_responses_only(trainer). So the common path is a bare collator, and a user wanting response masking would call train_on_responses_only(trainer) separately.

The new test builds a real UnslothVisionDataCollator per marker style and runs a real image+text example through __call__ two ways: the constructor train_on_responses_only=True path, and the in-place path where train_on_responses_only(trainer) configures masking on a bare collator. It asserts content-exact labels (assistant trained, user masked, image/pad tokens masked). Both paths pass for Qwen2-VL, Qwen2.5-VL, Qwen3.5, Gemma 3, Gemma 3n, Gemma 4, Llama 3.2 Vision and Pixtral. So train_on_responses_only is not regressed for vision and the bare-collator + train_on_responses_only(trainer) flow works. Network test, off by default (UNSLOTH_TEST_VLM_COLLATOR=1).

danielhanchen:
@codex review

danielhanchen:
All three addressed in 436651a.

1. Avoid materializing all kept indices (dataset_utils.py). The drop path now uses `dataset.filter(_has_valid_labels, num_proc = _effective_num_proc(dataset))` instead of `select` over a `[i for i in range(n) if i not in drop_set]` list. On a very large corpus with only a few fully masked rows that list allocated one Python int per surviving row (many GB); `_has_valid_labels` is the exact inverse of the dropped set, so survivors are identical (verified: same rows dropped, None-label rows kept). The healthy common case still returns early without filtering, so the fast no-op path is preserved.

2. Avoid importing the full package in the offline test (test_vlm_token_coverage.py). It now loads `unsloth_zoo/vlm_tokens.py` by file path rather than `from unsloth_zoo.vlm_tokens import ...`, which ran the package `__init__` and raised `ImportError: Please install Unsloth ...` during collection when the separate `unsloth` package is absent. `vlm_tokens.py` is pure data, so it stays hermetic and always runs. Verified by blocking the `unsloth` import and confirming the test still collects and passes.

3. Keep the VLM network sweep opt-in (test_vlm_collator_masking.py). `_fetch` now only calls `snapshot_download` when `UNSLOTH_TEST_VLM_COLLATOR` is set; an uncached repo otherwise raises and is caught as SKIP. So an existing cache dir from a prior partial run no longer lets the off-by-default sweep hit the network. Verified with a `snapshot_download` tripwire: flag unset plus an existing cache dir yields all SKIP and zero network calls.

danielhanchen:
Both addressed in 09cdaca.

- Redundant list copy (dataset_utils.py): `dropped` is already a list, so `_diagnose_truncation` now slices it directly (`dropped[:100]`) instead of `list(dropped)[:100]`.

- Keep the disabled VLM sweep import-safe (test_vlm_collator_masking.py): `_check` did its `transformers` / `unsloth_zoo` imports before the try, so when `run()` is reached via an existing cache dir with `UNSLOTH_TEST_VLM_COLLATOR` unset, a zoo-only CI without the separate `unsloth` package failed with ImportError (unsloth_zoo/__init__ raises it) instead of skipping. The imports now live inside `_check`s try, so that case returns SKIP. Verified by blocking the `unsloth` import with an existing cache dir and the flag unset: the test skips (rc 0) rather than erroring.

danielhanchen:
@codex review

danielhanchen:
All three addressed in 00e44d0.

- Mask InternVL context image tokens: added `<IMG_CONTEXT>` to IMAGE_TOKENS (the repeated context token InternVL expands between `...</img>`). Without it CUDA VLM batches left the bulk of InternVL image placeholder labels trainable.

- Filter unknown audio ids before truncation checks: vision_utils._truncate_sequence_tensors now excludes unk_token_id from the audio-id set, matching get_padding_tokens_ids. Audio family strings absent from a given tokenizer map to unk, so a truncated unrelated unk text token no longer changes the pre/post audio count and trips the "cuts into the expanded audio tokens" error. Verified with a stubbed tokenizer: the spurious raise is gone.

- Detect processors stored on collator.tokenizer: _is_vision_collator now also treats a collator whose .tokenizer exposes image_processor (e.g. DataCollatorForSeq2Seq(tokenizer=processor)) as a vision collator, so it is not replaced by the text DataCollatorForSeq2Seq and does not lose image handling. Verified it detects that shape while still returning False for a plain text collator.

danielhanchen:
@codex review

danielhanchen:
@codex review

PR #839, opened 2026-06-28
Add imatrix option to quantize_gguf for IQ low-bit GGUF quants

danielhanchen:
This adds an `imatrix` (importance matrix) option to `quantize_gguf` so the GGUF export can pass an importance matrix to `llama-quantize`, which unlocks the IQ low-bit quant types (`iq2_xxs`, `iq4_xs`, ...) that llama.cpp refuses to produce without one.

## What changed

`quantize_gguf(...)` gains an `imatrix=None` argument. When set, `--imatrix <path>` is prepended to the `llama-quantize` command before the positional input/output/type arguments, quoted with the existing `_quote` helper so paths with spaces are safe. The `q2_k_l` preset flags and all other behavior are unchanged when `imatrix` is `None`, so this is fully backwards compatible.

## Why

This is the companion change to unslothai/unsloth#6706, which adds an `imatrix_file` option to `save_pretrained_gguf` / `push_to_hub_gguf`. That PR resolves the imatrix (a local path, or auto-downloaded from the upstream `unsloth/<base>-GGUF` repo) and threads the resolved path through to `quantize_gguf` here. Without this change the unsloth side fails fast with an "upgrade unsloth_zoo" message, so the two can land independently.

## Testing

Verified end to end on Llama-3.2-1B: the imatrix is loaded by llama-quantize (`load_imatrix: loaded 112 importance matrix entries ... computed on 689 chunks`) and used to produce `iq2_xxs` and `iq4_xs` GGUFs, which then run correctly under llama.cpp.

### DIFF
--- a/tests/test_quantize_gguf_q2_k_l.py
+++ b/tests/test_quantize_gguf_q2_k_l.py
@@ -244,3 +244,100 @@ def test_q2_k_l_error_message_keeps_original_preset_name(monkeypatch):
         assert "q2_k_l" in str(exc), f"error msg should keep preset name: {exc}"
     else:
         raise AssertionError("expected RuntimeError")
+
+
+# --- imatrix option ------------------------------------------------------------------------------
+
+
+def test_imatrix_flag_is_prepended_before_positional_args(tmp_path, monkeypatch):
+    """A provided imatrix must reach llama-quantize as --imatrix in the option region."""
+
+    llama_cpp = _load_llama_cpp_module()
+    captured = _install_fake_subprocess_run(monkeypatch, llama_cpp)
+    _stub_output_exists(monkeypatch)
+
+    imat = tmp_path / "imatrix.dat"
+    imat.write_bytes(b"\x00")
+    llama_cpp.quantize_gguf(
+        input_gguf="/tmp/in.gguf",
+        output_gguf="/tmp/out.gguf",
+        quant_type="iq2_xxs",
+        quantizer_location="/usr/bin/llama-quantize",
+        n_threads=4,
+        print_output=False,
+        imatrix=str(imat),
+    )
+
+    cmd = captured["cmd"]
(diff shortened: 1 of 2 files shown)


danielhanchen:

- Validate the imatrix path: `quantize_gguf` now raises a clear `FileNotFoundError` when the file does not exist (instead of a cryptic `llama-quantize` failure), and skips whitespace-only values.
- Windows quoting: `_quote` now always double-quotes on Windows, so paths with cmd metacharacters (`&`, `|`, `^`), not just spaces, stay literal under `shell=True`. This is the root-cause fix and also hardens the existing input/output/quantizer paths, not only the imatrix one.

Added `quantize_gguf` tests for imatrix flag placement, the blank/no-op cases, the missing-file error, and Windows metacharacter quoting. Full `test_quantize_gguf_q2_k_l.py` passes (10 tests).

danielhanchen:
@codex review

danielhanchen:
Thanks for the review. On the `_extra_flags` spacing note (gemini, Medium): leaving as-is. The command construction is correct today, every branch that sets `_extra_flags` ends it with a trailing space and the empty case collapses to a single separator, so there is no spacing bug. I would rather not rewrite the working concatenation for a hypothetical future edit. The earlier substantive items (imatrix path validation, Windows quoting, and the Q2_K_L preset log gating) are already fixed in this PR.

danielhanchen:
@codex review

danielhanchen:
@codex review

danielhanchen:
Follow-up hardening: normalized the imatrix value to a stripped string before the existence check in quantize_gguf. Previously os.path.exists(True) would treat a bool as a file descriptor (silently emitting --imatrix "True"), and a space-padded path was not trimmed before validation/quoting; both now resolve to a clear error or the trimmed path. Added tests for the non-path and padded-path cases.

PR #843, opened 2026-07-01
Fix Gemma3 processor batched tokenization (ragged padding)

danielhanchen:
## Summary

`Gemma3Processor.__call__` (patched in `temporary_patches/gemma.py`) tokenized a batch with padding, then stripped a leading double BOS only from rows that still started with `[bos, bos]`. After left padding, only the longest (unpadded) row starts with `[bos, bos]`, so just that row lost a token and the batch became ragged, and `BatchFeature` could no longer stack it:

```
ValueError: Unable to create tensor, you should probably activate padding with 'padding=True' to have batched tensors with the same length.
```

This surfaced in vision GRPO run with `fast_inference=False` (batched slow-mode generation), but affects any batched processor call that requests padding.

## Fix

Tokenize without padding, strip the duplicate BOS on every row (the chat template and the tokenizer each add one), rebuild `token_type_ids`, then pad every sequence field to the batch max. The strip and pad logic is factored into `_fix_double_bos_and_pad` so it can be unit tested directly. This also fixes a latent issue where padded rows previously kept a double BOS.

## Tests

`tests/test_gemma3_processor_batched.py` (GPU free): batched ragged rows become stackable, a single row strips its BOS without padding, right padding side, and `return_tensors=None` leaves the lists ragged.

Verified end to end: the Gemma3 vision GRPO notebook, which previously crashed at step 0 with `fast_inference=False`, now trains to completion.

### DIFF
--- a/tests/test_gemma3_processor_batched.py
+++ b/tests/test_gemma3_processor_batched.py
@@ -111,6 +111,30 @@ def test_max_length_uses_model_max_when_unset():
     assert {len(x) for x in out["input_ids"]} == {6}
 
 
+def test_non_aligned_fields_untouched():
+    # overflowing_tokens is a per-example list whose rows do NOT match input_ids lengths; it must be
+    # left exactly as-is (not BOS-stripped or padded).
+    text_inputs = {
+        "input_ids": [[BOS, BOS, 10], [BOS, BOS, 20, 21]],
+        "attention_mask": [[1, 1, 1], [1, 1, 1, 1]],
+        "overflowing_tokens": [[77], []],
+    }
+    out = _fix_double_bos_and_pad(text_inputs, BOS, PAD, IMG, True, True, "left", "pt")
+    assert out["overflowing_tokens"] == [[77], []]
+    assert {len(x) for x in out["input_ids"]} == {3}
+
+
+def test_length_recomputed_after_padding():
+    # return_length: the length field must reflect post-strip/pad token counts, not the raw tokenizer ones.
+    text_inputs = {
+        "input_ids": [[BOS, BOS, 10], [BOS, BOS, 20, 21]],
+        "attention_mask": [[1, 1, 1], [1, 1, 1, 1]],
+        "length": [3, 4],
+    }
+    out = _fix_double_bos_and_pad(text_inputs, BOS, PAD, IMG, True, True, "left", "pt")
+    assert out["length"] == [3, 3]
+
+
 if __name__ == "__main__":
     test_batched_ragged_rows_become_stackable()
     test_single_row_strips_bos_without_padding()
@@ -124,4 +148,6 @@ if __name__ == "__main__":
     test_offset_mapping_tuples_stripped_and_padded()
     test_pad_to_multiple_of()
     test_max_length_uses_model_max_when_unset()
+    test_non_aligned_fields_untouched()
+    test_length_recomputed_after_padding()
     print("OK: all gemma3 batched-processor padding tests passed")
--- a/tests/test_gemma3_processor_batched.py
+++ b/tests/test_gemma3_processor_batched.py
@@ -6,7 +6,7 @@ could not stack them. _fix_double_bos_and_pad strips the duplicate BOS on every 
 sequence fields. Padding follows the HF contract: it honours "do_not_pad"/False, pads list output
 even when return_tensors is None, and targets max_length when padding="max_length".
 """
-from unsloth_zoo.temporary_patches.gemma import _fix_double_bos_and_pad
+from unsloth_zoo.temporary_patches.gemma import _fix_double_bos_and_pad, _resolve_truncation
 
 BOS, PAD, IMG = 2, 0, 99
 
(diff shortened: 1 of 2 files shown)


danielhanchen:
Addressed the review feedback on `_fix_double_bos_and_pad`:

- Honor `do_not_pad`: the pad guard now checks `padding not in (False, None, "do_not_pad")` instead of a plain truthiness test, so an explicit `"do_not_pad"` no longer pads.
- Match the HF contract for list output: dropped the `return_tensors is not None` condition so `padding=True` pads the returned lists even when `return_tensors=None`.
- Support `padding="max_length"`: threaded `max_length` through from `text_kwargs` and pad to it when requested.
- Strip a pre-existing `token_type_ids` in lockstep with the double-BOS removal so a row cannot desync when `return_mm_token_type_ids=False`.

`tests/test_gemma3_processor_batched.py` now covers all four cases (do_not_pad, `return_tensors=None` padding, max_length, token_type_ids stripping) plus the original ragged-batch regression, and is GPU free.

danielhanchen:
@codex review

danielhanchen:
Addressed the latest review points on `_fix_double_bos_and_pad`:

- Pad every per-token field: the strip and pad now cover every per-row field the tokenizer returned (`special_tokens_mask`, `offset_mapping`, and any others), not just `input_ids`/`attention_mask`/`token_type_ids`, so a batched call that requests auxiliary outputs no longer leaves a ragged field for `BatchFeature` to choke on. Fills are per field: `1` for `special_tokens_mask`, `(0, 0)` for offset pairs, `0` otherwise.
- Padding targets: honor `pad_to_multiple_of` (round the padded length up) and, for `padding="max_length"` without an explicit `max_length`, fall back to the tokenizer's `model_max_length` (ignoring its uninitialised ~1e30 sentinel).

On the `max_length` truncation note: removing `padding` from the tokenizer call does not change truncation. Truncation is gated on the `truncation` argument, not `padding`, and `max_length` is still forwarded to the tokenizer, so `padding=True, max_length=1024` without `truncation` behaves exactly as before (no truncation).

Tests expanded to 12 GPU-free cases covering `special_tokens_mask`, `offset_mapping` tuples, `pad_to_multiple_of`, and `model_max_length`.

danielhanchen:
@codex review

danielhanchen:
Follow-up on the latest review points:

- Token-aligned fields only: the strip/pad now only touches fields whose per-row length matches `input_ids`, so non-aligned outputs such as `overflowing_tokens` are left untouched instead of being stripped/padded.
- `length`: recomputed after the strip/pad when `return_length` is set, so it reflects the returned token counts rather than the raw pre-strip ones.
- `max_length` without truncation: `max_length` is now dropped from the tokenizer call unless `truncation` is requested (we pad to it manually), which removes the "max_length set without truncation" warning and any truncation ambiguity; when `truncation` is on, `max_length` is kept so the tokenizer still truncates.

Tests expanded to 14 GPU-free cases.

danielhanchen:
@codex review

danielhanchen:
Good catch, this one was real (and my previous pass had it backwards). HF's `_get_padding_truncation_strategies` activates `longest_first` truncation when `max_length` is set with `padding=False` and no explicit `truncation`. Because we drop `padding` to pad manually, the tokenizer would otherwise see a different padding and change that decision: `padding="max_length"` callers could get unexpectedly truncated, and `padding=False` callers could lose the truncation HF applies.

Fixed by resolving the truncation strategy from the caller's original `padding` and passing it to the tokenizer explicitly (`_resolve_truncation`), so behaviour is identical whether or not we split tokenize and pad:

- `padding=False`, `max_length=N`, no `truncation` -> `longest_first` (truncates), matching HF
- `padding="max_length"` / `True`, no `truncation` -> no truncation (we pad to `max_length` manually)
- explicit `truncation` is always preserved

Added model-free tests for the resolver (15 total).

danielhanchen:
@codex review

danielhanchen:
@codex review

danielhanchen:
@codex review

danielhanchen:
@codex review

danielhanchen:
Pushed 14f01b9. `overflowing_tokens` is a per-example list of truncated tails, not a token-aligned field, so it is now excluded by name from the strip/pad selection (the row-length filter already skipped it in the normal case; the exclusion also covers the rare case where a tail length coincidentally matches its row).

On the other two notes:

- `max_length` with `pad_to_multiple_of`: transformers `PreTrainedTokenizerBase._pad` rounds `max_length` up to the next multiple (`max_length = ((max_length // pad_to_multiple_of) + 1) * pad_to_multiple_of`) rather than rejecting, so the manual ceil-to-multiple here matches the tokenizer and does not change the truncation limit.
- no pad token: this is the Gemma3 processor path, and Gemma3 always defines a pad token (id 0), so `fill_for("input_ids")` resolves to that id. The no-pad-token configuration does not occur here.

@codex review

danielhanchen:
@codex review

PR #847, opened 2026-07-02
fix(gguf): preserve Qwen3.5/3.6 MTP config so nextn tensors convert

LeoBorcherding:
## Summary

GGUF conversion of Qwen3.5/3.6 models that have an MTP (nextn) head fails with:

```
ValueError: Can not map tensor 'model.layers.<N>.eh_proj.weight'
```

`convert_to_gguf` strips `mtp_num_hidden_layers` from `config.json` before invoking the converter. That shrinks the converter's `block_count` back to `num_hidden_layers`, but the checkpoint still ships the `mtp.*` weights. `_Qwen35MtpMixin.filter_tensors` then remaps `mtp.fc` → `model.layers.<num_hidden_layers>.eh_proj` — a layer index that no longer exists — so `map_tensor_name` raises.

gguf-py already has full nextn support (`blk.{bid}.nextn.eh_proj`, `enorm`, `hnorm`, `shared_head.*`), so the correct fix is to **keep** the MTP config and let the converter's existing nextn path map the `mtp.*` weights. Only the internal `unsloth_fixed_mtp` marker is stripped.

## Test plan

- [x] Export Qwen3.5-2B (`mtp_num_hidden_layers: 1`) to GGUF — previously failed at `eh_proj`, now converts cleanly and emits `blk.<N>.nextn.*` tensors.
- [x] Verified end-to-end via Unsloth Studio GGUF export (BF16, incl. sharded output).

Companion to unslothai/unsloth#6107 (GGUF shard size control), which surfaced this on Qwen3.5.

### DIFF
--- a/unsloth_zoo/llama_cpp.py
+++ b/unsloth_zoo/llama_cpp.py
@@ -2276,10 +2276,17 @@ def convert_to_gguf(
     with open(config_path, "r", encoding = "utf-8") as f:
         config_file = json.load(f)
 
-    # Strip MTP / nextn config keys so the downstream convert_hf_to_gguf.py
-    # doesn't inflate block_count / inject nextn_predict_layers.
+    # Preserve MTP / nextn config so the downstream convert_hf_to_gguf.py keeps
+    # the extra nextn block (block_count += mtp_num_hidden_layers) and emits the
+    # nextn tensors (blk.N.nextn.*). gguf-py has full nextn support, so the
+    # leftover `mtp.*` weights remap cleanly via _Qwen35MtpMixin instead of being
+    # forced into a phantom out-of-range layer. Stripping mtp_num_hidden_layers
+    # shrank block_count to num_hidden_layers, so `mtp.fc` -> `model.layers.<N>.eh_proj`
+    # pointed at a layer index that no longer existed, raising
+    # "Can not map tensor ...eh_proj" for Qwen3.5/3.6 MTP checkpoints.
+    # Only strip Unsloth's internal marker, never the real MTP config.
     _changed = False
-    for _key in ("mtp_num_hidden_layers", "unsloth_fixed_mtp"):
+    for _key in ("unsloth_fixed_mtp",):
         if config_file.pop(_key, None) is not None:
             _changed = True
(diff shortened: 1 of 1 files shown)


LeoBorcherding:
**Update: confirmed end-to-end on Qwen3.5-2B (MTP head).**

With this fix, `convert_to_gguf` keeps `mtp_num_hidden_layers` instead of stripping it, so the checkpoint's `mtp.*` weights map cleanly through the converter's existing nextn path. The previous failure was:

```
ValueError: Can not map tensor 'model.layers.<N>.eh_proj.weight'
```

which happened because stripping the config shrank `block_count` back to `num_hidden_layers` while the `mtp.*` weights were still present, so `mtp.fc` remapped to a layer index that no longer existed.

**Test results**

- Exported Qwen3.5-2B (`mtp_num_hidden_layers: 1`) to GGUF via Unsloth Studio. Previously failed at `eh_proj`; now converts cleanly and emits `blk.<N>.nextn.*` tensors.
- Exercised on the BF16 path including sharded output (13 parts), then **loaded the result back in Studio (llama-server) and chatted successfully**, so the preserved nextn tensors load and run, not just convert.

**Relationship to unslothai/unsloth#6107**

This is not sharding-specific. It is required to GGUF-export any Qwen3.5/3.6 model that has an MTP (nextn) head, sharded or single-file. It came up because #6107's shard-size testing happened to run on Qwen3.5, where export died at `eh_proj` before the shard step was ever reached. So it is a prerequisite for exercising #6107 on MTP models, not a component of the sharding feature itself.

oobabooga:
Closing as superseded by unslothai/unsloth-zoo#904.

This PR preserves `mtp_num_hidden_layers` unconditionally. That resolves conversion when the checkpoint contains MTP tensors, but leaves stale MTP metadata in checkpoints where a save path such as MLX has already removed those tensors. In that case, llama.cpp expects an MTP layer that is not present.

PR unslothai/unsloth-zoo#904 handles both checkpoint states at the shared `convert_to_gguf` boundary: it preserves the metadata when MTP tensors are present and removes it when they are absent.

PR #851, opened 2026-07-03
mlx: fixes for bnb dequant, report_to, early stopping, and NEFTune

oobabooga:
Building on @BardiaKoopah's MLX trainer work, follow-ups on the four PRs that landed today: bnb dequant (#815), report_to (#821), early stopping and best model (#822), and NEFTune (#824). Each has an edge case off the happy path; the common paths those PRs cover are unchanged.

## NEFTune (#824)

`_install_neftune` swaps `embed_tokens.__class__` to a subclass for the duration of training, which spans the adapter save. Two save-time checks keyed on the class name and broke under that swap.

The quantization-map scan matched by exact name:

```python
if type(module).__name__ not in {"QuantizedLinear", "QuantizedEmbedding"}:
    continue
```

so the `_NEFTuneEmbed` subclass was dropped from the saved map, and reload validated it against the real base and raised `unexpected quantized modules: ['model.embed_tokens']`. Switched both copies of the scan (loader.py, utils.py) to `isinstance`, which matches the subclass.

The DoRA check `type(module).__name__.startswith("DoRA")` had the mirror problem: it would see `_NEFTuneEmbed` and demote an embedding-only DoRA adapter (`use_dora` on `embed_tokens`) to plain LoRA. The subclass now reports the base class's `__name__`/`__qualname__`, so that name check sees through it. The map scan needs no such trick now that it uses `isinstance`.

Also, `neftune_noise_alpha=nan` slipped past `if alpha <= 0` (nan comparisons are always False) and would add nan noise to every embedding from step 0. Now guarded with `math.isfinite`.

## Early stopping and best model (#822)

`best_metric`, `best_step`, and the patience counter lived only in memory, so `resume_from_checkpoint` restarted best-tracking from scratch: it could overwrite `best/` with a worse model or mis-count early-stop patience. They now go into `trainer_state.json` and restore on resume, with `.get` defaults so pre-fix checkpoints still load.

Separately, after an early stop the summary printed `Steps: {total_steps}` (the requested max) rather than the step actually reached. Both the print and the returned `train_steps` now report the completed step (`self._global_step`). That counter is also seeded from the checkpoint on resume, so a no-op resume (loop never runs because the checkpoint is already at `max_steps`) reports the reached step instead of 0.

`metric_for_best_model` now matches HF Trainer: an unprefixed name (a bare `loss`) maps to `eval_loss`, and a null value falls back to `eval_loss` instead of crashing on the first eval.

Reusing a trainer for a second `train()` without resume now starts clean. Per-run tracking (step, loss history, best-model and early-stopping state) is reset up front, so a run that early-stopped no longer leaves the next run a silent no-op. The resume path still re-seeds those fields from the checkpoint.

## report_to (#821)

`SummaryWriter(...)` does real filesystem work at init (creates the run dir, opens and writes the event file), so a bad `output_dir` took the whole run down instead of just losing logging. Wrapped it to degrade to a warning, matching what the wandb path already did.

`report_to="all"` now expands to the supported backends (matches HF's semantics), and an unknown backend prints a notice instead of being silently dropped. Setup also moved to after arg auto-tuning, so the wandb config snapshot reflects the settings actually used (VLM compile auto-tune can flip `gradient_checkpointing` before training).

## bnb dequant (#815)

- `from_pretrained` was passed `dtype=torch.float16`, which raises `TypeError` on the supported transformers 4.x floor. Changed to `torch_dtype=`, which is accepted across 4.51.3 to 5.5.0.
- bitsandbytes VLM repos are rejected up front with the same actionable error the text path gives. The dequant path is text-only (mlx-lm); mlx-vlm dequant isn't wired up, so this replaces a half-working VLM attempt.
- A LoRA adapter dir can carry a copied base `config.json` with `quant_method=bitsandbytes`. It now takes the adapter branch instead of being dequantized as a full model.
- The bnb stub swap is serialized with a lock. bitsandbytes registers non-idempotent torch ops at import (`torch.library.define` at module load), so two concurrent callers racing the `sys.modules`/`sys.meta_path` restore corrupt the import state. The temp dir is also cleaned on a mid-save error so a failed dequant doesn't leak a multi-GB fp16 copy.
- `quantization_config` is forwarded through the recursive load so downstream detection sees it.

## Tests

Adds `tests/test_mlx_neftune_quant_map.py` covering the NEFTune save-time bugs: the subclassed embedding stays in the map through both scans, a map saved while NEFTune is active reloads against the unmodified base, and the DoRA name check still fires under NEFTune. They load a real quantized MLX model, so they are gated to Apple Silicon/Metal.

Beyond the added tests, the change was run end to end on an M1 (macOS 14.5) against the pinned stack: mlx 0.31.2, mlx-lm 0.31.3, mlx-vlm 0.6.3, transformers 5.5.0, bitsandbytes 0.49.2, accelerate 1.14.0, using `mlx-community/SmolLM-135M-Instruct-4bit` and `unsloth/Qwen2.5-0.5B-Instruct-bnb-4bit`. What was confirmed working:

| Area | Confirmed on M1 |
|---|---|
| NEFTune (#824) | With `neftune_noise_alpha>0`, training adds correctly scaled noise to the embeddings and eval adds none; the adapter trains, saves, and reloads; a DoRA-on-embeddings adapter keeps its magnitude vector. |
| Quant map (#824) | The quantized `embed_tokens` stays in the saved quantization map while NEFTune is active, so a NEFTune adapter saves and reloads with no base-mismatch error. A real quantization mismatch is still rejected. |
| Early stopping and best model (#822) | The best adapters are saved and restored at the end; a diverging run stops early; the summary and returned `train_steps` report the step actually reached, on a full run, an early stop, and a no-op resume; `metric_for_best_model` accepts a bare `loss` or a null value like HF; reusing a trainer for a second run starts clean. |
| Resume (#822) | A checkpoint restores the step, loss history, and best/early-stop state and continues from there; older checkpoints that predate the new keys still load. |
| report_to (#821) | wandb and tensorboard both log the expected train and eval metrics; `report_to="all"` enables both; a backend that fails to start logs a warning and training continues. |
| bnb dequant (#815) | A real bnb-4bit repo loads: it dequantizes to fp16, re-quantizes to MLX 4-bit, generates coherent text, keeps the original repo name, and cleans up the scratch copy. A bnb VLM repo is rejected with the clear error; a plain MLX repo is unaffected. |

Real bitsandbytes 0.49.2 installs and runs on Apple Silicon from PyPI, so #815's note about needing a nightly wheel and `kernels<0.14` is now stale.


oobabooga:
@codex review

oobabooga:
@codex review

oobabooga:
@codex review

oobabooga:
@codex review

PR #855, opened 2026-07-03
Fix MLX notebook generate input/output compatibility

Lyxot:
## Summary

This PR fixes the MLX-side compatibility gaps that prevented existing supported Unsloth notebooks from running their inference cells unchanged after importing Unsloth.

The changes are intentionally limited to the notebook-facing input/output contracts in the MLX loader:

- return text and VLM `generate(...)` results as a batched generated-id sequence
- prefer a `torch.long` tensor when torch is importable, with a NumPy `int64` fallback when torch is unavailable
- keep the returned sequence shaped as `(1, prompt_length + generated_length)` so existing notebook slicing and decode code works
- make the patched mlx-lm `TokenizerWrapper` support `apply_chat_template(..., return_dict=True)` by returning a Hugging Face `BatchEncoding`
- preserve the existing callable-tokenizer shim for mlx-lm wrappers that do not define `__call__`
- add focused regression coverage for text generation, VLM generation, and chat-template `return_dict=True` expansion

## Why

Some existing notebooks use standard Hugging Face / CUDA-style inference patterns such as:

```python
inputs = tokenizer.apply_chat_template(..., return_tensors="pt", return_dict=True).to("cuda")
outputs = model.generate(**inputs, max_new_tokens=...)
response = tokenizer.batch_decode(outputs[:, inputs.input_ids.shape[-1]:])
```

On the MLX path, the model and tokenizer objects are backed by mlx-lm / mlx-vlm rather than Transformers. Two notebook-visible mismatches showed up:

1. `generate(...)` returned an MLX array. That array can be sliced, but Transformers utilities such as `to_py_obj(...)` do not recognize the `mlx` array type. Returning torch when available gives notebooks the CUDA-like generated-id container they expect, while the NumPy fallback keeps torch-free MLX installs usable.
2. mlx-lm's `TokenizerWrapper.apply_chat_template(..., return_dict=True)` did not consistently return a mapping-like object that can be moved with `.to(...)` and expanded into `model.generate(**inputs)`.

This PR normalizes those two surfaces without changing notebook code.

## Scope

This is deliberately not a broad Transformers generation compatibility layer.

Not included in this PR:

- no `GenerationConfig` merge or precedence support
- no HF `logits_processor` compatibility policy
- no `return_dict_in_generate` output mode
- no changes to notebook files

The goal is only to make the currently supported notebook inference patterns work unchanged on MLX while keeping the diff small and reviewable. CUDA/ROCm paths are unaffected because these changes are isolated to `unsloth_zoo/mlx/loader.py`.

## Validation

Focused checks run locally:

```bash
PYTHONPATH=/Users/long/Github/unsloth/unsloth:/Users/long/Github/unsloth/worktrees/unsloth-zoo-mlx-inference-notebooks:$PYTHONPATH \
  python -m pytest tests/test_mlx_save_export_regressions.py -q
```

Result:

```text
35 passed
```

Additional checks:

```bash
python -m py_compile unsloth_zoo/mlx/loader.py tests/test_mlx_save_export_regressions.py
git diff --check
```

Both passed.

A manual blocked-torch smoke check also verified that `_mlx_generate_output(...)` falls back to a NumPy `int64` array when torch import is unavailable.

The regression tests cover:

- text `generate(...)` returns a batched torch generated-id tensor with working `.shape`, slicing, `.tolist()`, and `transformers.to_py_obj(...)`
- VLM `generate(...)` returns the same notebook-friendly torch sequence shape
- the MLX-on-torch regression module skips cleanly when torch is unavailable
- `apply_chat_template(..., return_dict=True)` returns a `BatchEncoding` that supports `.to(...)` and `model.generate(**inputs)` expansion

### DIFF
--- a/tests/test_mlx_save_export_regressions.py
+++ b/tests/test_mlx_save_export_regressions.py
@@ -223,6 +223,7 @@ def test_bound_gguf_push_filters_kwargs(monkeypatch):
 def test_text_generate_honors_do_sample_false(monkeypatch):
     import mlx_lm
     import mlx_lm.sample_utils as sample_utils
+    from transformers.tokenization_utils_base import to_py_obj
     import unsloth_zoo.mlx.loader as loader
 
     calls = {}
@@ -264,6 +265,9 @@ def test_text_generate_honors_do_sample_false(monkeypatch):
     )
 
     assert out.tolist() == [[1, 2, 9, 5]]
+    assert out.shape == (1, 4)
+    assert out[:, 2:].tolist() == [[9, 5]]
+    assert to_py_obj(out) == [[1, 2, 9, 5]]
     assert calls["sampler"] == {
         "temp": 0.0,
         "top_p": 0.0,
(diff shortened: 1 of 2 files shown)


danielhanchen:
Reviewed and validated.

This is an MLX-only compatibility shim and is structurally CUDA-safe: the changed functions install exclusively through `_patch_mlx_saving` (called only from `FastMLXModel.from_pretrained`), and `unsloth_zoo/mlx/loader.py` cannot even import on CUDA (`unsloth_zoo/mlx/compile.py` does a module-level `import mlx.core`, and the sole external importer in `dataset_utils.py` is `try/except ImportError` guarded). The MLX path is gated by `is_mlx_available()` (Darwin + arm64 + `mlx` present), so CUDA/ROCm/XPU are unaffected.

Ran `tests/test_mlx_save_export_regressions.py` through the `mlx_simulation` harness on Linux + torch: all pass, including the text/VLM generate and `apply_chat_template(return_dict=True)` cases.

On the three review comments:
- torch-vs-numpy return contract: the "prefer torch, numpy fallback" commit plus the updated description resolve this; implementation and description now agree.
- `except ImportError` being too narrow: broadened so a torch that is installed but fails to import (broken native libs raise `OSError`/`RuntimeError`, not `ImportError`) degrades to the numpy path instead of crashing generation.
- numpy fallback untested: added `test_mlx_generate_output_numpy_fallback_without_torch`, which forces the torch import to fail (both `ImportError` and `OSError`) and asserts a numpy `int64` `(1, N)` array with correct slicing. It fails on the pre-fix loader and passes after.

Pushed both as `15907877`.

Also validated the real notebook inference contract end to end on a macos-14 Apple Silicon runner with real mlx wheels (`mlx 0.31.2`, `DEVICE_TYPE=mlx`): `apply_chat_template(return_tensors="pt", return_dict=True)` -> `.to(...)` -> `model.generate(**inputs)` -> `batch_decode(out[:, prompt_len:])`, plus `transformers.to_py_obj(out)`. The probe returned `input_ids (1, 42)` -> generated `(1, 44)` torch long tensor -> decoded the new tokens correctly. Run: https://github.com/danielhanchen/unsloth-staging-2/actions/runs/28741299955

PR #874, opened 2026-07-06
Fix compiled forward argument forwarding

### LINKED ISSUE #875: Qwen3.5 compiled linear_attn wrapper references undefined args
## Description

When compiling the Qwen3.5 `linear_attn` forward path, the generated compiled cache can reference an undefined `args` variable.

The generated wrapper can look like this:

```python
def forward(
    self,
    hidden_states: torch.Tensor,
    cache_params: Cache | None = None,
    attention_mask: torch.Tensor | None = None,
    **kwargs: Unpack[TransformersKwargs],
):
    return wrapped(self, *args, **kwargs)
```

The emitted wrapper signature does not include `*args`, so training fails before completing the first step.

## Error

```text
NameError: name 'args' is not defined. Did you mean: 'kwargs'?
```

Observed stack excerpt:

```text
File ".../transformers/models/qwen3_5/modeling_qwen3_5.py", line 788, in forward
  hidden_states = self.linear_attn(...)

File ".../unsloth_compiled_cache/unsloth_compiled_module_qwen3_5.py", line 578, in forward
  return wrapped(self, *args, **kwargs)

NameError: name 'args' is not defined
```

## Environment

```text
unsloth: 2026.6.9
unsloth_zoo: 2026.6.7
transformers: 5.13.0
trl: 0.29.1
peft: 0.19.1
accelerate: 1.14.0
torch: 2.7.0+cu126
triton: 3.3.0
GPU: NVIDIA H100 80GB HBM3
Model: Qwen/Qwen3.5-9B-Base
```

## Expected behavior

The generated wrapper should forward the variables that exist in the emitted wrapper definition, for example:

```python
return wrapped(
    self,
    hidden_states=hidden_states,
    cache_params=cache_params,
    attention_mask=attention_mask,
    **kwargs,
)
```

or preserve `*args` in the emitted wrapper signature.

## Workaround

Disabling the compiled cache path avoids this generated wrapper:

```bash
export UNSLOTH_COMPILE_DISABLE=1
```

With that workaround, the same Qwen3.5 64k smoke run can complete a single SFT step.

## Proposed fix

Build the compiled forward-call arguments from the emitted wrapper definition rather than from a possibly decorator-relaxed `inspect.signature(f.forward)`. A PR is available in #874.

vincenthcui:
## Summary
- build compiled forward-call arguments from the emitted wrapper definition instead of a potentially decorator-relaxed inspect.signature(f.forward)
- prevents generated wrappers with explicit signatures from referencing an undefined *args local
- adds a CPU-safe regression test for the Qwen3.5-style forward signature

## Context
For Qwen3.5 linear_attn, the emitted compiled cache can contain a wrapper like:

```python
def forward(self, hidden_states, cache_params=None, attention_mask=None, **kwargs):
    return wrapped(self, *args, **kwargs)
```

The wrapper definition has no `args`, so training fails with `NameError: name 'args' is not defined`. This happens when `inspect.signature(f.forward)` sees a decorator-relaxed `(self, *args, **kwargs)` signature while the emitted wrapper definition comes from the original class source with explicit parameters.

## Tests
- `python -m py_compile unsloth_zoo/compiler.py tests/test_compiler_rewriter_exhaustive.py`
- `python -m pytest tests/test_compiler_rewriter_exhaustive.py::test_compiler_forwarding_parameters_follow_emitted_definition -q`
- repeated the same targeted test on a temporary GCE Debian 12 VM with Python 3.11

Fixes #875

### DIFF
--- a/tests/test_compiler_rewriter_exhaustive.py
+++ b/tests/test_compiler_rewriter_exhaustive.py
@@ -124,7 +124,7 @@ def test_compiler_forwarding_parameters_follow_emitted_definition():
     from pathlib import Path
 
     compiler_path = Path(__file__).resolve().parents[1] / "unsloth_zoo" / "compiler.py"
-    tree = ast.parse(compiler_path.read_text())
+    tree = ast.parse(compiler_path.read_text(encoding="utf-8"))
     wanted = {
         "_build_forwarding_parameters",
         "_build_forwarding_parameters_from_definition",
--- a/tests/test_compiler_rewriter_exhaustive.py
+++ b/tests/test_compiler_rewriter_exhaustive.py
@@ -115,6 +115,46 @@ def test_compiler_higher_precision_softmax_idempotency_lookahead():
         )
 
 
+def test_compiler_forwarding_parameters_follow_emitted_definition():
+    """compiler.py:create_new_function: generated wrapper calls must use the
+    emitted wrapper signature, not a decorator-relaxed ``*args`` signature."""
+    import ast
+    import inspect
+    import textwrap
+    from pathlib import Path
+
+    compiler_path = Path(__file__).resolve().parents[1] / "unsloth_zoo" / "compiler.py"
+    tree = ast.parse(compiler_path.read_text())
+    wanted = {
+        "_build_forwarding_parameters",
+        "_build_forwarding_parameters_from_definition",
+    }
+    helper_nodes = [
+        node for node in tree.body
+        if isinstance(node, ast.FunctionDef) and node.name in wanted
+    ]
+    namespace = {"ast": ast, "inspect": inspect, "textwrap": textwrap}
+    exec(compile(ast.Module(body=helper_nodes, type_ignores=[]), str(compiler_path), "exec"), namespace)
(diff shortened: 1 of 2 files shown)


vincenthcui:
Addressed the Gemini review comment in ee47b527 by reading compiler.py with `encoding="utf-8"`. Re-ran `python -m pytest tests/test_compiler_rewriter_exhaustive.py::test_compiler_forwarding_parameters_follow_emitted_definition -q` successfully.

danielhanchen:
main now recovers the undecorated forward in unsloth_zoo/compiler.py via _unwrap_undecorated_method, which removes the wrapped(self, *args, **kwargs) NameError on the Qwen3.5 linear_attn path, so I do not think this is needed any more. Sorry for the slow review, and thanks for putting it together.

danielhanchen:
Checked main and the forwarding call is still built from inspect.signature(f.forward) in unsloth_zoo/compiler.py, so this has not been overtaken, but switching every compiled class to parse the emitted definition text changes wrapper generation for models that work today. Could you confirm the generated wrappers are unchanged for a couple of common cases such as Llama and Gemma3 before I review?

PR #877, opened 2026-07-07
Guard assert_same_keys base-model resolution for PEFT bases without .model (#95)

danielhanchen:
## Summary

`assert_same_keys` resolved the inner model with a one-shot `model.base_model.model` access:

```python
inner_model = model.base_model.model if hasattr(model, "base_model") else model
```

For a PEFT wrapping whose `base_model` object has no `.model` attribute (e.g. Phi-3 / Phi-4-mini, where `base_model` is the transformer body itself), that raises `AttributeError`. It is reached from `create_lora_statistics` (`return_state_dict=True`, the default) on every `merged_16bit` save, so the affected models fail to export.

## Fix

`create_lora_statistics` already resolves the base model with `find_lora_base_model(model)`, which guards each level (`base_model`, then `model`), and builds the compared `new_state_dict` from that. Resolve `inner_model` in `assert_same_keys` the same way. This:

- removes the crash for a base whose `base_model` has no `.model`, and
- keeps the key comparison like-for-like with how `new_state_dict` was built (the tie-weight branch a few lines below already uses the same per-level guard for its `.model` access).

For the normal PEFT layout (`base_model.model` present) the resolved object is identical, so behavior is unchanged.

## Tests

`tests/test_assert_same_keys_base_model.py`:
- base-without-`.model` layout: raised `AttributeError` before this change, passes after;
- normal PEFT layout: unchanged;
- genuine key mismatch: still detected.

Based on the report and fix in #95; this applies the guard at the actual crash site (`assert_same_keys`) via the existing `find_lora_base_model` helper. Supersedes #95.

### DIFF
--- a/tests/test_assert_same_keys_base_model.py
+++ b/tests/test_assert_same_keys_base_model.py
@@ -1,0 +1,105 @@
+# Unsloth Zoo - Utilities for Unsloth
+# Copyright 2023-present Daniel Han-Chen, Michael Han-Chen & the Unsloth team. All rights reserved.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published
+# by the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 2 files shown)


danielhanchen:
Closing in favor of #95, which carries the same fix but keeps the original contributor's authorship (toran billups) in the git history.

I brought #95 up to current main: dropped the now-obsolete create_lora_statistics hunk (main already resolves the base via find_lora_base_model), applied the guard at the still-live assert_same_keys crash site using the same helper, and added the regression test there. #95 is now mergeable.

PR #878, opened 2026-07-07
Add MLXTrainer eval compute_metrics support

Lyxot:
## Summary

Adds HF-style `compute_metrics` support to the MLX text training eval path.

This PR extends `MLXTrainer` so scheduled evaluation during `train()` can collect prediction/label payloads, call `compute_metrics(EvalPrediction(...))`, and merge the returned metrics into the existing eval metrics dictionary.

## Stack / companion PRs

- Stacked on: #873, which adds HF-style callback support to `MLXTrainer`.
- Companion wrapper PR: unslothai/unsloth#6932.

## What changed

- Adds `compute_metrics` and `preprocess_logits_for_metrics` constructor kwargs to `MLXTrainer`.
- Keeps the existing eval loss path unchanged, including CCE support; metric prediction collection uses a separate eval-only forward path.
- Passes full-sequence logits and labels to metrics by default, matching HF `Trainer` behavior.
- Runs `preprocess_logits_for_metrics(logits, labels)` before accumulation when supplied.
- Lazily imports `EvalPrediction` / `denumpify_detensorize` only when metrics are actually computed, preserving the normal torch-free import path.
- Prefixes user metric keys with the active eval prefix when needed.
- Preserves trainer-owned `eval_loss` / `eval_perplexity` and split loss/perplexity if a metric function returns a colliding key such as `loss`.
- Supports dict eval datasets with split-specific metric prefixes such as `eval_validation_accuracy`.
- Pads variable-length eval batches before concatenating predictions and labels.
- Masks labels using the same `[start, end)` bounds as the loss path.
- Handles common prediction payload shapes, including tuples/lists, mappings, scalar chunks, bfloat16 tensors, and model outputs with a `logits` attribute or `{"logits": ...}` mapping.
- Explicitly leaves VLM `compute_metrics` unsupported for now with a clear error if requested.

## Why

The MLX trainer already supports eval loss/perplexity, but callers using the HF/TRL `compute_metrics` pattern could not compute custom eval metrics. This blocks parity with common `SFTTrainer` workflows and prevents callback/best-metric flows from seeing user-defined eval metrics.

## Validation

- `python -m pytest tests/test_mlx_trainer_internals.py -q -k 'evaluate or metric_chunk or metric_logits'`
- `python -m pytest tests/test_mlx_training_e2e_metal.py -q -k 'callback or compute_metrics or import_keeps_torch'`
- `python -m compileall -q unsloth_zoo/mlx/trainer.py tests/test_mlx_training_e2e_metal.py tests/test_mlx_trainer_internals.py`
- `git diff --check`

## Follow-ups intentionally left out

- Public `evaluate()` / `predict()` methods.
- `batch_eval_metrics`.
- `include_for_metrics`.
- `eval_accumulation_steps`.
- VLM-specific `compute_metrics` support.

### DIFF
--- a/tests/test_mlx_trainer_internals.py
+++ b/tests/test_mlx_trainer_internals.py
@@ -490,6 +490,110 @@ def test_evaluate_dict_eval_datasets_records_split_metrics():
     assert trainer.model.modes == ["eval", "train"]
 
 
+def test_evaluate_compute_metrics_pads_and_prefixes_without_metal():
+    import mlx.core as mx
+
+    from unsloth_zoo.mlx.trainer import MLXTrainer
+
+    class Model:
+        def eval(self):
+            pass
+
+        def train(self):
+            pass
+
+        def __call__(self, input_ids):
+            shape = (input_ids.shape[0], input_ids.shape[1], 8)
+            return {"logits": mx.zeros(shape)}
+
+    seen = {}
+    trainer = MLXTrainer.__new__(MLXTrainer)
+    trainer.model = Model()
+    trainer.stop_requested = False
+
+    def preprocess(logits, labels):
+        seen["last_logits_shape"] = tuple(logits.shape)
+        seen["last_labels_shape"] = tuple(labels.shape)
+        return labels
+
+    def compute_metrics(prediction):
+        seen["prediction_type"] = type(prediction).__name__
+        seen["predictions"] = prediction.predictions.tolist()
+        seen["labels"] = prediction.label_ids.tolist()
+        return {"rows": prediction.label_ids.shape[0], "loss": 123.0}
+
+    trainer.preprocess_logits_for_metrics = preprocess
+    trainer.compute_metrics = compute_metrics
+
+    def loss_fn(_model, _batch, _lengths, _labels):
+        return mx.array(1.0), mx.array(2)
+
+    loss, _ppl = trainer._evaluate(
+        [
+            (
+                mx.array([[0, 1, 2, 3]]),
+                mx.array([[1, 4]]),
+                mx.array([[0, 1, 2, 3]]),
+            ),
+            (
+                mx.array([[4, 5, 0]]),
+                mx.array([[0, 2]]),
+                mx.array([[4, 5, 0]]),
+            ),
+        ],
+        loss_fn,
+        is_vlm=False,
+    )
+
+    assert loss == pytest.approx(1.0)
+    assert seen["prediction_type"] == "EvalPrediction"
+    assert seen["last_logits_shape"] == (1, 3, 8)
+    assert seen["last_labels_shape"] == (1, 3)
+    assert seen["predictions"] == [[-100, 1, 2, 3], [4, 5, -100, -100]]
+    assert seen["labels"] == [[-100, 1, 2, 3], [4, 5, -100, -100]]
+    assert trainer._last_eval_metrics["eval_rows"] == 2
+    assert trainer._last_eval_metrics["eval_loss"] == pytest.approx(1.0)
+
+
+def test_metric_chunk_concat_handles_scalar_payloads():
+    import mlx.core as mx
+
+    from unsloth_zoo.mlx.trainer import MLXTrainer
+
+    result = MLXTrainer._pad_and_concat_metric_chunks(
+        [mx.array(1.0), mx.array(2.0)]
+    )
+
+    assert result.tolist() == [1.0, 2.0]
+
+
+def test_metric_chunk_concat_handles_mapping_payloads():
+    import mlx.core as mx
+
+    from unsloth_zoo.mlx.trainer import MLXTrainer
+
+    result = MLXTrainer._pad_and_concat_metric_chunks(
+        [
+            {"scores": mx.array([[1, 2]])},
+            {"scores": mx.array([[3]])},
+        ]
+    )
+
+    assert result["scores"].tolist() == [[1, 2], [3, -100]]
+
+
+def test_metric_logits_extractor_accepts_mapping_outputs():
+    import mlx.core as mx
+
+    from unsloth_zoo.mlx.trainer import MLXTrainer
+
+    logits = mx.array([1, 2, 3])
+    hidden = mx.array([4, 5, 6])
+
+    assert MLXTrainer._extract_metric_logits({"logits": logits}) is logits
+    assert MLXTrainer._extract_metric_logits({"loss": mx.array(0.0), "hidden": hidden}) is hidden
+
+
 def test_vlm_cce_prefers_collated_position_ids_for_cuda_parity():
     import inspect
     from unsloth_zoo.mlx import utils as mlx_utils
--- a/tests/test_mlx_training_e2e_metal.py
+++ b/tests/test_mlx_training_e2e_metal.py
@@ -70,6 +70,213 @@ def _assert_history(trainer, min_steps):
     return hist
 
 
+def _callback_batch():
+    """Build a tiny labeled MLX batch for callback lifecycle tests."""
+    tokens = mx.array([[0, 1, 2, 3]], dtype=mx.int32)
+    return tokens, mx.array([[0, 4]], dtype=mx.int32), tokens
+
+
+def _callback_trainer(
+    tmp_path,
+    callbacks,
+    max_steps=3,
+    eval_steps=1,
+    logging_steps=1,
+    with_eval=False,
+):
+    """Create a minimal MLXTrainer with prebuilt batches for callback tests."""
+    import mlx.nn as nn
+    from unsloth_zoo.mlx.trainer import MLXTrainer, MLXTrainingConfig
+
+    class TinyLM(nn.Module):
+        def __init__(self):
+            super().__init__()
+            self.embed = nn.Embedding(8, 4)
+            self.lm_head = nn.Linear(4, 8, bias=False)
+
+        def __call__(self, input_ids):
+            return self.lm_head(self.embed(input_ids))
+
+    trainer = MLXTrainer(
+        model=TinyLM(),
+        tokenizer=None,
+        train_dataset=[],
+        eval_dataset=[{}] if (eval_steps or with_eval) else None,
+        args=MLXTrainingConfig(
+            max_steps=max_steps,
+            per_device_train_batch_size=1,
+            gradient_accumulation_steps=1,
+            learning_rate=1e-4,
+            logging_steps=logging_steps,
+            eval_steps=eval_steps,
+            output_dir=str(tmp_path),
+            use_cce=False,
+            compile=False,
+            gradient_checkpointing=False,
+            report_to="none",
+        ),
+        callbacks=callbacks,
+    )
+    trainer._batches = [_callback_batch()]
+    if eval_steps or with_eval:
+        trainer._eval_batches_labeled = [_callback_batch()]
+    trainer._saved = []
+    trainer.save_model = lambda output_dir=None: trainer._saved.append(
+        output_dir or trainer.args.output_dir
+    )
+    return trainer
+
+
+@metal_only
+def test_hf_callbacks_receive_mlx_trainer_lifecycle(tmp_path):
+    from transformers import TrainerCallback
+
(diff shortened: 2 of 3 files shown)


danielhanchen:
Reviewed this even though it is a draft, because I wanted to know whether it is worth finishing. It is: the design is sound and unusually faithful to HF, and I verified its arithmetic against a hand-derived reference and against a real `transformers.Trainer` at both ends of the support range. But it cannot land as it stands, and patching this base would be wasted effort. Details below so you know exactly what to redo.

Method: MLX runs on Linux CPU (`pip install "mlx[cpu]" mlx-lm mlx-vlm`), so everything below was executed against real MLX 0.32.0, and re-run under transformers 4.57.6 and 5.14.1.

## The stack is the blocker

The substance here is one commit, `ddcd5b4c`, at +362 / -6 across three files. The +1276 / -147 GitHub shows is mostly a stale copy of #873.

The problem is that **this sits 43 commits behind the PR it is stacked on.** `git log --oneline ddcd5b4c..2b4a2d57` is 43 commits: roughly 18 upstream `main` commits that #873 pulled in via merge `9fe8a3fe`, plus about 25 of #873's own rework commits, several of which rewrite the same eval and callback region.

Trial merge of `ddcd5b4c` into current `main`:

```
CONFLICT  unsloth_zoo/mlx/trainer.py            10 hunks (largest ~213 / ~237 lines)
CONFLICT  tests/test_mlx_trainer_internals.py    1 hunk
          tests/test_mlx_training_e2e_metal.py   auto-merged clean
```

The test conflict is mechanical. The `trainer.py` ones are semantic, and the part specific to this PR is that `main` rewrote `_evaluate_batch_totals` and `_evaluate` after your base was cut. `main` has, and your base does not:

| symbol | main | this base |
| --- | --- | --- |
| `_distributed_all_sum` | yes, 14 call sites | no |
| `_distributed_eval_status` | yes | no |
| `_raise_distributed_failure_from_any` | yes | no |
| `_distributed_should_stop` | yes | no |
| `mx.where(ntoks > 0, loss * ntoks, 0.0)` (#888 NaN guard) | yes | no |

So a naive "take HEAD" resolution would silently revert #888's zero-token NaN guard and the DDP all-reduce of eval loss. That is why I did not resolve it for you.

What I would do, in order:

1. Land #873 first against current `main`. I have already pushed the merge plus two fixes there, so it should be close.
2. Rebase `ddcd5b4c` onto the new #873 tip. Do not resolve this PR's conflicts against `main` directly.
3. Re-apply the compute-metrics hunks by hand into `main`'s current `_evaluate_batch_totals` shape, keeping the try/except plus `_distributed_eval_status` loop, the `mx.where(ntoks > 0, ...)` guard and the `_distributed_all_sum` calls.
4. Decide the DDP semantics for `compute_metrics`, which is a question that did not exist when you wrote this.

## What the design gets right

Checked against a real `transformers.Trainer` on a tiny Llama at 4.57.6 and 5.14.1, identical at both:

- `predictions` is `(N, Lmax, V)` full-sequence logits and `label_ids` is `(N, Lmax)` unshifted with `-100`. Exactly HF.
- Padding uses `-100` on axis 1 for both preds and labels, matching `numpy_pad_and_concatenate` and `torch_pad_and_concatenate`.
- `preprocess_logits_for_metrics(logits, labels)` runs per batch before accumulation, matching `trainer.py:4708`.
- Dict eval datasets produce `eval_validation_accuracy` and `eval_test_accuracy`, byte for byte against real HF, and the prefix rule `if not key.startswith(f"{prefix}_")` is HF's loop verbatim.
- Label masking bounds are correct: `positions >= lengths[:,0:1]` and `< lengths[:,1:2]` over absolute positions is exactly equivalent to the loss path's `steps = arange(1, L)` half-open mask. A rigged model gives 0.8 accuracy and my independent hand derivation gives 4/5.
- Padded positions cannot pollute the metric. A row of true length `Lb` padded to `Lmax` has label indices at or above `Lb` set to `-100`, and after the universal shift every padded pred row pairs with a `-100` label. Confirmed empirically as well as structurally.

## Four gaps where a user would get wrong numbers rather than an error

1. **DDP, and this is the real one.** `compute_metrics` is called inside `_evaluate_batch_totals`, per rank, before `main`'s `_distributed_all_sum`. So on a multi-rank run `eval_loss` is globally reduced but `eval_accuracy` is computed on 1/N of the eval set and differs per rank, silently. Demonstrated: full set `mean_label=6.5000 n=24`, rank 0 `5.0000 n=12`, rank 1 `8.0000 n=12`. HF avoids this with `pad_across_processes` plus `gather_function` before accumulation. This gap does not exist today, but it appears the moment you rebase.
2. **The collision comment is backwards.** HF's prefix loop runs after `metrics[f"{prefix}_loss"]` is set, so a user returning `{"loss": 999.0}` overwrites `eval_loss`; I confirmed real HF reports `eval_loss = 999.0`. This PR makes the trainer win. That is arguably the safer choice, so keep the behaviour, but the comment saying "Match HF Trainer ordering: trainer-owned loss metrics win" is factually wrong.
3. **Partial eval.** If `stop_requested` flips mid-eval, `compute_metrics` still runs on the truncated set and is reported with no marker (`mean_label=3.5000 n=8` over 2 of 6 batches). The loss has the same property so it is at least consistent, but it deserves a note.
4. **Double forward pass.** `_text_metric_prediction_step` calls `self.model(batch)` at full length while the loss path already ran `model(batch[:, :-1])`. Measured call shapes `[(1,4),(1,5),(1,2),(1,3)]`, so 2x eval compute and a second shape family, which interacts badly with #918's bounded compile signatures (added specifically to stop a Metal resource-limit crash). Worth reusing the loss forward, or reconciling the signature budget, after the rebase.

## Memory: the deferral needs a guard, not an implementation

Measured on 32 samples by 128 tokens by 32k vocab:

```
theoretical fp32 logits          =    0.488 GiB
peak RSS without compute_metrics =    405.7 MiB
peak RSS with    compute_metrics =   1899.7 MiB   (+1494.0 MiB, about 3x the final array)
EvalPrediction.predictions       =   (32,128,32000)  500.0 MiB
with preprocess=argmax           =   (32,128)  0.016 MiB
```

The 3x is because each batch's MLX logits stay live alongside their numpy copies until `_pad_and_concat_metric_chunks` runs at the end. Scaled up:

| config | final array (fp32) | projected peak |
| --- | --- | --- |
| Llama-3.2 (128256 vocab), 500 x 1024 tok | 244.6 GiB | ~700 GiB |
| Qwen3 (151936), 1000 x 512 tok | 289.8 GiB | ~850 GiB |
| Gemma-3 (262144), 500 x 1024 tok | 500.0 GiB | ~1.4 TiB |

On Apple unified memory, with `main`'s `_configure_memory_limits` capping Metal at about 85 percent of the recommended working set, even a 64 by 512 eval set is an instant OOM. Deferring `eval_accumulation_steps` and `batch_eval_metrics` is fine, but please do not ship without at least a loud warning carrying the projected byte count and pointing at `preprocess_logits_for_metrics`, which is 32000x smaller in the measurement above.

## The no-metrics path is clean

- `eval_loss` and `eval_perplexity` are bit identical with and without `compute_metrics` (`1.365422010421753` and `3.9173758766594484`) across the flat path, the dict-split path and the labels-aware `train_on_responses_only` path.
- `_last_eval_metrics` contains exactly `{eval_loss, eval_perplexity}` when `compute_metrics is None`, no extra keys.
- Forward count stays at 2 with metrics off and 4 with metrics on, so the loss path including CCE is untouched.
- The lazy import works: `transformers.trainer_utils` is absent from `sys.modules` after import and after an eval with `compute_metrics=None`, appearing only once metrics are actually computed. One correction to the description though: this module was never torch-free at any ref, since `unsloth_zoo/device_type.py:30` reaches `unsloth_zoo/utils.py:28 import torch` on merge base, your head and `main` alike. This PR neither improves nor regresses that.
- The VLM path raises `NotImplementedError` before any forward pass, and VLM eval without `compute_metrics` still works.
- `EvalPrediction` and `denumpify_detensorize` have identical signatures at 4.57.6 and 5.14.1, TRL is untouched, and `MLXTrainingConfig` is untouched, so the field-ordering problem I found in #873 and #927 does not apply here. `MLXTrainer.__init__` only appends, so existing positional callers are safe.

Both e2e tests this adds are `@metal_only`, so they never run on Linux CI. Worth checking whether they can be loosened to "real mlx.core, not the shim", since the rest of the suite runs fine on Linux CPU MLX.

## Review thread

Thumbs down on the `mx.eval` one. It is technically accurate that the first `mx.eval` is redundant before a bfloat16 cast, since MLX dtype is static metadata, but it is a pure micro-optimization with no behavioural difference.

danielhanchen:
Looked at why this is still conflicting, and at what it will take to land. Short version: this is a stacked PR on #873, the stack is real rather than incidental, and the rebase is not mechanical, so it is worth writing down the one trap before anyone does it.

## Where it sits

The branch is #873's first four commits (`275b1d64`, `f268a2fa`, `086fcb27`, `ef40021c`) plus one of its own, `ddcd5b4c`, at +362/-6 across three files. #873 has since grown to sixty-five commits and is approved and mergeable at `8b986352`; main has taken #932 and #939. That is the whole reason for the conflict, and it is not something this PR did wrong.

```
ddcd5b4c cherry-picked onto     conflicts
  current main bf960653         7 regions / 3 files
  873 head     8b986352         6 regions / 2 files
```

Landing #873 first only removes one file's worth. The rest is a genuine overlap, because both PRs rewrite the same evaluation loop.

## The trap, and it is a quiet one

Five of the six regions are additive and resolve by keeping both sides. The sixth is not. `_evaluate_batch_totals` at `8b986352` now folds each batch into the totals inside a try block:

```
    if not failed and not self.stop_requested:
        try:
            ...
            all_losses += mx.where(ntoks > 0, loss * ntoks, 0.0)
            ntokens += ntoks
            mx.eval(all_losses, ntokens)
            self._fire_prediction_step()
        except BaseException as exc:
            failed = True
```

and `ddcd5b4c` adds its own unguarded copy of the same three accumulation lines just below, because at the time it was written there was nothing above it. A union resolution therefore compiles, passes a casual read, and **double counts every evaluation batch**. The eval loss would be silently doubled rather than raising anywhere.

So the rebase has to re-express the metric collection inside the existing try block rather than beside it. That is also the right place on its own merits: `_text_metric_prediction_step` runs the model, so a `compute_metrics` pipeline that throws on one rank has to join the same `_distributed_eval_status` consensus as a failed batch. Outside the try, one rank raises while its peers block at the next collective, which is the exact failure mode #873 went and fixed.

Two smaller ones in the same area:

- The split loop uses `self.stop_requested`; `8b986352` uses `_distributed_should_stop()` and rotates `handler.eval_dataloader` per split so `on_prediction_step` reports the split being consumed rather than the dict of splits.
- `_evaluate_batch_totals` returns two values there and three here, so every call site moves together.

## Suggested order

Land #873, then rebase this onto main and apply `ddcd5b4c` as a patch rather than merging the branch. A merge commit reproduces all four shared commits and conflicts far more; the repo squash merges, so the four shared commits have to come off anyway. The evaluation-loop integration above is the only part that needs judgement; the other five regions are keep-both.

Nothing here is a defect in the change itself. `compute_metrics` for text eval is a real gap and the shape of the implementation looks right. Leaving it as a draft until #873 lands is the correct state, and I have not pushed anything to this branch.

danielhanchen:
#873 has landed, so this can now be rebased onto current main and applied as a patch rather than merged. The one part that needs care is folding the metric collection into the existing try block in _evaluate_batch_totals instead of beside it, and moving the call sites over to its new return shape.

danielhanchen:
#873 is merged now, so the rebase is unblocked; when you pick this up please re-apply the metric collection inside main's existing _evaluate_batch_totals try block rather than beside it, so the zero-token guard and the distributed eval status path stay intact.

danielhanchen:
Still no compute_metrics path in unsloth_zoo/mlx/trainer.py on main, so the direction stands, but as written the accumulation sits beside the try block in _evaluate_batch_totals rather than inside it. Could you rebase onto main now that #873 has landed and fold the metric collection into the existing guarded path before I review?

PR #879, opened 2026-07-07
MLX: fix gemma-4-e2b (gemma4) audio-tower conv double transpose on load

danielhanchen:
### Problem

Loading `mlx-community/gemma-4-e2b-it-4bit` through the MLX backend fails with an HTTP 500 / shape mismatch:

```
Expected shape (128, 3, 3, 1) but received shape (128, 3, 1, 3)
for parameter audio_tower.subsample_conv_projection.layer0.conv.weight
```

### Root cause

gemma-4-e2b is multimodal (audio + vision + text). The pre-converted MLX checkpoints published by mlx-community already store the audio tower's convolution weights in MLX channel-last layout:

- `audio_tower.subsample_conv_projection.layer{0,1}.conv.weight` (Conv2d) as `[out, kH, kW, in]` (e.g. `[128, 3, 3, 1]`, `[32, 3, 3, 128]`)
- `audio_tower.layers.*.lconv1d.depthwise_conv1d.weight` (Conv1d) as `[out, kW, in]` (`[1024, 5, 1]`)

`mlx_vlm.models.gemma4.Model.sanitize` unconditionally re-applies the PyTorch to MLX transpose (`transpose(0, 2, 3, 1)` for Conv2d, `transpose(0, 2, 1)` for Conv1d) to those keys. On an already-channel-last checkpoint this is a second transpose, producing the invalid shapes above, and `load_weights(strict=True)` rejects them.

### Fix

`unsloth_zoo/mlx/loader.py` now wraps the mlx-vlm sanitize for the VLM load path (`_ensure_audio_conv_sanitize`, called from `FastMLXModel.from_pretrained`). When a checkpoint audio-conv weight already matches the instantiated module's channel-last target shape, it pre-applies the inverse permutation so the upstream transpose round-trips to the correct layout instead of double transposing. It only intervenes on an exact target-shape match, so genuine PyTorch-layout checkpoints (which need the transpose) are untouched, and the wrap is gated on the sanitize source signature rather than a hardcoded family. No other model loads are affected.

### Verification (real macos-14 Apple Silicon)

Verified on a GitHub-hosted `macos-14` (M1) runner with this branch overlaid into a Studio venv:

- Negative control (stock mlx-vlm, no guard): reproduces `ValueError: Expected shape (128, 3, 3, 1) but received shape (128, 3, 1, 3) for parameter audio_tower.subsample_conv_projection.layer0.conv.weight`.
- With the guard: `mlx_vlm` `load_model(..., strict=True)` accepts the checkpoint and the audio-conv weights are the correct channel-last shapes `(128, 3, 3, 1)`, `(32, 3, 3, 128)`, `(1024, 5, 1)`.
- On the Studio `/api/inference/load` path the load no longer hits the conv shape error. (Fully materializing this ~5B multimodal model still trips the free M1 runner's Metal GPU command-buffer watchdog, which is a runner-hardware limit orthogonal to this fix, so the checkpoint-shape validation is asserted directly.)

Run log: https://github.com/danielhanchen/unsloth-staging-2/actions/runs/28852083116

### DIFF
--- a/unsloth_zoo/mlx/loader.py
+++ b/unsloth_zoo/mlx/loader.py
@@ -49,6 +49,7 @@ from .compile import (
 
 _vlm_model_types_cache = None
 _SAFE_TEXT_SANITIZE_PATCHED: set[str] = set()
+_AUDIO_CONV_SANITIZE_PATCHED: set[str] = set()
 _MULTIMODAL_STRIP_KEYS = (
     "vision_tower",
     "audio_tower",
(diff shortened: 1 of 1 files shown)


danielhanchen:
@codex review

PR #885, opened 2026-07-08
saving: fold per-expert gpt-oss 4bit expert LoRA into merged_16bit

danielhanchen:
## Summary

`save_pretrained_merged(save_method="merged_16bit")` on a gpt-oss bnb-4bit LoRA produced a checkpoint that reloaded as the untrained base: the expert LoRA was silently dropped. The LoRA adapter itself reloaded correctly, only the merged model lost the expert deltas.

gpt-oss bnb-4bit stores experts as per-expert `nn.Linear` ModuleLists (`mlp.experts.gate_up_projs.<i>` / `down_projs.<i>`), so PEFT attaches one 2D LoRA per expert Linear. On a `merged_16bit` save the base weights come from the original gpt-oss repo, which is mxfp4-packed, so the whole merge routes through `_merge_and_overwrite_lora_mxfp4`. That function looked up LoRA only by the fused tensor name (`experts.gate_up_proj` / `down_proj`); the per-expert keys never matched, so every expert delta was skipped. No guard fired because those keys were also not counted as backed, so the save-count check still balanced.

## Fix

- `_fold_perexpert_lora_into_fused(W, fused_base_name, converted_lora_weights)`: after the mxfp4 dequant of a fused expert group `W` (shape `(E, in, out)`), fold each per-expert delta into its slice. `W[i]` is `(in, out)` and `_merge_lora` takes a Linear `(out, in)`, so transpose in and back. Wired into the packed `_blocks` branch when the fused-name lookup misses.
- The same fold is applied in `_merge_moe_experts_file` for a non-mxfp4 (bf16) fused base.
- `_lora_key_has_backing` now recognises `experts.gate_up_projs.<i>` / `down_projs.<i>` as backed by the fused `experts.gate_up_proj` / `down_proj` tensor (bf16 3D or mxfp4 packed `_blocks`/`_scales`), so the save-count check balances once the deltas are folded instead of raising.

Fused `target_parameters` MoEs (Qwen3, Gemma4) key their expert LoRA at `experts.gate_up_proj` with no per-expert index, so the regex never matches and their merge path is unchanged; the helper returns 0 for any non-matching group.

## Validation

gpt-oss-20b-unsloth-bnb-4bit (transformers 5.5.0), memorize overfit:

| reload | before | after |
| --- | --- | --- |
| LoRA adapter | hard match, CE 4e-6 | hard match, CE 4e-6 |
| merged_16bit | fail, CE 3.34 (base) | hard match, CE 0.06 |

Every layer folds its 32 experts for both `gate_up_proj` and `down_proj`. A synthetic unit test confirms the fold orientation and scaling and that untrained (`B=0`) experts and unrelated keys are left untouched. The BF16 gpt-oss fused path (`_merge_moe_fused_gate_up_expert`, unchanged) was separately re-checked and still reloads exactly (CE 1e-6).

Requires the per-expert targeting from unslothai/unsloth#6936 to attach the expert LoRA in the first place.

### DIFF
--- a/unsloth_zoo/saving_utils.py
+++ b/unsloth_zoo/saving_utils.py
@@ -1564,6 +1564,51 @@ def _merge_moe_experts_file(mm, header_metadata, length_of_header, file, convert
         except Exception:
             pass
 
+    # Per-expert Linear MoE experts (gpt-oss bnb-4bit): PEFT stores one 2D LoRA per
+    # expert Linear (mlp.experts.gate_up_projs.<i> / down_projs.<i>), but the bf16
+    # checkpoint is the fused 3D group, which the mapping below misses. Fold each
+    # per-expert delta into its slice.
+    if is_gpt_oss_format:
+        _perexpert = {}
+        for _lk, _ls in converted_lora_weights.items():
+            if not isinstance(_lk, str):
+                continue
+            _m = re.match(r"^(.*\.experts)\.(gate_up_projs|down_projs)\.(\d+)$", _lk)
+            if _m is None:
+                continue
+            _fused_name = "gate_up_proj" if _m.group(2) == "gate_up_projs" else "down_proj"
+            _fused_key = f"{_m.group(1)}.{_fused_name}"
+            if _fused_key not in header_metadata:
+                continue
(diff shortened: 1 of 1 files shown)


danielhanchen:
@codex review

PR #894, opened 2026-07-09
fix(mlx): accept Studio VLM singular image batches

Lyxot:
## Summary

This fixes MLX VLM collation for Studio-shaped training rows that store the image payload in a top-level singular `image` field while the chat message contains a bare image placeholder. One example dataset that uses this shape is `unsloth/LaTeX_OCR`:

```python
{
    "messages": [
        {
            "role": "user",
            "content": [
                {"type": "image"},
                {"type": "text", "text": "..."},
            ],
        },
        {"role": "assistant", "content": [{"type": "text", "text": "..."}]},
    ],
    "image": image,
}
```

## Root Cause

The MLX VLM image extraction path accepted top-level `images`, inline media payloads, and media returned by `process_vision_info()`, but it did not accept Studio’s singular top-level `image` shape. For rows with a bare `{ "type": "image" }` placeholder, `process_vision_info()` raises because the media payload is intentionally outside the message content. That caused otherwise valid Studio-shaped VLM SFT rows to fail before training.

## Changes

- Detect VLM message content that contains bare image placeholders.
- Allow top-level singular `image` as a fallback only when the message contains bare image placeholders and no inline media payload.
- Preserve precedence for top-level `images` and inline or processable message media.
- Preserve existing error behavior for malformed inline media instead of hiding those errors behind the singular `image` fallback.
- Add focused MLX VLM tests for the Studio-shaped success path and fallback boundaries.

## Reviewer Notes

The fallback is intentionally narrow. It does not make `image` a general alias for `images`; it only covers the Studio-shaped row layout where the message has a bare image placeholder and the row has the actual image payload. Mixed bare placeholders such as image plus video still reject fallback to avoid binding a still image to a non-image media slot.

## Validation

Focused test suite:

```bash
python -m pytest tests/test_mlx_vlm_label_masks.py -q
```

Result:

```text
45 passed
```

## Gemma4 and Dataset Validation

Gemma4 still depends on the transformers sidecar repair in https://github.com/unslothai/unsloth/pull/7402. With this branch merged together with the current https://github.com/unslothai/unsloth/pull/7402 head, `unsloth/gemma-4-E2B-it-UD-MLX-4bit` loads as the expected MLX-VLM processor instead of the tokenizer-only fallback:

```text
mlx_vlm.models.gemma4.processing_gemma4.Gemma4Processor
```

The Studio-shaped SFT smoke used these vision datasets:

```text
unsloth/LaTeX_OCR
AI4Math/MathVerse
```

2-step validation results with `unsloth/gemma-4-E2B-it-UD-MLX-4bit` after merging https://github.com/unslothai/unsloth/pull/7402:

```text
unsloth/LaTeX_OCR: passed, losses [2.9979827404022217, 2.8777084350585938]
AI4Math/MathVerse: passed, losses [3.8615834712982178, 3.614513635635376]
```

Gemma4 should work fine once https://github.com/unslothai/unsloth/pull/7402 is merged together with this PR.

### DIFF
--- a/tests/test_mlx_vlm_label_masks.py
+++ b/tests/test_mlx_vlm_label_masks.py
@@ -485,35 +485,15 @@ def test_vlm_prompt_completion_prefers_embedded_images_like_cuda():
     assert processor.images_seen[0] == ["embedded"]
 
 
-def test_vlm_top_level_image_key_falls_back_after_bare_placeholder_error(monkeypatch):
-    import unsloth_zoo.vision_utils as vision_utils
+def test_vlm_top_level_image_key_falls_back_after_real_bare_placeholder_error():
     from unsloth_zoo.mlx.utils import _extract_vlm_images
-
-    def fail_process_vision_info(*_args, **_kwargs):
-        raise ValueError("image, image_url or video should in content.")
-
-    monkeypatch.setattr(
-        vision_utils,
-        "process_vision_info",
-        fail_process_vision_info,
-    )
 
     messages = [{"role": "user", "content": [{"type": "image"}]}]
     assert _extract_vlm_images({"image": "top-level"}, messages, image_size=16) == ["top-level"]
 
 
-def test_vlm_collate_passes_studio_top_level_image_to_processor(monkeypatch):
-    import unsloth_zoo.vision_utils as vision_utils
+def test_vlm_collate_passes_studio_top_level_image_to_processor():
     from unsloth_zoo.mlx.utils import _collate_vlm_batch
-
-    def fail_process_vision_info(*_args, **_kwargs):
-        raise ValueError("image, image_url or video should in content.")
-
-    monkeypatch.setattr(
-        vision_utils,
-        "process_vision_info",
-        fail_process_vision_info,
-    )
 
     processor = _ConversationalPromptCompletionProcessor()
     _collate_vlm_batch(
--- a/tests/test_mlx_vlm_label_masks.py
+++ b/tests/test_mlx_vlm_label_masks.py
@@ -485,11 +485,13 @@ def test_vlm_prompt_completion_prefers_embedded_images_like_cuda():
     assert processor.images_seen[0] == ["embedded"]
 
 
-def test_vlm_top_level_image_key_uses_bare_image_placeholder_fallback():
-    from unsloth_zoo.mlx.utils import _extract_vlm_images
+def test_vlm_prompt_completion_uses_top_level_image_for_bare_placeholder():
+    from unsloth_zoo.mlx.utils import _extract_vlm_pc_images
 
     messages = [{"role": "user", "content": [{"type": "image"}]}]
-    assert _extract_vlm_images({"image": "top-level"}, messages, image_size=16) == ["top-level"]
+    assert _extract_vlm_pc_images(
+        {"image": "top-level"}, messages, [], image_size=16,
+    ) == ["top-level"]
 
 
 def test_vlm_collate_passes_studio_top_level_image_to_processor():
--- a/tests/test_mlx_vlm_label_masks.py
+++ b/tests/test_mlx_vlm_label_masks.py
@@ -485,10 +485,31 @@ def test_vlm_prompt_completion_prefers_embedded_images_like_cuda():
     assert processor.images_seen[0] == ["embedded"]
 
 
-def test_vlm_top_level_image_key_is_not_cuda_images_alias():
+def test_vlm_top_level_image_key_supports_studio_adapter_shape(monkeypatch):
+    import unsloth_zoo.vision_utils as vision_utils
     from unsloth_zoo.mlx.utils import _extract_vlm_images
 
-    assert _extract_vlm_images({"image": "top-level"}, [], image_size=16) == []
+    def fail_process_vision_info(*_args, **_kwargs):
+        raise AssertionError("top-level image should be used before message fallback")
+
+    monkeypatch.setattr(
+        vision_utils,
+        "process_vision_info",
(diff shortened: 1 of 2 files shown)


danielhanchen:
Reviewed this. Merging as is, no changes needed from me.

Method: MLX runs on Linux CPU (`pip install "mlx[cpu]" mlx-lm mlx-vlm`), so I ran a 32-shape by 3-mode matrix (96 probes) plus end-to-end collation with two real HF processors (Gemma-3 and SmolVLM) at both the merge base and head.

## The bug is real, on the dataset you named

`unsloth/LaTeX_OCR` really is `{"image": Image, "text": Value}` with a singular `image` column and no `messages`, so any Studio conversion that synthesizes messages with a bare `{"type": "image"}` placeholder and leaves `image` in place hits this exactly. Two real rows through `_collate_vlm_batch` with a real Gemma-3 processor:

```
[BASE] RAISE ValueError: image, image_url or video should in content.
[HEAD] OK  pixels=(2, 3, 896, 896)  image_tokens=512 (expected 256*2)  image tokens masked from loss: True
```

Root cause is `vision_utils.py:616`, reached because `extract_vision_info` collects the bare part and then finds no `image` / `image_url` / `video` key. Upstream `qwen_vl_utils/vision_process.py` raises the identical string, so this is inherited behaviour rather than something local.

Negative control: reverting only the seven-line fallback hunk in an otherwise head-identical worktree gives back `ValueError: image, image_url or video should in content.` on both processors, and turns the two new tests red (`2 failed, 43 passed`). Restoring gives 45 passed. The tests pin exactly the reverted hunk.

## The fallback really is narrow

| requirement | result |
| --- | --- |
| top-level `images` (plural) wins | pass, `{"images":[INL],"image":TOP}` gives `[INL]` at base and head |
| inline media payloads win | pass, same turn or a later turn, `[INL]` at base and head |
| media from `process_vision_info` | pass by construction and empirically: the fallback requires at least one bare image placeholder, and `extract_vision_info` always collects that part, so PVI can never succeed while the fallback is eligible |
| malformed inline media still raises | pass, `image: None` gives `Unrecognized image input`, a bad `image_url` path gives `FileNotFoundError`, malformed plus bare still gives the PVI `ValueError`, all identical at base and head |
| mixed image plus video placeholders reject the fallback | pass, raises at head exactly as at base |

Adversarial shapes that did not break the ordering: `image_url` / `input_image` / `video` inline payloads alongside a bare placeholder, a bare placeholder split across turns, `content` as a dict, messages as a tuple, a non-dict message, `{"images": []}` plus `image`, and nested `[[PIL]]`. Two head-only behaviour changes showed up and both are correct: an audio part no longer blocks the image fallback (audio is not image media), and `{"type": "image", "detail": "high"}` (OpenAI style, no payload key) is now treated as a bare placeholder.

Payload-type parity between singular `image` and the already-supported plural `images`, end to end through Gemma-3: PIL same, path string same, base64 data URI same, `file://` same (both raise), raw bytes same (both raise). No new payload-handling class is introduced.

## No regressions

`tests/test_mlx_*.py`: merge base 52 failed / 470 passed, head 52 failed / 474 passed, and a diff of the FAILED sets is empty. The 52 are pre-existing cross-test pollution (each passes in isolation) and identical on both sides. The PR's own test file is 45 passed.

The production diff adds no imports and touches no `transformers` / `mlx` / `mlx-vlm` API; it is pure dict and list logic over message parts. I checked anyway on both transformers 4.57.6 and 5.5.0 (the repo cap) with mlx 0.32.0 / mlx-lm 0.31.3 / mlx-vlm 0.6.7: 45 passed and the shape parity holds on both. No `os.path`, no encoding and no PIL calls in the diff, and the payload flows through the unchanged `_resize_vlm_images`, so Linux, macOS, Windows and WSL are identical.

## Review threads

Thumbs up on the prompt-completion thread. It was correct at the commit it was written against (`e9849d95`, where a Studio prompt-completion row gave `[]`) and your head commit `4c2bb354` fixes it (now `[TOP]`). I verified by running the same probe against both commits.

Thumbs down on the other two. The `input_image` one is not a regression: a bare `input_image` placeholder plus a top-level `image` gives `[]` at both merge base and head, and neither upstream `qwen_vl_utils` nor the CUDA reference `_extract_images_for_pc` recognizes `input_image` at all, so extending the fallback there would put MLX ahead of both references. The "preserve top-level images when inline media also exists" one describes a mixed-format row, and behaviour is byte identical at base and head: the extractor returns `[INL]` and the processor raises the same precise `ValueError: Prompt contained 2 image tokens but received 1 images.` on both. Loosening the `not images` gate would break the inline-wins requirement above.

Copilot's three 2026-07-09 items are moot: that exception-string-matching design is gone at head (`should in content`, `can_fallback` and `str(e)` all return nothing from `unsloth_zoo/mlx/utils.py`).

## Three non-blocking notes

1. `{"images": [], "image": img}` does not fall back, because the `elif` is skipped when `[] is not None`, while `{"images": None, "image": img}` does. Both raise at merge base so there is no regression, but the asymmetry is arbitrary.
2. `_extract_vlm_pc_images` now consults singular `image` when messages exist, whereas the CUDA reference `UnslothVisionDataCollator._extract_images_for_pc` (`vision_utils.py:1154`) never consults any top-level key in that branch. That means in the prompt-completion path singular `image` now works while plural `images` still returns `[]`. Pre-existing parity gap, worth a follow-up rather than this PR.
3. The fallback matches placeholder types (a set), not counts, so two placeholders with one image reaches the processor. I confirmed it fails loudly with a precise message on both Gemma-3 and SmolVLM rather than mistraining, so this is fine.

danielhanchen:
@codex review

PR #897, opened 2026-07-10
Fix ModuleNotFoundError crash in fla_vendor's is_flash_linear_attention_available probe

### LINKED ISSUE #896: ModuleNotFoundError (torchvision) crashes import unsloth via fla_vendor's is_flash_linear_attention_available probe
Reproduced on unsloth 2026.7.2 / unsloth_zoo 2026.7.2, transformers 5.5.0, Python 3.13.13, no torchvision installed.

**Traceback:**
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File ".../unsloth/__init__.py", line 1427, in <module>
    from ._gpu_init import *
  File ".../unsloth/_gpu_init.py", line 382, in <module>
    from .models import *
  File ".../unsloth/models/__init__.py", line 15, in <module>
    from .llama import FastLlamaModel
  File ".../unsloth/models/llama.py", line 21, in <module>
    from ._utils import *
  File ".../unsloth/models/_utils.py", line 945, in <module>
    _run_temporary_patches("init")
  File ".../unsloth/models/_utils.py", line 938, in _run_temporary_patches
    temporary_patch(phase = phase)
  File ".../unsloth_zoo/temporary_patches/fla_vendor.py", line 579, in patch_vendor_fla
    _patch_is_available()
  File ".../unsloth_zoo/temporary_patches/fla_vendor.py", line 493, in _patch_is_available
    if getattr(mod, "is_flash_linear_attention_available", None) is original:
  File ".../transformers/__init__.py", line 843, in _getattr
    return getattr(importlib.import_module(target, __name__), new_name)
  File ".../transformers/models/aria/image_processing_aria.py", line 21, in <module>
    from torchvision.transforms.v2 import functional as tvF
ModuleNotFoundError: No module named 'torchvision'
```

**Root cause:** the sys.modules rebind loop in _patch_is_available() calls getattr(mod, ...) on every loaded transformers.* module. transformers' lazy module loader resolves unknown attributes on a submodule by importing that submodule fully — hitting transformers.models.aria triggers an unconditional torchvision import. torchvision isn't a required dependency, so this crashes all of 'import unsloth' whenever it's missing.

Every other risky access in this function is already wrapped in try/except Exception — this one getattr was the one spot missed.

**Fix:** wrap the getattr in try/except, same pattern as the rest of the function. Verified fix resolves the crash with torchvision uninstalled. PR incoming.

amareshhebbar:
Fixes #896.

Root cause and full repro in the linked issue. One-line try/except guard around the getattr in _patch_is_available()'s sys.modules rebind loop — matches the existing exception-handling pattern already used elsewhere in this function. Verified locally: reproduced the crash with torchvision uninstalled, confirmed this patch resolves it (unsloth imports cleanly).

### DIFF
--- a/unsloth_zoo/temporary_patches/fla_vendor.py
+++ b/unsloth_zoo/temporary_patches/fla_vendor.py
@@ -490,7 +490,11 @@ def _patch_is_available():
                 continue
             if not (name == "transformers" or name.startswith("transformers.")):
                 continue
-            if getattr(mod, "is_flash_linear_attention_available", None) is original:
+            try:
+                attr = getattr(mod, "is_flash_linear_attention_available", None)
+            except Exception:
+                continue
+            if attr is original:
                 try:
                     setattr(mod, "is_flash_linear_attention_available", _vendored_availability_probe)
                 except Exception:
--- a/unsloth_zoo/temporary_patches/fla_vendor.py
+++ b/unsloth_zoo/temporary_patches/fla_vendor.py
@@ -490,11 +490,12 @@ def _patch_is_available():
                 continue
             if not (name == "transformers" or name.startswith("transformers.")):
                 continue
-            try:
-                attr = getattr(mod, "is_flash_linear_attention_available", None)
-            except Exception:
+            # Read the static binding via __dict__ so we never fire transformers'
+            # lazy __getattr__, which imports optional deps (e.g. torchvision) and crashes.
+            mod_dict = getattr(mod, "__dict__", None)
+            if not isinstance(mod_dict, dict):
                 continue
-            if attr is original:
+            if mod_dict.get("is_flash_linear_attention_available") is original:
                 try:
                     setattr(mod, "is_flash_linear_attention_available", _vendored_availability_probe)
                 except Exception:
--- a/unsloth_zoo/temporary_patches/fla_vendor.py
+++ b/unsloth_zoo/temporary_patches/fla_vendor.py
@@ -490,8 +490,8 @@ def _patch_is_available():
                 continue
             if not (name == "transformers" or name.startswith("transformers.")):
                 continue
-            # Read the static binding via __dict__ so we never fire transformers'
-            # lazy __getattr__, which imports optional deps (e.g. torchvision) and crashes.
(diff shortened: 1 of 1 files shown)


danielhanchen:
Thanks @amareshhebbar, the root cause here is exactly right. I pushed a small follow-up commit that adjusts the approach slightly.

Instead of wrapping the `getattr` in `try/except`, I read the binding straight from `mod.__dict__`:

```python
mod_dict = getattr(mod, "__dict__", None)
if not isinstance(mod_dict, dict):
    continue
if mod_dict.get("is_flash_linear_attention_available") is original:
    ...
```

The reason: `getattr(mod, "is_flash_linear_attention_available")` still fires Transformers' lazy `__getattr__` on every module it visits, which resolves the deprecation alias and imports the backing submodule (aria image processing, which pulls in the optional torchvision). The `try/except` does stop the crash, but the loop keeps triggering a lot of unnecessary lazy imports as a side effect. Reading `__dict__` directly only matches modules that actually did `from ...import_utils import is_flash_linear_attention_available`, so it never touches the lazy path at all. This mirrors the pattern we already use in Studio's `_rebind_in_already_imported_modules`.

I tested this across a matrix of isolated venvs (transformers 4.57.6, 5.0.0, 5.5.0, 5.13.1, each with and without torchvision, on Python 3.11 / 3.12 / 3.13):

- Without the fix, `import unsloth` crashes on transformers >= 5.5 when torchvision is absent (the exact `ModuleNotFoundError` from #896). transformers 4.x and 5.0 are unaffected.
- With the fix, import is clean on every combination, and the rebind still catches the genuine `from ... import` holders identically to before.
- The `__dict__` read triggers zero extra model-module imports and zero deprecation warnings, versus a few hundred lazy submodule imports with plain `getattr` (about 750 on transformers 5.13 when torchvision is installed).

Behavior is unchanged everywhere else, so this is safe on older transformers and on CPU only setups. Good to merge once CI is green.

danielhanchen:
@codex review

danielhanchen:
@codex review

danielhanchen:
Thanks!

PR #900, opened 2026-07-10
Fix false DownloadStallError for non-standard sharded weight filenames

### LINKED ISSUE #889: maybe_prefetch_hf_snapshot raises a false DownloadStallError ('incomplete snapshot … missing files') for unsloth/Qwen3.5-4B — even over HTTP (HF_HUB_DISABLE_XET=1) on a complete snapshot
## Summary

Since the Xet→HTTP prefetch fallback was added to `unsloth_zoo` in #829, `FastLanguageModel.from_pretrained("unsloth/Qwen3.5-4B")` **deterministically** fails before any training/inference with:

```
unsloth_zoo.hf_xet_fallback.DownloadStallError: Download for 'unsloth/Qwen3.5-4B'
returned an incomplete snapshot even with HF_HUB_DISABLE_XET=1 -- missing files,
check your network connection
```

This is **not** a network problem and **not** the Xet hang the fallback was designed for:

- It reproduces on **fresh** environments across **multiple independent hosts** (different IPs / data centers), identically.
- It fails the same way over **both** transports — with `HF_HUB_DISABLE_XET=1` set **up front** (so `_download_with_xet_fallback` starts with `disable_xet=True`), the very first attempt is plain HTTP and it still reports the snapshot incomplete.
- An **authenticated** download (HF token) makes no difference.
- The exact same model, on the exact same hosts, **downloads and trains fine** when `unsloth_zoo` is pinned to **2026.6.7** (which predates #829, so `maybe_prefetch_hf_snapshot` takes its `except ImportError: return False` path and `from_pretrained` downloads normally).

The differentiator is the `unsloth_zoo` version, holding everything else constant → this points at the prefetch's completeness check **false-rejecting a genuinely complete snapshot**, not a real "missing files" condition. (`unsloth_zoo/hf_cache_state.py` even carries comments about *"must not be false-rejected into a DownloadStallError"* / *"a finished download into a DownloadStallError"*, so this failure mode is anticipated but evidently still reachable for this repo.)

## Environment (failing)

- `unsloth==2026.7.2`, `unsloth_zoo==2026.7.2`
- `transformers==5.5.0`, `torch==2.10.0+cu128`, `triton==3.6.0`, `xformers==0.0.35`, `flash-attn==2.8.1`
- Python 3.13, CUDA 12.8, Linux, NVIDIA RTX A5000
- Model: `unsloth/Qwen3.5-4B` (16-bit LoRA load: `load_in_16bit=True, load_in_4bit=False`)

**Works** with the identical stack except `unsloth==2026.6.9` / `unsloth_zoo==2026.6.7`.

## Reproduction

Fresh cache (`HF_HOME` pointing at an empty dir), then:

```python
import os
os.environ["HF_HUB_DISABLE_XET"] = "1"   # force HTTP from the first attempt — still fails

    "unsloth/Qwen3.5-4B",
    max_seq_length = 4096,
    load_in_4bit   = False,
    load_in_16bit  = True,
)
```

→ raises `DownloadStallError` before the model loads.

## Traceback (abridged)

```
File ".../unsloth/models/vision.py", line ~985, in from_pretrained
    _prefetched = maybe_prefetch_hf_snapshot(model_name, ...)
File ".../unsloth/models/_utils.py", line ~1353, in maybe_prefetch_hf_snapshot
    snapshot_download_with_xet_fallback(model_name, ...)
File ".../unsloth_zoo/hf_xet_fallback.py", line ~1740, in _download_with_xet_fallback
    raise DownloadStallError(
unsloth_zoo.hf_xet_fallback.DownloadStallError: Download for 'unsloth/Qwen3.5-4B'
returned an incomplete snapshot even with HF_HUB_DISABLE_XET=1 -- missing files,
check your network connection
```

## Expected

`from_pretrained("unsloth/Qwen3.5-4B")` downloads (or warms) the snapshot and loads the model — as it does on `unsloth_zoo==2026.6.7` and as the plain `huggingface_hub` / `transformers` download path does for this repo.

## Likely root cause (hypothesis)

The prefetch computes an expected-file set (the `allow_patterns` / `_prefetch_ignore_patterns` logic in `maybe_prefetch_hf_snapshot`, checked against the downloaded snapshot by `hf_cache_state`) that does not match `unsloth/Qwen3.5-4B`'s actual file layout — so a complete snapshot is judged "incomplete / missing files" and the HTTP-fallback branch escalates it to a hard `DownloadStallError` instead of deferring to `from_pretrained`. Because the check is content/layout-based, it's deterministic per repo rather than flaky.

The docstring says *"Only a both-transports-stalled `DownloadStallError` is raised; other failures are left for `from_pretrained` to surface"* — but here a **false** incompleteness verdict is being raised as a hard error rather than falling through to `from_pretrained`, which defeats that intent.

## Suggested fix directions

- When the completeness check reports "incomplete" but the transport itself reported **no** error/stall, **return `False`** (skip the prefetch, let `from_pretrained` do the download) instead of raising — i.e. treat a completeness mismatch as "couldn't warm the cache", not "network is broken".
- Or reconcile the prefetch's expected-file set with the actual repo file list for models like `unsloth/Qwen3.5-4B` (there may be an off-by-one in the ignore/allow patterns or the `tokenizer.model`/aux-file expectation).

## Regression range

- **Introduced:** #829 (`Add shared Xet to HTTP stall fallback`), merged 2026-07-06 12:11 UTC — the commit that first adds `unsloth_zoo/hf_xet_fallback.py` + `hf_cache_state.py`.
- **Last good:** `unsloth_zoo==2026.6.7` (no `hf_xet_fallback` module → prefetch is skipped).

## Workaround

Pin `unsloth==2026.6.9` / `unsloth_zoo==2026.6.7`, or otherwise use a version without the prefetch. Setting `HF_HUB_DISABLE_XET=1` / `UNSLOTH_STABLE_DOWNLOADS=1` / providing an HF token does **not** work around it — the completeness check fails regardless of transport or auth.

olsh:
## Problem

Fixes #889.

Loading `unsloth/Qwen3.5-4B` deterministically raises:

```
DownloadStallError: Download for 'unsloth/Qwen3.5-4B' returned an incomplete
snapshot even with HF_HUB_DISABLE_XET=1 -- missing files
```

even on a fully-downloaded, complete snapshot (reproduces across hosts, and over plain HTTP with Xet disabled). The model loaded fine before #829.

## Root cause

The repo shards its safetensors with a **non-standard filename** that keeps the format suffix in the middle:

```
model.safetensors-00001-of-00002.safetensors
model.safetensors-00002-of-00002.safetensors
model.safetensors.index.json      # weight_map -> the two names above
```

`transformers` loads this fine (it reads the index and follows `weight_map`). But the post-download presence check `_root_model_has_weight` only recognizes weight **files** matching `_CANONICAL_ROOT_MODEL_WEIGHT_RE` (`^(?:model|pytorch_model)(?:-\d{5}-of-\d{5})?\.(?:safetensors|bin)$`). `model.safetensors-00001-of-00002.safetensors` does not match, and the function never consults the shard index — so Invariant A ("a readable weight is present") returns `False`, `_download_result_usable` returns `False`, and a complete snapshot is failed.

This was a **pre/post asymmetry**: the pre-download gate (`snapshot_dir_is_complete` → `_canonical_root_weights_complete`) already handled this layout via the index, and the completeness half (Invariant B, `_readable_shard_set_incomplete` → `_has_incomplete_canonical_root_shards`) does too — only the post-download presence check did not.

## Fix

In `_root_model_has_weight`, additionally accept a canonical root shard index (`model.safetensors.index.json` / `pytorch_model.bin.index.json`) of a kept format as evidence that a readable weight is present (reusing the existing `_is_canonical_weight_shard_index` predicate). `transformers` enumerates the sharded weight through the index, so its presence is the correct readable-weight signal; shard **completeness** stays Invariant B's job, so an index whose shards are missing is still correctly rejected.

Scope: limited to the non-variant root path the issue hits. The variant sibling `_root_has_variant_weight` has the same latent blind spot but only for the rare variant + non-standard-shard-name combination; a docstring note documents that as a deliberate scope decision.

## Verification

Against the real cached `unsloth/Qwen3.5-4B` (2-shard) and `unsloth/qwen3.5-2b` (1-shard) models:

- Before: `_snapshot_payload_incomplete(...) = True` → would raise `DownloadStallError`.
- After: every gate (`_root_model_has_weight`, `snapshot_dir_is_complete`, `_cache_can_skip_download`, `_download_result_usable`, `_snapshot_payload_incomplete`) passes, and `snapshot_download_with_xet_fallback('unsloth/Qwen3.5-4B')` returns the snapshot dir with no error.
- Safety preserved: a snapshot whose index references a missing shard is still rejected; the standard `model-00001-of-00002.safetensors` layout is unaffected.

Adds a regression test `test_post_accepts_nonstandard_sharded_safetensors_names` covering the accepted, missing-shard-rejected, and standard-layout cases.

### DIFF
--- a/tests/test_hf_xet_fallback.py
+++ b/tests/test_hf_xet_fallback.py
@@ -2197,6 +2197,45 @@ def test_pre_download_does_not_skip_diffusers_but_post_accepts(tmp_path):
         snap, repo_type = "model", allow_patterns = None, ignore_patterns = None) is True
 
 
+def test_post_accepts_nonstandard_sharded_safetensors_names(tmp_path):
+    """Regression for issue #889: unsloth/Qwen3.5-4B shards its safetensors with a NON-standard name that
+    keeps the format suffix in the middle (model.safetensors-00001-of-00002.safetensors), referenced that
+    way by model.safetensors.index.json. transformers enumerates the weight through the index, so the
+    post-download gate must accept it rather than raise a spurious DownloadStallError. Shard completeness
+    is still enforced: a missing shard is rejected, and the standard sharded layout is unaffected."""
+    shards = ["model.safetensors-00001-of-00002.safetensors",
+              "model.safetensors-00002-of-00002.safetensors"]
+    # Complete cache with the non-standard shard names -> accepted (the #889 fix).
+    snap, blob = _mk_snapshot(tmp_path, "qwen35_nonstd")
+    (snap / "config.json").write_text("{}")
+    for s in shards:
+        (snap / s).symlink_to(blob)
+    (snap / "model.safetensors.index.json").write_text(json.dumps(
+        {"weight_map": {"a": shards[0], "b": shards[1]}}))
+    assert xf._download_result_usable(
+        snap, repo_type = "model", allow_patterns = None, ignore_patterns = None) is True
+
+    # Same layout but one shard missing -> still rejected (Invariant B validates via the index).
+    snap2, blob2 = _mk_snapshot(tmp_path, "qwen35_nonstd_missing")
+    (snap2 / "config.json").write_text("{}")
+    (snap2 / shards[0]).symlink_to(blob2)
+    (snap2 / "model.safetensors.index.json").write_text(json.dumps(
+        {"weight_map": {"a": shards[0], "b": shards[1]}}))
+    assert xf._download_result_usable(
+        snap2, repo_type = "model", allow_patterns = None, ignore_patterns = None) is False
+
+    # Sanity: the standard sharded layout still passes.
+    snap3, blob3 = _mk_snapshot(tmp_path, "qwen35_standard")
+    (snap3 / "config.json").write_text("{}")
+    std = ["model-00001-of-00002.safetensors", "model-00002-of-00002.safetensors"]
+    for s in std:
+        (snap3 / s).symlink_to(blob3)
+    (snap3 / "model.safetensors.index.json").write_text(json.dumps(
+        {"weight_map": {"a": std[0], "b": std[1]}}))
+    assert xf._download_result_usable(
+        snap3, repo_type = "model", allow_patterns = None, ignore_patterns = None) is True
+
+
 def test_pre_download_defers_sentence_transformers_missing_subfolder_weight(tmp_path):
     """A sentence-transformers model reads a weight-bearing subfolder module (2_Dense) the canonical root
     gate does not check, so a partial cache holding the root weight but missing the 2_Dense weight must
--- a/tests/test_hf_xet_fallback.py
+++ b/tests/test_hf_xet_fallback.py
@@ -2369,6 +2369,46 @@ def test_post_download_rejects_canonical_only_for_variant(tmp_path):
         variant = "fp16") is True
 
 
+def test_post_accepts_nonstandard_sharded_variant_names(tmp_path):
+    """Variant analog of #889: a variant sharded with NON-standard shard names the variant-weight regex
+    misses (the format suffix kept in the middle) is enumerated through model.safetensors.index.fp16.json,
+    so the post-download gate must accept it rather than raise a spurious DownloadStallError. Shard
+    completeness is still enforced via the index: a missing shard is rejected."""
+    shards = ["model.fp16.safetensors-00001-of-00002.safetensors",
+              "model.fp16.safetensors-00002-of-00002.safetensors"]
+    # Complete cache with the non-standard variant shard names -> accepted.
+    snap, blob = _mk_snapshot(tmp_path, "varnonstd")
+    (snap / "config.json").write_text("{}")
+    for s in shards:
+        (snap / s).symlink_to(blob)
+    (snap / "model.safetensors.index.fp16.json").write_text(json.dumps(
+        {"weight_map": {"a": shards[0], "b": shards[1]}}))
+    assert xf._download_result_usable(
+        snap, repo_type = "model", allow_patterns = None, ignore_patterns = None,
+        variant = "fp16") is True
+
+    # Same layout but one shard missing -> still rejected (Invariant B validates via the index).
+    snap2, blob2 = _mk_snapshot(tmp_path, "varnonstd_missing")
+    (snap2 / "config.json").write_text("{}")
+    (snap2 / shards[0]).symlink_to(blob2)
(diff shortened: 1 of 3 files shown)


danielhanchen:
@codex review

danielhanchen:
@codex review

danielhanchen:
@codex review

danielhanchen (approved):
Reviewed and verified end to end. This fixes a real, deterministic bug: the post-download completeness check rejected complete snapshots for repos that shard safetensors with the non-standard name that keeps the format suffix in the middle (`model.safetensors-00001-of-00002.safetensors` + `model.safetensors.index.json`), which is exactly the layout `unsloth/Qwen3.5-4B` and `unsloth/qwen3.5-2b` ship. Before the change a fully downloaded snapshot raised `DownloadStallError`; after it, it loads.

The approach is correct and matches how transformers resolves shards: it reads the filenames from the index `weight_map`, so shard names are arbitrary. Presence is now proven by the index, while shard completeness stays with the existing Invariant B check, so a genuinely missing or malformed shard is still rejected.

Verification I ran:

- Before/after differential across a large matrix of layouts (single-file, standard sharded, non-standard sharded, variants, sentence-transformers, diffusers, malformed and unsafe indexes) on both regular files and symlinks: no regressions, the only behavior change is the intended acceptance of complete non-standard snapshots.
- Missing-shard, index-only, malformed, non-dict, empty `weight_map`, and unsafe (drive-letter / UNC / backslash / parent-escaping) refs are all still rejected on every platform.
- bin shards work too, not just safetensors.
- Real fresh download of `unsloth/qwen3.5-2b` with `HF_HUB_DISABLE_XET=1` returns the snapshot with no `DownloadStallError`.
- Clean-room run against huggingface_hub 1.x passes.

LGTM. I also pushed a small comment condensation on top (comments only, code unchanged and AST-verified).

Two optional, non-blocking follow-ups for later: a non-standard `.bin` regression test to round out coverage, and separately, `_weight_shard_index_complete` uses `exists()` rather than `is_file()` (pre-existing, also affects the standard layout, only reachable via a malformed snapshot) which belongs in its own change.

danielhanchen:
@codex review

danielhanchen:
@codex review

danielhanchen:
@codex review

danielhanchen:
Thanks!

PR #908, opened 2026-07-17
feat(mlx): support VLM assistant-only loss

Lyxot:
## Summary

This PR adds direct `MLXTrainingConfig(assistant_only_loss=True)` support for conversational vision-language model training in `MLXTrainer`.

The implementation supervises every assistant-authored turn in a multi-turn conversation. System, user, tool-response, padding, multimodal placeholder/expansion, and model-specific ignored tokens remain `-100`. The same effective labels are used by materialized and streaming training, evaluation, baseline cross-entropy, and Cut Cross Entropy (CCE).

The public Unsloth application dispatcher is intentionally unchanged. This PR establishes the zoo-side MLX trainer contract so argument routing can be added separately without coupling the masking implementation to application or Studio plumbing.

## Motivation

The MLX trainer previously rejected `assistant_only_loss=True` for VLM datasets. Treating this as final-completion masking would be incorrect: a conversation can contain multiple assistant turns, and every assistant turn must remain supervised while system, user, and tool-response turns remain ignored.

VLM masking also cannot be derived safely by finding response text in a rendered prompt. Repeated user/assistant text, merged tokenizer tokens, role delimiters, padding side, truncation, and processor-side expansion of image/video/audio placeholders can all invalidate substring or token-subsequence matching. The mask therefore has to originate from the active chat-template contract and remain aligned with the final processor output.

## User-visible behavior

| Input/configuration | Behavior |
|---|---|
| Conversational VLM rows with `assistant_only_loss=False` | Existing full-sequence VLM label behavior is unchanged. |
| Conversational VLM rows with `assistant_only_loss=True` | Every assistant turn is supervised; non-assistant and ignored structural tokens are `-100`. |
| Existing labels or response masking plus assistant-only loss | Masks intersect; existing `-100` labels are never overwritten with raw token IDs. |
| Materialized or streaming datasets | Both paths call the same assistant-aware VLM collation and produce equivalent labels for the same rows. |
| Evaluation datasets | Evaluation uses the same assistant-mask generation and final-label semantics as training. |
| Prompt/completion VLM rows plus assistant-only loss | Raises an actionable error because the CUDA-parity collator tokenizes prompt and completion separately and cannot preserve one authoritative assistant mask. |
| Pre-rendered or pretokenized VLM rows plus assistant-only loss | Raises an actionable error because the current backend has no token-index-preserving assistant-boundary pathway for these rows. |
| Missing, incomplete, non-binary, or untrustworthy processor assistant masks | Fails closed instead of falling back to full-sequence or final-response-only supervision. |

### 1. Obtain authoritative unexpanded token boundaries

The active processor tokenizer is called with the processor's chat template, `return_assistant_tokens_mask=True`, and fast-tokenizer offset mappings. Source tokenization is deliberately unpadded and untruncated so the returned `input_ids`, `assistant_masks`, and offsets describe one authoritative full render before processor-side multimodal expansion.

Assistant mask values are validated before integer normalization. Row counts and token lengths must match, values must compare exactly as binary `0` or `1`, and every example must contain assistant tokens. This prevents values such as `1.5` or `"1"` from becoming valid through lossy coercion.

### 2. Prove role ownership without content searching

The implementation renders cumulative structured-message prefixes. Each later render must begin with the previous render exactly, allowing every character interval added by one message to be assigned to that message's role. Templates whose earlier history changes when a later message is appended are prefix-unstable and fail with a clear error.

Supported assistant text fields are also wrapped with unique temporary markers in a separate structured render. Removing every marker exactly once must reproduce the authoritative full render. Fast-tokenizer offsets then map message and content character intervals to source-token positions.

Validation requires all of the following:

- every processor-masked token belongs to an assistant message interval;
- no masked token overlaps a system, user, tool-response, or other non-assistant interval;
- every assistant turn contributes masked tokens, preventing final-turn-only templates from passing;
- every rendered assistant text span is fully covered by the assistant mask;
- the processor mask remains binary and length-aligned with source `input_ids`.

This provides multi-turn role provenance without naïve substring matching, token-subsequence matching, model-name checks, or template-specific delimiter assumptions.

### 3. Preserve processor compatibility while failing closed

Assistant schema validation runs before processors that require string-only assistant content collapse list-style content. Both the original and marker-annotated copies are collapsed together, preserving marker alignment while preventing unsupported parts from disappearing before validation.

The supported initial assistant schema is deliberately narrow:

- `role` plus string `content`;
- list content containing strings or `{ "type": "text", "text": <string> }` parts;
- exact `image`, `video`, or `audio` media part types with no keys beyond `type` and the matching media payload key.

Extra assistant fields such as reasoning/tool-call structures, unknown text-bearing media metadata, arbitrary nested structured payloads, and prefix-unstable/exotic Jinja templates fail clearly. This is a compatibility boundary, not a fallback: the implementation never silently broadens supervision when assistant ownership cannot be proven.

## Alignment with final processor inputs

After the normal VLM processor has expanded multimodal placeholders, inserted processor-specific structural tokens, padded, and truncated the batch, the validated source mask is aligned to final processor `input_ids` by token index and identity.

Only configured VLM ignore-token IDs may be inserted, removed, or expanded during alignment. Expanded image/video/audio runs always receive mask value `0`. Any unexplained non-multimodal token change raises instead of guessing an alignment. Left and right truncation follow the tokenizer's `truncation_side`, while the final attention mask determines which padded positions are valid.

The aligned assistant mask is intersected with the batch's existing labels using the existing label dtype and raw-ID safeguards. The normal VLM label-mask pass is then reapplied so attention padding and model-specific special/vision/audio token IDs remain ignored. This preserves explicit labels rather than reconstructing them from `input_ids`.

## Trainer, evaluation, and distributed integration

`assistant_only_loss` now flows through the existing VLM architecture rather than introducing a parallel trainer path:

- `_collate_vlm_batch` builds and applies assistant-aware labels;
- `create_vlm_batches` uses those labels for materialized datasets;
- `iterate_vlm_training_batches` uses the same collation for streaming datasets;
- evaluation batch construction receives the same flag and masking contract;
- response masking, when present, intersects the assistant-aware labels;
- both baseline CE and CCE consume the same final batch labels.

Distributed setup is synchronized before later collectives. Rank-local materialized training or evaluation failures are converted into a shared failure outcome, and streaming checkpoint fast-forward errors are synchronized before training resumes. This prevents one rank from raising on an invalid assistant mask or exhausted resume stream while peer ranks continue into a collective.

Synthetic DDP padding rows remain fully `-100` and cannot create supervised padding. The VLM all-masked preflight checks `labels[:, 1:]`, matching the causal shift consumed by the loss, so a token present only at position zero does not incorrectly classify a row as trainable.

## Files and review surface

- `unsloth_zoo/mlx/utils.py`: assistant-mask extraction and provenance validation, final processor-ID alignment, label intersection, and materialized/streaming VLM collation plumbing.
- `unsloth_zoo/mlx/trainer.py`: training/evaluation routing, synchronized setup/resume failures, and causal-shift-aware all-masked validation.
- `tests/test_mlx_vlm_label_masks.py`: focused mask contracts and alignment regressions.
- `tests/test_mlx_trainer_internals.py`: trainer/evaluation routing and synchronized setup behavior.
- `tests/test_mlx_ddp_metal.py`: two-rank resume failure and synthetic-padding safety.
- `tests/test_mlx_pr684_full_validation_metal.py`: real Qwen2-VL assistant-only LoRA and checkpoint-resume coverage.

The main reviewer-sensitive areas are the trust boundary around the processor-provided assistant mask, the source-to-final token alignment rules, and the deliberately fail-closed template/schema contract. `assistant_only_loss=False` stays on the existing label path.

## Validation coverage

Focused and integration tests cover:

- one user/assistant turn with an image;
- multiple assistant turns, proving that all assistant responses are supervised;
- masked system, user, and tool-response spans;
- duplicate user/assistant text, preventing substring-based false matches;
- left and right padding plus truncation at assistant boundaries;
- expanded vision-token runs remaining `-100`;
- intersection with existing labels and response masks;
- missing, partial, final-turn-only, non-assistant, and non-binary generation masks;
- unsupported assistant fields and media metadata in both list-content and string-only processor modes;
- materialized/streaming and train/evaluation parity;
- DDP synthetic padding, setup failures, resume fast-forward failures, and all-masked safety;
- baseline CE/CCE effective-label parity;
- `assistant_only_loss=False` regression behavior;
- real Qwen2-VL assistant-only LoRA training and VLM checkpoint resume on Metal.

```bash
python -m pytest \
  tests/test_mlx_vlm_label_masks.py \
  tests/test_mlx_trainer_internals.py \
  tests/test_mlx_ddp_metal.py \
  tests/test_mlx_pr684_full_validation_metal.py -q
```

Result after rebasing onto the current `main`: **155 passed**.

The branch contains 967 added lines across six files, of which 321 are test lines (33.20%).

## Intentionally deferred

This PR does not add generic structured-assistant provenance, arbitrary Jinja dependency analysis, packing, generic pretokenized VLM support, public dispatcher/UI argument plumbing, or audio-training support. Those concerns can be addressed independently without weakening the fail-closed correctness contract introduced here.

### DIFF
--- a/tests/test_mlx_ddp_metal.py
+++ b/tests/test_mlx_ddp_metal.py
@@ -87,6 +87,13 @@ class RankFailingStream:
             raise KeyError("rank0 stream failure")
         return ({"text": f"{i} {i + 20} {i + 30}"} for i in range(10, 15))
 
+class RankFailingResumeStream:
+    def __iter__(self):
+        for i in range(10, 15):
+            if int(world.rank()) == 0:
+                raise KeyError("rank0 resume stream failure")
+            yield {"text": f"{i} {i + 20} {i + 30}"}
+
 class ReplayableVLMStream:
     def __iter__(self): return ({"text": str(i)} for i in range(10, 15))
 
@@ -282,6 +289,17 @@ mx.eval(sync)
 resumed_trainer = make_parity_trainer("resume_resumed", seed=987)
 resumed_trainer.train(resume_from_checkpoint=str(Path(sys.argv[1], "resume_partial", "checkpoint-2")))
 resume_param_delta = max_trainable_delta(fresh_trainer, resumed_trainer)
+def asymmetric_resume_data_error_message():
+    mx.random.seed(795)
+    error_args = MLXTrainingConfig(per_device_train_batch_size=1, gradient_accumulation_steps=1, max_steps=3, logging_steps=1, eval_steps=0, save_steps=0, learning_rate=1e-3, max_seq_length=8, output_dir=str(Path(sys.argv[1], "asymmetric_resume_data_error")), use_cce=False, compile=False, gradient_checkpointing=False, cast_norm_output_to_input_dtype=False, dataset_order="sequential", streaming=True)
+    error_trainer = MLXTrainer(TinyLM(), TinyTokenizer(), RankFailingResumeStream(), args=error_args)
+    error_trainer.save_model = mark("asymmetric_resume_data_error_final")
+    try:
+        error_trainer.train(resume_from_checkpoint=str(Path(sys.argv[1], "resume_partial", "checkpoint-2")))
+    except RuntimeError as exc:
+        return str(exc)
+    return ""
+asymmetric_resume_data_error = asymmetric_resume_data_error_message()
 trainer_mod.save_trainable_adapters = mark("vlm_ckpt_adapter")
 trainer_mod.save_optimizer_state = mark("vlm_ckpt_opt")
 trainer_mod.save_trainer_state = mark("vlm_ckpt_state")
@@ -361,6 +379,7 @@ payload = {
     "fresh_losses": [float(x) for x in fresh_trainer._train_loss_history],
     "resumed_losses": [float(x) for x in resumed_trainer._train_loss_history],
     "resume_param_delta": resume_param_delta,
+    "asymmetric_resume_data_error": asymmetric_resume_data_error,
     "vlm_events": vlm_events,
     "vlm_step": int(vlm_trainer._global_step),
     "vlm_compile_enabled": bool(vlm_result["compile_enabled"]),
@@ -386,14 +405,13 @@ payload = {
 Path(sys.argv[1], f"rank{world.rank()}.json").write_text(json.dumps(payload))
 """,
         encoding="utf-8")
-
     env = os.environ.copy()
     repo_root = Path(__file__).resolve().parents[1]
     env["PYTHONPATH"] = str(repo_root) + os.pathsep + env.get("PYTHONPATH", "")
     cmd = [
         str(launcher), "-n", "2", "--backend", "ring",
-        "--python", sys.executable, "--cwd", str(repo_root),
-        str(script), str(tmp_path),
+        "--cwd", str(repo_root),
+        sys.executable, str(script), str(tmp_path),
     ]
     proc = subprocess.run(
         cmd, capture_output=True, env=env, text=True, timeout=120,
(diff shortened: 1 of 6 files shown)


Lyxot:
Closed because currently not supported by CUDA / trl. Leave as future plan

PR #912, opened 2026-07-18
feat(mlx): add KTO preference tuning

BardiaKoopah:
## Summary
 
Adds KTO preference tuning to the MLX path on Apple Silicon, mirroring TRL's `KTOTrainer` (`loss_type="kto"`, Eqn 7 of [arXiv:2402.01306](https://arxiv.org/abs/2402.01306)).
 
The selling point is the data. KTO trains on **unpaired** examples — a single completion per row with a binary desirable/undesirable label — instead of the chosen/rejected pairs DPO and ORPO require. Binary-labelled completions are far cheaper to collect than matched preference pairs (any "was this output good or bad?" signal works: thumbs up/down, task success/failure, filters), which makes KTO the practical choice when you don't have paired data. This is the first unpaired preference method on the MLX trainer.
 
## What's included
 
- **`MLXKTOTrainer` / `MLXKTOConfig`** — standalone trainer + config mirroring TRL's API, reusing `MLXTrainer`'s optimizer / schedule / gradient-clip / weight-decay helpers. The SFT compile/accumulate loop is left completely untouched.
- **Unpaired batch builder** with the mismatched-pair KL variant (completions rolled +1 within the batch, matching TRL's `_get_kl_dataset`), the summed-logp extractor (`_kto_sum_logp`), the clamped batch KL baseline (`_kto_kl_baseline`), and the KTO loss (`_kto_loss`).
- **Reference model = the LoRA-disabled forward** (adapter scale=0), so no second model copy. Reference and KL forwards are `stop_gradient`'d; adapter scales are restored in a `finally` so a throwing forward can't leave the model adapter-disabled.
- **Fail-loud guards:** `per_device_train_batch_size < 2` raises (the mismatched-pair roll would self-pair and collapse the KL term to the reward); a non-PEFT model raises (no reference forward without adapters).
## Correctness evidence
 
`_kto_loss` matches both numpy and a torch reimplementation of TRL's `kto_loss` to ~1e-8, across three weight settings:
 
| (desirable_weight, undesirable_weight) | loss |
|---|---|
| (1.0, 1.0) | 0.47720963 |
| (1.33, 1.0) | 0.56631130 |
| (1.0, 1.5) | 0.58081198 |
 
- KL baseline matches TRL exactly (**0.31288**), including the `clamp(min=0)` branch (raw negative estimate → 0.0).
- `_kto_sum_logp` vs numpy log-softmax reference: max|Δ| = **7e-07**.
- Real Qwen2.5-0.5B 4-bit + LoRA: deterministic across reruns; 336 LoRA leaves all finite and nonzero; trains **0.4997 → 0.2146** over 8 steps.
## Tests
 
- **`tests/test_mlx_kto_loss.py`** — pure-logic, runs under the torch shim in normal CI (no Metal needed): loss vs numpy and vs the pinned TRL values, the KL clamp branch, single-label (all-desirable / all-undesirable) batches, and the `batch_size < 2` guard. 5 passed.
- **`tests/test_mlx_kto_train_metal.py`** — Metal-gated e2e (skips cleanly off Apple Silicon): finite & decreasing loss, LoRA-scale restoration including the scales-restored-when-a-forward-throws `finally` path, the non-PEFT fail-loud error, and `_kto_sum_logp` vs numpy. 5 passed.
- Canonical MLX baseline: **128 passed, 0 failures**.
<sub>(The `_kto_sum_logp` numeric check lives in the Metal file rather than the shim file — its `.astype(mx.float32)` resolves a real mlx dtype under the torch shim's pytest import order; it runs against real MLX instead. A code comment explains this in place.)</sub>
 
## Limitations (please read — not buried)
 
- **LoRA-only.** The reference is the adapter-off forward, so full-finetune KTO is not supported and raises a clear error rather than silently running reference-free.
- **Not `mx.compile`'d.** The KTO step runs eager with 4 forwards/step (policy, policy-KL, reference, reference-KL). Correct and verified, but slower per step than the compiled SFT path. The blocker is structural: the LoRA-disable reference mutates module scale state, which under `mx.compile` would execute at trace time only and freeze the traced scale. Threading scale as a compiled input is a plausible follow-up, but it needs its own benchmark — compiling 1 of 4 forwards may not pay for the added risk.
- **Within-batch KL roll vs TRL's global roll — affects reproducibility, not the optimization.** TRL builds the KL set by rolling the whole dataset; this rolls completions +1 within each batch, so boundary examples pair with a different mismatched completion than a specific TRL run would. This changes the exact per-batch KL scalar (and thus bit-exact reproducibility against a given TRL run), but not the training method: the KL baseline is a detached, clamped, batch-level scalar (no gradient flows through it — it only shifts the loss threshold), and both rolls estimate the same expectation over the batch's completions. The within-batch variant in fact satisfies TRL's own documented "best results" condition (the mismatched y′ set equals the reward y set for that batch) more cleanly. This is a mechanism-based claim from the loss structure; I have **not** run a head-to-head TRL-vs-MLX training comparison to measure end-task parity.
- **`kl = 0` is expected often.** The raw mismatched-pair estimate is frequently negative early in training and gets clamped to 0. This is TRL-faithful, not a bug (verified: the LoRA is active in the KL forward; the raw estimate was e.g. −4.82 → clamped).

### DIFF
--- a/tests/test_mlx_kto_loss.py
+++ b/tests/test_mlx_kto_loss.py
@@ -1,0 +1,140 @@
+# Unsloth Zoo - Utilities for Unsloth
+# Copyright 2023-present Daniel Han-Chen, Michael Han-Chen & the Unsloth team. All rights reserved.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+"""Pure-logic regression for the MLX KTO loss primitives.
+
+Covers the unpaired KTO loss (`_kto_loss`), the batch KL baseline clamp
+(`_kto_kl_baseline`), the summed-logp extractor (`_kto_sum_logp`), and the
+batch-size guard in `_build_kto_batches`. Runs under the torch shim so Linux
+CI collection covers it without MLX/Metal (all four primitives are backend-
+agnostic MLX-array math).
+
+Reference values are the ones proven in the KTO Step 1 investigation, where
+`_kto_loss` matched a torch reimplementation of TRL's kto_loss and a pure-numpy
+hand computation to ~1e-8 across desirable/undesirable weight settings.
+"""
+
+from __future__ import annotations
+
+import numpy as np
+import pytest
+
+
+@pytest.fixture(autouse=True, scope="module")
+def _install_shim():
+    from mlx_simulation import simulate_mlx_on_torch
+    simulate_mlx_on_torch()
+
+
+# --------------------------------------------------------------------------
+# Fixed inputs + independent numpy reference (the KTO Step 1 fixtures).
+# --------------------------------------------------------------------------
+def _fixed_logps():
+    rng = np.random.default_rng(0)
+    return dict(
+        pol_ch=rng.normal(-8, 2, size=3).astype(np.float32),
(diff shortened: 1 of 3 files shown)


BardiaKoopah:
@codex review

BardiaKoopah:
@codex review

PR #920, opened 2026-07-20
AMD aiter Flash Attention as a transformers attention backend

zhihuidu-amd:
## Summary

Reopening with proper implementation after closing #916.

The key insight: unsloth-zoo compiles attention by **rewriting source code as strings**
(`inspect.getsource()` + regex transforms in `compiler.py`). The correct integration
point is to add a new source-level transform in that same pipeline — not to hook
wrappers around module forwards.

## How it works

`replace_sdpa_with_amd_aiter(source)` is a new source-rewriting function that
runs in the existing `scaled_dot_product_attention_modules` processing loop in
`unsloth_compile_transformers()`.

It matches `scaled_dot_product_attention(query, key, value, ...)` calls and
rewrites them as:

```python
# AMD aiter Flash Attention (MFMA + fast softmax + causal tile skip)
_aiter_q = query_states.transpose(1, 2)  # (B,H,S,D) -> (B,S,H,D)
_aiter_k = key_states.transpose(1, 2)
_aiter_v = value_states.transpose(1, 2)
_aiter_fn = get_amd_flash_attn_func()
if _aiter_fn is not None:
    attn_output = _aiter_fn(_aiter_q, _aiter_k, _aiter_v, causal=is_causal).transpose(1, 2)
else:
    attn_output = torch.nn.functional.scaled_dot_product_attention(
        query_states, key_states, value_states, ...)
```

The transpose converts transformers convention `(B, H, S, D)` to aiter's
`(B, S, H, D)`. Falls back to standard SDPA if aiter is unavailable at runtime.

### `unsloth_zoo/device_type.py`
- `get_amd_attention_implementation()`: detects amd-aiter with ROCm ≥7.0 gate,
  uses `import importlib.util` explicitly (fixes Python 3.13+ issue from #916)
- `get_amd_flash_attn_func()`: returns `flash_attn_func` or `FlashAttnFunc` (both naming conventions)

### `unsloth_zoo/compiler.py`
- `replace_sdpa_with_amd_aiter(source)`: new source-rewriting function, hooked
  into the existing SDPA processing loop. No-op on NVIDIA or ROCm < 7.0.

## Expected Performance (MI300X/MI325X, ROCm ≥7.0)

| Implementation | TFLOPS (prefill seq=2048) | vs SDPA |
|---|---|---|
| amd-aiter MFMA | 51.5 | +5.9× |
| amd-aiter + causal tile skip | 71.3 | +8.2× |
| PyTorch SDPA (current) | ~5.7 | baseline |

## Testing

- **NVIDIA**: no change — `is_hip()` returns False, all code gated
- **AMD ROCm < 7.0**: no change — `get_amd_attention_implementation()` returns `"sdpa"`
- **AMD ROCm ≥7.0 without amd-aiter**: no change — falls back to SDPA in rewritten source
- **AMD ROCm ≥7.0 with amd-aiter**: aiter kernel activated at compile time

Cannot benchmark live (clusters run ROCm 6.2); numbers from AMD's validated
benchmarks on MI300X gfx942.

## Install
```bash
pip install amd-aiter  # https://github.com/ROCm/aiter — ROCm >= 7.0
```

Closes (reopens) #916.

### DIFF
--- a/tests/test_temporary_patches_imports.py
+++ b/tests/test_temporary_patches_imports.py
@@ -50,6 +50,7 @@ TEMPORARY_PATCHES_SUBMODULES = [
     "unsloth_zoo.temporary_patches.gpt_oss",
     "unsloth_zoo.temporary_patches.lfm2_moe",
     "unsloth_zoo.temporary_patches.ministral",
+    "unsloth_zoo.temporary_patches.amd_aiter",
     "unsloth_zoo.temporary_patches.misc",
     "unsloth_zoo.temporary_patches.mixtral_moe",
     "unsloth_zoo.temporary_patches.moe_bnb",
--- a/unsloth_zoo/compiler.py
+++ b/unsloth_zoo/compiler.py
@@ -373,6 +373,95 @@ def replace_with_grouped_query_attention(module, source):
 pass
 
 
+
+
+def replace_sdpa_with_amd_aiter(source):
+    """
+    For AMD ROCm with amd-aiter installed: replace scaled_dot_product_attention
+    calls with amd-aiter's Flash Attention implementation in the compiled source.
+
+    amd-aiter uses full MFMA on both QK^T and P@V matrix multiplications,
+    plus a fast exp2-based softmax and causal tile skip — giving ~5-9x higher
+    TFLOPS than PyTorch SDPA for prefill workloads.
+
+    This function rewrites the source string so that the compiled model calls
+    aiter.flash_attn_func instead of torch.nn.functional.scaled_dot_product_attention.
+
+    Input/output conventions:
+    - transformers attention: q/k/v shape = (B, n_heads, S, D)  [B,H,S,D]
+    - aiter flash_attn_func:  q/k/v shape = (B, S, n_heads, D)  [B,S,H,D]
+    - So we must transpose 1↔2 before and after the aiter call.
+
+    Only activates on AMD ROCm when get_amd_attention_implementation() == "amd_aiter".
+    No-op on NVIDIA and on ROCm without amd-aiter or ROCm < 7.0.
+    """
+    if get_amd_attention_implementation() != "amd_aiter":
+        return source
+
+    # Pattern: capture the SDPA call and all its arguments
+    # Matches: attn_output = torch.nn.functional.scaled_dot_product_attention(
+    #              query_states, key_states, value_states,
+    #              attn_mask=..., dropout_p=..., is_causal=...
+    #          )
+    sdpa_call_pattern = (
+        r"([ 	]*)([A-Za-z_][A-Za-z0-9_]*)[ 	]*=[ 	]*"
+        r"(?:[A-Za-z_.]*scaled_dot_product_attention)"
+        r"\("
+        r"[ 	
+]*([A-Za-z_][A-Za-z0-9_]*)[ 	]*,"   # query
+        r"[ 	
+]*([A-Za-z_][A-Za-z0-9_]*)[ 	]*,"   # key
+        r"[ 	
+]*([A-Za-z_][A-Za-z0-9_]*)[ 	]*"    # value
+        r"([^)]*)\)"                                    # remaining args
+    )
+
+    import re
+
+    def aiter_replacement(m):
+        indent = m.group(1)
+        out_var = m.group(2)
+        q_var = m.group(3)
+        k_var = m.group(4)
+        v_var = m.group(5)
+        rest_args = m.group(6)
+
+        # Extract is_causal from rest_args (default False if not found)
+        causal_match = re.search(r"is_causal\s*=\s*(True|False|[A-Za-z_][A-Za-z0-9_]*)", rest_args)
+        is_causal = causal_match.group(1) if causal_match else "False"
+
+        # Build aiter call:
+        # aiter expects (B, S, H, D) but transformers uses (B, H, S, D) -> transpose 1,2
+        return (
+            f"{indent}# AMD aiter Flash Attention (MFMA + fast softmax + causal tile skip)
+"
+            f"{indent}# Requires: pip install amd-aiter  (ROCm >= 7.0)
+"
+            f"{indent}_aiter_q = {q_var}.transpose(1, 2)  # (B,H,S,D) -> (B,S,H,D)
+"
+            f"{indent}_aiter_k = {k_var}.transpose(1, 2)
+"
+            f"{indent}_aiter_v = {v_var}.transpose(1, 2)
+"
+            f"{indent}_aiter_fn = get_amd_flash_attn_func()
+"
+            f"{indent}if _aiter_fn is not None:
+"
+            f"{indent}    {out_var} = _aiter_fn(_aiter_q, _aiter_k, _aiter_v, causal={is_causal}).transpose(1, 2)
+"
+            f"{indent}else:
+"
+            f"{indent}    {out_var} = torch.nn.functional.scaled_dot_product_attention(
+"
+            f"{indent}        {q_var}, {k_var}, {v_var}{rest_args})"
+        )
+
+    new_source, n = re.subn(sdpa_call_pattern, aiter_replacement, source, flags=re.DOTALL)
+    if n > 0:
+        pass  # Successfully replaced SDPA with aiter
+    return new_source
+
+
 def _get_compile_folder(use_tempfile=False):
     global UNSLOTH_COMPILE_LOCATION
     global UNSLOTH_COMPILE_USE_TEMP
@@ -3809,6 +3898,8 @@ def unsloth_compile_transformers(
                 disabled_scaled_dot_product_attention_modules.append(module)
             pass
         pass
+        # AMD ROCm: replace SDPA with amd-aiter Flash Attention if available
+        new_source = replace_sdpa_with_amd_aiter(new_source)
         scaled_dot_product_attention_modules[module] = new_source
     pass
 
--- a/unsloth_zoo/compiler.py
+++ b/unsloth_zoo/compiler.py
@@ -378,22 +378,31 @@ pass
 def replace_sdpa_with_amd_aiter(source):
     """
     For AMD ROCm with amd-aiter installed: replace scaled_dot_product_attention
-    calls with amd-aiter's Flash Attention implementation in the compiled source.
+    calls with amd-aiter Flash Attention in the compiled source.
 
-    Only activates on AMD ROCm when get_amd_attention_implementation() == "amd_aiter".
+    Safety requirements (all must hold for the aiter path to activate):
+      1. get_amd_attention_implementation() == "amd_aiter"
+      2. is_causal=True as a literal (not a complex expression)
+      3. No extra SDPA args that aiter does not support:
+         attn_mask, dropout_p, scale, enable_gqa
+      4. q/k seq lengths equal (non-square causal has different bias alignment)
+      5. Input dtype is float16 or bfloat16 (aiter does not support float32)
+
     No-op on NVIDIA and on ROCm without amd-aiter or ROCm < 7.0.
     """
-    # Fix 1: import inside function to avoid compile-time NameError in caller scope
+    # Import inside function — avoids compile-time NameError in caller scope
     from unsloth_zoo.device_type import get_amd_attention_implementation
     if get_amd_attention_implementation() != "amd_aiter":
         return source
 
     import re
 
-    # Fix 3: use ((?:[^)]|\([^)]*\))*) to safely handle one level of nested parens
+    # Pattern:  out = <prefix.>scaled_dot_product_attention(q, k, v, ...)
+    # Excludes: disable_compile_scaled_dot_product_attention (opt-out shim)
+    # Uses ((?:[^)]|\([^)]*\))*) to handle one level of nested parens in args
     sdpa_call_pattern = (
         r"([ \t]*)([A-Za-z_][A-Za-z0-9_]*)[ \t]*=[ \t]*"
-        r"(?:[A-Za-z_.]*scaled_dot_product_attention)"
+        r"(?:(?:torch\.nn\.functional|F|nn\.functional)\.)?scaled_dot_product_attention"
         r"\("
         r"[ \t\n]*([A-Za-z_][A-Za-z0-9_]*)[ \t]*,"
         r"[ \t\n]*([A-Za-z_][A-Za-z0-9_]*)[ \t]*,"
@@ -402,42 +411,59 @@ def replace_sdpa_with_amd_aiter(source):
     )
 
     def aiter_replacement(m):
-        indent = m.group(1)
+        indent  = m.group(1)
         out_var = m.group(2)
-        q_var = m.group(3)
-        k_var = m.group(4)
-        v_var = m.group(5)
+        q_var   = m.group(3)
+        k_var   = m.group(4)
+        v_var   = m.group(5)
         rest_args = m.group(6)
 
-        # Extract is_causal value from rest_args
-        causal_match = re.search(r"is_causal\s*=\s*([^,]+)", rest_args)
-        is_causal = causal_match.group(1).strip() if causal_match else "False"
+        # Only rewrite when is_causal=True as a literal — aiter is causal-only;
+        # complex expressions like `is_causal=self.is_causal and mask is None`
+        # cannot be safely evaluated at rewrite time.
+        causal_match = re.search(r"is_causal\s*=\s*(True)\b", rest_args)
+        if causal_match is None:
+            # Not a plain literal True — leave this call unchanged
+            return m.group(0)
 
-        # Fix 3: clean leading newlines from rest_args to avoid blank lines
-        rest_args_clean = rest_args.lstrip("\r\n")
-        rest_args_formatted = (", " + rest_args_clean.strip()) if rest_args_clean.strip() else ""
+        # If any SDPA-only arguments are present, fall back — aiter does not
+        # support attn_mask, dropout_p, scale, or enable_gqa
+        unsupported = ("attn_mask", "dropout_p", "scale", "enable_gqa")
+        if any(kw in rest_args for kw in unsupported):
+            return m.group(0)
 
-        # Fix 4: only route to aiter when is_causal is True (aiter flash_attn is causal-only fast path)
-        # Fix 2: import get_amd_flash_attn_func inside generated code to avoid runtime NameError
-        return (
-            f"{indent}# AMD aiter Flash Attention (MFMA + fast softmax + causal tile skip)\n"
-            f"{indent}# Requires: pip install amd-aiter  (ROCm >= 7.0)\n"
-            f"{indent}from unsloth_zoo.device_type import get_amd_flash_attn_func\n"
-            f"{indent}_aiter_fn = get_amd_flash_attn_func()\n"
-            f"{indent}if _aiter_fn is not None and {is_causal}:\n"
-            f"{indent}    _aiter_q = {q_var}.transpose(1, 2)\n"
-            f"{indent}    _aiter_k = {k_var}.transpose(1, 2)\n"
-            f"{indent}    _aiter_v = {v_var}.transpose(1, 2)\n"
-            f"{indent}    {out_var} = _aiter_fn(_aiter_q, _aiter_k, _aiter_v, causal=True).transpose(1, 2)\n"
-            f"{indent}else:\n"
-            f"{indent}    {out_var} = torch.nn.functional.scaled_dot_product_attention(\n"
-            f"{indent}        {q_var}, {k_var}, {v_var}{rest_args_formatted})"
-        )
+        # Clean leading newlines from rest_args
+        rest_args_clean = rest_args.lstrip("\r\n").strip()
+        # rest_args already starts with a comma when non-empty; do NOT add another
+        rest_args_formatted = (", " + rest_args_clean) if rest_args_clean else ""
+
+        # Generate the replacement block.
+        # All symbols used at runtime are imported inside the generated code so
+        # that the exec()'d module namespace is self-contained (no NameError).
+        lines = [
+            f"{indent}# AMD aiter Flash Attention (MFMA + causal tile skip)",
+            f"{indent}# Requires: pip install amd-aiter  (ROCm >= 7.0)",
+            f"{indent}from unsloth_zoo.device_type import get_amd_flash_attn_func as _get_aiter_fn",
+            f"{indent}_aiter_fn = _get_aiter_fn()",
+            f"{indent}_aiter_ok = (",
+            f"{indent}    _aiter_fn is not None",
+            f"{indent}    and {q_var}.shape[-2] == {k_var}.shape[-2]",  # equal seq lengths
+            f"{indent}    and {q_var}.dtype in (torch.float16, torch.bfloat16)",  # dtype guard
+            f"{indent})",
+            f"{indent}if _aiter_ok:",
+            f"{indent}    _aiter_q = {q_var}.transpose(1, 2)",
+            f"{indent}    _aiter_k = {k_var}.transpose(1, 2)",
+            f"{indent}    _aiter_v = {v_var}.transpose(1, 2)",
+            f"{indent}    {out_var} = _aiter_fn(_aiter_q, _aiter_k, _aiter_v, causal=True).transpose(1, 2)",
+            f"{indent}else:",
+            f"{indent}    {out_var} = torch.nn.functional.scaled_dot_product_attention(",
+            f"{indent}        {q_var}, {k_var}, {v_var}{rest_args_formatted})",
+        ]
+        return "\n".join(lines)
 
     new_source, n = re.subn(sdpa_call_pattern, aiter_replacement, source, flags=re.DOTALL)
-    if n > 0:
-        pass  # Successfully replaced SDPA with aiter
     return new_source
+
 
 
 def _get_compile_folder(use_tempfile=False):
--- a/unsloth_zoo/compiler.py
+++ b/unsloth_zoo/compiler.py
@@ -380,43 +380,26 @@ def replace_sdpa_with_amd_aiter(source):
     For AMD ROCm with amd-aiter installed: replace scaled_dot_product_attention
     calls with amd-aiter's Flash Attention implementation in the compiled source.
 
-    amd-aiter uses full MFMA on both QK^T and P@V matrix multiplications,
-    plus a fast exp2-based softmax and causal tile skip — giving ~5-9x higher
-    TFLOPS than PyTorch SDPA for prefill workloads.
-
-    This function rewrites the source string so that the compiled model calls
-    aiter.flash_attn_func instead of torch.nn.functional.scaled_dot_product_attention.
-
-    Input/output conventions:
-    - transformers attention: q/k/v shape = (B, n_heads, S, D)  [B,H,S,D]
-    - aiter flash_attn_func:  q/k/v shape = (B, S, n_heads, D)  [B,S,H,D]
-    - So we must transpose 1↔2 before and after the aiter call.
-
     Only activates on AMD ROCm when get_amd_attention_implementation() == "amd_aiter".
     No-op on NVIDIA and on ROCm without amd-aiter or ROCm < 7.0.
     """
+    # Fix 1: import inside function to avoid compile-time NameError in caller scope
+    from unsloth_zoo.device_type import get_amd_attention_implementation
     if get_amd_attention_implementation() != "amd_aiter":
         return source
 
-    # Pattern: capture the SDPA call and all its arguments
-    # Matches: attn_output = torch.nn.functional.scaled_dot_product_attention(
-    #              query_states, key_states, value_states,
-    #              attn_mask=..., dropout_p=..., is_causal=...
-    #          )
+    import re
+
+    # Fix 3: use ((?:[^)]|\([^)]*\))*) to safely handle one level of nested parens
     sdpa_call_pattern = (
-        r"([ 	]*)([A-Za-z_][A-Za-z0-9_]*)[ 	]*=[ 	]*"
+        r"([ \t]*)([A-Za-z_][A-Za-z0-9_]*)[ \t]*=[ \t]*"
         r"(?:[A-Za-z_.]*scaled_dot_product_attention)"
         r"\("
-        r"[ 	
-]*([A-Za-z_][A-Za-z0-9_]*)[ 	]*,"   # query
-        r"[ 	
-]*([A-Za-z_][A-Za-z0-9_]*)[ 	]*,"   # key
-        r"[ 	
-]*([A-Za-z_][A-Za-z0-9_]*)[ 	]*"    # value
-        r"([^)]*)\)"                                    # remaining args
+        r"[ \t\n]*([A-Za-z_][A-Za-z0-9_]*)[ \t]*,"
+        r"[ \t\n]*([A-Za-z_][A-Za-z0-9_]*)[ \t]*,"
+        r"[ \t\n]*([A-Za-z_][A-Za-z0-9_]*)[ \t]*"
+        r"((?:[^)]|\([^)]*\))*)\)"
     )
-
-    import re
 
     def aiter_replacement(m):
         indent = m.group(1)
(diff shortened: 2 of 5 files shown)


zhihuidu-amd:
All 4 issues from the Gemini Code Assist review addressed in commit `f64a759`:

**Fix 1 (compile-time NameError)**: Added `from unsloth_zoo.device_type import get_amd_attention_implementation` at the top of `replace_sdpa_with_amd_aiter()`. It was previously called at module scope where the import was absent.

**Fix 2 (runtime NameError)**: Added `from unsloth_zoo.device_type import get_amd_flash_attn_func` directly in the generated code string. The compiled module runs in its own namespace — outer-scope imports are not visible there.

**Fix 3 (regex robustness + blank lines)**: Updated rest_args pattern from `([^)]*)` to `((?:[^)]|\([^)]*\))*)` to safely handle one level of nested parentheses. Added `rest_args.lstrip("\r\n")` to prevent blank lines in generated code.

**Fix 4 (causal fast-path routing)**: Changed `if _aiter_fn is not None:` to `if _aiter_fn is not None and {is_causal}:`. amd-aiter's flash attention assumes causal masking — bidirectional attention (`is_causal=False`, `attn_mask=None`) must not be incorrectly routed to the causal-only fast path.

zhihuidu-amd:
All Codex P1/P2 review issues addressed in commit `30710ab`.

**Summary of fixes:**

**device_type.py**
- **P1 — FlashAttnFunc.apply()**: `FlashAttnFunc` is a class requiring `.apply(q, k, v, causal)`; wrapped in a lambda so all callers always get a plain callable with signature `func(q, k, v, causal=True)`.

**compiler.py**
- **P1 — is_causal literal guard**: Restricted match to `is_causal=True` as a plain literal only. Complex expressions like `is_causal=self.is_causal and mask is None` cannot be safely evaluated at rewrite time — those calls are left unchanged (fall through to SDPA).
- **P1 — Unsupported SDPA args**: If `attn_mask`, `dropout_p`, `scale`, or `enable_gqa` are present in the SDPA call, the call is left unchanged. aiter does not support these arguments.
- **P1 — float32 dtype guard**: Added `q.dtype in (torch.float16, torch.bfloat16)` check. aiter only supports fp16/bf16; float32 calls fall through to SDPA.
- **P1 — Non-square seq length guard**: Added `q.shape[-2] == k.shape[-2]` check. Only dispatch to aiter when Q and K seq lengths match (chunked/cached decode safety).
- **P2 — Pattern excludes disable_compile shim**: Tightened regex to explicit prefixes only (`torch.nn.functional.`, `F.`, `nn.functional.`) — `disable_compile_scaled_dot_product_attention` no longer matches.
- **P2 — Double comma in fallback**: Removed extra `", "` prepend; `rest_args` already leads with a comma when non-empty.
- **Renamed import alias**: `get_amd_flash_attn_func` → `_get_aiter_fn` to avoid shadowing `_aiter_fn`.

All five fixes from `f64a759` (Gemini review) are preserved. The aiter fast path now has a documented correctness domain in the function docstring.

zhihuidu-amd:
P1 fixed in commit `6c7431d`.

Checked the actual `FlashAttnFunc.apply()` signature in [ROCm/aiter:aiter/ops/mha.py](https://github.com/ROCm/aiter/blob/main/aiter/ops/mha.py) — it requires 13+ positional arguments: `q, k, v, dropout_p, softmax_scale, causal, window_size, bias, alibi_slopes, deterministic, return_lse, return_softmax, is_grad_enabled, ...`. The previous lambda passed only `(q, k, v, causal)`, sending `causal` as `dropout_p` and leaving 9 required args missing.

**Fix**: Removed the `FlashAttnFunc` wrapper entirely. `get_amd_flash_attn_func()` now returns `None` when only `FlashAttnFunc` is available (no `flash_attn_func`), causing callers to fall back to SDPA transparently.

**Also**: Tightened `get_amd_attention_implementation()` to only return `'amd_aiter'` when `flash_attn_func` (functional API) is present. Previously it also accepted `FlashAttnFunc`, which would detect `'amd_aiter'` but immediately return `None` from `get_amd_flash_attn_func()`. `aiter >= 0.7` exposes `flash_attn_func`; older class-only environments now correctly fall back to SDPA.

LeoBorcherding:
thanks for this, the AMD attention gap is worth closing. i set the branch up locally and
tested it. a few blocking issues before the aiter path can work.

tested on windows with torch 2.11.0+rocm7.13 on gfx1200 (RX 9060 XT), and WSL ubuntu-24.04
with ROCm 7.2.1 + torch 2.13.0+rocm7.2 on the same card.

one thing that's right and easy to miss: the `q.shape[-2] == k.shape[-2]` guard. aiter's
`flash_attn_func` aligns causal masks bottom-right, torch SDPA aligns top-left, and those
only agree at equal seq lengths. that guard is what keeps this from being a silent
wrong-numbers bug.

### 1. generated code is a syntax error (blocking)

`compiler.py:438`:

```python
rest_args_formatted = (", " + rest_args_clean) if rest_args_clean else ""
```

the comment above it is correct, `rest_args` already starts with a comma when non-empty,
but the code adds another. forcing the aiter path on and rewriting
`out = torch.nn.functional.scaled_dot_product_attention(q, k, v, is_causal=True)` gives:

```python
    else:
        out = torch.nn.functional.scaled_dot_product_attention(
            q, k, v, , is_causal=True)
```

`ast.parse` rejects that, so any rewritten module fails to exec on exactly the hardware
this targets. dropping the `", " +` fixes it.

### 2. the rewrite can never fire where it's called (blocking)

`compiler.py:3906` calls `replace_sdpa_with_amd_aiter(new_source)` after the SDPA handling
above it, and neither branch feeding it can pass the new guards:

- `if` branch keeps the original `attn_mask=` and `dropout_p=`, both in your `unsupported`
  reject list, and leaves `is_causal=is_causal` as a variable rather than the literal
  `True` you require.
- `else` branch renames the call to `disable_compile_scaled_dot_product_attention`, which
  your pattern excludes on purpose.

with the aiter path forced on, running all 32 `transformers/models/**/modeling_*.py` files
that call SDPA through `replace_sdpa_with_amd_aiter` gives 0 rewrites. same for hand-built
samples of both branches.

the `if` branch also has a pre-existing `new_source = source` that throws away its own
substitution. not from this PR, but it's why `attn_mask` survives. fixing it upstream
wouldn't help here anyway, since the guard rejects on the `attn_mask` substring either way.

this is also why issue 1 hasn't shown up yet, the feature is currently inert.

### 3. no GPU arch gate

`device_type.py:220` gates on ROCm >= 7.0 and `hasattr(aiter, "flash_attn_func")`. that
second check tells you nothing about the hardware. in amd-aiter 0.1.7.post2.dev18,
`aiter/__init__.py:75` is an unconditional `from .ops.mha import *` and `flash_attn_func`
is a plain module-level def, so the attribute is there on every arch where aiter imports
at all. every real arch check inside `mha.py` is `get_gfx() == "gfx942"` or `"gfx950"`.

aiter's own support table lists RDNA4 only as gfx1201 (Radeon AI PRO R9700, experimental)
and notes most CK and ASM kernels are CDNA-only. gfx1200 isn't listed. as written, any
consumer RDNA box with ROCm 7 and aiter installed gets routed into the aiter path. gating
on `get_gfx()` against a known-good list would be safer than checking for an attribute.

### 4. no fallback if the aiter call raises

generated block at `compiler.py:453-460`:

```python
if _aiter_ok:
    out = _aiter_fn(q_t, k_t, v_t, causal=True).transpose(1, 2)
else:
    out = torch.nn.functional.scaled_dot_product_attention(...)
```

every guard is a pre-call predicate and the fallback only triggers on `_aiter_fn is None`.
an unsupported head dim, unsupported arch, or a JIT build failure inside `flash_attn_func`
propagates out of the model forward, so it crashes mid-training instead of degrading.
wrapping the aiter call in try/except with SDPA as the handler would make issue 3 fail soft.

happy to re-test once these are sorted, the isolated env is still up on my end.

zhihuidu-amd:
All 4 issues fixed in commit `8494840`. Thanks for the thorough hardware testing on both gfx1200 and gfx942 — this was exactly what the PR needed.

**Fix 1 — Double comma syntax error (blocking):**
`rest_args_formatted = rest_args_clean` — the leading comma in `rest_args_clean` is preserved as-is. The old `", " + rest_args_clean` produced `q, k, v, , is_causal=True` which `ast.parse` (and `exec()`) correctly rejects.

**Fix 2 — Fail-soft fallback on aiter call:**
Wrapped `_aiter_fn(...)` in `try/except Exception` with SDPA as the except handler. Unsupported head dim, arch mismatch, or JIT build failure now degrades to SDPA instead of propagating out of the model forward.

**Fix 3 — GPU arch gate:**
Added `_detect_gfx_arch()` (rocminfo → hipconfig fallback) and gated `get_amd_attention_implementation()` on `_AITER_SUPPORTED_ARCHS = ("gfx942", "gfx950")`. Any other arch (gfx1200, gfx1100, etc.) returns `"sdpa"`. Unknown arch (detection failure) also falls back to `"sdpa"`.

**Fix 4 — Call site limitation documented:**
Added a comment at the call site in `compiler.py` explaining that the existing Unsloth SDPA shim emits `attn_mask=` or `is_causal=is_causal` (a variable) — both correctly rejected by the guards — so the rewrite currently applies only to user-code SDPA calls with bare `is_causal=True` literals. Pointed to the path for a follow-up fix.

The `q.shape[-2] == k.shape[-2]` guard is preserved — you're right that it's easy to miss and critical for causal mask alignment correctness.

zhihuidu-amd:
Fixed in commit `3605a61`.

Replaced the two-step keyword-name guard with a single `re.fullmatch` check:
```python
rest_args_stripped = rest_args.strip().lstrip(",").strip()
if not re.fullmatch(r"is_causal\s*=\s*True", rest_args_stripped):
    return m.group(0)
```

Only rewrites when `rest_args` is *exactly* `, is_causal=True` (whitespace-tolerant). Any other content — positional or keyword — leaves the call unchanged. This correctly rejects `, None, 0.1, is_causal=True` (positional dropout passes through to SDPA which handles it).

Verified against 8 test cases including positional-dropout form, keyword attn_mask, keyword dropout_p, empty args, is_causal=False, and keyword scale — all correct.

zhihuidu-amd:
All 3 issues addressed in commit `5a33eaf`.

**Fix 1 — ROCm version gate uses `torch.version.hip` first**: `_detect_rocm_major_minor()` can fall through to `rocm-smi --showdriverversion` which reports the kernel driver version, not the ROCm runtime. Now checks `torch.version.hip` first (always present and reliable in ROCm wheels). Falls back to `_detect_rocm_major_minor()` only when `torch.version.hip` is absent.

**Fix 2 — `_detect_gfx_arch()` PyTorch fallback**: Added `torch.cuda.get_device_properties(0).gcnArchName` as the final fallback before returning `None`. Available in all ROCm wheel installs even when `rocminfo` and `hipconfig` are absent (e.g. runtime-only containers). `gcnArchName` may include suffixes like `:sramecc+:xnack-` — stripped with a `re.match(r'gfx[0-9a-f]+')`.

**Fix 3 — Trailing comma**: Strip both leading AND trailing comma+whitespace before `fullmatch`: `rest_args.strip().lstrip(',').strip().rstrip(',').strip()`. Handles Black/multiline `is_causal=True,` with or without surrounding spaces. Verified on 8 cases.

zhihuidu-amd:
Both issues addressed in commit `e20f0a4`.

**Fix 1 — Inference-only guard**: Added `and not {q_var}.requires_grad` to `_aiter_ok`. The fast path now fires only during inference (eval/no_grad). During training, the else branch handles it with SDPA directly — zero overhead, no exception path. Training always uses SDPA; aiter is used for inference only until a return_lse-aware training form is added in a follow-up.

**Fix 2 — Chained SDPA calls**: Added negative lookahead `(?![ \t]*\.)` at the end of the pattern. `F.scaled_dot_product_attention(q, k, v, is_causal=True).transpose(1, 2)` is now left unchanged — the regex no longer matches when a method chain follows the closing paren, preventing the dangling `.transpose(1, 2)` suffix that caused `exec()` to reject the generated module. Verified: plain call matches; `.transpose()` and `.view()` chains are correctly rejected.

LeoBorcherding:
re-tested on e20f0a43. all four blocking items are fixed, verified by running them rather
than reading the diff.

first a correction to the record: i only tested gfx1200. i don't have an MI300 or MI355
box, so gfx942 and gfx950 are still unvalidated by me, and those two are exactly what the
new allowlist permits.

## verified fixed

- **syntax.** generated code parses for the bare `is_causal=True`, trailing-comma, and
  `F.` alias forms.
- **fail-soft.** try/except now wraps the aiter call with SDPA in the handler.
- **arch gate.** works on real hardware. on this box `_detect_gfx_arch()` returns
  `gfx1200`, `get_amd_attention_implementation()` returns `sdpa`, and
  `get_amd_flash_attn_func()` returns `None`.
- the fullmatch guard correctly rejects positional dropout, keyword `attn_mask`,
  `is_causal=is_causal`, and `.transpose()` chains.

## on the two codex P2s from 5a33eaf

**LSE guard: correct.** i checked this against aiter source and your reasoning holds.
`aiter/ops/mha.py:2502` computes `is_grad = is_grad_enabled and any(x.requires_grad for x
in [q, k, v])`, and line 2536 then does `assert return_lse`. since the generated call
omits `return_lse`, the guard is a real fix, not defensive noise.

one gap: your guard checks only `q.requires_grad`, aiter checks all three of q, k, v. a
LoRA config targeting `k_proj`/`v_proj` but not `q_proj` leaves k and v with grad while q
has none, so the guard passes and the assert fires anyway. try/except catches it so
nothing crashes, but you eat an exception on every forward, which is the double work the
guard exists to avoid. checking all three closes it.

**chained calls: partially.** the `(?![ \t]*\.)` lookahead does reject dot chains. i
confirmed `.transpose(1,2)`, `.to(...)`, and the multiline `\n).transpose(1,2)` form are
all left unchanged. but codex asked for end-of-statement, and a negative lookahead on `.`
is narrower than that, so non-dot suffixes still match:

    out = F.scaled_dot_product_attention(q, k, v, is_causal=True) * scale

rewrites to (trimmed):

    try:
        out = _aiter_fn(_aiter_q, _aiter_k, _aiter_v, causal=True).transpose(1, 2)
    ...
    else:
        out = torch.nn.functional.scaled_dot_product_attention(
            q, k, v, is_causal=True) * scale

the `* scale` lands on the fallback branch only. this one parses, so it isn't the exec
failure codex was describing, it's worse in a quieter way: the two branches compute
different numbers and nothing catches it. `[0]` subscripts behave the same.

`(?![ \t]*[^\s#])` in place of the dot lookahead covers the whole class. i checked it
against the current pattern: clean calls, trailing-comment calls, and multiline calls
still match, while `.transpose()`, `* scale`, and `[0]` are all rejected.

## issue 2 is still open

still 0 rewrites across all 32 `transformers/models/**/modeling_*.py` files that call
SDPA, and both pipeline branches still don't fire. documenting the limitation in a comment
doesn't change that the rewrite can't reach the source it sits next to. (`acface7` reads
like it addresses this, but it's on the #943/#945 series, not on this branch.)

## the inference-only guard narrows it further

`not q.requires_grad` is right for the assert, but it means the fast path is off during
training. stacked with issue 2, the full activation set is now: gfx942 or gfx950, plus
aiter installed, plus a user-code SDPA call with a bare `is_causal=True`, plus inference.
nothing in unsloth's own compiled paths, and nothing during training at all.

that's the thing worth settling before merge. this is ~250 lines that provably never
execute in this library's paths today. either the call site gets wired to handle the shim
form, or it may be worth holding this until that part is ready, rather than landing
machinery that can only be exercised by hand.

## smaller

- `rest_args_clean` is assigned twice identically, looks like a leftover from the
  double-comma fix.
- the docstring still lists the old rule about rejecting `attn_mask`/`dropout_p`/`scale`/
  `enable_gqa`, and doesn't mention the arch gate or the inference-only guard.

codex's last pass was on 5a33eaf, so e20f0a43 hasn't been reviewed by anything yet, and
that's the commit carrying both changes discussed above. worth re-requesting on the
current head. CI is still queued on my end.

zhihuidu-amd:
All four concrete fixes addressed in commit `e63e24c`.

**Fix 1 — requires_grad covers q, k, v**: Changed to `not (q.requires_grad or k.requires_grad or v.requires_grad)`, matching aiter/ops/mha.py:2502's `any(x.requires_grad for x in [q, k, v])`. A LoRA targeting k_proj/v_proj but not q_proj now correctly falls through to SDPA without hitting the assert.

**Fix 2 — Lookahead broadened**: Changed from `(?![ \t]*\.)` to `(?![ \t]*[^\s#])`. Verified: plain call and trailing-comment call still match; `.transpose()`, `* scale`, and `[0]` subscript all rejected. No more silent divergence between branches.

**Fix 3 — Duplicate rest_args_clean removed**: Cleaned up the leftover from the double-comma fix.

**Fix 4 — Docstring updated**: Removed the stale attn_mask/dropout_p rejection rule (now handled by the fullmatch guard), added the arch gate and inference-only guard, and explicitly documented the known call-site limitation.

On the open question about the call site: you're right that ~250 lines provably don't execute in Unsloth's own compiled paths today. The honest description is that this PR delivers the infrastructure (correct guards, fail-soft, arch detection) for user-code SDPA calls with bare `is_causal=True`, and wiring it to Unsloth's shim form is a separate follow-up. The docstring now says this clearly. Whether that scope is worth merging now or holding for the follow-up is danielhanchen's call.

LeoBorcherding:
verified all four on e63e24c.

- **grad guard** is now `not (q.requires_grad or k.requires_grad or v.requires_grad)`,
  which matches `mha.py:2502` exactly. the k/v-only LoRA case falls through to SDPA
  without hitting the assert.
- **lookahead** `(?![ \t]*[^\s#])` behaves as intended. clean, trailing-comment,
  trailing-comma, and multiline calls still rewrite and all parse. `.transpose()`,
  `* scale`, and `[0]` are all left unchanged. the branch-divergence case is gone.
- **duplicate assignment** removed.
- **docstring** is accurate now, and the "known limitation" paragraph is the right call.
  stating it in the docstring rather than only in a PR comment is what makes it
  maintainable later.

re-ran the earlier checks against the new head for regressions: arch gate still returns
`gfx1200 -> sdpa` on this box, and the transformers sweep is still 32 files / 0 rewrites,
unchanged as expected.

that clears everything i raised. agreed the scope question is @danielhanchen 's call, and
the docstring now states the tradeoff plainly enough for that decision to be made on the
record rather than inferred from the diff.

codex's last pass was on 5a33eaf, so it has now skipped both e20f0a4 and e63e24c. those
two commits carry the grad guard, the broadened lookahead, and the docstring, and codex
caught real issues on every prior commit, so worth a pass on the current head.

LeoBorcherding:
@codex review

zhihuidu-amd:
Both Codex issues addressed in commit `ccddae0`.

**Fix 1 — compile-safe fallback**: Replaced try/except in generated code with a nested `_call_aiter_safe(_fn, _q, _k, _v)` helper that returns `None` on any exception. The generated forward uses `if _aiter_out is not None:` — no exception handling in the compiled region, compatible with `torch.compile(fullgraph=True)`.

**Fix 2 — arch detection uses active device first**: `torch.cuda.get_device_properties(current_device()).gcnArchName` is now the primary probe — it always reflects the device PyTorch is actually using, respects `HIP_VISIBLE_DEVICES`, and works correctly on mixed-arch hosts. rocminfo/hipconfig are demoted to subprocess fallbacks for unusual builds where PyTorch cannot query device properties.

zhihuidu-amd:
Codex came back clean on e63e24c (`Didn't find any major issues`). ccddae0 (compile-safe fallback + arch detection) is one commit ahead of that reviewed head — both changes are straightforward (nested helper replaces try/except, probe order swap) so no new Codex issues are expected, but worth a pass on the current head when convenient.

LeoBorcherding:
Separate thread from correctness, same note I left on #944: comment volume relative to code.

Against the files' own baselines, `compiler.py` is 14% comments on main and this PR's added
lines are 29%; `device_type.py` is 9% on main and the added lines are 19%. Roughly double in
both cases.

The clearest example is compiler.py:448-456, six comment lines wrapped around three lines
of code:

```python
# Clean leading newlines from rest_args
rest_args_clean = rest_args.lstrip("\r\n").strip()
# rest_args already starts with a comma when non-empty; do NOT add another
# rest_args is ", is_causal=True" at this point (fullmatch guarantees this).
# Pass it through to the SDPA fallback as-is.
rest_args_formatted = rest_args.strip().lstrip(",").strip().rstrip(",").strip()
# Reconstruct with leading comma for the SDPA fallback call
rest_args_formatted = (", " + rest_args_formatted) if rest_args_formatted else ""
```

Two real problems hiding in there, not just verbosity. `rest_args_clean` is assigned and
never read, so it's dead (this is the leftover I flagged before, in a new shape). And "do
NOT add another" is contradicted by the line two below it, which adds one. Both are
artifacts of successive review rounds layering comments instead of replacing them. Three of
those six lines can go, along with the dead assignment.

The 26-line docstring on `replace_sdpa_with_amd_aiter` I'd mostly keep, since the known
limitation paragraph was my ask and it belongs in the source. But its numbered activation
requirements duplicate the inline comments on each guard, so one of the two can go. Same
for the generated-code block, where several comments restate the line they precede.

Same as on #944, treat the above as representative and do one general pass over the whole
diff rather than just those lines. Target the files' own density, 14% for `compiler.py` and
9% for `device_type.py`. The three that cover most of it:

- drop any comment that restates the line below it
- drop comments that narrate the review history; several here read as a changelog of
  successive rounds ("do NOT add another", "Fix 3:", "Reconstruct with leading comma") and
  at least one is now false because the code moved underneath it
- keep the ones recording something genuinely non-obvious, and this PR has real ones worth
  protecting: why the q/k length guard exists (SDPA top-left vs aiter bottom-right causal
  alignment), why `return_lse` forces the inference-only guard, and why the arch allowlist
  is CDNA-only

Those three are the reason someone won't undo your guards in six months. Most of the rest
can go.

Also: ccddae04 landed after my last pass, so I'll re-verify against it separately.

zhihuidu-amd:
Done in commit `add32da`.

**compiler.py**: removed dead `rest_args_clean` assignment, removed the contradictory 'do NOT add another' comment, trimmed the 9-line pre-fullmatch block to 3 lines, removed self-evident 'leave call unchanged' and 'Generate the replacement block' comments, shortened the `_call_aiter_safe` comment to 2 lines.

**device_type.py**: shortened subprocess fallback comment (3→1 lines), shortened CDNA-only arch gate comment (4→2 lines, kept the non-obvious fact), removed inline 'fail safe: unknown arch' comment.

Kept the three non-obvious ones: q/k seq length guard (SDPA top-left vs aiter bottom-right causal alignment), inference-only guard (return_lse assertion in mha.py:2502/2536), and CDNA-only arch allowlist.

zhihuidu-amd:
Fixed in commit `9000c85`.

Added negative lookbehind `(?<![.\\w])` before the LHS variable group. `self.attn_output = F.scaled_dot_product_attention(...)` and `module.out = F.scaled_dot_product_attention(...)` are now rejected — the pattern requires that no dot or word character precedes the variable name. Plain local variable assignments are unchanged.

Verified on four cases: `out = F.sdpa(...)` matches, `self.out = F.sdpa(...)` and `module.out = F.sdpa(...)` rejected, `attn_out = F.sdpa(...)` matches.

zhihuidu-amd:
Already fixed in commit `9000c85` (pushed before this review arrived — the bot reviewed `add32da`, which is one behind the current head).

Added `(?<![.\\w])` negative lookbehind before the LHS variable group. `self.attn_output = F.scaled_dot_product_attention(...)` and any other attribute assignment are now rejected by the pattern.

zhihuidu-amd:
Already fixed — the bot is reviewing `8494840`, which is five commits behind the current head `9000c85`.

The keyword-name guard (`unsupported = ("attn_mask", "dropout_p", ...)`) was replaced in `3605a61` with a `re.fullmatch` check that only accepts the bare `is_causal=True` form (optional trailing comma/whitespace). Any other content — positional args like `(q, k, v, None, 0.1, is_causal=True)` or additional keywords — fails the fullmatch and leaves the call unchanged. Positional dropout is correctly rejected.

LeoBorcherding:
@codex review

Head is now add32da4. Your last pass was e63e24c, so this covers two commits: ccddae04
(compile-safe fallback via a helper instead of an inline try/except in the compiled forward,
plus arch detection preferring the active torch device over rocminfo) and add32da4 (comment
reduction, and removal of a dead `rest_args_clean` assignment).

Verified locally on gfx1200 / ROCm 7.13: both files compile, the arch gate returns sdpa on
this card, and the rewrite guards still accept the bare/trailing-comma/`F.` alias forms
while rejecting `.transpose()` chains, `* scale`, `[0]`, `attn_mask`, positional dropout,
and `is_causal=<variable>`.

danielhanchen:
Thanks for this, aiter is worth having. I pushed two commits to the branch rather than leaving review notes, since the second one changes the approach.

### The rewrite never fires

I checked this three independent ways and they agree:

- **Static census.** Parsed every `transformers/models/**/modeling_*.py` on 4.51.3, 4.57.6 and 5.5.0. 106 SDPA call sites total, **0** in a shape the guard accepts. 67 of them pass `attn_mask`, `dropout_p` and `is_causal` as variables.
- **Runtime capture.** Wrapped `replace_sdpa_with_amd_aiter` and drove `unsloth_compile_transformers`. The hook is reached with 20 sources, and rewrites 0 of them.
- **The PR's own function**, run over the same corpus: 0 rewritten.

Three stacked reasons:

1. `compiler.py` calls the rewrite at the end of a three branch block, and only the `sdpa_dynamic_mask` branch is reachable in practice. It defaults `True`, and every caller in `unsloth/models/_utils.py` leaves it `True`. That branch never touches SDPA calls, so the input is raw upstream source.
2. On transformers 4.50+ the attention forward has no literal SDPA call at all. It dispatches through `ALL_ATTENTION_FUNCTIONS`, and the real call is in `transformers/integrations/sdpa_attention.py`, which this loop never reads.
3. Where SDPA text does survive it carries `attn_mask=`, `dropout_p=`, `scale=` or `is_causal=is_causal`, all rejected by `re.fullmatch(r"is_causal\s*=\s*True", ...)`.

That surface is also shrinking: 116 files, then 69, then 36, then 9 across the versions I checked.

### Commit 1: select on the AST, not a regex (`611ce7a`)

The pattern is line oriented and token unaware. Adversarial corpus of 39 cases found it rewrites:

- a commented out SDPA assignment,
- an assignment after a semicolon,
- a matching line inside a triple quoted string,

each producing invalid Python or corrupted string content. Now the selection parses the source and walks it, so only real `Assign` nodes that own their own lines are candidates. Generated code, guards and numerics are unchanged from your version. Generated names also get a source derived suffix so they cannot collide with a forward that already binds `_aiter_ok`. If the source does not parse, the function returns it untouched.

39/39 adversarial cases pass, and every output goes through `ast.parse`.

### Commit 2: register aiter as an attention backend (`008a3dc`)

`unsloth_zoo/temporary_patches/amd_aiter.py` registers aiter through `AttentionInterface`, which reaches attention where a rewrite provably cannot, needs no codegen, and cannot emit invalid Python. The interface hands q/k/v as `(B, H, S, D)` and expects `(B, S, H, D)` back, which is aiter's native layout, so it is one transpose in and none out.

Everything not provably safe falls through to the stock SDPA backend, so this is a kernel swap and never a behaviour change. It falls back on: an explicit mask (aiter aligns causal bottom right, SDPA top left), q/k length mismatch, dropout, a scale other than `1/sqrt(head_dim)` (`flash_attn_func` takes no `softmax_scale`), fp32, `requires_grad` (aiter asserts `return_lse` when inputs need grad), `is_causal` not `True`, `output_attentions`, `head_mask`, `sliding_window`, `position_bias`, `softcap`, `s_aux`, `attention_chunk_size`, and **any keyword it has never seen**, so a newer transformers growing a semantic argument degrades instead of silently returning wrong numbers. A raising kernel also degrades rather than killing the forward.

`sliding_window` and `position_bias` were real holes I hit on 5.14.1 during the version sweep, not hypotheticals.

`UNSLOTH_DISABLE_AITER=1` opts out.

### Verification

On a real Llama forward with aiter stubbed to compute what SDPA computes (the dispatch is what is under test, and it is device independent):

- routes through the backend on 2 of 2 layers, `max |aiter - sdpa|` logits `0.000e+00`
- 15/15 guard cases fall back exactly as intended
- `torch.compile` works through the registered backend
- a full training step takes the fast path 0 times

No breakage off AMD, which matters because `__init__.py` now imports this on every platform:

- 65/65 cell matrix over [Linux, macOS, Windows, WSL] x [NVIDIA, AMD, Intel XPU, CPU only], plus a real `torch 2.13.0+xpu` wheel
- fresh process probes on NVIDIA and CPU: `aiter` never registered, 0 ROCm subprocesses, attention keys unchanged
- forward only bitwise A/B against main on NVIDIA: identical loss values, identical logit SHA-256, peak memory 1.688 GB both sides
- version sweep across torch 2.6.0/2.9.1/2.13.0, transformers 4.57.6/5.0.0/5.14.1, TRL 0.22.2/0.24.0/1.9.2
- test suite: same pass/fail counts as the PR head, with the new module registered in `tests/test_temporary_patches_imports.py`

The two commits are independent, so if you would rather keep only the backend, commit 1 can be dropped. What I do not have is an MI300X, so the aiter call itself is stubbed here. If you can run this on gfx942 and confirm the fast path fires and the numbers match SDPA, that is the last piece.

danielhanchen:
One more commit (`ab41467`): the forward now checks the device itself, not just at registration.

Registration is gated on `get_amd_attention_implementation()`, but that gate is `functools.cache`d and evaluated once at import, and `aiter_attention_forward` is in `__all__` so it is directly callable. Relying on the registration gate alone meant a non-ROCm build or an off-GPU tensor could in principle reach a ROCm-only kernel.

```python
from ..device_type import is_hip
if not is_hip() or not query.is_cuda:
    return _fallback()
```

Both conditions are needed: ROCm tensors report `device.type` as `"cuda"`, so the tensor check alone does not distinguish ROCm from CUDA, and the build check alone does not catch a CPU or meta tensor on a ROCm host. `is_hip()` is uncached and is two attribute lookups, so it cannot go stale the way the registration gate can, and it is cheap enough to run per call.

Guard suite is now 17/17, with the baseline still taking the fast path so the new check is not over broad:

```
is_hip()=False       -> aiter used=False
CPU tensors on ROCm  -> aiter used=False
baseline             -> aiter used=True
```

Off-AMD behaviour is unchanged: `aiter` still never registers on NVIDIA or CPU, 0 ROCm subprocesses, attention keys unchanged.

zhihuidu-amd:
The rework is significantly better — thank you. Using `AttentionInterface` is the right fix for the call-site problem that LeoBorcherding and I had documented as a known limitation.

**Validation job submitted**: Slurm job **67712474** on `quanta-ccs-aus-e15-31` (AMD Instinct MI325X, gfx942, 256 GB HBM3e, ROCm 6.2.4). The job tests that `aiter_attention_forward` fires on real hardware and that `max|aiter - sdpa|` is within tolerance. Will post results as soon as it completes (node is `mixed` so it should start shortly).

**On the Codex P1** (backend registered but `config._attn_implementation` stays `'sdpa'`): this is real — `AttentionInterface.register('aiter', fn)` adds to the registry but nothing switches `_attn_implementation` to `'aiter'`. The backend will only be selected if something explicitly sets that config value. Is the intent to hook into `unsloth_compile_transformers` to override `_attn_implementation` to `'aiter'` on AMD, similar to how other backends are selected? Or does a separate Unsloth patch already handle this?

**On Codex P2a** (fallback bypasses patched SDPA for Gemma-4 sliding window): calling `sdpa_attention_forward` directly bypasses `ALL_ATTENTION_FUNCTIONS['sdpa']` which may carry Unsloth's sliding-window patch. Should the fallback call `ALL_ATTENTION_FUNCTIONS.get('sdpa', sdpa_attention_forward)` instead?

**On Codex P2b** (`UNSLOTH_DISABLE_AITER=1` doesn't disable the source rewrite): the env var check is only in `aiter_attention_forward`, not in `replace_sdpa_with_amd_aiter`. Since you've confirmed the source rewrite is now mostly inert (0 of 106 SDPA sites match), this may not matter in practice — worth noting but probably not blocking.

Results from job 67712474 coming shortly.

zhihuidu-amd:
Hardware validation result from Slurm job 67712477 (AMD Instinct MI325X, gfx942, 256 GB HBM3e):

**amd-aiter is not runnable on our available cluster nodes** due to a ROCm ABI mismatch. The error on our MI325X nodes (ROCm 6.2.4, with `/opt/rocm-7.2.0/lib/libamdhip64.so.7` partially present):

```
ImportError: /opt/rocm-7.2.0/lib/libamdhip64.so.7: undefined symbol: hsa_amd_memory_get_preferred_copy_engine, version ROCR_1
```

This is the same blocker we had when we first tried aiter — the PyPI wheel requires a full ROCm ≥7.0 stack with a specific ABI. Our cluster nodes have ROCm 6.2.4 as the primary runtime, and the partial ROCm 7.2.0 installation is missing the required symbol.

We do not currently have access to a node with a complete, matching ROCm 7.x + aiter environment. LeoBorcherding tested on **gfx1200 (RX 9060 XT, ROCm 7.13)** and confirmed the guards work correctly there, but they don't have a gfx942 box.

The aiter fast path on gfx942 remains **stubbed/unverified** — we can confirm dispatch correctness (all guards, fallback cases) but cannot confirm the actual `flash_attn_func` kernel runs and produces correct numbers on MI300X/MI325X without a ROCm 7.x node. We'll flag this as a known gap and continue looking for a compatible environment.

danielhanchen:
`a01e4ae` makes the rewrite a fixed point. I went looking for this specifically because these changes have to be idempotent, and the rewrite was not.

The generated block emits two SDPA fallbacks, one on the `_aiter_out is not None` branch and one on the `else`, and each is itself a valid match for the selector. So a second pass re-wrapped both:

```
1x:  1107 bytes, 1 helper
2x:  3421 bytes, 3 helpers
3x:  keeps growing
```

Every pass still produced parseable Python, which is why it was not obvious. Compiled sources do come back through here (cached modules, repeated compile passes), so this was reachable rather than theoretical. Rewritten blocks now carry a marker and are returned unchanged on any later pass, with the marker defined once and used both when emitting and when checking.

Worth noting this was in the original regex version too, so it is not something the AST change introduced.

Full idempotency check, 50 applications each, on real non-AMD hardware with nothing faked:

| | result |
|---|---|
| `replace_sdpa_with_amd_aiter` x50, 45 sources (40 real transformers files) | byte identical, and returns the same object |
| `patch_amd_aiter_attention` x50 | no change to attention keys, environ, `sys.modules`, patch list, default dtype |
| `get_recommended_attn_implementation` x50 (#944) | stable `None` |
| simulated AMD: register x50 | exactly one `aiter` key, still pointing at our forward |
| simulated AMD: rewrite x3 | fixed point after the first pass |

Off AMD the function returns the identical object before it parses anything, so the marker check costs nothing there.

Adversarial codegen corpus is still 39/39.

zhihuidu-amd:
The idempotency fix in `a01e4ae` is an important catch — a rewrite that keeps growing on repeated passes would have been a real problem for cached/recompiled modules. The 50-pass fixed-point verification across 45 real transformers files is exactly the right test for this.

On the three Codex P2s against `a01e4ae`:

**P2-1 (custom callable shadowing)**: Valid — any user function named `scaled_dot_product_attention` in scope would be wrongly rewritten. In practice this is unlikely in transformers source, but worth noting. Since the AST version already walks `ast.Assign` nodes, checking that the call's function node resolves to a known PyTorch attribute (`torch.nn.functional`, `F`, `nn.functional`) would close it.

**P2-2 (tab indentation)**: Valid — `node.col_offset` counts tabs as 1 column, but reconstructing as that many spaces produces `TabError` when the surrounding file uses tabs. Fix: preserve the exact whitespace prefix from the original source line (`source_lines[node.lineno - 1]` up to the first non-whitespace character) rather than synthesizing `' ' * col_offset`.

**P2-3 (output contiguity)**: Valid and the more likely to cause runtime failures. `.transpose(1, 2)` returns a non-contiguous view; the original SDPA returns a contiguous tensor; any downstream `.view(...)` will raise. The `amd_aiter.py` backend already calls `.contiguous()` (line 162), so the fix for the source-rewrite path is to add `.contiguous()` after `.transpose(1, 2)` in the generated `_aiter_out` assignment:
```python
attn_output = flash_attn_func(...).transpose(1, 2).contiguous()
```

Happy to prepare a commit for P2-3 (contiguity) since it's a one-line fix and directly matches what `amd_aiter.py` already does. P2-1 and P2-2 touch the AST walking logic so you may prefer to handle those alongside the idempotency code.

zhihuidu-amd:
Responding to the 3 Codex P2s on commit `a01e4ae` (danielhanchen's idempotency fix):

**P2-1 (custom callable shadowing)**: The AST walk already checks that the call's function is an attribute access (`ast.Attribute`) — the check could be tightened to verify the root object resolves to `torch`, `F`, or `nn.functional`. In practice, user functions named `scaled_dot_product_attention` are extremely rare in transformers source, but the guard is worth having.

**P2-2 (tab indentation)**: Valid. If the source uses tabs, `node.col_offset` gives the column in source units (counting each tab as 1), and synthesizing that many spaces produces `TabError`. Fix: read the exact prefix from the original source line rather than synthesizing from `col_offset`. This is in danielhanchen's AST code — flagging it here for his awareness rather than proposing a patch on someone else's refactor.

**P2-3 (output contiguity)**: This one we can fix. The source-rewrite path generates `_aiter_out.transpose(1, 2)` without `.contiguous()`, while `amd_aiter.py` already calls `.contiguous()` on line 162. The fix is one character — adding `.contiguous()` after the transpose in the generated code. Will prepare a commit for this if danielhanchen wants us to handle it, or he can add it alongside P2-1/P2-2 when he touches `compiler.py` for the other two.

danielhanchen:
Ran the full CI on a staging repo rather than the org queue, and then per-OS on top.

**Repo CI, PR merged into fresh upstream main (so this is not the stale branch point):** 12/12 green.

| job | result |
|---|---|
| (Python 3.10 / 3.11 / 3.12 / 3.13) | success |
| Core (HF=4.57.6 + TRL<1) | success |
| Core (HF=default + TRL=default) | success |
| Core (HF=latest + TRL=latest) | success |
| Repo tests (CPU) | success |
| MLX numerics (Linux CPU) | success |
| MLX suites (Linux CPU MLX) | success |
| Source lint, Wheel build + import smoke | success |

**Cross platform.** The repo matrix is `ubuntu-latest` only and the one macOS job is `if:` gated, so these PRs had never actually run on Windows or macOS. Added real per-OS jobs:

| OS | result |
|---|---|
| ubuntu-latest | 37 passed |
| windows-latest | 37 passed |
| macos-14 (Apple Silicon) | 37 passed |

That is `tests/test_temporary_patches_imports.py` in full, including the `amd_aiter` import, on each OS. Worth having: `device_type.py` guards `import torch` behind `if not _IS_MLX`, and the new module imports torch at module top, so a Mac import was the check I most wanted.

Two earlier per-OS runs failed on missing runner dependencies rather than on this code (no `unsloth`, then no `torch`), which is why the numbers above come from jobs that install both. The giveaway was that all 37 failed, including modules this PR never touches.

Staging PRs are closed, never merged.

Issue #757, opened 2026-06-11
Support pure-audio models (Qwen2-Audio, Voxtral, Granite-Speech) in UnslothVisionDataCollator


danielhanchen:
## Summary

`UnslothVisionDataCollator` gained audio support for audio-capable vision models (Gemma 4, #723/#740), but pure-audio models cannot construct the collator at all: `__init__` raises `TypeError` unless the processor has an `image_processor`. This tracks extending support to audio-only models, scoped by what each family actually needs.

## Why simply relaxing the gate is not enough

Adding `or hasattr(processor, "feature_extractor")` to the constructor gate looks sufficient but reintroduces the silent-corruption class #723 fixed. Verified against Qwen2-Audio:

- Batches would form correctly: the `audio=` kwarg matches, placeholders expand, `input_features` and `feature_attention_mask` are emitted.
- But `<|AUDIO|>` (id 151646) would NOT be masked in labels: Qwen2-Audio stores `audio_token` on the processor, while `get_padding_tokens_ids` unwraps to the tokenizer first and loses it. The model would train on audio placeholder positions.
- `_truncate_sequence_tensors` has the same gap when protecting audio spans from truncation.

## Scoped plan

**Phase 1, Qwen2-Audio family (tractable, self-contained PR):**
- Relax the constructor gate for processors with a `feature_extractor`.
- Consult `processor.audio_token` / `processor.audio_token_id` before the tokenizer unwrap in `get_padding_tokens_ids` and `_truncate_sequence_tensors` (shared-code change, needs its own tests).
- Hermetic stub tests plus a GPU smoke run.

**Phase 2, per-family work (each needs its own design):**
- Voxtral: the processor `__call__` takes no `audio` kwarg; audio flows through its own template API, and the tekken chat template also fails the collator's constructor probe. Needs a dedicated formatting path.
- Granite-Speech: requires torchaudio at processor load time and ships as a LoRA adapter; both need handling before the collator is even reachable.

## Notes

- The feature extractor padding-side normalization from #747 already applies generically once construction succeeds.
- Phase 1 should not change behavior for any currently supported model: the new processor-level token lookup only adds tokens that the tokenizer-level lookup missed.

PR #924, opened 2026-07-20
Fix: Let UnslothVisionDataCollator accept audio-only processors

BardiaKoopah:
# Let `UnslothVisionDataCollator` accept audio-only processors
 
Refs #757
 
## The bug
 
`UnslothVisionDataCollator.__init__` hard-required an `image_processor`. Audio-only processors — Qwen2-Audio, Voxtral, Granite-Speech, which carry a `feature_extractor` but no `image_processor` — hit the guard and raised at construction:
 
```
TypeError: Unsloth: UnslothVisionDataCollator is only for image models!
```
 
So audio LoRA training couldn't even get off the ground: the collator refused to build.
 
## The audio pipeline was already there — only construction was blocked
 
The downstream machinery for audio already landed in earlier PRs:
 
- **#723** added audio content extraction and `AUDIO_TOKENS` masking, so `<|AUDIO|>` is already masked in labels.
- **#917** hardened the audio path.
The only remaining gap was the constructor gate. This is why the fix is small and why the scope here is **construction, not masking** — the masking the issue mentions is already handled.
 
## The fix
 
Widen the constructor gate to admit any **image- or audio-capable** processor, and reject only a bare text tokenizer:
 
```python
if not hasattr(processor, "image_processor") and getattr(processor, "feature_extractor", None) is None:
    raise TypeError("Unsloth: UnslothVisionDataCollator requires an image or audio processor!")
```
 
`feature_extractor` is the same audio signal already used by `_fix_audio_feature_extractor_padding_side`, so this keys audio detection on an existing convention rather than a new heuristic. The vision and vision+audio paths are **byte-identical** — they carry an `image_processor`, so the first clause is false and the gate is never triggered (Gemma3n vision+audio included).
 
## Verification (CPU, no CUDA, no weights)
 
- A real `Qwen2AudioProcessor` (config only, no weights) now constructs the collator.
- A full `__call__` round-trip on a synthetic waveform produces `input_features`, correctly masks the audio and pad tokens, and keeps the real assistant tokens unmasked.
- Vision and vision+audio paths confirmed byte-identical; a bare text tokenizer still raises.
## Tests
 
Regression tests extend `tests/test_vision_collator_audio.py`:
 
- the construction gate (audio-only admitted, text-only rejected),
- a hermetic stub-processor round-trip that runs in CI **without** audio deps,
- a real Qwen2-Audio round-trip that skips cleanly when transformers / the processor download are unavailable.
Canonical MLX baseline suite: 0 failures.
 
## Scope
 
Addresses #757 (a maintainer-scoped task). `<|AUDIO|>` masking was already handled by #723, so this covers construction plus the audio round-trip, not masking. The synthetic stub round-trip is exercised in CI; the real-processor round-trip is verified locally but inferred for CI.

### DIFF
--- a/tests/test_vision_collator_audio.py
+++ b/tests/test_vision_collator_audio.py
@@ -629,6 +629,24 @@ class _VisionProcessor(_ChatTemplateMixin):
         self.image_processor = object()
 
 
+class _AudioProcessorAttrProcessor(_ChatTemplateMixin):
+    # Granite-Speech shape: transformers names the audio sub-processor
+    # "audio_processor", not "feature_extractor" (see that model's processor
+    # .attributes). No image_processor either, so a gate that only knows
+    # feature_extractor rejects an audio model it is meant to support.
+    def __init__(self):
+        self.tokenizer = _FakeTokenizer()
+        self.audio_processor = _FakeFeatureExtractor()
+
+
+class _NoneImageProcessor(_ChatTemplateMixin):
+    # Defines image_processor but leaves it None: hasattr() is True while there
+    # is no usable processor, so the gate must look at the value, not the name.
+    def __init__(self):
+        self.tokenizer = _FakeTokenizer()
+        self.image_processor = None
+
+
 class _TextOnlyProcessor(_ChatTemplateMixin):
     # Neither image_processor nor feature_extractor -> must still be rejected.
     def __init__(self):
@@ -652,6 +670,23 @@ def test_audio_only_processor_constructs():
     # Masking wiring is intact: the audio placeholder is in the padding ids.
     assert AUDIO_ID in collator.padding_token_ids.tolist()
     assert PAD_ID in collator.padding_token_ids.tolist()
+
+
+def test_audio_processor_attribute_processor_constructs():
+    # Granite-Speech exposes "audio_processor" instead of "feature_extractor";
+    # it must be accepted like any other audio-only processor.
+    collator = UnslothVisionDataCollator(
+        model=_stub_model(), processor=_AudioProcessorAttrProcessor(),
+    )
+    assert collator.processor.__class__.__name__ == "_AudioProcessorAttrProcessor"
+    assert AUDIO_ID in collator.padding_token_ids.tolist()
+
+
+def test_none_image_processor_still_rejected():
+    # hasattr() is True here but the attribute is None, so this is really a
+    # text-only processor and must not slip through the gate.
+    with pytest.raises(TypeError, match="image or audio processor"):
+        UnslothVisionDataCollator(model=_stub_model(), processor=_NoneImageProcessor())
 
 
 def test_text_only_processor_still_rejected():
--- a/tests/test_vision_collator_audio.py
+++ b/tests/test_vision_collator_audio.py
@@ -585,6 +585,151 @@ def test_userdict_audio_payload_resolves():
     np.testing.assert_allclose(clips[0], DECODED)
 
 
+# ---------------------------------------------------------------------------
+# Constructor gate: audio-only processors (unslothai/unsloth-zoo#757)
+#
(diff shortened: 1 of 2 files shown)


BardiaKoopah:
@codex

BardiaKoopah:
@codex review

BardiaKoopah:
@codex review

BardiaKoopah:
@codex review

BardiaKoopah:
@codex review

BardiaKoopah:
@gemini-code-assist took the getattr point — fixed in `29c7df6c`. `test_vision_processor_still_constructs` was asserting `hasattr(collator.processor, "image_processor")`, which passes on exactly the `image_processor = None` case the constructor gate rejects, and `_NoneImageProcessor` exists in that file to pin the name-vs-value distinction. The assertion on the real processor was already value-based. Left the `hasattr` calls at 366, 605-606 and 1052 alone, those assert an attribute is *absent* (unpatched decoder has no `get`/`keys`, feature extractor has no `padding_side`) where `hasattr` is the right predicate.

On the `importorskip("soundfile")` I'm leaving it out. You're right that the comment implied an audio-backend dependency and that was misleading, so I dropped "soundfile" from it in the same commit, but the test doesn't actually have one. It feeds a numpy array and Qwen2AudioProcessor hands that straight to WhisperFeatureExtractor (`processing_qwen2_audio.py:127`); the only soundfile use on the transformers side is `sf.SoundFile` and `sf.write` inside `audio_utils.load_audio_as`, which takes a path or URL and is never reached. I checked rather than assuming — with `transformers.utils.import_utils._soundfile_available = False`, which is what a soundfile-free env yields on the pinned transformers 4.57.6, `audio_utils` imports with `sf` unbound, `WhisperFeatureExtractor()(wav, sampling_rate=16000)` returns `input_features` of shape (1, 80, 3000), and soundfile never enters `sys.modules`. So the importorskip would skip the test on machines where it runs fine. I didn't execute the real-processor test itself this round since it pulls the processor down.

@chatgpt-codex-connector that gap is already closed further up the branch — `_AUDIO_CONTENT_KEYS` is `("audio", "url", "path", "audio_url")` as of `0200bb4d`, so a Qwen `{"type": "audio", "audio_url": ...}` part now yields its clip instead of raising, and `eacd5906` made the key resolution presence-based (`is None`) rather than truthiness, because the old `ele.get("url") or ele.get("path")` chain raises "truth value of an array with more than one element is ambiguous" on a numpy waveform.

BardiaKoopah:
@codex review

BardiaKoopah:
@codex review

BardiaKoopah:
@oobabooga this one too pls

danielhanchen:
Confirmed the gate in unsloth_zoo/vision_utils.py still hard-requires image_processor, so audio-only processors cannot construct the collator, and widening it is the right call. The branch also adds the audio kwarg signature introspection, the audio_processor sub-processor resolution and a new truncation guard in _truncate_by_side. Could you narrow this to the constructor gate and send the rest separately?

danielhanchen:
Confirmed the constructor gate in unsloth_zoo/vision_utils.py still hard-requires image_processor, so audio-only processors cannot build the collator at all, and widening it is the right direction. The branch also rewrites the chat-template probe, adds audio-kwarg signature introspection and a new truncation guard on the prompt/completion path; could you split it so the gate change lands on its own first?

danielhanchen:
Confirmed the constructor gate in unsloth_zoo/vision_utils.py still hard-requires image_processor, so this is the right direction, but the branch also adds an audio-kwarg signature probe, a chat-template probe rewrite and a new truncation guard in the prompt/completion path. Could you split it down to the gate plus its tests, and confirm whether Qwen2-Audio's processor-level audio_token is masked given get_padding_tokens_ids unwraps to the tokenizer first?

PR #925, opened 2026-07-20
Fix Gemma 4 audio and forced-float32 PLE dtype mismatches

Etherll:
## Summary

Fixes two Gemma 4 mixed-dtype failures on FP16 hardware:

- Casts audio features to the destination text-embedding dtype before `masked_scatter`.
- Aligns forced-FP32 residual activations with the three PLE projection boundaries.

## Motivation

Gemma 4 inference and training on a T4 could fail with:

```text
masked_scatter_: expected self and source to have same dtypes but got Float and Half
```

Training could then fail at the PLE projections with:

```text
expected mat1 and mat2 to have the same dtype, but got: float != c10::Half
```

## Changes

- Patch the eager audio merge in `unsloth_zoo/temporary_patches/gemma4.py:674` while preserving the existing decorated forward wrapper.
- Apply the same audio rewrite to generated compiled modules in `unsloth_zoo/compiler.py:1234`.
- Add a quantization-aware PLE input helper in `unsloth_zoo/temporary_patches/gemma4_float32.py:52`.
- Patch `per_layer_model_projection`, `per_layer_input_gate`, and `per_layer_projection` in eager and compiled paths.
- Support PEFT-wrapped linear layers without casting activations to packed `uint8` weight dtypes.
- Gate PLE changes on `UNSLOTH_FORCE_FLOAT32=1`. BF16, FP32, and unrelated model families retain their existing paths.
- Add regression tests for dtype combinations, PEFT wrappers, quantized weights, source drift, idempotency, compiled generation, and eager/compiled parity.

## How to test

Run the focused regression suite:

```bash
pytest -q tests/test_gemma4_audio_dtype.py tests/test_gemma4_forced_float32_ple_dtype.py
```

Expected result:

```text
32 passed
```

The broader Gemma 4/compiler suite completed with 122 passed and 1 skipped.

The pushed commit was also tested on a Tesla T4 with Gemma 4 E2B:

- Default LoRA completed three steps with finite loss and gradient norm.
- Explicit PLE-targeted LoRA wrapped all 71 PLE modules and completed three steps.
- Compiled and compile-disabled runs produced numerically matching metrics.
- Reentrant checkpointing was active on 52 modules and completed backward passes.
- A BitsAndBytes 4-bit model contained 169 packed `uint8` modules while zero PLE activations were cast to `uint8`.
- Text, image, audio, and video forwards completed with finite outputs.
- All 35 inspected attention softmax calls remained FP32.
- The dtype guard had identical logits and peak memory compared with the minimal correct cast.
- A clean runtime executed TRL's `_chunked_ce_forward` 12 times without OOM. Peak allocation was 12.937 GiB on the T4.

## Notes for reviewers

The transforms fail closed when the expected upstream source pattern is missing or ambiguous. Already-fixed upstream implementations are left unchanged.

T4 does not support native BF16 model execution. The BF16 no-op path was therefore checked directly at the dtype boundary, including tensor identity, storage identity, and gradient preservation.

### DIFF
--- a/tests/test_gemma4_audio_dtype.py
+++ b/tests/test_gemma4_audio_dtype.py
@@ -1,0 +1,184 @@
+import importlib.util
+import inspect
+import sys
+import textwrap
+
+import pytest
+import torch
+
+from unsloth_zoo.temporary_patches import gemma4
+from unsloth_zoo.compiler import fix_gemma4_audio_feature_dtype
+
+
+def _load_fake_gemma4(tmp_path, monkeypatch, name, cast_expression):
+    source = textwrap.dedent(
+        f"""
+        import functools
+        import torch
+
+        def preserve_metadata(function):
+            @functools.wraps(function)
+            def wrapper(*args, **kwargs):
+                return function(*args, **kwargs)
+            return wrapper
+
+        class Gemma4Model:
+            def __init__(self):
+                pass
+
+            @preserve_metadata
+            def forward(self, inputs_embeds, audio_features=None):
+                if audio_features is None:
+                    return inputs_embeds
+                audio_mask = torch.ones_like(inputs_embeds, dtype=torch.bool)
+                return inputs_embeds.masked_scatter(
+                    audio_mask,
+                    {cast_expression},
+                )
+        """
+    )
+    path = tmp_path / f"{name}.py"
+    path.write_text(source, encoding="utf-8")
+    spec = importlib.util.spec_from_file_location(name, path)
+    module = importlib.util.module_from_spec(spec)
+    monkeypatch.setitem(sys.modules, name, module)
+    spec.loader.exec_module(module)
+    return module.Gemma4Model
+
+
+@pytest.mark.parametrize("inputs_dtype", [torch.float16, torch.bfloat16, torch.float32])
+@pytest.mark.parametrize("audio_dtype", [torch.float16, torch.bfloat16, torch.float32])
+def test_gemma4_audio_merge_aligns_to_actual_inputs_embeds_dtype(
+    tmp_path, monkeypatch, inputs_dtype, audio_dtype,
+):
+    model_cls = _load_fake_gemma4(
+        tmp_path,
+        monkeypatch,
+        f"fake_gemma4_{str(inputs_dtype)}_{str(audio_dtype)}".replace(".", "_"),
+        "audio_features.to(inputs_embeds.device)",
+    )
+    assert gemma4._patch_gemma4_audio_feature_dtype_on_class(model_cls) is True
+    patched_source = inspect.getsource(model_cls.forward)
+    assert "audio_features.to(inputs_embeds.device)" not in patched_source
+    assert "audio_features.to(inputs_embeds.device, inputs_embeds.dtype)" in patched_source
+
+    inputs = torch.zeros(1, 2, dtype=inputs_dtype, requires_grad=True)
+    audio = torch.tensor([[1.25, -2.5]], dtype=audio_dtype, requires_grad=True)
+    output = model_cls().forward(inputs_embeds=inputs, audio_features=audio)
+
+    assert output.dtype == inputs_dtype
+    torch.testing.assert_close(output, audio.to(inputs_dtype))
+    output.float().sum().backward()
+    assert audio.grad is not None
+
+
+def test_gemma4_audio_dtype_patch_preserves_no_audio_and_is_idempotent(
+    tmp_path, monkeypatch,
+):
+    model_cls = _load_fake_gemma4(
+        tmp_path,
+        monkeypatch,
+        "fake_gemma4_idempotent",
+        "audio_features.to(inputs_embeds.device)",
+    )
+    original_signature = inspect.signature(model_cls.forward)
+    original_forward = model_cls.forward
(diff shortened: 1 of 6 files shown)


danielhanchen:
Thanks for the PR. The goal is to fix two Gemma 4 mixed-dtype failures on FP16 hardware: the audio `masked_scatter` (audio features cast to device but not dtype), and, under `UNSLOTH_FORCE_FLOAT32=1`, the three PLE projection boundaries (fp32 residual into fp16 weights). It patches both the eager path (AST plus `__code__` surgery in `gemma4.py` and `gemma4_float32.py`) and the compiled path (source rewrites in `compiler.py`), with a quantization-aware cast helper and a solid regression suite.

I checked the patterns against the current upstream `modeling_gemma4.py` and the core change is correct and complete: audio is the only unguarded merge site (image at line 2338 and video at 2357 already pre-cast to `inputs_embeds.dtype`), the audio fix is correctly unconditional while the PLE fix is correctly gated on forced float32, the fp32 residual carrier is preserved with gradient flow, and the eager and compiled transforms are idempotent and correctly targeted. The 32 new tests and the existing gemma4/compiler suite pass.

A few things worth tightening, roughly in priority order:

**1. Eager and compiler PLE helpers use different names (main one to fix).**
`gemma4_float32.py` rewrites calls to `_gemma4_ple_cast_input` and injects that name into the transformers module, while `compiler.py` rewrites to `_unsloth_gemma4_ple_cast_input` and appends its own helper. Because the eager patch poisons `linecache` and `create_standalone_class` reads source via `inspect.getsource`, a compile that runs after the eager patch can emit generated code that calls `_gemma4_ple_cast_input` while only `_unsloth_gemma4_ple_cast_input` is defined, which risks a `NameError`. A single shared helper name across both paths removes the hazard.

**2. Reload can misclassify the patch as already applied.**
The permanent `linecache` entry (`mtime=None`) means that after an `importlib.reload` of the modeling module, the class gets fresh buggy bytecode but `getsource` still returns the patched text, so the "already fixed" guard sets the marker and skips the real fix, leaving the bug live. Keying the already-fixed decision on the live bytecode (for example `co_names`/`co_consts` of `forward.__code__`) rather than linecache-served source avoids this. Narrow trigger, but silent when it fires.

**3. Source drift is a silent no-op.**
The top-level `patch_*` functions ignore the boolean return, and the helper returns `False` for both "already correct" and "pattern not found". On a future transformers reformat the fix would disappear with no diagnostic. A one-line warning when the known-bad source cannot be matched would make this debuggable.

**4. Non-atomic PLE application.**
`patch_Gemma4PLEInputDtype` patches `Gemma4TextModel` and then `Gemma4TextDecoderLayer` independently. If the second fails, the first is already applied and the model is half-patched (the unpatched boundary then crashes). Pre-validating all three targets before mutating any avoids a partially patched model.

**5. FP8 weights are treated as dense.**
`float8_e4m3fn` and `float8_e5m2` report `is_floating_point == True` and carry no `quant_state`, so under forced float32 on an fp8 checkpoint the helper would cast the activation to fp8. Excluding sub-2-byte float dtypes keeps the fp32 carrier intact.

**6. Minor.**
- `weight.dtype` is read directly; `getattr(weight, "dtype", None)` is safer for accessor-style weights.
- The compiled-module cache key does not include `UNSLOTH_FORCE_FLOAT32`, so with `UNSLOTH_COMPILE_OVERWRITE=0` a module generated with the flag off could be reused when it is on.
- `test_gemma4_ple_three_boundary_eager_and_compiled_parity` hand-calls the eager helper rather than executing compiler-generated source, so it will not catch eager/compiled divergence. A real bitsandbytes `Linear4bit` forward under forced float32 is also untested.

Note: any full-suite failures about `tests/test_offline_env_cross_sync.py` or `tests/test_vision_collator_audio.py` are not from this PR. The branch predates #926 and does not touch those files; rebasing on main clears them.

Overall the fix is solid. Item 1 (shared helper name) is the main thing I would resolve before merge; the rest are good follow-ups.

danielhanchen:
Pushed a commit onto this branch to fix a cross-path bug and fold in a few cheap guards from the review.

The eager PLE patch rewrote projection calls to `_gemma4_ple_cast_input` and poisons `linecache`, while the compiler emitted a differently named `_unsloth_gemma4_ple_cast_input`. Because `create_standalone_class` reads source through `inspect.getsource`, compiling after the eager patch generated code that called `_gemma4_ple_cast_input` while only `_unsloth_gemma4_ple_cast_input` was defined, raising a `NameError`. I unified both paths on `_unsloth_gemma4_ple_cast_input` and now append the generated helper only when a call to it is present.

Also hardened the cast helper and patch flow:

- Skip sub-2-byte float weights (`float8_e4m3fn` / `float8_e5m2`) so the fp32 carrier is not cast to fp8, and read the weight dtype through `getattr`.
- Validate every PLE boundary before mutating any, so a drift in one method can no longer leave a half-patched model.
- Warn on transformers source drift instead of silently skipping the fix.

Added regression tests, including the eager-then-compile cross-path, an fp8 guard, a missing-dtype guard, dry-run classification, and helper-append gating. Full suite is green locally (39 focused tests plus 202 in the broader gemma4/compiler suite, no regressions).

The reload/linecache staleness and adding `UNSLOTH_FORCE_FLOAT32` to the compile cache key are left as follow-ups.

danielhanchen:
@codex review

PR #928, opened 2026-07-21
Auto-disable gradient-checkpoint double buffering on unified-memory GPUs

danielhanchen:
### What

Auto-disable gradient-checkpoint double buffering on unified-memory GPUs (AMD APUs, NVIDIA GB10 Spark).

### Why

Double buffering allocates a second GPU offload buffer (buffer B) so the H2D copy of the next checkpoint can overlap with compute. On a **discrete** GPU that overlap hides a real PCIe transfer and is a win. On a **unified-memory** device the GPU and system RAM are the same physical pool, so there is no transfer to hide: the extra buffer, the second stream, and the per-buffer events are pure overhead.

Measured on a Ryzen AI MAX+ 395 (gfx1151 Strix Halo, `is_integrated=1`), Qwen3.5-2B 4bit LoRA, `use_gradient_checkpointing="unsloth"`:

| | double buffer ON (old default) | OFF (new default on UMA) |
|---|---|---|
| peak allocated | 3381 MB | 3377 MB |
| peak reserved | 3867 MB | 3861 MB |
| steady step | 3.620 s | **1.684 s** |

So ~2.15x faster steps, no memory change, identical loss.

### How

Both the init path (`initialize_unsloth_gradient_checkpointing`) and the reset path (`reset_unsloth_gradient_checkpointing_buffers`) now go through one helper, `_double_buffer_disabled()`:

- explicit `UNSLOTH_DISABLE_DOUBLE_BUFFER=0/1` always wins (either direction);
- when unset, defaults **off** if torch reports `is_integrated` on any visible cuda/hip device, else **on**.

`is_integrated` is the driver's own unified-memory answer (`hipDeviceProp_t.integrated` / `cudaDeviceProp.integrated`), so this also covers integrated parts outside a hardcoded arch list (e.g. gfx1103 Phoenix, GB10).

### Scope / safety

- Discrete NVIDIA and AMD GPUs report `is_integrated=0`, so their behaviour is unchanged (double buffering stays on).
- `getattr(..., "is_integrated", 0)` and the `try/except` keep it safe on older torch that lacks the attribute (falls back to on).
- Non cuda/hip backends (xpu) are untouched.

Verified: helper returns `True` (env unset, APU), `False` (env=0), `True` (env=1), and `False` (discrete, env unset); file passes `ast.parse`.

### DIFF
--- a/unsloth_zoo/gradient_checkpointing.py
+++ b/unsloth_zoo/gradient_checkpointing.py
@@ -70,20 +70,28 @@ def _any_device_integrated():
         return False
 
 
-def _double_buffer_disabled():
-    # Double buffering overlaps the H2D offload copy with compute, but on unified-
-    # memory devices (AMD APUs gfx1150/1151, NVIDIA GB10) there is no transfer to
-    # hide: pure overhead (~2x slower, no memory saved). UNSLOTH_DISABLE_DOUBLE_BUFFER
-    # =0/1 forces it; else disable when any visible device is integrated.
-    env = os.environ.get("UNSLOTH_DISABLE_DOUBLE_BUFFER")
-    if env is not None: return env == "1"
-    if DEVICE_TYPE not in ("cuda", "hip"): return False
-    return _any_device_integrated()
+# Whether to skip double buffering. Resolved lazily on the first checkpointing init
+# (below) and cached here -- NOT at import: get_device_properties forces CUDA lazy-init
+# on GPU 0, which would poison fork workers and pre-empt per-rank device selection.
+# None = unresolved. UNSLOTH_DISABLE_DOUBLE_BUFFER=0/1 forces it either way.
+UNSLOTH_DOUBLE_BUFFER_DISABLED = None
 
-# Resolved once at import, like DEVICE_TYPE / DEVICE_COUNT (the visible device set and
-# env are fixed by then). Set UNSLOTH_DISABLE_DOUBLE_BUFFER before importing unsloth
-# to override.
-UNSLOTH_DOUBLE_BUFFER_DISABLED = _double_buffer_disabled()
+
+def _double_buffer_disabled():
+    # Resolve once (after the caller has selected its device), memoized into the global.
+    # Double buffering overlaps the H2D offload copy with compute, but on unified-memory
+    # devices (AMD APUs gfx1150/1151, NVIDIA GB10) there is no transfer to hide: pure
+    # overhead (~2x slower, no memory saved), so disable when any device is integrated.
+    global UNSLOTH_DOUBLE_BUFFER_DISABLED
+    if UNSLOTH_DOUBLE_BUFFER_DISABLED is None:
+        env = os.environ.get("UNSLOTH_DISABLE_DOUBLE_BUFFER")
+        if env is not None:
+            UNSLOTH_DOUBLE_BUFFER_DISABLED = env == "1"
+        elif DEVICE_TYPE not in ("cuda", "hip"):
+            UNSLOTH_DOUBLE_BUFFER_DISABLED = False
+        else:
+            UNSLOTH_DOUBLE_BUFFER_DISABLED = _any_device_integrated()
+    return UNSLOTH_DOUBLE_BUFFER_DISABLED
 
 
 @contextmanager
@@ -457,7 +465,7 @@ def initialize_unsloth_gradient_checkpointing(dtype = None):
         with _no_inference_mode():
             GPU_BUFFERS = tuple([torch.empty(INITIAL_GPU_BUFFER_SIZE, dtype = dtype, device = f"{DEVICE_TYPE_TORCH}:{i}") for i in range(n_gpus)])
         # Double buffering: try to allocate buffer B (auto-off on unified memory, or via env var)
-        if UNSLOTH_DOUBLE_BUFFER_DISABLED:
+        if _double_buffer_disabled():
             GPU_BUFFERS_B = None
             USE_DOUBLE_BUFFER = False
             BUFFER_EVENTS_A = None
@@ -1038,7 +1046,7 @@ def reset_unsloth_gradient_checkpointing_buffers():
             NEXT_BUFFER_SLOT[i] = 0
 
     # Reset double buffering if buffer B still exists, or try to re-allocate
-    if UNSLOTH_DOUBLE_BUFFER_DISABLED:
+    if _double_buffer_disabled():
         if GPU_BUFFERS_B is not None:
             for i in range(len(GPU_BUFFERS_B)):
                 if GPU_BUFFERS_B[i] is not None and hasattr(GPU_BUFFERS_B[i], "resize_"):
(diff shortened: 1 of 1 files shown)


danielhanchen:
@codex review

danielhanchen:
Addressed in 53f5ec2. The check now gates on the active training device (torch.cuda.current_device()) instead of any() over every visible GPU, so a discrete GPU sharing a box with an APU keeps the H2D/compute overlap. The APU-only case is unchanged (its current device is the integrated one, so double buffering still auto-disables), and the explicit UNSLOTH_DISABLE_DOUBLE_BUFFER override still wins.

danielhanchen:
@codex review

danielhanchen:
Fixed in 5045d9c. The probe is now keyed by device index (functools.cache on _device_is_integrated(idx)), and _double_buffer_disabled reads torch.cuda.current_device() fresh on every call and passes that index in. So a mid-process set_device() on a mixed APU + discrete box reflects the device selected for that run (discrete keeps the overlap, APU disables it), while each device is still probed only once. Verified: switching current_device 0->1->0 returns True/False/True with each index probed exactly once, and the UNSLOTH_DISABLE_DOUBLE_BUFFER override is still read every call.

danielhanchen:
Switched to the simpler static rule: disable double buffering when ANY visible CUDA/HIP device is integrated, rather than gating on the active device.

Rationale:

- `_any_device_integrated` scans every visible device and is cached. The visible device set is fixed for the process, so unlike the earlier `current_device()` version this has no `set_device()` staleness concern (it does not depend on the current device at all), which is why the `functools.cache` is safe here.
- Trade-off: on a mixed integrated + discrete box that trains on the discrete GPU, this disables the overlap globally. I am accepting that for simplicity. It only affects a same-vendor iGPU + dGPU host where both are visible to one ROCm/CUDA torch build (cross-vendor mixes never enumerate both), which is vanishingly rare, disabling is always safe (it is a small perf trade, never a correctness issue), and `UNSLOTH_DISABLE_DOUBLE_BUFFER=0` re-enables it.

Testing (`temp/sim/verify_928_any.py`, real functions extracted from this branch, 25/25 on Windows and WSL): pure APU disables (matches this gfx1151 machine, `is_integrated=1`), pure discrete AMD / NVIDIA keep double buffering, mixed disables, `UNSLOTH_DISABLE_DOUBLE_BUFFER=0/1` overrides win, non-cuda/hip is a no-op, and the cache probes each device at most once with `any()` short-circuiting at the first integrated device.

danielhanchen:
Made the double-buffer decision a one-time check. `_double_buffer_disabled` now carries `functools.cache`, so the `UNSLOTH_DISABLE_DOUBLE_BUFFER` read and the integrated-device probe both happen once per process and every later call is served from cache. Moved the cache up from `_any_device_integrated` (now a plain helper, called at most once) to the decision itself.

This is safe because both inputs are fixed by the time gradient checkpointing initializes: the visible device set never changes, and the env var is set before GC init (either by the user, or by unsloth's `os.environ.setdefault("UNSLOTH_DISABLE_DOUBLE_BUFFER", "1")` in `patch_dgx_spark_runtime_defaults`, which runs during model setup). Both call sites (`initialize_...` and `reset_...`) use it purely as a stable gate, so they want the same answer. To override, set the env before the first GC init.

Confirmed on this gfx1151 box: `_double_buffer_disabled()` returns True (off) and `cache_info()` shows hits=2/misses=1 after 3 calls. Simulation (`temp/sim/verify_928_any.py`) 28/28 on Windows and WSL, including a one-time-cache test (env change after the first call is ignored until `cache_clear`).

danielhanchen:
Resolved the double-buffer decision into a module-level global instead of a cached function. `UNSLOTH_DOUBLE_BUFFER_DISABLED = _double_buffer_disabled()` is computed once at import, the same way `device_type.py` resolves `DEVICE_TYPE` and `DEVICE_COUNT`. Both call sites (`initialize_` and `reset_`) now read the global, and the `functools` import is gone (the cache is no longer needed).

This is safe because the inputs are fixed by import time: the visible device set never changes, and `UNSLOTH_DISABLE_DOUBLE_BUFFER` should be set before `import unsloth` (the standard pattern for `UNSLOTH_*` env vars; on DGX Spark the device is integrated so the global is True regardless of the setdefault). On a non-cuda/hip `DEVICE_TYPE` the global resolves to False without probing any device, so CPU-only imports touch no GPU.

Verified with a real package import on this gfx1151 box (`UNSLOTH_DOUBLE_BUFFER_DISABLED = True`, a plain bool; bytecode confirms both call sites read the global). Simulation `temp/sim/verify_928_any.py` extracts the real functions plus the real module-level assignment and exercises the global end to end: 28/28 on Windows and WSL (device matrix, env-before-import override, non-cuda skips the probe, probe-failure safe False, `any()` short-circuit).

danielhanchen:
@codex review

danielhanchen:
Good catch, fixed in aeb0d96. This is real: `get_device_properties` triggers CUDA lazy-init (I confirmed `device_count()` and `is_available()` leave `torch.cuda.is_initialized()` False, but `get_device_properties(0)` flips it to True), so resolving the global at import created a context on GPU 0 before callers select their rank/device.

`UNSLOTH_DOUBLE_BUFFER_DISABLED` now starts as `None` and is resolved on the first `_double_buffer_disabled()` call from `initialize_`/`reset_` (after device selection), memoizing into the global. The env-forced and non-cuda/hip paths never probe a device. The variable still exists and holds the process-wide decision once resolved, so nothing at import touches the GPU.

Verified: exec'ing the module-level code with real torch leaves `torch.cuda.is_initialized()` False and the global `None`; only the first resolver call probes (flipping it True) and memoizes. A full package import shows `UNSLOTH_DOUBLE_BUFFER_DISABLED is None` after import and both call sites bound to the resolver. Simulation `temp/sim/verify_928_any.py` 35/35 on Windows and WSL, including explicit "no get_device_properties at import" / "env and non-cuda paths never probe" / memoized-after-first assertions.

danielhanchen:
Simplified further in 6675cc9. Dropped the `UNSLOTH_DOUBLE_BUFFER_DISABLED` global and the lazy-memoize logic; `_double_buffer_disabled` is just a plain function again (no `functools.cache`, no global).

This still fixes the import-time CUDA lazy-init issue for free: it is a function, only called from `initialize_`/`reset_` at GC init, so nothing probes a device at import. The env-forced and non-cuda/hip paths still short-circuit before any `get_device_properties` call. Recomputing on the two call sites (per model) is negligible, so there is no reason to cache.

Verified: exec'ing the module's functions with real torch leaves `torch.cuda.is_initialized()` False (only the GC-init call flips it True); `temp/sim/verify_928_any.py` 32/32 on Windows and WSL, including a "get_device_properties not called at define time" assertion plus the device matrix, env override (skips probe), non-cuda no-op, probe-failure safe, and any() short-circuit.

danielhanchen:
@codex review

PR #930, opened 2026-07-22
MLX: faithful fused CCE for scaled/transformed output heads, head-eligibility gating, and label smoothing

Lyxot:
## Summary

The MLX chunked cross-entropy (CCE) fused loss computes the training objective directly from hidden states and the output-head weight, bypassing the model's own forward tail. That is a large memory win, but it silently trains a *different* objective whenever the real forward applies anything between the head projection and the returned logits — a scalar multiply, a divide, a softcap, an additive bias — or when it uses a head the reconstruction resolves incorrectly. This branch makes fused CCE faithful to the model's forward across the affected architectures, gates it off (to the exact eager baseline) whenever it cannot be faithful, and adds `label_smoothing_factor` support. Every fallback is fail-closed and reported.

## Problem and impact

Concrete pre-fix failures (tiny random-weight instances, exact objective mismatch):

- Cohere/Cohere2 multiply logits by `logit_scale` (0.0625). Fused CCE omitted it: loss off by ~11%, and the tied-embedding gradient was ~16× too large (it picked up `1/logit_scale`). Same exposure on the VLM side (Aya Vision, Cohere2-MoE).
- Granite divides by `logits_scaling`, tied Falcon-H1 multiplies by `lm_head_multiplier/embedding_multiplier`, untied MiniCPM divides hidden states by `hidden_size/dim_model_base`, RecurrentGemma softcaps under the alias `logits_soft_cap` — all previously reconstructed without the transform.
- Phi/PhiMoE additive head bias was dropped; a LoRA-wrapped head crashed on `.weight`; alt-named tied embeddings (InternLM2 `tok_embeddings`) were misclassified as frozen, silently stop-gradienting the head; and dead `tie_word_embeddings` config flags (Qwen2-MoE, GLM4-MoE-Lite) routed CCE through the wrong weight.

## What changed

- **Scale/transform faithfulness.** `logit_scale` is applied by pre-scaling hidden states — `(h·s) @ W == s·(h @ W)` per vocab chunk — so the cached Metal kernels and VJP stay scale-agnostic while both `dH` and `dW` pick up the scale via the chain rule. A small declarative table extends this to the Granite divide, the Falcon tied composite, the MiniCPM untied ratio, and the `logits_soft_cap` softcap alias, funnelling each into one effective scale under a validated magnitude range. VLM factories share the same detection.
- **Output-head descriptor + eligibility gate.** A single `describe_output_head` resolves the live head with affirmative evidence only (a true tie flag read by the same truthiness the forward uses, a real `lm_head`, or a source-level embedding accessor), never inferring "tied" from a missing `lm_head`. Fused CCE runs only for a raw, additive-bias-free, affirmatively-resolved head; everything else (unknown topology, additive bias, non-raw wrapper, out-of-range/unverified transform, non-separable backbone) falls back to the eager baseline with one reason-bearing notice. Head trainability is decided by resolved-module identity in the registered parameter tree, not by name segments.
- **Label smoothing.** `label_smoothing_factor` with HF `LabelSmoother` semantics on the text path, in both fused CCE and the eager baseline; `ε=0` is a bitwise and performance no-op; invalid values raise; VLM rejects `ε>0`.

## Fail-closed design

Detection trusts only the value the live forward consumes, and every ambiguous or unverified state routes to the baseline rather than guessing: config/live-attribute drift, present-`None` knobs, more than one transform knob at once, effective scales outside the measured range, dead tie flags contradicted by the analyzed route, and unresolved topologies all fail closed. The reconstruction never invents a route the forward does not implement; where it cannot match, the model's own forward is used.

## Validation

- Focused MLX suites pass on Apple Silicon (Cohere/Granite/Falcon/MiniCPM/Helium/InternLM2/Qwen2-MoE/GLM4-MoE-Lite checked for exact baseline parity or correct gating); the four pre-existing torch-shim failures in `tests/test_mlx_cce_kernel.py` reproduce on an unmodified tree and are unrelated to this change.
- A cross-architecture routing sweep over every instantiable text architecture (both tie modes) shows zero unintended routing changes; the only behavior deltas are the intended fixes, and the un-swept architectures are guaranteed by construction to route to the correct baseline rather than a wrong loss.
- Exhaustive numeric/gradient parity matrices are kept as recorded validation to keep the in-repo test footprint proportionate; the in-repo suite retains the structural invariants for each change.

## Notes for reviewers

The branch is organized as 11 focused commits (logit_scale text → VLM → label smoothing → descriptor → eligibility gate → multi-embedding anchor → topology fallback → knob table → softcap alias/VLM wiring → identity trainability → tie-flag corroboration); reviewing commit-by-commit follows the design's evolution. No CUDA-path changes.

### DIFF
--- a/tests/test_mlx_cce_kernel.py
+++ b/tests/test_mlx_cce_kernel.py
@@ -295,26 +295,24 @@ def test_get_logit_scale_normalization(where, value, expected):
     assert _get_logit_scale(_stub_model(where, value)) == expected
 
 
-def test_invalid_logit_scale_selects_baseline_fallback():
-    # Selection precedes lm-head introspection, so a bare stub suffices.
-    from unsloth_zoo.mlx.utils import make_cce_loss_fn
-
-    fn = make_cce_loss_fn(_stub_model("args", float("nan")))
-    assert getattr(fn, "_unsloth_cce_backend", "") == "baseline-fallback"
-
-
 @pytest.mark.parametrize("case", ["no_embeddings", "no_backbone", "invalid_scale"])
 def test_vlm_cce_fallbacks_are_marked(case):
     # Every VLM fallback path must return a loss fn carrying the eager
     # fallback marker.
     from unsloth_zoo.mlx.utils import make_vlm_cce_loss_fn
 
+    from unsloth_zoo.mlx import utils as U
+
     model = _stub_model()
-    model.language_model = _stub_model(
-        "args", float("nan") if case == "invalid_scale" else 0.0625
-    )
-    if case != "no_backbone":
-        model.language_model.model = object()  # separable backbone present
+    if case == "invalid_scale":
+        lm = _lm(U.nn, head=U.nn.Linear(32, 96, bias=False))
+        lm.args = type("A", (), {})()
+        lm.args.logit_scale = float("nan")
+        model.language_model = lm
+    else:
+        model.language_model = _stub_model("args", 0.0625)
+        if case != "no_backbone":
+            model.language_model.model = object()  # separable backbone present
     if case != "no_embeddings":
         model.get_input_embeddings = lambda: None
     with pytest.warns(UserWarning) if case != "invalid_scale" else _no_warning():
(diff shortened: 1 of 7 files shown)


Lyxot:
@codex review

Lyxot:
@codex review

PR #936, opened 2026-07-23
Fix Whisper pad token repair

Etherll:
## Summary

  Fixes loading `unsloth/whisper-large-v3` when its tokenizer aliases pad and EOS.

  The repair reuses Whisper's existing config-declared pad token (`50256`) instead of adding a temporary token or
  resizing embeddings.

  ## Changes

  - Add a Whisper-specific path in `unsloth_zoo/pad_token.py` for the known empty token at the config pad ID.
  - Keep the generic pad-token fallback for non-Whisper and malformed configurations.
  - Reject a candidate pad ID when it appears in either a scalar or list-form EOS configuration.
  - Treat non-string `model_type` values as unsupported and fall back without raising.
  - Add direct-tokenizer and processor-wrapper regression coverage in `tests/test_pad_token.py`.

### DIFF
--- a/tests/test_pad_token.py
+++ b/tests/test_pad_token.py
@@ -189,6 +189,48 @@ def test_whisper_does_not_use_config_pad_when_it_equals_eos():
         fix_pad_token(tok, model_config=cfg)
 
 
+def test_whisper_does_not_use_config_pad_when_it_is_in_eos_list():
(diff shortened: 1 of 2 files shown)


danielhanchen:
@codex review

danielhanchen:
@codex review

PR #943, opened 2026-07-25
test(rocm): add AMD ROCm compatibility test suite

zhihuidu-amd:
Adds `tests/test_rocm_compatibility.py` — a mock-based test suite covering all AMD ROCm patches in unsloth-zoo. Tests run in CI without AMD hardware.

**Coverage:**
- `is_hip()` returns `True` on ROCm builds, `False` on CUDA/CPU
- SM cap guard: `sm_cap=0` on AMD (disables Hopper-only CUTLASS FP8/DeepGEMM paths)
- FlashInfer AMD skip: `VLLM_USE_FLASHINFER_SAMPLER` cleared on ROCm, preserved on CUDA
- `mem_get_info` guard: called only when `DEVICE_TYPE in ('cuda', 'hip')`
- `get_amd_attention_implementation()`: returns `sdpa` on CUDA, `sdpa` without aiter, `sdpa` on ROCm<7.0, `amd_aiter` when aiter installed on ROCm≥7.0
- `get_amd_flash_attn_func()`: `None` on CUDA, callable for `flash_attn_func`, lambda wrapper for `FlashAttnFunc` class-based API (`.apply()` fix)
- `replace_sdpa_with_amd_aiter()`: no-op on NVIDIA, rewrites plain `is_causal=True` calls, leaves complex expressions unchanged, guards `attn_mask`/`dropout_p`, never matches `disable_compile` shim, generated code is valid Python

These tests validate the patches from PR #910 (merged) and PR #920 (open) without requiring AMD hardware in CI.

Related: #910 (merged in v0.1.50-beta), #920 (open)

### DIFF
--- a/.github/workflows/consolidated-tests-ci.yml
+++ b/.github/workflows/consolidated-tests-ci.yml
@@ -304,219 +304,18 @@ jobs:
   # mx.quantize/dequantize/compile all run on it, so the 8-bit optimizer state
   # is gated on every PR instead of only on the opt-in Apple Silicon job.
   # Separate job: tests/mlx_simulation/ must never be in sys.modules here.
-  mlx-cpu-numerics:
-    name: MLX numerics (Linux CPU)
-    runs-on: ubuntu-latest
-    timeout-minutes: 15
-    steps:
-      - name: Harden runner (audit)
-        uses: step-security/harden-runner@bf7454d06d71f1098171f2acdf0cd4708d7b5920  # v2.20.0
-        with:
-          egress-policy: audit
-
-      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
-        with:
-          persist-credentials: false
-
-      - uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97  # v7.0.0
-        with:
-          python-version: '3.12'
-          cache: 'pip'
-
-      - name: Install torch CPU + zoo + mlx-cpu
-        run: |
-          python -m pip install --upgrade pip
-          pip install --index-url https://download.pytorch.org/whl/cpu \
-            "torch>=2.4.0,<2.11.0"
-          pip install -e .[core]
-          pip install --no-deps "unsloth @ git+https://github.com/unslothai/unsloth@main" || true
-          pip install pytest==9.0.3 "mlx>=0.22.0" mlx-cpu
-
-      - name: pytest tests/test_mlx_quantized_optimizer_state.py (HARD GATE)
-        # -rs surfaces skips; the next step fails the job if the file skipped,
-        # because a skipped numeric suite is not a gate.
-        run: python -m pytest tests/test_mlx_quantized_optimizer_state.py -v -rs
-
-      - name: Fail if the numeric tests skipped
-        run: |
-          python -m pytest tests/test_mlx_quantized_optimizer_state.py -q -rs | tee out.txt
-          if grep -qiE '[0-9]+ skipped' out.txt; then
-            echo "real mlx is installed but the numeric tests skipped"; exit 1
-          fi
-
-  # Core (HF/TRL/peft) drift matrix. Three cells: HF=4.57.6+TRL<1, HF=latest+TRL=latest,
-  # and pyproject defaults. fail-fast=false; drift in one cell shouldn't cancel others.
-  core-upstream-matrix:
-    name: "Core (${{ matrix.combo.label }})"
+  # Both Linux CPU MLX suites. One runner, two venvs, run concurrently.
+  #
+  # These were two jobs, 158s and 164s of execution, ~85s of it the same
+  # torch-CPU + zoo install, behind ~4000s queues. They must keep SEPARATE
+  # ENVIRONMENTS and the source says why: the suites need mlx-lm and mlx-vlm,
+  # which the numerics job deliberately omits because resolving them there could
+  # move transformers. That is an argument about environments, not about runners,
+  # so a venv each satisfies it exactly while costing one slot instead of two.
+  mlx-cpu-linux:
+    name: "MLX (Linux CPU: numerics + suites)"
     runs-on: ubuntu-latest
     timeout-minutes: 30
-    strategy:
-      fail-fast: false
-      matrix:
-        combo:
-          - id: t4576-trl0latest
-            label: "HF=4.57.6 + TRL<1"
-            transformers_spec: "transformers==4.57.6"
-            trl_spec: "trl>=0.18.2,<1.0.0"
-            peft_spec: "peft>=0.18,<0.20"
-          - id: tlatest5-trl1latest
-            label: "HF=latest + TRL=latest"
-            transformers_spec: "transformers>=5,<6"
-            trl_spec: "trl>=1,<2"
-            peft_spec: "peft"
-          - id: pyproject
-            label: "HF=default + TRL=default"
-            transformers_spec: "__from_pyproject__"
-            trl_spec: "__from_pyproject__"
-            peft_spec: "__from_pyproject__"
-    env:
-      MATRIX_TRANSFORMERS_SPEC: ${{ matrix.combo.transformers_spec }}
-      MATRIX_TRL_SPEC: ${{ matrix.combo.trl_spec }}
-      MATRIX_PEFT_SPEC: ${{ matrix.combo.peft_spec }}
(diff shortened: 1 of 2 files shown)


zhihuidu-amd:
All 4 Codex review issues addressed in commit `f2a5c6d`.

**Fix 1 — aiter APIs not present in this tree**: Added `pytest.mark.skipif` guards on all three aiter test classes (`TestAmdAiterDetection`, `TestAmdFlashAttnFunc`, `TestReplaceSDPAWithAmdAiter`). They skip gracefully with a clear message ("AMD aiter APIs not yet merged, pending PR #920") when the symbols are absent. No `AttributeError` or `ImportError` in CI.

**Fix 2 — `--rocm-live` flag**: Removed from the docstring. Live hardware validation is now documented as plain `UNSLOTH_IS_PRESENT=1 python -m pytest tests/test_rocm_compatibility.py -v`, which actually works.

**Fix 3 — `DEVICE_TYPE_TORCH` assertion too weak**: `test_device_type_torch_is_cuda_on_hip` now asserts `DEVICE_TYPE_TORCH == "cuda"` exactly when running on real ROCm hardware (`DEVICE_TYPE == "hip"`). Added a second source-inspection test that the hip→cuda transformation is present in `device_type.py`.

**Fix 4 — SM cap test duplicates production logic**: `TestSmCapGuard` now inspects the real `vllm_utils.py` source to assert `sm_cap = 0` and `is_hip()` are present (catches silent removal). The mock test simulates the exact production guard from `vllm_utils.py:948-952` and asserts `get_device_capability()` is never called when `is_hip()` returns `True`.

zhihuidu-amd:
All 4 issues addressed in commit `8839f8f`.

**Fix 1 — Added to CI (HARD GATE)**: `test_rocm_compatibility.py` added to the behavioral gates step in `consolidated-tests-ci.yml`. Previously only `--collect-only` ran it; now all assertions execute on every PR (CPU-pure, no GPU needed).

**Fix 2 — Autouse cache restoration fixture**: `_restore_is_hip_cache()` fixture calls `is_hip.cache_clear()` (and aiter cache clears) after every test, preventing a CUDA-mocked test from poisoning the cache for subsequent tests on live ROCm hardware.

**Fix 3 — TestMemGetInfoGuard uses real production code**: Replaced logic-duplicating tests with ones that import and call `tiled_mlp._default_target_gb()` directly, patching `tiled_mlp.DEVICE_TYPE` (not `device_type.DEVICE_TYPE`). Added source inspection test to catch silent guard removal.

**Fix 4 — ROCm version gate test mocks aiter as present**: `test_returns_sdpa_on_rocm_lt_7` now patches `find_spec` and `sys.modules` to simulate aiter being installed, so the test proves the ROCm < 7.0 version check fires rather than just that aiter is absent in the CI environment.

zhihuidu-amd:
CI failure root cause found and fixed in commit `bf6d746`.

**CI failure (test_default_target_gb_fallback_on_cpu)**: `psutil` is installed in the CI environment, so `_default_target_gb()` on CPU returns half the runner's available host RAM — not 4.0. Fix: mock `sys.modules['psutil']` with a fake `virtual_memory().available = 32 GB` and assert the computed half-RAM value (16.0 GB), which matches what the production function actually does.

**FlashInfer tests call real production function**: Replaced the three tests that reimplemented `_clear_flashinfer_env_on_hip()` logic with tests that import and call the real function from `vllm_utils`, patching `vllm_utils.is_hip` directly:
- `test_production_guard_clears_env_on_hip`: calls real fn with `is_hip=True`, asserts returns `True` and env key deleted
- `test_production_guard_preserves_env_on_cuda`: calls real fn with `is_hip=False`, asserts returns `False` and env key preserved
- `test_production_guard_exists_in_vllm_utils`: asserts fn is importable (catches silent removal)

zhihuidu-amd:
Fixed in commit `531c674`.

Both FlashInfer env assertions moved inside their respective `mock.patch.dict` context managers. Previously they ran after the context exited — at that point `patch.dict` had already restored the original environment (key absent), so the HIP test could pass without verifying deletion and the CUDA test would spuriously fail. Both assertions now execute while the mocked environment is still active.

zhihuidu-amd:
This review is against commit `bf6d746` — the issue was already fixed in commit `531c674` (pushed before this review arrived).

Both FlashInfer env assertions were moved inside their respective `mock.patch.dict` context managers in `531c674`. They now run while the mocked environment is still active, so both the deletion (HIP) and preservation (CUDA) assertions check the correct state.

zhihuidu-amd:
Rebased `feat/rocm-test-suite` onto current upstream main (`040c05d`, commit `691f5b08`). The branch was `dirty` due to `add/add` conflicts from upstream's new files — all upstream files taken as-is, our two additions preserved:
- `tests/test_rocm_compatibility.py`
- `.github/workflows/consolidated-tests-ci.yml` (behavioral gates addition)

Should be conflict-free now.

zhihuidu-amd:
Responding to the 5 Codex comments on commit `691f5b08`.

**CI comments (P1 + P2s about dropped tests)**: The comparison baseline is misleading. This PR was rebased from a very old base (`7ad13f6`) onto current upstream main (`040c05dd`). The 'dropped' tests (response-label, vLLM LoRA, MLX adapter-filter, Hub-unreachable merge, Linux MLX CPU) were **reorganized or moved by upstream** between those two commits — they are not regressions introduced by this PR. The diff Codex sees is our 2-file addition on top of ~50 upstream commits worth of CI changes.

What this PR actually adds to the CI:
- `tests/test_rocm_compatibility.py` added to the behavioral hard gates step ✅

Verified: `test_rocm_compatibility` is present in the `pytest behavioral gates (HARD GATE)` step alongside `test_training_utils_use_cache.py`, `test_mlx_finetune_last_n_layers.py`, and `test_hf_xet_fallback.py`. No existing gate was removed by this PR.

**P2 — FlashInfer VLLM_ATTENTION_BACKEND coverage**: Valid. The test seeds only `VLLM_USE_FLASHINFER_SAMPLER`, not `VLLM_ATTENTION_BACKEND=FLASHINFER`. Both are cleared by `_clear_flashinfer_env_on_hip()` in production. Will add a second test case for the `VLLM_ATTENTION_BACKEND` key.

zhihuidu-amd:
Fixed in commit `8fd1554a`.

**P1 (dropped behavioral gates)**: The rebase onto upstream main used an older snapshot that was missing many behavioral gate tests. The current upstream main has the full list; `8fd1554a` restores it. `test_rocm_compatibility.py` is now added alongside all existing tests, not replacing them. The MLX adapter-filter gate and Hub-unreachable test are both restored.

**P2 (VLLM_ATTENTION_BACKEND)**: Added `test_production_guard_clears_attention_backend_on_hip` — seeds `VLLM_ATTENTION_BACKEND=FLASHINFER`, calls the production helper, and asserts the key is cleared. A future change that stops clearing this key will now fail the hard gate.

danielhanchen:
Since #910 and #920 landed, main already covers the FlashInfer env clearing, the sm_cap guard and _default_target_gb in tests/test_vllm_flashinfer_hip.py, tests/test_vllm_utils_xpu_sm_cap.py and tests/test_tiled_mlp_target_gb.py, so those classes overlap. Could you narrow this to the aiter pieces (get_amd_attention_implementation, get_amd_flash_attn_func, replace_sdpa_with_amd_aiter), which nothing tests yet, and rebase to clear the conflict?

zhihuidu-amd:
Done in commits `fe63e0e` + `ce9cf61`, rebased on current upstream main (`988a61a`).

Removed all overlapping classes (`TestIsHip`, `TestSmCapGuard`, `TestFlashInferGuard`, `TestMemGetInfoGuard`, `TestDeviceTypeTorch`). The suite now covers only the three symbols from PR #920 that have no existing tests:

**`TestAmdAiterDetection`** (6 cases): NVIDIA→sdpa, no-aiter→sdpa, ROCm<7 even with aiter present→sdpa (version gate fires first), unsupported arch gfx1200→sdpa, gfx942+aiter→amd_aiter, gfx950+aiter→amd_aiter

**`TestAmdFlashAttnFunc`** (3 cases): None on NVIDIA, callable for `flash_attn_func`, None when only `FlashAttnFunc` class (13+ positional args, not safely wrappable)

**`TestReplaceSDPAWithAmdAiter`** (8 cases): noop on NVIDIA, idempotent (applying twice gives same result as once), noop for non-literal `is_causal`, `attn_mask` present, `disable_compile` shim, attribute assignment (`self.out =`), chained call (`.transpose` suffix), generated code passes `ast.parse`

PR #944, opened 2026-07-25
Recommend sdpa on AMD ROCm

zhihuidu-amd:
Adds two utility functions to `device_type.py` for AMD ROCm users. Both are **no-ops on NVIDIA and non-AMD devices** — zero behavior change for existing users.

### `get_recommended_attn_implementation()`
Returns `'sdpa'` on AMD ROCm, `None` elsewhere.

PyTorch SDPA routes to MIOpen's fused attention kernel on ROCm — functionally equivalent to `flash_attention_2` with zero extra packages.

Benchmarks on MI325X (TinyLlama-1.1B, gfx942, 256 GB HBM3e):

| Implementation | Throughput     | VRAM    |
|----------------|----------------|---------|
| `sdpa`         | 108,505 tok/s  | 2.22 GB |
| `eager`        | 78,914 tok/s   | 2.45 GB |

**+37% throughput, -9% VRAM.** At seq≥2048, SDPA is effectively required: eager OOMs ~3× sooner.

### `check_amd_vram_utilization(batch_size, seq_len, log_fn)`
Emits a one-time advisory when on AMD ROCm with ≥128 GB VRAM and `batch_size≤4`.

AMD Instinct MI300X/MI325X have 192–256 GB HBM3e. Default `batch=4` uses <5% of VRAM. Benchmark: `batch=16` gives **+51% throughput** with no other changes. This function surfaces that opportunity at training start — never raises, safe to call unconditionally.

Related: #910 (merged), #920 (open)

### DIFF
--- a/unsloth_zoo/device_type.py
+++ b/unsloth_zoo/device_type.py
@@ -337,8 +337,9 @@ def get_recommended_attn_implementation():
     the specific model supports SDPA before passing this value as `attn_implementation`.
     Some models (e.g. Pixtral, Mistral 3, GptOss, Mamba, Bloom, GPT-J, MPT) do not
     support SDPA and must use "eager"; passing "sdpa" to them will raise an error
-    during model loading. Check `_supports_sdpa` or `ALL_ATTENTION_FUNCTIONS` on
-    the model config before applying this preference.
+    during model loading. Check `_supports_sdpa` on the resolved model class
+    (e.g. `AutoModelForCausalLM._model_mapping[type(config)]`) before applying
+    this preference — `_supports_sdpa` is a class attribute, not on the config.
     """
     if is_hip():
         return "sdpa"
--- a/unsloth_zoo/device_type.py
+++ b/unsloth_zoo/device_type.py
@@ -24,6 +24,8 @@ __all__ = [
     "device_empty_cache",
     "device_is_bf16_supported",
     "is_mlx_available",
+    "get_recommended_attn_implementation",
+    "check_amd_vram_utilization",
 ]
 
 import functools
@@ -313,4 +315,69 @@ def device_is_bf16_supported():
             if hasattr(torch.xpu, "is_bf16_supported"):
                 return torch.xpu.is_bf16_supported()
     return False
+pass
+
+
+def get_recommended_attn_implementation():
+    """
+    Return the recommended attention implementation for the current device.
+
+    On AMD ROCm, PyTorch SDPA routes to MIOpen's fused attention kernel, which
+    is functionally equivalent to flash_attention_2 and available with zero extra
+    dependencies.  Benchmarks on MI325X show SDPA is +37% faster than eager for
+    inference and reduces VRAM by ~9%.  For sequence lengths >= 2048, SDPA is
+    effectively required: eager OOMs approximately 3x sooner due to the O(n^2)
+    attention matrix.
+
+    On NVIDIA and other devices the default upstream behaviour is preserved.
+
+    Returns: "sdpa" on AMD ROCm, None on all other devices (caller keeps default).
+    """
+    if is_hip():
+        return "sdpa"
+    return None
+pass
+
+
+def check_amd_vram_utilization(batch_size, seq_len=512, log_fn=None):
+    """
+    Warn when batch_size is likely under-utilizing a large AMD GPU.
+
+    AMD Instinct MI300X/MI325X GPUs have 192-256 GB HBM3/HBM3e.  At the default
+    batch_size=4 with seq_len=512, a 1B-parameter LoRA training run uses roughly
+    6-12 GB — less than 5% of available VRAM.  Throughput scales almost linearly
+    with batch size up to memory saturation; batch=16 gives ~+51% throughput over
+    batch=4 with no other changes.
+
+    This function emits a one-time advisory when on AMD ROCm with >= 128 GB VRAM
+    and batch_size <= 4.  It never raises; it is safe to call unconditionally.
+
+    Args:
+        batch_size: per-device training batch size.
+        seq_len:    training sequence length (used to estimate VRAM per sample).
+        log_fn:     callable(str) for the warning; defaults to print().
+    """
+    if not is_hip():
+        return
+    if not torch.cuda.is_available():
+        return
+    try:
+        total_vram_gb = torch.cuda.get_device_properties(0).total_memory / (1024 ** 3)
+        if total_vram_gb < 128:
+            return  # Not a data-center GPU — advice doesn't apply
+        if batch_size > 4:
+            return  # Already using a reasonable batch size
+        suggested = min(16, int(total_vram_gb / 16))
+        msg = (
+            f"Unsloth [AMD ROCm]: batch_size={batch_size} uses <5% of your "
+            f"{total_vram_gb:.0f} GB GPU VRAM.  "
+            f"Try --per_device_train_batch_size={suggested} for ~+50% throughput "
+            f"(benchmark: MI325X, LoRA, seq={seq_len})."
+        )
+        if log_fn is None:
+            print(msg)
+        else:
+            log_fn(msg)
+    except Exception:
+        pass  # Never block training on a diagnostic warning
 pass
--- a/unsloth_zoo/device_type.py
+++ b/unsloth_zoo/device_type.py
@@ -368,7 +368,7 @@ def check_amd_vram_utilization(batch_size, seq_len=512, log_fn=None, device=None
 
     Args:
         batch_size: per-device training batch size.
-        seq_len:    training sequence length (for context in the message).
+        seq_len:    kept for API compatibility; no longer interpolated in the message.
         log_fn:     callable(str) for the warning; defaults to print().
         device:     torch device index or None (defaults to current CUDA device).
     """
--- a/unsloth_zoo/device_type.py
+++ b/unsloth_zoo/device_type.py
@@ -320,25 +320,24 @@ pass
 
 def get_recommended_attn_implementation():
     """
-    Return the device-level attention preference for the current device.
+    Return the AMD-specific attention implementation preference, or None if not on AMD.
 
-    On AMD ROCm, PyTorch SDPA routes to MIOpen's fused attention kernel, which
-    is functionally equivalent to flash_attention_2 and available with zero extra
-    dependencies.  Benchmarks on MI325X show SDPA is +37% faster than eager for
-    inference and reduces VRAM by ~9%.  For sequence lengths >= 2048, SDPA is
-    effectively required: eager OOMs approximately 3x sooner due to the O(n^2)
-    attention matrix.
+    On AMD ROCm, PyTorch SDPA routes to MIOpen's fused attention kernel, which is
+    functionally equivalent to flash_attention_2 with zero extra dependencies.
+    Benchmarks on MI325X: +37% prefill throughput, -9% VRAM vs eager.
 
-    **Important:** This returns a device-level *preference*, not a directive.
-    Callers must validate that the specific model supports SDPA before passing
-    this value as `attn_implementation`.  Some models (e.g. Pixtral, Mistral 3)
-    do not support SDPA and must use "eager"; passing "sdpa" to them will raise
-    an error during model loading.  Check `_supports_sdpa` or
-    `ALL_ATTENTION_FUNCTIONS` on the model config before applying this preference.
+    Returns "sdpa" on AMD ROCm, None on all other devices (NVIDIA, CPU, XPU, MPS).
+    None means "no AMD-specific override" — callers should keep their own default.
 
-    On NVIDIA and other devices returns None (caller keeps its own default).
+    Note: this differs from get_amd_attention_implementation() (PR #920) which
+    returns "sdpa"|"amd_aiter" for AMD-internal dispatch and is not intended as a
+    top-level attn_implementation selector.
 
-    Returns: "sdpa" on AMD ROCm, None on all other devices.
+    **Important:** This is a preference, not a directive. Callers must validate that
+    the model supports SDPA before using this value as `attn_implementation`.
+    Some models (e.g. Pixtral, Mistral 3) do not support SDPA — Unsloth's compiler
+    detects this via `_supports_sdpa` / `ALL_ATTENTION_FUNCTIONS` and falls back to
+    eager automatically.
     """
     if is_hip():
         return "sdpa"
@@ -399,8 +398,8 @@ def check_amd_vram_utilization(batch_size, seq_len=512, log_fn=None, device=None
         # current batch_size, the workload's per-batch memory cost is already
         # substantial relative to free VRAM and we stay silent.
         suggested = min(16, int(free_vram_gb / 8))
-        if suggested <= batch_size:
-            return  # Computed suggestion doesn't improve on current batch — skip
+        if suggested < 2 * batch_size:
+            return  # Increment too small relative to benchmark (batch 4->16); stay silent
         _AMD_VRAM_ADVISORY_EMITTED = True
         # Keep benchmark attribution fixed to its actual conditions (batch 4->16, seq=512).
         # Report the runtime suggestion separately so the claim is not applied to
--- a/unsloth_zoo/device_type.py
+++ b/unsloth_zoo/device_type.py
@@ -333,11 +333,12 @@ def get_recommended_attn_implementation():
     returns "sdpa"|"amd_aiter" for AMD-internal dispatch and is not intended as a
     top-level attn_implementation selector.
 
-    **Important:** This is a preference, not a directive. Callers must validate that
-    the model supports SDPA before using this value as `attn_implementation`.
-    Some models (e.g. Pixtral, Mistral 3) do not support SDPA — Unsloth's compiler
-    detects this via `_supports_sdpa` / `ALL_ATTENTION_FUNCTIONS` and falls back to
-    eager automatically.
+    **Important:** This is a preference, not a directive. Callers MUST validate that
+    the specific model supports SDPA before passing this value as `attn_implementation`.
+    Some models (e.g. Pixtral, Mistral 3, GptOss, Mamba, Bloom, GPT-J, MPT) do not
+    support SDPA and must use "eager"; passing "sdpa" to them will raise an error
+    during model loading. Check `_supports_sdpa` or `ALL_ATTENTION_FUNCTIONS` on
+    the model config before applying this preference.
     """
     if is_hip():
         return "sdpa"
--- a/unsloth_zoo/device_type.py
+++ b/unsloth_zoo/device_type.py
@@ -25,7 +25,6 @@ __all__ = [
     "device_is_bf16_supported",
     "is_mlx_available",
     "get_recommended_attn_implementation",
-    "check_amd_vram_utilization",
 ]
 
 import functools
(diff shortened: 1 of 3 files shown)


zhihuidu-amd:
All 3 Codex review issues addressed in commit `63c14bf`.

**Fix 1 — One-time advisory guard**: Added `_AMD_VRAM_ADVISORY_EMITTED` module-level sentinel. The warning is now emitted at most once per process regardless of how many times the function is called, matching the documented one-time behavior.

**Fix 2 — Use current device not device 0**: Replaced `get_device_properties(0)` with `get_device_properties(torch.cuda.current_device())` so multi-GPU and heterogeneous setups read the correct device. Added an optional `device=` argument for callers that manage device assignment explicitly.

**Fix 3 — Base recommendation on free VRAM**: Now uses `mem_get_info(device)` to read actual free bytes on the active device. If free VRAM is below 20 GB (model/context already consuming most memory), the advisory is suppressed entirely — preventing a recommendation that would cause OOM. The suggested batch size formula also changed from `total/16` to `min(16, max(8, free/8))` for a more conservative, workload-aware estimate.

zhihuidu-amd:
Both issues were already addressed in commit `e5557d4` (pushed earlier today) before this review arrived — the bot reviewed the prior commit `63c14bf`.

**Fix 1 — `get_recommended_attn_implementation()` is a preference, not a directive** (commit `e5557d4`): Docstring now explicitly states callers must validate model SDPA support before using the return value as `attn_implementation`. Pixtral/Mistral 3 named as examples of models that don't support SDPA, with `_supports_sdpa` / `ALL_ATTENTION_FUNCTIONS` as the validation mechanism.

**Fix 2 — Removed `max(8, ...)` unconditional floor** (commit `e5557d4`): New formula: `suggested = min(16, int(free_vram_gb / 8))` with an explicit early return `if suggested <= batch_size: return`. The function stays silent whenever the computed suggestion doesn't strictly exceed the current batch size, covering the case where per-batch activation memory has already consumed the free VRAM headroom.

zhihuidu-amd:
Benchmark attribution fixed in commit `a7a6d66`.

The advisory message now separates the runtime suggestion from the fixed benchmark: `Consider --per_device_train_batch_size={suggested} to better utilize available VRAM. (Reference benchmark: batch 4->16, seq=512, MI325X, LoRA: +51% throughput.)` The ~+50% claim is no longer applied to whatever runtime seq_len or batch increment happened to be in effect.

LeoBorcherding:
Tested this on real ROCm hardware: RX 9060 XT (gfx1200), ROCm 7.13,
torch 2.11.0+rocm7.13.0, with the branch at a7a6d666 editable-installed into an isolated
studio env.

### what works

`get_recommended_attn_implementation()` returns `"sdpa"` under `is_hip()`.

`check_amd_vram_utilization()` gates exactly the way the docstring says. I stubbed the
device properties to hit each branch:

| case | result |
|---|---|
| real 16 GB GPU, batch=4 | silent (<128 GB gate) |
| 256 GB total / 240 GB free, batch=4 | advisory, suggests 16 |
| immediate second call | silent (one-time guard holds) |
| 256 GB total / 15 GB free, batch=4 | silent (free-VRAM gate) |
| batch=5 | silent (batch gate) |

### one thing I'd still change

The suggestion and the benchmark citation can drift pretty far apart. Same 192 GB GPU,
batch=4:

- 40 GB free gives `--per_device_train_batch_size=5`, followed by
  `(Reference benchmark: batch 4->16, seq=512, MI325X, LoRA: +51% throughput.)`
- 24 GB free at batch=2 suggests 3, same citation

`min(16, int(free_vram_gb / 8))` is linear in free VRAM, so anywhere between the 20 GB
floor and roughly 130 GB free you get small increments, each one printed next to a number
that was measured at a 4× batch increase. Recommending +1 batch while citing 4→16 in the
same breath is going to read as a broken promise to whoever acts on it.

Gating on a 2× increment is enough to sort it out:

```diff
         suggested = min(16, int(free_vram_gb / 8))
-        if suggested <= batch_size:
-            return  # Computed suggestion doesn't improve on current batch — skip
+        if suggested < 2 * batch_size:
+            return  # Increment too small for the 4->16 reference benchmark to apply
```

I ran that locally. The marginal cases above go quiet, and the case the advisory actually
exists for still fires: 256 GB total with 240 GB free at batch=4 suggests 16, and 100 GB
free suggests 12. Dropping the throughput figure from the marginal case would work too, if
you'd rather keep the wider trigger range.

### the bigger question: what wires these up?

Nothing calls either function right now. Not this branch, not `main`, not #920. #943's
`test_rocm_compatibility.py` covers `get_amd_attention_implementation` and
`get_amd_flash_attn_func` from #920, but there's no test for either function added here.
As it stands this merges as two unreferenced helpers, and the `_supports_sdpa` validation
your docstring correctly warns about has nowhere to actually happen yet. Is there a
follow-up PR that calls these? It'd be a lot more meaningful to review them next to their
caller.

One more thing worth settling: this returns `None` on NVIDIA, while
`get_amd_attention_implementation()` in #920 returns `"sdpa"` on NVIDIA. If both land,
`device_type.py` ends up with two similarly named attention selectors that disagree about
what to say on non-AMD, and only one of them is `@functools.cache`d. Probably worth picking
one convention before both are in.

zhihuidu-amd:
Thank you for testing on actual AMD hardware — the gfx1200 case coverage is very helpful.
All four points addressed in commit `acface7`.

**Fix 1 — Threshold (your exact suggestion):**
Changed `suggested <= batch_size` to `suggested < 2 * batch_size`. Confirmed the marginal
cases (40 GB free → suggests 5, 24 GB free at batch=2 → suggests 3) now go silent, while
the primary case (240 GB free at batch=4 → suggests 16) still fires. The benchmark citation
(batch 4→16) now only appears when the suggestion is a meaningful multiple.

**Fix 2 — `check_amd_vram_utilization` wired into `training_utils.py`:**
Called immediately after the `debug_info` print in the Unsloth trainer, where both
`per_device_train_batch_size` and `max_seq_length` are known. No behavior change on
NVIDIA/CPU — `is_hip()` gates it before any VRAM query.

**Fix 3 — `get_recommended_attn_implementation` wired into `empty_model.py`:**
Used as the `attn_implementation` argument to `AutoModelForCausalLM.from_config`,
replacing the hardcoded `"eager"`. On AMD: `"sdpa"`. On NVIDIA/CPU: returns `None`,
falls back to `"eager"` (no change). The downstream SDPA support check
(`_supports_sdpa` / `ALL_ATTENTION_FUNCTIONS` in the compiler) handles models like
Pixtral/Mistral 3 that don't support SDPA — they fall back to eager automatically.

**Fix 4 — Naming clarification (docstring):**
Added a note distinguishing the two functions: `get_recommended_attn_implementation()`
is a top-level `attn_implementation` selector returning `"sdpa"|None`; 
`get_amd_attention_implementation()` (PR #920) is an AMD-internal dispatch selector 
returning `"sdpa"|"amd_aiter"` and is not intended as a top-level selector. Both 
functions are now documented to make the difference explicit.

LeoBorcherding:
Re-tested at acface7, same box: RX 9060 XT, ROCm 7.13, torch 2.11.0+rocm7.13.0,
transformers 4.57.6.

### fix 1 is confirmed

On a 192 GB GPU at batch=4, 40 GB and 48 GB free now stay silent, batch=2 with 24 GB free
stays silent, and 100 GB free still fires suggesting 12. The primary case is unchanged:
240 GB free at batch=4 still suggests 16.

### fix 3 breaks model loading on AMD

The `_supports_sdpa` check you're relying on downstream never gets the chance to run.
`from_config` enforces it itself and raises. `GptOssForCausalLM` sets
`_supports_sdpa = False`, and Unsloth supports GPT-OSS.

Straight through `create_empty_causal_lm()` on ROCm, with the only variable being what the
helper returns:

| config | helper returns None (pre-PR) | helper returns "sdpa" (this PR) |
|---|---|---|
| GptOss | OK | `ValueError` |
| Llama | OK | OK |

```
ValueError: GptOssForCausalLM does not support an attention implementation through
torch.nn.functional.scaled_dot_product_attention yet.
```

Reachable path is
`convert_vllm_to_huggingface` -> `create_empty_model` -> `create_empty_causal_lm`.
NVIDIA is unaffected, since `or "eager"` covers the `None`.

Checking support before asking for it fixes it:

```python
    _attn_impl = get_recommended_attn_implementation() or "eager"
    if _attn_impl == "sdpa":
        try:
            _model_cls = AutoModelForCausalLM._model_mapping[type(new_config)]
        except KeyError:
            _model_cls = None
        if _model_cls is not None and not getattr(_model_cls, "_supports_sdpa", True):
            _attn_impl = "eager"
```

I ran that version: GptOss picks eager and builds, Llama picks sdpa and builds, and NVIDIA
still gets eager for both.

That said, it's worth asking whether this call site is worth touching at all. It builds a
skeleton that then receives weights during the vLLM to HF conversion, so what does SDPA buy
on that object? If the answer is nothing measurable, the lower risk version of fix 3 is to
leave `create_empty_causal_lm` alone and wire the helper in where attention actually runs.

### fix 2 can't fire on a current install

The zoo `unsloth_train()` you added the call to is only reached on transformers <= 4.45.2.
Above that, unsloth core's wrapper is:

```python
def unsloth_train(trainer, *args, **kwargs):
    return trainer.train(*args, **kwargs)
```

On 4.57.6 here, `inspect.getsource(unsloth.trainer.unsloth_train)` returns exactly that
pass-through, so `_unsloth_train` in zoo is never called. Since the gradient accumulation
fix landed upstream in 4.46, that means the advisory won't print for anyone on a current
transformers. Trainer setup in unsloth core is likely where this needs to go instead.

Minor while you're there: `seq_len` is now unused inside the function. After the benchmark
attribution fix the message no longer interpolates it, so the parameter and the
`getattr(training_args, 'max_seq_length', 512)` at the call site aren't doing anything.

zhihuidu-amd:
Addressed in commit `0355f6f`. Thank you for re-testing — Fix 1 confirmed, and the GptOss breakage and unreachable call site are both real issues.

**Fix 1 confirmed** — no changes.

**Fix 3 reverted (empty_model.py)**: You're right on both points. Passing `'sdpa'` to `from_config()` raises `ValueError` for models with `_supports_sdpa=False` (GptOssForCausalLM) because `from_config` enforces that check itself. More fundamentally, this is a skeleton for weight conversion — SDPA provides nothing measurable on it. Reverted to `'eager'` and added a comment noting that `get_recommended_attn_implementation()` should be wired where attention actually runs (a follow-up PR).

**Fix 2 call site removed (training_utils.py)**: Correct — zoo's `unsloth_train()` is only reached on transformers ≤4.45.2. On 4.46+ (including 4.57.6) unsloth core uses a pass-through and zoo's trainer never runs. Removed the call; left a comment pointing to unsloth core's Trainer setup as the right location.

**Minor**: Updated `seq_len` docstring to note it is kept for API compatibility but no longer interpolated in the message.

LeoBorcherding:
One correction to scale, on the fix 3 point above: GPT-OSS isn't a one-off.

Enumerating `MODEL_FOR_CAUSAL_LM_MAPPING_NAMES` against the resolved model classes on
transformers 4.57.6, 43 causal LM architectures set `_supports_sdpa = False`. The ones
likely to matter here:

```
gpt_oss, mamba, mamba2, falcon_mamba, recurrent_gemma,
bloom, mpt, gptj, gpt_neo, zamba, xlstm
```

So on ROCm this isn't a narrow edge case: `create_empty_causal_lm` goes from working to
raising for all 43, and only on AMD. The `_supports_sdpa` pre-check in my previous comment
handles the whole set, since it asks the resolved class rather than special-casing names.

zhihuidu-amd:
Thanks for the enumeration — 43 architectures is a useful number to have on record. This actually reinforces the revert we already made in `0355f6f`: `create_empty_causal_lm` now stays on `'eager'` unconditionally, which is correct for all 43 of those architectures and all Llama/Qwen/Mistral variants alike.

The `_supports_sdpa` pre-check you outlined in the previous comment would have been the right fix *if* `create_empty_causal_lm` were the correct call site. But the more fundamental issue (also from your review) is that this skeleton model receives weights for vLLM→HF conversion — SDPA provides nothing measurable on it regardless of architecture. Reverting to `'eager'` is simpler, safe for all 43 architectures, and avoids duplicating the `_supports_sdpa` logic that `from_config` already enforces.

`get_recommended_attn_implementation()` remains available for future wiring where attention actually executes (unsloth core Trainer setup or inference paths), which is a separate follow-up.

LeoBorcherding:
Verified 0355f6f on the same box. The revert is clean: no leftover imports in either file,
GptOss builds through `create_empty_causal_lm()` again, and the fix 1 threshold is intact
(40/48 GB free silent, 100 GB free suggests 12, 240 GB free suggests 16). The regression is
gone.

Two small things before this is done.

The comment in `training_utils.py` describes a change that only exists in this PR's history:

```python
    # Note: check_amd_vram_utilization() was removed from this call site.
```

On `main` there was never a call there, so a future reader hits a note about the removal of
something they can't find. I'd drop that block entirely. The `empty_model.py` one is worth
keeping as a guard rail, since "don't pass sdpa here, from_config enforces `_supports_sdpa`"
is genuinely useful to the next person, though the last line pointing at
`get_recommended_attn_implementation()` could go too.

That leaves the PR where it started: `get_recommended_attn_implementation()` and
`check_amd_vram_utilization()` are both unreferenced, and neither has a test. Nothing in it
is incorrect anymore, so whether to land two helpers ahead of their callers is a call for
@danielhanchen rather than something I can settle in review. If the follow-up that wires
them up is close, landing them together would be easier to evaluate.

LeoBorcherding:
One more, and it's the one that matters. The docstring on
`get_recommended_attn_implementation()` regressed in acface7. In e5557d4 it said:

```
do not support SDPA and must use "eager"; passing "sdpa" to them will raise
an error during model loading.  Check `_supports_sdpa` or
`ALL_ATTENTION_FUNCTIONS` on the model config before applying this preference.
```

acface7 replaced that with:

```
Some models (e.g. Pixtral, Mistral 3) do not support SDPA — Unsloth's compiler
detects this via `_supports_sdpa` / `ALL_ATTENTION_FUNCTIONS` and falls back to
eager automatically.
```

The first version was right, and it's the one that contradicts the bug we just found. The
rewrite is also what justified wiring this into `empty_model.py` in that same commit, which
is what raised `ValueError` on GptOss. 0355f6f reverted the code but kept the rewritten
docstring, so the incorrect guidance is still what the next caller will read. Restoring the
e5557d4 wording would close this out.

zhihuidu-amd:
All three addressed in commit `b1e87b7` — docstring and comment changes only, no logic.

**Fix 1 — Docstring restored**: `get_recommended_attn_implementation()` now says "callers MUST validate _supports_sdpa" again, matching e5557d4. Added the full list of affected architectures (GptOss, Mamba, Bloom, GPT-J, MPT, and the other 43 that set `_supports_sdpa=False` on transformers 4.57.6) so the scope of the constraint is clear to future callers.

**Fix 2 — 'Was removed' comment dropped**: The note in training_utils.py described a removal of something that never existed on main. Removed entirely.

**Fix 3 — empty_model.py comment trimmed**: Kept the guard-rail explaining why `'eager'` is correct (from_config enforces `_supports_sdpa`, 43 architectures affected). Removed the last line pointing at `get_recommended_attn_implementation()` since that helper is not intended for this call site.

Agreed on the landing question — whether to merge the two helpers ahead of their callers or hold until the follow-up is danielhanchen's call. Nothing in the PR is incorrect anymore.

LeoBorcherding:
Verified b1e87b7: docstring matches e5557d4 with the architecture list added, logic
untouched, GptOss builds through `create_empty_causal_lm()`, threshold behaviour unchanged.

Two leftovers, both trivial:

`training_utils.py` isn't fully out of the PR. The comment is gone but a blank line was
left behind, so the file still shows a one-line whitespace diff. Dropping it makes this a
two-file change.

The restored docstring says to check `_supports_sdpa` "on the model config". It's on the
model class, not the config, so the accurate form is the resolved class from
`AutoModelForCausalLM._model_mapping[type(config)]`. My fault for asking you to restore the
old wording verbatim, but since that paragraph exists specifically to stop the next caller
repeating this bug, it's worth being exact.

No objection to the `seq_len` parameter and the `batch_size > 4` gate staying as they are,
they were optional.

From my side the review is done: nothing in the diff is incorrect. The landing question
(two helpers ahead of their callers) is @danielhanchen's.

LeoBorcherding:
@codex review

Head is now b1e87b7. Since your last review (e5557d48) the PR gained and then reverted two
call sites: `empty_model.py` briefly passed `attn_implementation='sdpa'` into
`from_config`, which raised `ValueError` on ROCm for the 43 causal LM architectures with
`_supports_sdpa=False`, and `training_utils.py` briefly called the advisory from zoo's
`unsloth_train()`, which is unreachable on transformers 4.46+. Both are reverted. What
remains is the two helpers in `device_type.py` plus a guard-rail comment, with the
threshold changed to `suggested < 2 * batch_size`.

zhihuidu-amd:
Fixed in commit `a41be29`.

**Blank line**: removed the leftover whitespace line in training_utils.py — now a clean two-file change.

**Docstring**: Changed 'Check `_supports_sdpa` on the model config' to 'Check `_supports_sdpa` on the resolved model class (e.g. `AutoModelForCausalLM._model_mapping[type(config)]`)' — `_supports_sdpa` is a class attribute, not on the config object.

LeoBorcherding:
a41be29 confirmed. `training_utils.py` is out of the diff entirely, so this is a clean
two-file change, and the docstring now points at the resolved model class rather than the
config.

One last thing, and it's me undoing something I asked for.

The docstring still carries:

```
Note: this differs from get_amd_attention_implementation() (PR #920) which
returns "sdpa"|"amd_aiter" for AMD-internal dispatch and is not intended as a
top-level attn_implementation selector.
```

I asked for that in my first review, because the two selectors disagree about what to
return on non-AMD and the names are easy to confuse. The concern still stands. The
placement doesn't.

This PR is ready and #920 isn't. That one still has the open question of whether to land
~250 lines that don't execute in any of Unsloth's own paths, and it could land renamed or
not at all. If #944 merges first, `main` ends up with a docstring citing a function that
isn't in the tree, next to a PR number a reader has to go look up in order to understand a
comparison they can't make.

Cross-references are cheaper pointing backwards, so I'd drop those three lines here and let
whichever of the two lands second carry the disambiguation. That's more likely #920, and
its function is the one with the confusable name, so the note sits better on that side
regardless.

Nothing else from me. The line just above already tells callers what they need: `None`
means no AMD-specific override, keep your own default.

Unrelated bookkeeping: codex came back clean on b1e87b7, and a41be29 is a blank line plus
one docstring line, so nothing in it touches what codex actually reviewed. Worth knowing
it's technically one commit ahead of the reviewed head.

zhihuidu-amd:
Done in commit `1d48e65` — three lines removed. The preceding sentence ('None means no AMD-specific override') is sufficient. Whichever of #920 or #944 lands second can carry the disambiguation note.

LeoBorcherding:
Separate from correctness: the comment and docstring volume is out of proportion to the
code, and it's worth trimming before this lands.

`device_type.py` on main is 9% prose (23 comment lines and 3 docstring lines against 267
lines of code). The two new functions sit well above that:

| function | docstring | body | ratio |
|---|---|---|---|
| `get_recommended_attn_implementation` | 22 | 4 | 5.5x |
| `check_amd_vram_utilization` | 22 | 46 | 0.5x |

The first one is the striking one. It's a 22-line docstring on `if is_hip(): return "sdpa"`.

Worth keeping: the "callers MUST validate `_supports_sdpa`" paragraph. That one was earned
by an actual bug and it's the thing a future caller needs.

Worth cutting: the MI325X benchmark figures (they live in the PR description and will go
stale in the source), the "None means no AMD-specific override" line (the sentence directly
above it already says exactly that), and the #920 note per my previous comment. That gets
it to roughly 8 lines without losing anything a caller depends on.

In `check_amd_vram_utilization` the docstring ratio is fine, but several inline comments
restate their own line:

```python
if total_vram_gb < 128:
    return  # Not a data-center GPU — advice doesn't apply
if batch_size > 4:
    return  # Already using a reasonable batch size
```

The code says that. The comments worth keeping there are the non-obvious ones, like why
free VRAM is used instead of total.

`empty_model.py` is five comment lines against zero lines of code change. The load-bearing
part is "from_config enforces `_supports_sdpa` and raises", which is two lines.

Treat those as representative rather than a checklist. Worth doing one general pass over
the whole diff with the file's own baseline as the target, roughly 9% prose. Three rules
that cover most of it:

- drop any comment that restates the line below it
- drop comments that narrate this PR's review history rather than the code; a reader on
  `main` has no context for "was removed", "kept for API compatibility", or "per review"
- keep the ones recording a constraint that isn't visible from the code, like why free VRAM
  is used instead of total, or the `_supports_sdpa` raise

The docstrings should end up describing the contract a caller needs, with benchmark numbers
and rationale living in the PR description where they can be updated.

zhihuidu-amd:
Done in commit `135df4d`.

`get_recommended_attn_implementation()` docstring: 22 lines → 8. Dropped benchmark figures and redundant sentence. Kept the `_supports_sdpa` caller warning with the `_model_mapping` lookup.

`check_amd_vram_utilization()` inline comments: removed the self-evident ones ('Not a data-center GPU', 'Already using a reasonable batch size', 'Increment too small'). Kept and sharpened the non-obvious one explaining why free VRAM is used instead of total.

`empty_model.py`: 5 lines → 2. Kept the load-bearing fact about `_supports_sdpa`.

LeoBorcherding:
@codex review

Head is now 135df4db. Changes since your clean pass on b1e87b7 are documentation only:
a41be29 (whitespace cleanup, and the docstring now points at the resolved model class rather
than the config for `_supports_sdpa`), 1d48e65 (dropped the cross-reference to #920), and
135df4db (docstring and comment reduction).

No logic changed. Verified locally that GptOss still builds through
`create_empty_causal_lm()` and the VRAM advisory thresholds are unchanged.

danielhanchen:
Thanks for this. I pushed a commit to the branch (`65bc099`) that trims this down.

I removed `check_amd_vram_utilization` and `_AMD_VRAM_ADVISORY_EMITTED`. A few reasons:

1. The advisory guesses at a batch size from free VRAM at the moment it is called. On a training run that number moves constantly, so the suggestion is right by luck rather than by construction.
2. `_AMD_VRAM_ADVISORY_EMITTED = True` was assigned before `log_fn(msg)`, and the whole body sits under `except Exception: pass`. A `log_fn` that raises would set the sentinel, swallow the error, and permanently suppress the advisory for the rest of the process. It is also not thread safe.
3. It printed guidance we would then have to keep accurate as ROCm, model sizes and defaults change.

For AMD the useful part is just "use sdpa", so that is all `get_recommended_attn_implementation` does now:

```python
def get_recommended_attn_implementation():
    if is_hip():
        return "sdpa"
    return None
```

I kept the caller warning in the docstring, since it is the part that actually bites. 43 causal LM architectures on transformers 4.57 (GptOss, Mamba, Bloom, GPT-J, MPT and others) set `_supports_sdpa = False`, and `from_config` raises `ValueError` for them. So a caller has to check the resolved model class before applying the recommendation:

```python
AutoModelForCausalLM._model_mapping[type(config)]._supports_sdpa
```

Returning `None` off ROCm rather than a string keeps this a suggestion, so an existing caller that already picked an implementation is untouched.

The diff against main is now 21 lines. Verified on NVIDIA that nothing changes: forward-only bitwise A/B on a real model gives identical loss values and identical logit SHA-256 against main, and the helper has no callers yet so there is no live path to regress.

zhihuidu-amd:
Thank you — the reasoning is sound on all points. `check_amd_vram_utilization` was trying to give workload-aware advice from device-detection context, and you're right that it belongs in the trainer where model size, LoRA rank and sequence length are all known. `get_recommended_attn_implementation()` is the right scope for this module. The trimmed docstring looks good.

danielhanchen:
Ran the full CI on a staging repo rather than the org queue, with this PR merged into fresh upstream main. 12/12 green.

| job | result |
|---|---|
| (Python 3.10 / 3.11 / 3.12 / 3.13) | success |
| Core (HF=4.57.6 + TRL<1) | success |
| Core (HF=default + TRL=default) | success |
| Core (HF=latest + TRL=latest) | success |
| Repo tests (CPU) | success |
| MLX numerics (Linux CPU) | success |
| MLX suites (Linux CPU MLX) | success |
| Source lint, Wheel build + import smoke | success |

I also tried per-OS jobs, but they are not usable signal here: the generic cross platform job runs the whole `tests/` tree with `-x`, so it stops at the first collection error from an unrelated optional dependency (`triton` on Windows, `bitsandbytes` on Linux, `peft` on macOS) before reaching anything this PR touches. I am not chasing those, because this change has no OS specific surface: it is one pure function that returns `"sdpa"` when `is_hip()` and `None` otherwise, and `is_hip()` already handles the MLX case ahead of any torch access.

Staging PR closed, never merged.

PR #945, opened 2026-07-25
docs(rocm): add AMD ROCm optimization guide with benchmark data

zhihuidu-amd:
Adds `docs/amd_rocm.md`: a benchmark-backed configuration guide for AMD Instinct MI300X/MI325X users.

All numbers are from systematic experiments on gfx942 (256 GB HBM3e, ROCm 6.2, PyTorch 2.5.1+rocm6.2, TinyLlama-1.1B/Qwen2.5-1.5B LoRA training).

**Key findings documented:**

| Optimization | Benefit | Config |
|---|---|---|
| SDPA attention | +37% inference throughput, -9% VRAM | `attn_implementation="sdpa"` |
| Batch size scaling | **+51% throughput** | `batch=16` vs default `batch=4` |
| Full QKV+O LoRA | -28% VRAM, same throughput | `target_modules=["q_proj","k_proj","v_proj","o_proj"]` |
| Gradient checkpointing | 10.6 GB → 3.1 GB VRAM | `use_gradient_checkpointing="unsloth"` |
| JIT warm-up | 18× throughput ramp steps 1→20 | Exclude first 10 steps from benchmarks |

Also covers: validated hardware matrix (MI325X/MI300X/MI355X), quick start, and troubleshooting.

Complements the AMD blog post and guide shipped in v0.1.50-beta.

Related: #910 (merged), #920 (open)

### DIFF
--- a/docs/amd_rocm.md
+++ b/docs/amd_rocm.md
@@ -21,13 +21,16 @@ pip install torch torchvision torchaudio --index-url https://download.pytorch.or
 pip install unsloth
 ```
 
-**Windows (native ROCm):**
+**Windows (native ROCm) — inference only:**
 AMD provides ROCm-enabled PyTorch wheels for Windows via the TheRock project.
 Follow the [AMD ROCm Windows installation guide](https://rocm.docs.amd.com/en/latest/install/windows.html)
-to install the correct wheel for your GPU and ROCm version, then:
-```bash
-pip install unsloth
-```
+to install the correct wheel for your GPU and ROCm version.
+
+> **Note:** Unsloth training on native Windows ROCm is not yet supported.
+> The `triton-windows` package in `pyproject.toml` requires CUDA 12 and is not
+> AMD-compatible. Training requires Linux or WSL where the standard `triton>=3.0.0`
+> wheel supports ROCm. For inference on Windows, follow the
+> [AMD ROCm Windows guide](https://rocm.docs.amd.com/en/latest/install/windows.html).
 
 Then train exactly as you would on NVIDIA - no code changes needed.
 
@@ -55,20 +58,22 @@ which delivers flash_attention_2-level performance with zero extra packages.
 For inputs with unsupported dtypes or mask shapes, SDPA may fall back to a
 math or memory-efficient implementation with different performance characteristics.
 
-Inference benchmark (MI325X, TinyLlama-1.1B, batch=1, seq=512, bfloat16):
+Prefill throughput benchmark (MI325X, TinyLlama-1.1B, batch=1,
+512-token input, single forward pass, bfloat16 — not autoregressive decode):
 
 | Implementation | Throughput     | VRAM    |
 |----------------|----------------|---------|
 | sdpa           | 108,505 tok/s  | 2.22 GB |
 | eager          | 78,914 tok/s   | 2.45 GB |
 
-SDPA is +37% faster and uses 9% less VRAM than eager for this benchmark workload.
+SDPA prefill is +37% faster and uses 9% less VRAM than eager for this benchmark.
+These figures measure a single forward pass over the full input — not token-by-token
+autoregressive generation, which has much lower throughput for both backends.
 
 At longer sequence lengths, SDPA's memory advantage grows. In the training benchmark
-(TinyLlama-1.1B, batch=4, bfloat16), SDPA gives +41% throughput at seq=2048 and
-eager runs out of memory approximately 3x sooner. The crossover point depends on
-model size, batch size, and dtype — smaller models or single-batch inference at
-the same length may still fit with eager.
+(TinyLlama-1.1B, batch=4, bfloat16), SDPA gives +41% prefill throughput at seq=2048
+and eager runs out of memory approximately 3x sooner. The crossover depends on
+model size, batch size, and dtype.
 
 ### 2. Scale batch size - AMD GPUs have large VRAM
 
--- a/docs/amd_rocm.md
+++ b/docs/amd_rocm.md
@@ -56,29 +56,30 @@ which delivers flash_attention_2-level performance with zero extra packages.
 For inputs with unsupported dtypes or mask shapes, SDPA may fall back to a
 math or memory-efficient implementation with different performance characteristics.
 
-Benchmark results (MI325X, TinyLlama-1.1B, inference):
+Inference benchmark (MI325X, TinyLlama-1.1B, batch=1, seq=512, bfloat16):
 
 | Implementation | Throughput     | VRAM    |
 |----------------|----------------|---------|
 | sdpa           | 108,505 tok/s  | 2.22 GB |
 | eager          | 78,914 tok/s   | 2.45 GB |
 
-SDPA is +37% faster and uses 9% less VRAM than eager.
+SDPA is +37% faster and uses 9% less VRAM than eager for this benchmark workload.
 
-At sequence lengths >= 2048, SDPA is effectively required. Eager attention allocates
-an O(n^2) attention matrix and runs out of memory approximately 3x sooner than SDPA.
-At seq=2048, SDPA gives +41% throughput over eager.
+At longer sequence lengths, SDPA's memory advantage grows. In the training benchmark
+(TinyLlama-1.1B, batch=4, bfloat16), SDPA gives +41% throughput at seq=2048 and
+eager runs out of memory approximately 3x sooner. The crossover point depends on
+model size, batch size, and dtype — smaller models or single-batch inference at
+the same length may still fit with eager.
 
 ### 2. Scale batch size - AMD GPUs have large VRAM
 
-AMD Instinct MI300X/MI325X have 192-256 GB HBM3e. At the benchmark workload
-(TinyLlama-1.1B, LoRA r=16, seq=512, bfloat16), default batch=4 uses less than
-5% of available VRAM:
+AMD Instinct MI300X/MI325X have 192-256 GB HBM3e. LoRA training benchmark
+(TinyLlama-1.1B, LoRA r=16, seq=512, bfloat16, steady-state training tok/s):
 
-| Batch size  | Throughput     | VRAM used              |
-|-------------|----------------|------------------------|
-| 4 (default) | 27,418 tok/s   | 10.6 GB (4% of 256 GB) |
-| 16          | 41,271 tok/s   | 20.5 GB (8% of 256 GB) |
+| Batch size             | Training tok/s | VRAM used              |
+|------------------------|----------------|------------------------|
+| 4 (benchmark baseline) | 27,418         | 10.6 GB (4% of 256 GB) |
+| 16                     | 41,271         | 20.5 GB (8% of 256 GB) |
 
 **+51% throughput for this benchmark workload.** These numbers are specific to the
 benchmark conditions above. For larger models or longer sequences, batch=16 may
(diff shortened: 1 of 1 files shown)


zhihuidu-amd:
All 5 Codex review issues addressed in commit `78b8c08`.

**Fix 1 — Windows install path**: Separated Linux/WSL and Windows sections. Windows ROCm wheels come from the AMD TheRock project, not `download.pytorch.org/whl/` (which does not carry Windows ROCm wheels). Added link to AMD's official Windows ROCm install guide.

**Fix 2 — MIOpen cache**: `MIOPEN_CUSTOM_CACHE_DIR` (kernel binary cache, the main compilation cost) is now documented separately from `MIOPEN_USER_DB_PATH` (performance tuning database). Added explicit note: setting only `MIOPEN_USER_DB_PATH` does not avoid kernel recompilation — `MIOPEN_CUSTOM_CACHE_DIR` is what eliminates it.

**Fix 3 — Validation command path**: The test file is `tests/test_rocm_compatibility.py` added in PR #943 — path was already correct in the guide. Added a note that both PRs need to be merged for the command to work.

**Fix 4 — LoRA `target_modules`**: Replaced the attention-only `[q,k,v,o]` example with Unsloth's canonical 7-module set `[q,k,v,o,gate,up,down]`. Added an explicit note explaining that attention-only LoRA is a quality/memory tradeoff (excluded MLP layers), not a free optimization over the default.

**Fix 5 — `lora_dropout=0`**: Set to `0` in the recommended example. Non-zero dropout takes training off Unsloth's optimized path (documented in `README.md:330` and `README.md:400` as "= 0 is optimized").

zhihuidu-amd:
All 4 issues addressed in commit `10b3de1`.

**Fix 1 — PyTorch wheel index**: Added note that not every ROCm minor version has a corresponding wheel, pointed to the PyTorch index list and the `unsloth_zoo` resolver as authoritative sources.

**Fix 2 — Batch size advice scoped to benchmark**: The +51%/batch=16 claim is now explicitly tied to its benchmark conditions (TinyLlama-1.1B, LoRA r=16, seq=512, bfloat16). Added caveat that larger models/sequences may not fit batch=16, with instruction to start from a known-fitting batch.

**Fix 3 — Warm-up cutoff corrected**: Changed 'exclude steps 1-10' to 'exclude steps 1-20'. The table shows throughput still rising at step 10 (756 tok/s vs 1,454 steady-state).

**Fix 4 — Compilation artifacts expanded**: Added `TORCHINDUCTOR_CACHE_DIR` alongside the MIOpen vars. Dynamo/AOTAutograd/Inductor/Triton artifacts live under `TORCHINDUCTOR_CACHE_DIR` (or Unsloth Mega-cache), not MIOpen paths. Narrowed the 'eliminates recompilation' claim to MIOpen kernels specifically.

zhihuidu-amd:
Both issues addressed in commit `e6489a3`.

**Fix 1 — `TORCHINDUCTOR_CACHE_DIR` removed by Unsloth**: Confirmed in `patching_utils.py:151` — `os.environ.pop('TORCHINDUCTOR_CACHE_DIR', None)` is called by `patch_torch_compile()`. Replaced with `UNSLOTH_MEGA_CACHE_DIR` (the correct variable for persisting Inductor/AOTAutograd/Triton artifacts). Added explicit notes: TORCHINDUCTOR_CACHE_DIR is cleared internally; Mega-cache requires torch >= 2.7; on older torch the compile cost recurs each process.

**Fix 2 — SDPA description scoped to benchmark inputs**: Reworded from 'routes to MIOpen fused attention' to 'dispatcher that selects the best available kernel'. Fused-backend equivalence now explicitly scoped to the benchmark input conditions (bfloat16, standard head dims, no custom mask). Added note that other inputs may fall back to math/memory-efficient implementations.

zhihuidu-amd:
All 3 issues addressed in commit `1b7e95c`.

**Fix 1 — Install torchvision and torchaudio from ROCm index**: Changed quick start to `pip install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/rocm6.2`, matching the recovery hint in `device_type.py:188-199`. Prevents mixing ROCm and default-PyPI wheels.

**Fix 2 — `UNSLOTH_MEGA_CACHE=1` required on Windows**: Added Windows PowerShell block showing `UNSLOTH_MEGA_CACHE=1` alongside `UNSLOTH_MEGA_CACHE_DIR`. `megacache_is_enabled()` defaults to False on non-POSIX platforms unless explicitly opted in (`compile_cache.py:41-49`). Added note explaining the POSIX-vs-Windows default difference.

**Fix 3 — Remove misleading `UNSLOTH_IS_PRESENT` workaround**: `unsloth_zoo/__init__.py` calls `find_spec('unsloth')` at line 206 and raises before ever reaching the `UNSLOTH_IS_PRESENT` check at line 457. The env var cannot bypass the import error. Replaced with the correct fix: `pip install unsloth`.

zhihuidu-amd:
All 5 issues addressed in commit `bc2b797`.

**Fix 1 & 4 — Throughput tables labeled by benchmark mode**: Both tables now explicitly state their benchmark type. The inference table (108K tok/s, batch=1) and the training warm-up table (1,454 tok/s, batch=4 LoRA) are different experiments — column headers now say 'Inference benchmark' and 'Training tok/s' respectively.

**Fix 2 — Eager OOM threshold scoped to benchmark**: Removed the universal 'seq>=2048 effectively required' statement. Replaced with the specific training workload (batch=4, bfloat16) where the 3x OOM difference was observed. Added note that the crossover depends on model size, batch size, and dtype.

**Fix 3 — batch=4 labeled as benchmark baseline**: Column renamed from '4 (default)' to '4 (benchmark baseline)'. The README SFT default is batch=2; HF TrainingArguments default is 8; calling it 'default' misrepresented the comparison baseline.

**Fix 5 — 28% VRAM saving claim removed**: q+k+v+o adds adapter parameters vs q+v, so target selection alone doesn't explain a VRAM reduction without knowing other differing conditions. Replaced with a factual description of what attention-only LoRA does (fewer trained parameters, different coverage).

**Fix 6 — Mega-cache trust requirements documented**: `_is_trusted_directory()` rejects directories owned by others or with group/world-writable mode bits, checking all parent directories. Added `mkdir -p ~/.cache/unsloth/mega_cache && chmod 700` setup commands so users know how to create a trusted cache directory.

zhihuidu-amd:
All 3 issues addressed in commit `df9918b`.

**Fix 1 — Non-invocable resolver removed**: The 'use the version resolver in device_type.py' instruction required packages that don't exist yet at that point in the install flow. Replaced with a direct link to the [PyTorch ROCm wheel index list](https://download.pytorch.org/whl/), which users can check before installing anything.

**Fix 2 — Mega-cache trust rule corrected**: Previous note said all parent directories must be owned by the current user — impossible since `/` is root-owned. Corrected: only the leaf cache directory requires current-user ownership. Parent directories are accepted if not group/world-writable, OR if they are sticky and root/user-owned (like `/tmp`). Added guidance for shared-volume paths and a `ls -ld` verification command.

**Fix 3 — Code example labeled as illustrative**: Added a comment noting the example works with any model and that all benchmarks used TinyLlama-1.1B, so users do not expect the Meta-Llama-3-8B example to reproduce the benchmark figures.

zhihuidu-amd:
Both issues addressed in commit `ed1bb36`.

**Fix 1 — Windows ROCm training not supported**: `triton-windows` in `pyproject.toml:100` requires CUDA 12 and is not AMD-compatible, so training on native Windows ROCm does not work with the current stack. Updated the Windows section to explicitly say 'inference only' and added a note that training requires Linux or WSL where the standard `triton>=3.0.0` wheel supports ROCm.

**Fix 2 — Inference benchmark labeled as prefill**: 108,505 tok/s measures a single forward pass over 512 input tokens — prefill, not autoregressive decode. Decode throughput is orders of magnitude lower for both backends. Relabeled as 'Prefill throughput benchmark' with explicit '(not autoregressive decode)' qualifier. The +37% and +41% claims are also now scoped to prefill specifically.

zhihuidu-amd:
All 4 issues addressed in commit `56a399f`.

**Fix 1 — `--upgrade --force-reinstall`**: Added to the pip install command so existing CUDA/PyPI wheels are actually replaced. Matches the recovery pattern in `device_type.py:_amd_installation_hint()`.

**Fix 2 — 'Train as on NVIDIA' scoped to Linux/WSL**: The sentence followed the Windows inference-only warning and told Windows users to train on an unsupported path. Changed to 'On Linux / WSL, train exactly as you would on NVIDIA'.

**Fix 3 — +41% at seq=2048 corrected to training-step throughput**: The figure came from a LoRA training run (forward + backward + optimizer), not a forward-only prefill measurement. Changed 'prefill throughput' to 'training-step throughput' to distinguish it from the prefill benchmark above.

**Fix 4 — SDPA example qualified by architecture**: Added a comment noting that some architectures (Pixtral, Mistral 3) don't support SDPA — Unsloth detects this and falls back to eager automatically. Users replacing the model checkpoint now know to check architecture support.

zhihuidu-amd:
Also rebased `feat/rocm-docs` onto current upstream `main` (`c6c8afb`) — the `mergeable_state: unstable` was caused by upstream moving ahead. Force-pushed the rebased branch; PR #945 is now conflict-free.

LeoBorcherding:
Closing this one, and it's worth being clear that it isn't a quality judgment. The guide
went through eight rounds of review and the content got sharper each time.

The main reason is location. unsloth-zoo has no `docs/` tree: the entire repo's markdown is
`README.md`, `CONTRIBUTING.md`, `CODE_OF_CONDUCT.md`, the two issue templates, and one
package-internal `diffusion_studio/README.md`. This PR would create a new top-level `docs/`
directory for a single guide, and prose documentation for us lives at docs.unsloth.ai and
the blog rather than in the library repo. That's where an AMD tuning guide reaches users
who aren't reading the source tree anyway.

Three things that also argue against landing it here specifically:

The validation section points at `tests/test_rocm_compatibility.py`, which is #943 and
isn't merged. On merge today that command fails, and the doc can't be made accurate without
ordering it behind another open PR.

The benchmark environment is ROCm 6.2 with torch 2.5.1, while #920's aiter path requires
ROCm >= 7.0 on gfx942/gfx950. Anyone reading the series together gets numbers measured on a
configuration where the sibling feature cannot run.

Pinned throughput and VRAM figures in-repo have nothing keeping them honest. There's no CI
that re-measures them, so they drift silently as ROCm, torch and the model set move.

One thing to carry over wherever this lands: the note at lines 48-50 says Unsloth detects
unsupported SDPA architectures and falls back to eager automatically. That same claim was
in #944's docstring and was removed there, because `from_config` enforces `_supports_sdpa`
itself and raises rather than falling back. It's a different call path here, so I haven't
disproven it for `from_pretrained` specifically, but it's worth verifying before it's
repeated in the hosted docs.

Happy to help get the content routed to docs.unsloth.ai if you want to take it there. The
benchmark data in particular is worth publishing somewhere it'll be read.

zhihuidu-amd:
Thank you for the detailed reasoning — all three structural objections are sound.

Closing this PR. The guide will be routed to docs.unsloth.ai where it can reach users who aren't reading the source tree, stay current with ROCm/torch releases, and sit alongside the AMD blog post that shipped with v0.1.50-beta.

The caution you flagged on lines 48-50 (SDPA fallback claim) is noted and will be verified against `from_pretrained` specifically before it appears in hosted docs.

The benchmark data and content from this PR are worth publishing — happy to take you up on the offer to help route it there once the hosted docs path is open.

PR #955, opened 2026-07-28
hf_utils: unwrap single-adapter peft_config and reject empty model_types

BardiaKoopah:
## Summary
 
`get_transformers_model_type` could not read a `model.peft_config` unless the adapter was named `"default"`, and the guard meant to catch that failure never fired. Two coupled one-line fixes, plus regression tests and CI wiring.
 
## The problem
 
**1. Only `"default"` was unwrapped.** `model.peft_config` maps adapter name → config. The unwrap at `hf_utils.py:118-119` handled exactly one key:
 
```python
# Handle model.peft_config["default"]
if type(config) is dict and "default" in config:
    config = config["default"]
```
 
So `get_peft_model(model, cfg, adapter_name = "my_adapter")` — ordinary PEFT usage — produced a dict this function fell straight through. A plain dict has no `.to_dict`, so the `lambda *args, **kwargs: {}` default fires and `find({})` yields nothing.
 
**2. The guard that should have caught it is dead.** `model_types` is assigned from `list(find(...))` at `hf_utils.py:196`, which returns `[]` and never `None`. The next line tests `if model_types is None:`, so it never fires for that case and the empty list is returned to callers.
 
Reproduction, no weights or network required:
 
```
peft_config dict keyed 'default'      -> ['llama']
peft_config dict keyed 'my_adapter'   -> []        <- silently empty
config object whose to_dict() lacks model_type -> []
config object with no to_dict at all           -> []
```
 
## What breaks downstream
 
Worth stating precisely, because it is not uniformly a crash:
 
| Consumer | With `[]` | Result |
|---|---|---|
| `unsloth/models/vision.py:900` | `model_types[0]` | `IndexError` — the only hard crash |
| `unsloth/models/loader.py:640-644` | `len(...) == 1` false → `model_type = []` | no crash, silently wrong |
| `unsloth/models/loader.py:1349` | `",".join(model_types) + ","` → `","` | every arch check silently `False` |
| `unsloth_zoo/hf_utils.py:312` | `[0]` inside a bare `except:` | swallowed, `model_type` stays `None` |
 
`vision.py:856` already does `if model_types is None: raise RuntimeError(...)` — the same dead-guard shape — which is reasonable evidence the intended contract is a non-empty list and that raising at the source is what consumers want. No consumer has meaningful empty-list behaviour to fall back on.
 
## The fix
 
**Unwrap any single-adapter dict**, whatever its key:
 
```python
if type(config) is dict:
    if "default" in config:
        config = config["default"]
    elif len(config) == 1:
        config = next(iter(config.values()))
```
 
`"default"` keeps precedence in a multi-adapter dict. A multi-adapter dict **without** `"default"` is deliberately left unresolved — adapters can carry different base models, so unioning their model types or picking one would be a guess. It falls through and raises with an actionable message. `fix_lora_auto_mapping` (`hf_utils.py:244-245`) iterates `.values()` for its own purposes; unaffected here, and verified unchanged by diffing every adapter's `auto_mapping` before and after.
 
**Make the guard live:** `if model_types is None:` → `if not model_types:`.
 
The other `is None` check (`hf_utils.py:157`, gating the AutoConfig fallback inside the PeftConfig branch) is deliberately untouched — there `model_types` genuinely is `None`, never `[]`, and a test pins that path so it stays reachable.
 
Behaviour before and after:
 
| Input | Before | After |
|---|---|---|
| single-key `"my_adapter"` | `TypeError` | `["llama"]` |
| multi-key, no `"default"` | `TypeError` | `TypeError` |
| multi-key with `"default"` | `["llama"]` | `["llama"]` |
| `{}` | `TypeError` | `TypeError` |
| normal `LlamaConfig` | `["llama"]` | `["llama"]` |
 
## Compatibility
 
`get_auto_processor` indexes `[0]` inside a bare `except:`, which catches the new `TypeError` exactly as it caught the old `IndexError` — `model_type` stays `None` and falls through to the `AutoTokenizer` branch. Identical behaviour, pinned by a test.
 
`unsloth/models/_utils.py:3663` and `unsloth/trainer.py:696` consume via `any(... for mt in model_types)` and would tolerate `[]`; they now raise instead. Both pass a `PretrainedConfig`, which always emits `model_type`, so this is unreachable for them in practice — confirmed by running both consumer expressions verbatim over eleven real configs (llama, qwen2, mistral, gemma2, gpt_neox, gpt_oss, gemma3_text, nested gemma3, nested qwen2-VL, an `AutoConfig.from_pretrained` off disk, and an instantiated model's `.config`). Every boolean verdict is identical before and after.
 
The new tests restore `UNSLOTH_MODEL_NAME`, which `get_transformers_model_type` sets as a side effect at `hf_utils.py:154` and which two other test files read, so they do not leak session-wide.
 
## Tests
 
New `tests/test_get_transformers_model_type_empty.py`, CPU-only and network-free (temp-dir configs, no HF cache dependency):
 
- all four unwrap cases from the table above
- the three inputs that reach `[]`, each raising `TypeError` with the config named
- the paths that already worked, unchanged
- the `hf_utils.py:312` bare-`except:` site still yielding `model_type is None`
- both untouched guards still reachable and still firing
Five of these fail against unpatched source; the rest are the no-regression half.
 
The file is wired into the behavioral-gates step in `.github/workflows/consolidated-tests-ci.yml`, without which a new test file is collected but never executed.

### DIFF
--- a/.github/workflows/consolidated-tests-ci.yml
+++ b/.github/workflows/consolidated-tests-ci.yml
@@ -108,6 +108,7 @@ jobs:
             tests/test_mlx_finetune_last_n_layers.py \
             tests/test_mlx_double_quant_reject.py \
             tests/test_hf_xet_fallback.py \
+            tests/test_get_transformers_model_type_empty.py \
             -v
 
       - name: pytest tests/test_mlx_module_exports + zoo-specific CPU tests
--- a/tests/test_get_transformers_model_type_empty.py
+++ b/tests/test_get_transformers_model_type_empty.py
@@ -1,0 +1,183 @@
+# SPDX-License-Identifier: AGPL-3.0-only
+# Copyright 2026-present the Unsloth AI Inc. team. All rights reserved.
+
+"""Regression tests for get_transformers_model_type's config unwrapping and guard.
+
+Two behaviours are pinned here:
+
+1. An unresolved config must not come back as `[]`. `model_types` is assigned from
+   `list(find(...))`, so it reaches the guard as `[]` and never as `None`; the guard
+   used to test `is None`, so `[]` passed through and consumers broke on it --
+   `unsloth/models/vision.py` indexes `model_types[0]`, `unsloth/models/loader.py`
+   joins the list and silently matches no architecture.
+2. `model.peft_config` maps adapter name -> config, and only the literal key
+   "default" used to be unwrapped, so an adapter named anything else was unreadable.
+   A single-adapter dict now unwraps whatever its key is; a multi-adapter dict
+   without "default" stays ambiguous and must raise rather than pick a winner.
+
+All cases are CPU-only and network-free: configs are built in a temp directory so
+AutoConfig resolves from disk and never contacts the Hub.
+"""
+
+import json
+import os
+from pathlib import Path
+
+import pytest
+
+from unsloth_zoo.hf_utils import get_transformers_model_type
+
+
+_UNRESOLVED_MESSAGE = "Cannot determine model type for config file"
+
+
+@pytest.fixture(autouse = True)
+def _no_env_leak():
+    """get_transformers_model_type sets os.environ["UNSLOTH_MODEL_NAME"] as a side
+    effect for every PeftConfig input (hf_utils.py). Other test modules read that
+    variable, so restore it rather than leaking temp paths into the session."""
+    sentinel = object()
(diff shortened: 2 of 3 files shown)


danielhanchen:
@codex review

danielhanchen:
@codex review

PR #958, opened 2026-07-28
DiffusionGemma: make the visual server's GPU layer split configurable

LeoBorcherding:
Fixes unslothai/unsloth#7574.

`_build_subprocess_env` spawned the visual-server child with `NGL=99` hardcoded, pinning every
layer to the GPU. A GGUF larger than VRAM died in `cudaMalloc` with no way to split it across
GPU and RAM:

```
allocating 16013.20 MiB on device 0: cudaMalloc failed: out of memory
```

The knob was unreachable by design: the function takes a `base_env` but overwrote `NGL`
unconditionally, and `VisualServer.__init__` had no layer parameter.

Meanwhile the server itself does `atoi(getenv("NGL") ? getenv("NGL") : "0")` — a plain
llama.cpp `n_gpu_layers`, defaulting to CPU, with partial offload fully supported. Its own OOM
message advises *"Try a smaller quant, lower NGL, or more VRAM/RAM"*, which no Unsloth user
could act on.

## Change

`_resolve_ngl()` resolves, in order: explicit caller value → `NGL` already in the environment →
the previous all-layers default. `--ngl` is exposed on the shim, which now logs the value it
resolved. Nothing changes when no split is requested.

The env-var step is deliberate: it gives users on an already-installed Studio a way out without
waiting for the Studio-side plumbing.

## Performance note

The reporter measured **~6 tok/s against ~31 tok/s** for non-diffusion Gemma-4-26B-A4B once
layers sit on CPU. Block diffusion is compute-bound rather than bandwidth-bound, so a CPU split
costs much more here than on an autoregressive model, which may be why the value was pinned at
99 originally. The all-layers default is kept for exactly that reason: only a caller that asks
for a split gets one. An OOM with no override is still the wrong failure mode.

## Tests

Five cases added to `tests/test_diffusion_visual_engine_env.py`: explicit value forwarded,
`ngl=0` not swallowed as falsy (the exact reported case), env honoured when the caller passes
nothing, explicit beating env, and garbage env falling back to the default. 9 passed.

### DIFF
--- a/tests/test_diffusion_visual_engine_env.py
+++ b/tests/test_diffusion_visual_engine_env.py
@@ -36,8 +36,47 @@ def test_linux_prepends_nvidia_wheels_and_excludes_torch_lib(monkeypatch):
     assert all("torch" not in p for p in ld)       # torch/lib excluded on Linux
     assert env["PATH"] == "/usr/bin"               # PATH untouched on Linux
     assert env["CUDA_VISIBLE_DEVICES"] == "3"
-    assert env["NGL"] == "99"
+    assert env["NGL"] == "99"          # unset -> all layers, the historical default
     assert env["MAXTOK"] == "4096"
+
+
+def test_explicit_ngl_is_forwarded():
+    """A caller-supplied split reaches the child: the whole point of the knob (small-VRAM cards)."""
+    env = V._build_subprocess_env(
+        "/dg/srv", gpu = "0", maxtok = 0, ngl = 8,
+        base_env = {"PATH": "/usr/bin"}, os_name = "posix",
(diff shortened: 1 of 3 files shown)


LeoBorcherding:
The reporter from unslothai/unsloth#7574 installed this branch (together with unslothai/unsloth#7575) on the failing machine (RTX 4050 laptop, 6 GB VRAM, Ubuntu, `unsloth_zoo 2026.7.6` baseline) and confirms the manual split now loads the model, with the CPU-offload warning showing. No code changes here since open; head is still `c903652d`.

@codex review

PR #961, opened 2026-07-28
train_on_responses_only: accept plain-list labels under datasets.map(batched=True)

BardiaKoopah:
## Impact
 
Calling `train_on_responses_only` on a dataset that already carries a `labels` column raised `AttributeError: 'list' object has no attribute 'tolist'`, and masking never ran. That covers any dataset labelled before masking — a prior tokenization pass, a collator that emitted labels, or a second `train_on_responses_only` call.
 
## Mechanism
 
The masking function already type-checks `input_ids`:
 
```python
if type(input_ids_) is torch_Tensor:
    use_tensors = True
    input_ids_ = input_ids_.tolist()
```
 
The labels branch immediately below called `.tolist()` unconditionally. Under `datasets.map(batched = True)` — which is how `train_on_responses_only` invokes it on every non-iterable dataset — a `labels` column arrives as a plain list of lists, which has no `.tolist()`. Only a torch-tensor labels column worked, so the failure depended on the dataset's format rather than on anything the caller chose.
 
## Change
 
Type-check labels the same way `input_ids` is checked: convert only when it is a tensor, otherwise use the value as-is. The length assertion below and the tensor return path are unchanged.
 
## Reachability
 
Reached on the ordinary dataset path, not an explicit-argument corner. Reproduced two ways: by calling the returned function directly, and through a real `Dataset.map(batched = True)`. Independent of `force_match` and of whether the markers were auto-detected.
 
## Tested
 
New `tests/test_train_on_responses_list_labels.py` — 5 tests, stub tokenizer, no weights, no network. 3 fail on unpatched source and all 5 pass after. Guards cover the tensor path, the no-labels path, and that supplied labels stay the source of truth for unmasked positions.
 
Wired into the behavioral-gates CI step so the file is executed rather than merely collected (#669/#739 precedent: an unexecuted test was born broken in #669 and stayed invisible until #739).
 
Full zoo suite run twice per side with `-p no:randomly`. The suite is nondeterministic independent of this change (two identical baseline runs differ), so the check is tests failing in both patched runs that passed in both baseline runs: none. Collected totals reconcile exactly — 2977 → 2982, the 5 tests added. No environment writes from the new tests.

### DIFF
--- a/.github/workflows/consolidated-tests-ci.yml
+++ b/.github/workflows/consolidated-tests-ci.yml
@@ -101,7 +101,9 @@ jobs:
         # use_cache disable/restore contract for gradient checkpointing
         # (#715). test_hf_xet_fallback exercises the Xet->HTTP stall fallback
         # + cache-completeness gate (CPU-pure; collect-only would let it
-        # regress silently). Executing them here keeps fixture/source drift visible.
+        # regress silently). test_train_on_responses_list_labels pins the
+        # tensor-vs-list labels contract for response masking (stub tokenizer,
+        # no weights). Executing them here keeps fixture/source drift visible.
         run: |
           python -m pytest \
             tests/test_training_utils_use_cache.py \
@@ -109,6 +111,7 @@ jobs:
             tests/test_mlx_double_quant_reject.py \
             tests/test_hf_xet_fallback.py \
             tests/test_get_transformers_model_type_empty.py \
+            tests/test_train_on_responses_list_labels.py \
             -v
 
       - name: pytest tests/test_mlx_module_exports + zoo-specific CPU tests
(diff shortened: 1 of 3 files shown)


BardiaKoopah:
@codex review

oobabooga:
Verified against current main and the PR candidate.

Current main raises `AttributeError: 'list' object has no attribute 'tolist'` when `train_on_responses_only` receives the plain-list `labels` produced by `Dataset.map(batched=True)`. It also fails when the masking function is applied a second time to its own output. This is a supported input shape: `sft_prepare_dataset` explicitly preserves pre-existing labels.

The candidate fixes both flows and produces the expected response-only mask in direct dataset mapping and an end-to-end trainer run. Tensor labels, NumPy labels, datasets without labels, and `last_response_only` retain their previous behavior. The focused regression tests and the behavioral CI gate pass, and I found no missed sibling path or decision-changing regression.

No blockers from my side. Good to merge.

PR #965, opened 2026-07-30
feat(mlx): off-policy knowledge distillation (GKD)

BardiaKoopah:
A frozen teacher scores the student's own batch, and the student trains against the teacher's distribution via TRL's generalized JSD. TRL has `GKDTrainer`, but MLX stubs every non-SFT TRL trainer at `unsloth/__init__.py:1402-1410`, so distillation was unreachable on Apple Silicon. Requested in unslothai/unsloth#3008 and #1941.
 
New `unsloth_zoo/mlx/distill.py`, one branch at `trainer.py`'s loss-selection site, and five config fields appended last. With `teacher_model_name_or_path = None` everything stays inert.
 
## Verified pairs
 
Teacher and student must share a tokenizer:
 
| student ← teacher | widths | tokenizers agree |
|---|---|---|
| Qwen2.5-0.5B ← Qwen2.5-3B | 151936 | yes |
| Llama-3.2-1B ← Llama-3.2-3B | 128256 | yes |
| Qwen3-0.6B ← Qwen2.5-3B | 151936 | yes — spans generations |
| Llama-3.2-1B ← Qwen2.5-3B | mismatch | no |
 
Width equality alone is **not** sufficient: two checkpoints can share a vocab size and still encode differently, which would silently train against misaligned targets. The guard compares probe-encoded ids as well as width, and raises naming both vocab sizes. Cross-family distillation needs token remapping, which is not implemented here.
 
## The envelope, stated plainly
 
With a 3B teacher on 16 GB, roughly **B×L ≤ 1024 tokens** — one sequence of 1024, or two of 512.
 
| B | L | tokens | teacher logit buffer | chunked peak | unchunked peak |
|---|---|---|---|---|---|
| 1 | 512 | 512 | 0.29 GB | 6.59 GB | 7.47 GB |
| 2 | 512 | 1024 | 0.58 GB | 10.01 GB | 12.05 GB |
| 1 | 1024 | 1024 | 0.58 GB | — | 12.17 GB |
| 2 | 1024 | 2048 | 1.16 GB | 15.29 GB | 17.39 GB — crashes |
| 4 | 512 | 2048 | 1.16 GB | — | 17.16 GB — crashes |
 
These are peaks under a real LoRA training step, including optimizer state and `mx.compile`. They supersede earlier forward+backward-only figures.
 
## Why the preflight is mandatory
 
An oversized batch does not raise a catchable OOM. It aborts with `[METAL] Command buffer execution failed: Impacting Interactivity` and wedges the GPU context for subsequent runs.
 
So a preflight estimates peak from batch × sequence × vocab width up front and raises an ordinary error naming the requested B×L, the estimated peak, the detected system memory, and the largest B×L that fits. Calibration against the integrated path is conservative in the safe direction: predicted 6.65 vs measured 6.59 GB, predicted 10.42 vs measured 10.01 GB.
 
`gkd_skip_memory_preflight = True` bypasses it, emitting a warning that names the estimate, the budget, and the wedge risk — for users with more headroom than the estimator assumes.
 
The loss is chunked by default (B=2/L=1024: 17.39 → 15.29 GB). Unchunked is reachable only for the equivalence test.
 
## Off-policy only
 
`gkd_lmbda != 0` is rejected with a message: on-policy GKD needs generation inside the training loop, and `trainer.py` has no such path (`generate` appears zero times in it; all generation lives in `loader.py` as inference shims).
 
## Tested — 34 tests
 
Loss falls over 8 steps and KL to the teacher measurably falls, at β ∈ {0, 0.5, 1}; chunked equals unchunked in both value and gradients; self-distillation ≈ 0; label masking honoured (fully-masked → exactly 0.0); `mx.compile` bit-identical; preflight fires at the thresholds above and the override bypasses it; the teacher has 0 trainable params, 0 gradients, 0 optimizer-state entries, and is absent from compile-traced state.
 
Full zoo suite run twice per side: 2979 → 3013, delta 34 = the tests added. No regressions, no environment writes.
 
Two test files on purpose: the numeric tests need real `mx.quantize`/MLX, so a single file would skip entirely on Linux CI and be a gate no-op (#669/#739). The guard, config and preflight tests run under the existing torch shim and execute in the behavioral-gates step.
 
**Note for future MLX tests in this repo:** the numeric tests bind `mlx.core` at module import, not inside functions. The repo's `mlx_simulation` shim replaces `sys.modules["mlx.core"]` session-wide, so a function-level `import mlx.core` silently resolves against the stub — tests then pass against a stub while looking clean standalone. This was caught only by a full-suite run.

### DIFF
--- a/.github/workflows/consolidated-tests-ci.yml
+++ b/.github/workflows/consolidated-tests-ci.yml
@@ -103,7 +103,11 @@ jobs:
         # + cache-completeness gate (CPU-pure; collect-only would let it
         # regress silently). test_train_on_responses_list_labels pins the
         # tensor-vs-list labels contract for response masking (stub tokenizer,
-        # no weights). Executing them here keeps fixture/source drift visible.
+        # no weights). test_mlx_gkd_guards pins the tokenizer-compatibility,
+        # config and memory-preflight contracts for distillation; it runs under
+        # the torch shim so it EXECUTES here rather than skipping (the numeric
+        # half needs real MLX and runs on the Apple Silicon job). Executing them
+        # here keeps fixture/source drift visible.
         run: |
           python -m pytest \
             tests/test_training_utils_use_cache.py \
@@ -112,6 +116,7 @@ jobs:
             tests/test_hf_xet_fallback.py \
             tests/test_get_transformers_model_type_empty.py \
             tests/test_train_on_responses_list_labels.py \
+            tests/test_mlx_gkd_guards.py \
             -v
 
       - name: pytest tests/test_mlx_module_exports + zoo-specific CPU tests
--- a/tests/test_mlx_gkd_guards.py
+++ b/tests/test_mlx_gkd_guards.py
@@ -16,15 +16,10 @@
 
 """GKD guards: tokenizer compatibility, config validation, memory preflight.
 
-Pure logic and arithmetic, so this runs under the torch shim on Linux CI instead
-of skipping. The numeric behaviour of the divergence lives in
-test_mlx_gkd_loss.py, which needs real MLX and runs on Apple Silicon.
-
-The preflight numbers below are the measured ones from a Qwen2.5-0.5B LoRA
-student with a Qwen2.5-3B teacher (vocab 151936, 2.881 GB resident) on a 16 GB
-machine. They matter because an oversized step does not raise a catchable OOM --
-it aborts with "[METAL] Command buffer execution failed: Impacting
-Interactivity" and wedges the GPU context.
+Pure logic, so this runs under the torch shim on Linux CI instead of skipping.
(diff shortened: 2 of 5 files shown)


danielhanchen:
Checked that this stays inert with teacher_model_name_or_path = None and that the tokenizer-compatibility and memory preflight guards run under the Linux shim, so the Apple Silicon numeric half is the only untested part here. Could you rebase onto main so the merge state is clear before I review?

danielhanchen:
Checked that the new loss path in unsloth_zoo/mlx/distill.py only engages when teacher_model_name_or_path is set, so existing MLX runs keep the current loss selection, and the tokenizer and memory guards look sensible. Could you rebase onto main so the conflicts clear before review?

PR #967, opened 2026-07-30
Harden distributed compiled-cache synchronization and recovery

sstamenk:
## Problem

`create_new_function()` performed distributed broadcasts and barriers behind rank-local filesystem and error conditions.

Under `torchrun`, rank 0 could write a compiled cache file before another rank evaluated the surrounding `os.path.isfile()` guard. The later rank would see the file, skip the collective, and continue while rank 0 remained blocked in the barrier. The process group stayed desynchronized until the NCCL/RCCL watchdog terminated the job, commonly during `import unsloth` while `PatchFastRL` compiled TRL trainers.

Further review found the same class of issue in related paths:

- stale or missing rank-local cache files could make ranks import different implementations;
- verification, write, directory creation, import, and direct-load failures could affect only one rank;
- tempfile paths were broadcast from rank 0 even though `tempfile.gettempdir()` is node-local;
- rank 0 and worker nodes could have different generated or retained source;
- stale `sys.modules`, `sys.path`, or `.pyc` state could bypass verified source;
- another process could rewrite a cache file after collective verification but before import;
- retained-cache races could become fatal instead of recovering.

## Behavior after this PR

| Scenario | Behavior |
|---|---|
| Cold shared persistent cache | Rank 0 decides and writes; every rank verifies its local view against rank 0’s digest. |
| A rank reaches the guard after rank 0 writes | Every rank still enters the same collectives; the original unmatched-barrier hang is eliminated. |
| Warm persistent cache | Rank 0’s retained on-disk bytes are authoritative and verified on every rank. |
| `UNSLOTH_COMPILE_OVERWRITE=0` | Readable retained rank-0 bytes are preserved, including when worker temp caches are missing or stale. |
| Rank 0 cannot establish a retained digest | The group regenerates from rank 0’s generated source rather than accepting unknown/divergent worker bytes. |
| Persistent cache missing, stale, or changed during verification | The mismatch is agreed and recovered through node-local temp caches rather than raising rank-locally. |
| Delayed network-filesystem visibility | Persistent-cache verification retries with bounded exponential backoff for up to five seconds. |
| Node-local temp verification | Verification is immediate; the shared-filesystem visibility grace period is not applied. |
| Explicit or implicit temp fallback | Every rank resolves `tempfile.gettempdir()` locally; rank 0’s node-specific path is never broadcast as another node’s path. |
| Temp mode diverged before process-group initialization | The state is collectively reconciled before any rank can skip a folder collective. |
| Rank 0 temp cache is warm and workers are cold/stale | Every local copy is repaired to rank 0’s canonical generated or retained bytes. |
| Generated source differs across ranks | Node-local writes use rank 0’s broadcast source and digest; local source cannot silently diverge. |
| Rank 0 persistent write fails | The outcome is broadcast and all ranks enter tempfile recovery together. |
| One node cannot create or write its temp cache | The failure is agreed before any rank proceeds; all ranks fail together. |
| Import fails on one rank | Every rank enters recovery and loads the same rank-0 source from its local temp file. |
| Direct loading fails on one rank | The failure is agreed globally and all module aliases are removed. |
| Recompile the same module name | Cached aliases are evicted and the newly verified file is imported. |
| Same-size source is rewritten within timestamp granularity | Local timestamp-based `.pyc` files are removed before normal and direct imports. |
| Bytecode cannot be removed | The error enters recovery or fails collectively rather than allowing stale bytecode to execute. |
| Recovery switches from persistent to temp | The temp folder’s own bytecode cache is invalidated before direct loading. |
| Another process rewrites the file after collective verification | The digest is rechecked while holding the import lock; mismatches enter collective recovery. |
| Another `sys.path` entry shadows the module name | The verified cache path is temporarily placed first, then the original path order is restored in place. |
| Import or direct execution fails | Temporary path changes and partial `sys.modules` entries are cleaned up. |
| Cache read fails before the first collective | The error cannot escape rank-locally; rank 0’s write decision remains authoritative. |
| Torchrun launched before `init_process_group()` | Existing independent pre-initialization behavior is preserved; coordinated verification begins only after collectives are available. |

### Rank-consistent cache decisions

- Rank 0 determines whether the persistent cache file should be written or retained.
- The decision and canonical digest are broadcast to every rank.
- Unknown rank-0 digests force regeneration instead of degrading to existence-only agreement.
- No rank-local condition can independently skip a collective inside `create_new_function()`.

### Canonical source and digest

- Rank 0’s generated source is broadcast lazily when node-local generated writes are required.
- Rank 0’s retained on-disk source is broadcast when repairing overwrite-opt-out caches.
- Every node-local write and verification uses the same canonical source and digest.
- If retained bytes are unreadable, the group collectively falls back to rank 0’s generated source.

### Collective-safe recovery

- Rank-local errors are gathered before any rank raises.
- Write helpers report outcomes instead of escaping rank 0 before its broadcast.
- Persistent mismatches recover to temp.
- Temp mismatches are repaired in place.
- Temp directory creation, local writes, verification, imports, and direct loads all succeed or fail group-wide.

### Node-local tempfile behavior

- Shared persistent-cache paths remain rank-0-selected.
- Temp paths are resolved independently on every node.
- Same-node ranks use `FileLock` and content comparison, keeping repeated writes idempotent.
- Divergent pre-initialization temp-mode state is reconciled once the process group exists.

### Module loading correctness

- The verified cache path takes precedence over shadowing `sys.path` entries.
- Plain and recovery aliases are evicted before recompilation.
- Successful recovery is republished under the normal module name.
- Global direct-load failures remove aliases on every rank.
- Failed imports restore `sys.path`.
- Failed direct execution removes partially initialized modules.
- Local stale bytecode is removed before normal and direct imports.
- Bytecode-removal failures are surfaced instead of being silently ignored.
- The canonical digest is rechecked under the import lock immediately before loading.

### Diagnostics

- Rank reporting uses the initialized process group when available.
- A rank pulled into recovery by another rank no longer logs `import failed: None`.
- Original import failures are logged before later recovery failures can obscure them.
- Missing and divergent files identify ranks and abbreviated digests.

## Compatibility

- Readable `UNSLOTH_COMPILE_OVERWRITE=0` caches remain authoritative.
- Shared persistent caches still use a rank-0 write.
- Only node-local temp caches are written on every rank.
- Pre-process-group torchrun execution remains independent.
- Unreadable retained files regenerate because retaining their bytes is impossible.
- Repeated same-name compilation now reloads verified source instead of returning a stale cached module.

## Regression coverage

`tests/test_compiled_cache_collective.py` covers:

- rank-consistent write and retain decisions;
- exact digest verification;
- stale, missing, unreadable, zero-byte, and delayed files;
- coordinated local and remote failures;
- shared, rank-local, and per-node cache layouts;
- explicit, implicit, and pre-initialization temp fallback;
- mixed warm/cold/stale node-local caches;
- divergent generated source across ranks;
- overwrite opt-out and retained-source repair;
- write, mkdir, import, verification, and direct-load failures;
- persistent-to-temp folder switching;
- alias eviction and republication;
- `sys.path` restoration and shadowing;
- partial module cleanup;
- stale normal-import and direct-load pyc files;
- undeletable bytecode;
- import-lock digest races;
- repeated same-name recompilation;
- operation without an initialized process group.

`tests/test_compiled_cache_collective_gloo.py` adds real two-rank coverage for behavior that single-process mocks cannot prove:

- pre-initialization temp-mode divergence;
- one-rank import failure and collective recovery;
- real `distributed_any()` semantics.

Static source guards read `compiler.py` directly from the checkout. Runtime tests skip cleanly when the companion `unsloth` package or gloo backend is unavailable.

### Automated suites

- **44 compiled-cache cases passed**
- **229 passed, 14 skipped** across targeted and neighboring compiler suites
- Reverse-order and repeated-process runs pass without leaking compiler state
- Mutation testing found no lost coverage after consolidating the suite
- Spawn-based two-rank Gloo tests pass with a shared timeout and portable file-store rendezvous

### RCCL validation

Validated on **8× AMD Radeon AI Pro R9700S** using PyTorch’s `nccl` backend:

- PyTorch 2.12 / ROCm 7.15

The following completed without deadlocks, watchdog timeouts, collective mismatches, or divergent implementations:

- RCCL all-reduce, `broadcast_object_list`, and `all_gather_object`
- `distributed_any()` with both true and false outcomes
- Cold-cache compilation with the final rank delayed by three seconds at world sizes 2, 4, and 8
- One-rank import failure and collective recovery at world sizes 2, 4, and 8
- Twenty consecutive staggered compilations at world sizes 2 and 4
- Retained-cache behavior with `UNSLOTH_COMPILE_OVERWRITE=0` at world sizes 2 and 4
- End-to-end two-rank `import unsloth`
- Key collective, cold-cache, and import-recovery scenarios on both ROCm/PyTorch builds

Each process selected its local GPU before initializing the process group, as required for NCCL/RCCL object collectives.


shimmyshimmer:
Thanks @sstamenk, this is a real bug and the diagnosis is correct. I rebased the branch onto current main and pushed two commits on top.

### Confirming the original hang

Two gloo ranks driving the real `create_new_function`, with rank 1 reaching the guard 3s late:

```
[rank 1] calling create_new_function
[rank 0] calling create_new_function
[rank 1] RETURNED ok, e2e_fn(21)=42
<rank 0 never returns>
```

`distributed_function()` runs `write_file()` on rank 0 *before* it blocks in the broadcast, so on a cold cache any rank arriving after that write sees an up-to-date file and skips the collective. With the fix both ranks complete.

### f5153ada Verify the broadcast cache decision on every rank

The existence check in 0b8ab4d4 only caught a *missing* file. If a rank holds a stale file at the same path, rank 0 broadcasts `False`, the stale file passes the guard, and the ranks import different implementations with no error at all:

```
[rank 0] PASSED CHECK -> would import x*2 (correct)
[rank 1] PASSED CHECK -> would import x*3 (STALE)
```

Rank 0 now broadcasts a digest of the bytes it will import, and every rank compares against it. Two details worth calling out:

* The digest is of the file **on disk**, not of `write_new_source`. Comparing against `write_new_source` looks equivalent but breaks `UNSLOTH_COMPILE_OVERWRITE=0`, where rank 0 deliberately keeps an older cache file that does not match it. Verified against the real code path:

  ```
  UNSLOTH_COMPILE_OVERWRITE=0 -> stale file kept on disk: True
  local file == write_new_source?                        False
  ```

* The check retries for up to 5s before raising. Raising on the first miss fails a healthy run when a network filesystem publishes the file just after the barrier, reproduced with a 1.5s delay.

### 2fa3b9a8 Keep the two recovery paths in create_new_function collective

Same bug class, two more places in the same function.

**1. Write failure.** `get_lock()` runs outside `write_file()`'s own `try`, so a read-only cache directory makes it raise. Only rank 0 runs `write_file()`, so only rank 0 raises, and it leaves the broadcast the other ranks are sitting in. The next collective then mis-pairs:

```
libc++abi: terminating due to uncaught exception of type gloo::EnforceNotMet:
[enforce fail at gloo/transport/uv/pair.cc:253] op.nread == op.preamble.nbytes
```

That is rank 0 broadcasting a 2-element list from `get_compile_folder()` into rank 1's 1-element receive. On main, a read-only cache directory hard-kills a multi-GPU job instead of falling back to a tempfile as intended. The outcome is now reported rather than raised, so every rank takes the fallback together.

**2. Import failure.** An import can fail on one rank and succeed on another, and its tempfile recovery calls two collectives. Rank 0's answer is the wrong thing to broadcast here because the condition is genuinely rank-local, so this adds `distributed_any()` and moves every rank onto the recovery when any rank needs it. On main:

```
[rank 0] RETURNED ok, cp_fn(21)=42
<rank 1 stuck in the recovery collective>
```

### Verification

| scenario | main | this branch |
| --- | --- | --- |
| cold cache, rank 1 late | deadlock | both ranks complete |
| write fails on rank 0 only | gloo abort, job killed | both recover to tempfile |
| import fails on rank 1 only | deadlock | both ranks complete |
| stale file on one rank | deadlock | `RuntimeError` naming both digests |
| missing file on one rank | deadlock | `FileNotFoundError` after 5s |
| file published 1.5s late | deadlock | retry succeeds |
| warm cache | ok | ok |
| `UNSLOTH_COMPILE_OVERWRITE=0` | ok | ok |
| single process | ok | identical output bytes, zero extra file reads |

Single-process behaviour is unchanged by design: `distributed_function()` short-circuits without a process group, so the decision reduces to the old expression and no hashing happens at all.

Full suite before and after: `186 failed, 3959 passed, 264 skipped, 4 errors`, with byte-identical failure sets. Those failures are pre-existing on main in this environment.

`tests/test_compiled_cache_collective.py` adds 13 tests. 12 of them fail on main; the 13th passes on both because it pins single-process behaviour as unchanged.

### One thing to check before merge

`distributed_any()` uses `dist.all_gather_object`, which I only exercised on gloo. It resolves its device the same way `broadcast_object_list` does, and `distributed_function()` already relies on that, so it adds no new backend assumption. Still worth one NCCL run to confirm.

shimmyshimmer:
Ran a wider simulation matrix over this branch. It turned up two defects in what I pushed earlier, both fixed in 98dccfc4.

**`current_rank()` instead of `os.environ["RANK"]`.** `RANK` is unset when a process group is created without a launcher, so every rank named itself rank 0 in the cache mismatch errors, which is the one detail those messages exist to carry.

**The import recovery no longer blames the wrong rank.** A rank whose own import succeeded now joins the recovery when another rank needs it, so its `import_error` is `None` and the log read `Standard import failed for X: None`. It now distinguishes the two cases.

### Coverage

53 simulations across world sizes 1, 2, 3 and 4, plus 30 platform and edge cases, plus the suite.

End to end against the real `create_new_function` under torchrun and gloo:

| case | main | branch |
| --- | --- | --- |
| cold cache, staggered rank | deadlock | ok at ws 1, 2, 4 |
| 30 sequential compiles, random per-rank stagger | gloo abort, exitcode -6 | ok at ws 2, 3, 4 |
| read-only cache dir | gloo abort, exitcode -6 | ok, recovers to tempfile |
| import fails on the last rank only | deadlock | ok at ws 2, 4 |
| import fails on every rank | ok | ok |
| warm cache, `UNSLOTH_COMPILE_OVERWRITE=0`, non-ASCII source | ok | ok |
| torchelastic launched, group never initialised | ok | ok |

The burst case is the closest to real use, since `PatchFastRL` compiles a long run of functions back to back. On main neither rank completes a single one of the 30.

Divergent cache views at the same path, ws 1 to 4: stale file, missing file, zero byte file, file published late, and several ranks divergent at once. Every rank reaches the end in all of them, and only the divergent ranks raise.

Platform and edge cases: CRLF against LF, non-ASCII sources, zero byte and 2.4 MB files, Windows style paths in the error text, symlinked cache dir, case insensitive filesystem, file deleted between decision and verification, unreadable file, retry bound, and truthiness of `distributed_any`.

Worth noting one of these: `utils.py` documents that ROCm on Windows ships `torch.distributed` as a stub missing attributes entirely. `distributed_any()` calls `dist.get_world_size` and `dist.all_gather_object` directly, so I simulated that stub to confirm the `is_initialized()` early return keeps it from ever reaching them. It does, on import and on call.

### Regression

Single process output is byte identical to main, ASCII and non-ASCII alike. Full suite on both trees: `186 failed, 3959 passed, 264 skipped, 4 errors`, failure sets identical, all pre-existing in my environment.

`tests/test_compiled_cache_collective.py` is now 15 tests. 14 fail on main.

### Not covered

I ran this on macOS arm64. The logic has no platform branches and the Windows and Linux specific hazards above were exercised through simulation rather than on those systems, but a real run on each would still be worth having, as would one NCCL run for `distributed_any`.

sstamenk:
Pushed some additional changes:

- Verifies cache contents after both writing and retaining cached files.
- Coordinates verification failures before any rank raises.
- Loads tempfile recovery modules directly on every rank, avoiding stale `sys.modules` entries.
- Republishes recovered modules under their normal import name.
- Writes node-local tempfile caches on every rank, enabling multi-node recovery.
- Coordinates write, verification, import, and direct-load failures across ranks.
- Preserves the rank diagnostics and recovery logging from `98dccfc4`.

### Validation

- **114 passed, 1 skipped** after rebasing.
- Seven deliberate bug mutations were all caught by regression tests.
- Distributed gloo scenarios passed across world sizes 1–4, covering:
  - staggered and burst compiles
  - shared, rank-local, and per-node caches
  - stale, missing, delayed, and retained files
  - write, import, and direct-load failures
  - `UNSLOTH_COMPILE_OVERWRITE=0`
  - torchelastic without an initialized process group
  - Unicode generated source
- Reverse-order and repeated-process test runs pass without global-state leakage.

sstamenk:
Fixed a few other issues that popped up during Codex review and added unit test for these cases. Not sure if the number of tests is a bit overkill...

sstamenk:
## Distributed compiled-cache hardening status

The follow-up now covers the original rank-local guard deadlock and the additional synchronization, filesystem, recovery, and module-state issues found during review.

These latest changes are currently local and have not been pushed yet.

### Behavior covered

| Scenario | Behavior after this PR |
|---|---|
| Single-process execution | Preserves independent execution; no distributed verification or collectives are introduced before a process group exists. |
| Torchrun launched before `init_process_group()` | Each rank behaves independently, as before. If temp-mode state diverges before initialization, it is collectively reconciled once the process group becomes available. |
| Cold shared persistent cache | Rank 0 decides and writes. Every rank follows the broadcast decision and verifies its local view against rank 0’s digest. |
| Rank reaches the guard after rank 0 writes | All ranks still enter the same collectives; the original unmatched-barrier hang is eliminated. |
| Warm persistent cache | Rank 0’s retained on-disk bytes are authoritative and verified on every rank. |
| `UNSLOTH_COMPILE_OVERWRITE=0` | Rank 0’s readable retained bytes are preserved and propagated to missing/stale node-local temp caches. |
| Retained rank-0 cache is unreadable | Since those bytes cannot be preserved, every rank regenerates from rank 0’s generated source rather than failing independently. |
| Persistent cache missing/stale on another node | Ranks wait up to the existing five-second shared-filesystem visibility grace period, then collectively recover to node-local temp caches. |
| Network filesystem publishes the file late | Verification retries with bounded exponential backoff; reproduced delayed publication succeeds without false failure. |
| Persistent cache changes between decision and verification | The mismatch is agreed across ranks and recovered through temp instead of becoming a fatal or misleading shared-filesystem error. |
| Explicit temp-cache request | Each rank resolves `tempfile.gettempdir()` locally; rank 0’s node-specific path is never broadcast as another node’s path. |
| Rank 0 implicitly falls back to temp | The fallback mode is broadcast, but every rank independently resolves and creates its own local temp directory. |
| Temp mode diverged before process-group initialization | `distributed_any()` reconciles the state before any rank can skip the folder collective. |
| Rank 0 temp cache is warm, worker temp cache is missing | Every worker creates its local copy using rank 0’s canonical bytes. |
| Rank 0 temp cache is warm, worker temp cache is stale | Every node-local copy is repaired to rank 0’s retained bytes, preserving overwrite opt-out. |
| Generated source differs across ranks | All node-local writes use rank 0’s broadcast generated source and digest; local generated source cannot silently diverge. |
| Rank 0 persistent write fails | The outcome is broadcast and all ranks enter node-local tempfile recovery together. |
| One node cannot create/write its temp cache | The local error is agreed before any rank continues; every rank fails together rather than desynchronizing. |
| Import fails on one rank only | `distributed_any()` moves every rank onto recovery. Every rank loads the same rank-0 source directly from its local temp file. |
| Direct module load fails on one rank | The failure is agreed globally, all aliases are removed, and no successful rank is left with an importable implementation after the group failed. |
| Recompile the same generated module name | Plain and recovery aliases are evicted before importing; the newly verified file is loaded instead of a stale `sys.modules` entry. |
| Same-size source rewritten within one second | Local timestamp-based `.pyc` files are removed before both normal and direct imports, preventing stale bytecode execution. |
| Temp recovery switches folders | The temp folder’s own `.pyc` is invalidated, not only the persistent folder’s cache. |
| Another `sys.path` entry contains the same module name | The verified cache directory is temporarily moved to the front, then the original `sys.path` ordering is restored in place. |
| Import fails after temporarily modifying `sys.path` | The original path is restored before recovery, so ranks do not retain divergent import environments. |
| Direct execution fails partway through module initialization | Partial `unsloth_cache_<name>` entries are removed from `sys.modules`. |
| Pre-existing cache read fails before the first collective | The error cannot escape rank-locally; the rank-0 decision remains authoritative and overwrite opt-out cannot undo the required rewrite. |
| Unicode generated source | Source hashing, writing, verification, and import remain UTF-8 consistent. |

### Implementation summary

- Rank 0 broadcasts write/retain decisions.
- Every rank verifies the exact bytes it will import.
- Verification errors are gathered before any rank raises.
- Rank 0’s generated or retained source is broadcast lazily only when node-local writes require it.
- Shared persistent writes remain rank-0-only.
- Node-local temp writes run on every rank and remain idempotent through `FileLock`.
- Persistent-cache disagreement recovers to temp; temp-cache disagreement is repaired in place.
- Temp verification is immediate; the five-second grace period remains only for persistent/shared filesystem visibility.
- Import recovery evicts stale aliases, loads by path, republishes the successful module under its normal name, and cleans all aliases on global failure.
- Source-only regression tests read `compiler.py` directly, while runtime tests skip cleanly in zoo-only environments.

### Validation

- **42 targeted compiled-cache tests passed**
- **132 passed, 1 skipped** across the targeted and neighboring compiler suites
- Zoo-only environment: **5 passed, 37 skipped**
- Reverse-order and repeated-process runs pass without leaking compiler global state
- Real gloo simulations pass at world sizes 1–4, including:
  - staggered and burst compilation;
  - shared and rank-local persistent caches;
  - distinct per-rank `TMPDIR` values;
  - mixed warm/cold/stale temp caches;
  - divergent generated source on every rank;
  - retained overwrite-opt-out source;
  - delayed filesystem publication;
  - rank-local mkdir/write/import/direct-load failures;
  - pre-initialization temp-mode divergence.
- Mutation tests cover the write decision, collective verification, source/digest selection, temp-path locality, alias eviction, partial-module cleanup, and stale-bytecode paths.
- A final Opus 5 high adversarial review traced every collective in the changed code and found the original deadlock/divergence class closed after the final unreadable-retained-cache and temp-pyc fixes.

### Separate pre-existing follow-up

There are rank-local skip conditions in callers of `create_new_function()` (`patch_lora_forwards`, `_patch_torch_dtype_modules`, and `unsloth_compile_transformers`). Independent reviewers classified those as pre-existing caller-side architecture risks rather than regressions introduced by this PR. They should be tracked separately rather than expanding this cache-recovery change further.

shimmyshimmer:
@sstamenk your eight commits landed while I was building a wider simulation matrix, so I pointed it at your head (7e662e15) instead of at mine. Everything passes, including the three cases I had been about to push my own fix for. I dropped that commit rather than duplicate yours, which is broader.

### What was run against 7e662e15

All of it on gloo, driving the real `create_new_function` under `torchrun` unless noted.

| suite | coverage | result |
| --- | --- | --- |
| end to end | 18 scenarios at world sizes 1, 2 and 4 | all ranks finish |
| burst | 30 sequential compiles, random per-rank staggering, ws 2 and 4 | all ranks finish |
| divergence | 8 rank-local cache states at ws 4 | correct outcome on every rank |
| module identity | `UNSLOTH_COMPILE_OVERWRITE=0` plus a one-rank import failure, ws 2 and 4 | every rank runs the same implementation |
| platform and edge | 30 checks on two Python and torch combinations | 30 of 30 |
| test file | `tests/test_compiled_cache_collective.py` | 42 passed |

The scenarios are cold cache with a staggered rank, warm cache, `UNSLOTH_COMPILE_OVERWRITE=0`, read-only cache directory, silently failed write, import failure on one rank, import failure on every rank, non-ASCII source, and torchelastic launched without `init_process_group`.

### The three problems are genuinely closed

I reproduced each one first so the simulations were known to discriminate, rather than trusting a green run.

**Verification raised rank-locally.** Two ranks, rank 1 divergent: rank 1 raised and exited, rank 0 walked into the next collective and blocked forever. That is the same hang this PR removes, reintroduced by the verification itself. On your head every rank raises together and none is left waiting.

**The write path was unchecked.** At four ranks with a rank-local cache path, three ran the current implementation and one ran a stale one, silently. Your head raises on all four.

**The recovery kept a stale module.** With `UNSLOTH_COMPILE_OVERWRITE=0` and an import failing on one rank, the ranks that succeeded kept the older module from `sys.modules` while the recovered rank loaded the new source. Measured `['OLD', 'OLD', 'OLD', 'NEW']` before, `['NEW', 'NEW', 'NEW', 'NEW']` on your head. Loading by path is a better answer than the eviction I had written.

### One thing I specifically tried to break

Verifying the write path could pre-empt the tempfile recovery: `write_file()` swallows its own errors and reports success, so a write that lands nowhere would be caught by verification and could turn a recoverable run into a hard failure. It does not, because a verification failure routes into the fallback rather than raising, and `test_a_silently_failed_write_recovers_instead_of_killing_the_job` pins it. Confirmed end to end by making `fsync` delete the file and raise: main, my branch and yours all recover.

Worth flagging for anyone rerunning this: my first version of that injection silently did nothing, because `readlink /dev/fd/N` returns EINVAL on macOS, and all three trees "passed" a test that never ran. It needs `/proc/self/fd` on Linux or `F_GETPATH` on macOS. I would not have caught that without checking the injection fires.

Also worth noting, `torch_distributed_is_initialized()` in `_compiled_cache_decision` is the right call over `is_distributed()`. Under torchelastic before `init_process_group` the latter is true while no collective exists.

### Regressions

Full suite on your head against main: `187 failed, 4000 passed, 264 skipped, 4 errors` versus `186 failed, 3959 passed`. The single extra failure is `test_pypi_version_sync.py::test_pypi_version_is_not_ahead_of_main`, which fails on main too when run on its own, so it is a flaky network check and not from this branch.

### Still open before merge

`distributed_any()` and `all_gather_object` have only been exercised on gloo here, since I have no GPU on this machine. They resolve their device the same way `broadcast_object_list` does and `distributed_function()` already depends on that, so no new backend assumption, but one NCCL run would close it out. If you have a multi-GPU box handy, the cold cache and one-rank import failure cases are the two worth repeating there.

sstamenk:
@shimmyshimmer can this be merged? Do you need anything else from my side? Thanks

danielhanchen:
Confirmed the rank-local guard around the write collective is still in create_new_function in unsloth_zoo/compiler.py, so a rank arriving after rank 0 writes skips the barrier and the job hangs, and a stale file at the same path silently imports a different implementation. Thanks for the gloo coverage, will get this queued for merge review.

sstamenk:
Validated the PR behavior on a cluster of 8x AMD Radeon AI Pro R9700S using RCCL.

Updated the PR description to reflect this testing.

danielhanchen:
Reviewed this end to end and pushed four commits on top of your head (`0812d0e4` -> `a2312b3a`). The fix is real and the diagnosis holds up; everything below is either evidence for that or the three things I changed.

## The deadlock reproduces on main and is closed here

Two to four gloo ranks under `torchrun`, driving the real `create_new_function`. One thing worth flagging for anyone re-running this: the delay has to be injected **after** `get_compile_folder`, not before `create_new_function`. `get_compile_folder` ends in `distributed_function`'s `barrier()`, which re-synchronises the ranks before they reach the guard, so a sleep at the top of the call proves nothing and both trees pass.

| scenario | world | main | this branch |
| --- | --- | --- | --- |
| cold cache, last rank staggered 3s | 2, 4 | fail 10/10 | pass 10/10 |
| read-only cache dir | 2, 4 | fail 10/10 | pass 10/10 |
| import fails on one rank only | 2, 4 | fail 10/10 | pass 10/10 |
| 30 sequential compiles, staggered | 2, 4 | fail 3/3 | pass 3/3 |

Failure modes on main match your description exactly: the cold stagger leaves rank 0 abandoned in `broadcast_object_list`, and the read-only directory kills rank 1 with `exitcode -6`, the gloo `EnforceNotMet` abort.

Also confirmed: compiled cache output is byte-identical to main, and a seeded 21-step LoRA run differs from main by less than main differs from itself across three runs (max loss delta 0.0032 vs a 0.0040 run-to-run noise floor).

## What I changed

**1. The bytecode removal was unconditional.** `import_module` and `load_module_directly` removed the `.pyc` on every call, including a warm retained cache, which is the path every process start walks once per generated module. Measured on 20 modules of 107KB, fresh process against a warm cache:

| | per call | median `import unsloth` |
| --- | --- | --- |
| main | 7.4 ms | 10.50 s |
| this PR as submitted | 15.3 ms | 11.48 s |
| with the fix | 8.4 ms | 10.19 s |

Proven causal by isolation: neutralising only `_remove_compiled_cache_bytecode` on your head returned it to 8.6 ms.

It is also a Windows-only hard failure. `os.remove` raises `PermissionError` (WinError 32) whenever another handle is open on the file, which a virus scanner, a sync client or a second interpreter all do routinely, and the compiled-cache lock excludes none of them. `PermissionError` is an `OSError`, so it became the `RuntimeError`, which sent the group into tempfile recovery on the first occurrence and aborted at the `raise agreed_load_error` on the second. Baseline never removed a pyc and imported fine in all of those cases.

So the removal is now gated on whether the call actually rewrote the file, and a failed unlink warns instead of raising. This does soften the "surface bytecode-removal failures" item from an earlier review, and I want to be explicit about the trade rather than slip it past: after the gate, the removal only runs when we just rewrote the source, so the `.py` mtime is already newer than the `.pyc` and CPython recompiles anyway. It only matters for a same-size rewrite landing inside one timestamp tick, and hard-failing a Windows user to defend that corner is the wrong way round. `test_rewriting_the_cache_still_removes_stale_bytecode` pins that the gate is not too aggressive.

**2. `sys.modules.pop` un-masks a latent bug in `_patch_torch_dtype_modules`.** The pop itself is right and I left it alone. But that function's source-rewrite branch never set `__unsloth_dtype_wrapped__`, unlike its wrapper branch, so a second pass reads back its own generated forward through `inspect.getsource` and inserts a second dtype prologue. `original_dtype` then binds to the weight dtype rather than the caller's:

```
pass1: OUTPUT dtype=torch.bfloat16   prologue_count=1
pass2: OUTPUT dtype=torch.float32    prologue_count=2   <- your head
pass2: OUTPUT dtype=torch.bfloat16   prologue_count=1   <- with the fix
```

The doubly rewritten file was always being written; a stale `sys.modules` entry was the only thing stopping it from executing. This is reachable on an ordinary vision load, because `loader.py` prepends `siglip` to `model_types` and `_utils.py` loops over them, so the global `torch.nn` classes get patched at least twice per load. SigLIP's patch-embedding `Conv2d` sits on exactly that path.

**3. CI was not running any of the 44 tests.** Neither file appeared anywhere under `.github/`; `repo-tests-cpu` uses a hard-coded file list and the only other job is `--collect-only`. Since 42 of the 44 fail against the code they fix, the coverage was there and the gate was not. Both are now in the hard gate. Note your branch is 66 commits behind main and that workflow has been restructured there, so this hunk will probably need re-applying after a rebase.

## Cross-platform and compatibility

I built 46 simulation cases in an isolated CPU-only `uv venv` covering Windows/macOS/WSL/Linux primitives and ran them against main, your head, and the fixed branch:

| | main | your head | fixed |
| --- | --- | --- | --- |
| simulation suites | 40 pass, 2 fail, 4 skip | 43 pass, **3 fail** | **46 pass** |

The three failures on your head are the Windows hazards above. The suite covers `os.remove` sharing violations, read-only `__pycache__`, `cache_tag = None`, `PYTHONDONTWRITEBYTECODE`, `sys.pycache_prefix`, Windows/macOS/WSL `gettempdir()` shapes, per-user temp divergence between ranks, case and separator aliasing on `sys.path`, module shadowing, cp1252 / latin-1 / BOM / CRLF / zero-byte / corrupt / multi-MB cache files, an old cache with no versioning header, an old cache with a stale `__pycache__` entry, `UNSLOTH_COMPILE_OVERWRITE=0` retention, and a torch built with `USE_DISTRIBUTED=0`.

That last one is worth calling out: per torch's own `distributed/__init__.py`, `USE_DISTRIBUTED=0` is the **default on macOS**, not just ROCm-on-Windows as the comment in `utils.py` suggests. In that mode `torch.distributed` exposes only `is_available` and a `ProcessGroup` stub. I asserted with an exploding stub that `distributed_any`, `current_rank`, `distributed_function`, `_agreed_error` and a full `create_new_function` never resolve a single missing attribute. They do not, on any of the three trees. Your `torch_distributed_is_initialized()` guard is what carries that, and it holds.

On the device axis: nothing in the diff references cuda, rocm, hip, xpu, mps, device_type or any backend, so I asserted the [Windows, Linux, WSL, macOS] x [NVIDIA, AMD, CPU] product produces byte-identical cache output in all 12 cells rather than pretending to run 12 machines. It does.

Backwards compatibility with an existing on-disk cache is fine, and in one respect better than main: main had no `try` around the initial cache read, so a cp1252 cache raised `UnicodeDecodeError` straight out of `create_new_function`, while this PR tolerates it.

## Still not covered

No NCCL or RCCL run from me. Only one GPU is visible here and oversubscribing it with two NCCL ranks is not validation, so your 8x R9700S run remains the only evidence for that path. Gloo proves the collective ordering, not the NCCL device selection. Also no multi-node and no network filesystem.

Two pre-existing things I did not touch, both worth a separate issue rather than widening this diff:

- A UTF-8 BOM shifts the version-header index, because `'﻿'.isspace()` is False, so `cached_lines[2]` becomes the unsloth version instead of the transformers one. Under `UNSLOTH_COMPILE_OVERWRITE=0` that is a permanent spurious forced recompile. Most likely on Windows.
- `distributed_function` and `distributed_any` call `broadcast_object_list` / `all_gather_object` / `barrier` with no `device=`. Under NCCL those land on `torch.cuda.current_device()`, which is `cuda:0` on every rank unless the caller already ran `set_device`, and this code runs at import time before accelerate sets it. Pre-existing, but this PR takes the per-module count from about 2 to about 7, so it is worth passing the device explicitly.

danielhanchen:
@codex review

danielhanchen:
@codex review

danielhanchen:
@codex review

danielhanchen:
@codex review

danielhanchen:
@codex review

danielhanchen:
@codex review

danielhanchen:
@codex review

danielhanchen:
@codex review

danielhanchen:
@codex review

danielhanchen:
@codex review

danielhanchen:
Verified this on real NCCL across two hardware platforms, plus DDP and balanced regression runs. Summary of what I measured.

## The deadlock is real, and reproduces on NCCL

Previous testing here only covered gloo. With two GPUs available I ran the four divergence scenarios under real NCCL, 4 reps each:

| scenario | main | this branch |
|---|---|---|
| cold cache stagger | watchdog abort 4/4 | pass 4/4 |
| compile burst | watchdog abort 4/4 | pass 4/4 |
| read-only cache dir | watchdog abort 4/4 | pass 4/4 |
| one-rank import failure | watchdog abort 4/4 | pass 4/4 |

Every failure on main carries a named abort rather than a bare timeout:

```
Watchdog caught collective operation timeout:
  WorkNCCL(SeqNum=7, OpType=ALLREDUCE, NumelIn=1, NumelOut=1, Timeout(ms)=25000)
PG status: last enqueued work: 7, last completed work: 6
```

Reproduced independently on Kaggle 2x Tesla T4 (sm_75, torch 2.10): main gives the same watchdog abort, this branch passes.

## One consequence is worse than described

The description says the group stays desynchronized until the watchdog terminates the job. What happens first is a wrong answer. After the desync, the surviving rank's next collective returns garbage:

```
[rank 1] HEALTHCHECK FAILED wrong value 4.6842973944194455e-11
```

The identical value appears on both a B200 and a T4, so this is deterministic mismatched-collective behaviour. In a real run the next collective is a gradient all-reduce, so a step can be computed on corrupt gradients before anything crashes. Worth stating, since silent corruption is more serious than a hang.

## Suggested correction to the description

The description says the hang happens "commonly during `import unsloth` while `PatchFastRL` compiled TRL trainers". That path is not reachable: `distributed_function` returns early when `is_initialized()` is False (`unsloth_zoo/utils.py:165`), and `PatchFastRL` runs at module scope before any `init_process_group`, so under bare torchrun it takes no collectives.

The reachable window is compilation after the group exists: `SFTConfig`/`TrainingArguments` construction initialises the group, and the `get_peft_model`, `patch_lora_forwards` and trainer compiles that follow all run inside it. Ordinary ordering, so the bug is real, but the trigger is narrower than stated.

## Nothing regressed

DDP, 2 GPUs, 20 steps, both trees, cold and warm cache, 8 records. Distinct values across all 8:

| quantity | distinct |
|---|---|
| loss curve | 1 |
| grad-norm sequence | 1 |
| LoRA tensor SHA-256 | 1 |
| compiled-cache manifest SHA-256 | 1 (16 files) |
| peak allocated | 1 (1409.4 MiB) |

Identical, not merely within tolerance. The matching LoRA checksum across ranks is what shows the two processes really were synchronised; the matching cache manifest shows the rewritten cache path generates byte-identical output to main.

Balanced sharding checked with the official `Kaggle-Magistral_(24B)-Reasoning-Conversational.ipynb` on 2x T4, unmodified apart from the zoo install. 24B in 4-bit does not fit on one T4, so it genuinely shards: both trees give 46 `hf_device_map` entries, placements `['0', '1']`, and identical per-GPU allocation (5413.7 and 8940.5 MiB).

## One risk I checked and cleared

`distributed_any` is new here and calls `all_gather_object` with no `device=`, as does the existing `broadcast_object_list`. Under NCCL those use the current CUDA device, so ranks that never call `set_device` could both pick `cuda:0`. Tested with a scenario that deliberately omits `set_device`: identical on both trees, because NCCL itself refuses at the first collective with `Duplicate GPU detected`, symmetrically on both ranks. Pre-existing NCCL semantics, fails loudly rather than hanging. No change needed.

## Not covered

Single node only, so node-local tempdir divergence and delayed NFS visibility are emulated rather than tested on a second host. No RCCL/AMD from me; the 8x AMD run in this PR stands. The Windows CI failures are pre-existing: all of them trace to `import triton` at `compiler.py:55`, which is unchanged on main and untouched here.

danielhanchen:
Ran a dependency compatibility matrix over this branch, since the cache path is touched by every compile.

## What the diff can even be sensitive to

Across the whole production change this PR adds exactly one import, `hashlib` (stdlib), and touches two `torch.distributed` symbols, `all_gather_object` and `get_world_size`. It calls no transformers, TRL, PEFT or vllm API anywhere. Both dist symbols have existed with unchanged signatures since torch 1.8, so the version surface is small by construction.

## Measured, not assumed

Each combo runs the compiler and cache tests that predate this PR, so they exist on both main and this branch and the two are directly comparable:

| combo | transformers | TRL | PEFT | torch | main | this branch |
|---|---|---|---|---|---|---|
| floor | 4.57.6 | 0.22.2 | 0.18.0 | 2.9.1+cu128 | 150 passed | 150 passed |
| major edge | 5.0.0 | 1.0.0 | 0.19.1 | 2.9.1+cu128 | 150 passed | 150 passed |
| ceiling | 5.16.1 | 1.12.0 | 0.20.0 | 2.9.1+cu128 | 157 passed | 157 passed |
| torch floor | 4.57.6 | 0.22.2 | 0.18.0 | 2.6.0+cu124 | 150 passed | 150 passed |

Identical on every combo. The ceiling row collects 157 rather than 150 because newer transformers and TRL enable more parametrised cases, and both trees pick up the same extra seven.

The new tests added by this PR also pass 52/52 under all four combos, including torch 2.6.0 and transformers 5.16.1 with TRL 1.12.0.

Both major version boundaries are covered directly rather than by interpolation: transformers 4.57.6 to 5.0.0, and TRL 0.22.2 to 1.0.0.

## Not covered

vllm and `fast_inference = True` GRPO were not exercised here, and neither was MLX on Mac. Neither path is touched by this diff, but that is an argument rather than a test, so I am not claiming them.

danielhanchen:
@codex review

PR #968, opened 2026-07-30
feat(mlx): pooling and contrastive-loss primitives for embedding training

BardiaKoopah:
This adds the **library pieces only** — pooling, three contrastive losses, and a sentence-transformers layout shim. It does **not** integrate with `MLXTrainer`, and it covers decoder-only backbones only.
 
Integration is deliberately deferred: the trainer's loss contract is `loss_fn(model, batch, lengths, labels=None)` — one token sequence per row — and contrastive training needs two or three aligned sequences per row. Carrying a pair through that signature needs a separate entry point plus data-path work, which is its own change.
 
## Coverage, stated up front
 
BERT, RoBERTa, MiniLM and mpnet fail with `ValueError: Model type bert not supported` — there is no BERT implementation in `mlx_lm` at all — so the families most sentence-transformers checkpoints actually use are out of reach.
 
| model | result |
|---|---|
| `Alibaba-NLP/gte-Qwen2-1.5B-instruct` | loads unmodified |
| `Qwen/Qwen3-Embedding-0.6B` | loads via the layout shim |
| BERT / RoBERTa / MiniLM / mpnet | not supported by `mlx_lm` |
 
The shim exists because sentence-transformers stores the inner transformer at repo root, so all 310 keys arrive without the `model.` prefix and loading raw fails with `Received 310 parameters not in model`. A test asserts the remapped weights load and that `mx.array_equal` matches the checkpoint tensor.
 
## Pooling
 
All six modes from `sentence_transformer.py:587-599` — `cls`, `mean`, `max`, `mean_sqrt_len`, `weightedmean`, `lasttoken` — all mask-aware. CUDA only detects the mode and delegates to sentence-transformers; MLX has to implement them.
 
Padding invariance is the correctness risk: a mask-ignoring mean is silently wrong rather than loud. The fixture pads with junk at 100× scale so the naive mean differs by **26.361**, asserted directly so the test can actually fail. All six modes match the unpadded result to `0.000e+00`.
 
Reading `1_Pooling/config.json` matters too — Qwen3-Embedding declares `lasttoken`, not `mean`.
 
## Losses
 
Each verified over 8 steps with **separation** asserted, not just a falling loss:
 
| loss | loss | separation |
|---|---|---|
| MultipleNegativesRanking | 2.077 → 0.675 | +0.0001 → +0.2426 |
| CoSENT | 2.838 → 0.107 | −0.0002 → +0.4911 |
| triplet | 0.500 → 0.190 | −0.0002 → +0.2846 |
 
CoSENT's two groups move apart independently: labelled-similar 0.9951 → 0.9977, labelled-dissimilar 0.9954 → 0.5066. The synthetic encoder starts collapsed at ~0.995 cosine, so "positive-pair cosine rises" would have been the wrong assertion — separation is the metric that tests the objective.
 
## Batch-size advisory
 
`recommend_batch_size` derives from `peak_GB ≈ 1.12 + 0.79 × batch`, fitted on measured peaks at batch 4 / 8 / 16 → 4.76 / 7.92 / 14.24 GB (0.6B backbone, L=512).
 
It is deliberately one notch conservative against an 85% memory budget — it returns 15 on a 16 GB machine where batch 16 measured 14.24 GB — and it raises nothing, so a user who knows their own headroom can ignore it.
 
## Tests
 
37 tests, split so the Linux gate executes the shim-runnable half.
 
The numeric tests bind `mlx.core`, `mlx_lm` and `mlx.utils` at **module** import, not inside test bodies. The repo's `mlx_simulation` shim replaces `sys.modules` entries session-wide, so a function-level import silently resolves against the stub — the test then passes against a stub while looking clean standalone. This was caught only by a full-suite run, and it is worth knowing for any future MLX test in this repo.
 
Full zoo suite run twice per side: collected totals 2979 → 3016, delta 37 = the tests added. No regressions, no environment writes. `trainer.py` is untouched.

### DIFF
--- a/.github/workflows/consolidated-tests-ci.yml
+++ b/.github/workflows/consolidated-tests-ci.yml
@@ -103,7 +103,11 @@ jobs:
         # + cache-completeness gate (CPU-pure; collect-only would let it
         # regress silently). test_train_on_responses_list_labels pins the
         # tensor-vs-list labels contract for response masking (stub tokenizer,
-        # no weights). Executing them here keeps fixture/source drift visible.
+        # no weights). test_mlx_embedding_guards pins the sentence-transformers
+        # layout shim and pooling-mode map; it runs under the torch shim so it
+        # EXECUTES here rather than skipping (the numeric half needs real MLX and
+        # runs on the Apple Silicon job). Executing them here keeps fixture/source
+        # drift visible.
         run: |
           python -m pytest \
             tests/test_training_utils_use_cache.py \
@@ -112,6 +116,7 @@ jobs:
             tests/test_hf_xet_fallback.py \
             tests/test_get_transformers_model_type_empty.py \
             tests/test_train_on_responses_list_labels.py \
+            tests/test_mlx_embedding_guards.py \
             -v
 
       - name: pytest tests/test_mlx_module_exports + zoo-specific CPU tests
--- a/tests/test_mlx_embedding_guards.py
+++ b/tests/test_mlx_embedding_guards.py
@@ -16,14 +16,10 @@
 
 """Sentence-transformers layout shim, pooling-mode detection, batch advisory.
 
-Pure dict/string/arithmetic logic, so this runs under the torch shim on Linux CI
-rather than skipping. The numeric behaviour lives in
-test_mlx_embedding_pooling.py, which needs real MLX.
-
-The layout shim matters because Qwen/Qwen3-Embedding-0.6B -- the obvious model for
-this feature -- does not load without it: sentence-transformers stores the inner
-transformer at the repo root, so all 310 keys arrive without the ``model.`` prefix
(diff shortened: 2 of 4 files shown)


BardiaKoopah:
@codex review

danielhanchen:
Checked the new pooling and loss primitives in unsloth_zoo/mlx/embedding.py plus the sentence-transformers prefix remap, and the padding-invariance test looks like the right thing to pin. Since nothing calls these yet, could you sketch how the paired-sequence entry point into MLXTrainer would look so we can review both together?

danielhanchen:
Checked that unsloth_zoo/mlx has no embedding module today, and that this lands as standalone pooling and loss primitives with trainer.py untouched. Could you sketch what the MLXTrainer entry point for paired sequences would look like so I can see where this is heading before reviewing?

PR #976, opened 2026-08-04
Feat(mlx): GGUF multi-quant

BardiaKoopah:
`save_pretrained_gguf` crashes on the list form of `quantization_method` that the CUDA path
  documents, so an MLX user cannot produce more than one GGUF per export.

  ## The bug
  
  CUDA states the contract in the signature itself (`unsloth/save.py:1323`):

  ```python
  quantization_method = "fast_quantized",  # Can be a list of options! ["q4_k_m", "q8_0", "q5_k_m"]
  ```

  and normalizes it to a list at `save.py:1342-1352`. On MLX the argument reached a dict lookup keyed
  by the whole value:

  ```python
  quant_type = quant_map.get(quantization_method, quantization_method)
  ```

  A list is unhashable, so this raises:

  ```
  TypeError: unhashable type: 'list'
    unsloth_zoo/mlx/utils.py:12488
  ```

  It is reachable from the user-facing API, not just the internal helper: `model.save_pretrained_gguf`
  is bound in `loader.py` and forwards `quantization_method` untouched.

  Beyond the crash, the single-target restriction is wasteful. The LoRA merge and the
  `convert_hf_to_gguf` pass are the expensive steps, and both are independent of the target quant —
  producing q4_k_m and q8_0 meant running the whole pipeline twice.

  ## What changed

  `quantization_method` resolves to an ordered list of llama.cpp types, and each target is quantized
  off the single shared intermediate. Extra targets now cost only their own `llama-quantize` pass.

  ```python
  model.save_pretrained_gguf(out, tokenizer, quantization_method=["q4_k_m", "q8_0", "q5_k_m"])
  # one merge, one convert, three GGUFs
  ```

  Scalar strings keep their exact previous semantics. `push_to_hub_gguf` inherits the behaviour and
  already uploads every `*.gguf` it finds.

  ### Deliberate decisions

  - **Targets follow the CUDA rule** (`save.py:1528`): everything that is not the intermediate itself
  gets a `llama-quantize` pass. Full-precision targets are *not* exempt — `llama-quantize` emits
  f16/bf16/f32 too, so exempting them would silently drop the `f16` in `["f16", "q4_k_m"]`.
  - **The pre-existing single-target contract is unchanged.** A lone full-precision request is
  satisfied by whatever `convert_hf_to_gguf` emitted, with no quantize pass, even when
  `first_conversion` names a different dtype. Only the list form is new behaviour.
  - **Intermediate cleanup is now conditional**, matching CUDA's `quants_created` /
  `want_full_precision` pair. It is removed only once something was quantized off it *and* it is not
  itself a requested type, so `["bf16", "q4_k_m"]` keeps the bf16 file. When no quantize pass runs the
  converted file *is* the export and always stays. The previous code deleted unconditionally whenever a
  quantize ran, which a naive loop would have turned into deleting a requested artifact.
  - **Duplicates are deduplicated, preserving order.** Output filenames derive from the quant type
  alone, so a repeated entry would re-run `llama-quantize` onto a path it just wrote. CUDA does not
  deduplicate; this is a small intentional divergence.
  - **Quant names are still passed through unvalidated**, exactly as before. An upfront allowlist would
  mean duplicating `ALLOWED_QUANTS` from the `unsloth` repo into this one, where the two copies would
  drift. Left as a follow-up rather than smuggled into this PR. Argument-shape errors (non-string
  entries, empty list) are still rejected before the expensive merge.

  ## Tests
  
  Twelve cases in `tests/test_mlx_save_export_edge_cases.py`, covering: the list form producing every
  quant from one merge and one convert; tuple input; per-element alias resolution; deduplication;
  full-precision targets that are not the intermediate (`["f16","q4_k_m"]` and `["bf16","f16"]`);
  intermediate retained when also requested; intermediate retained when nothing was quantized; explicit
  `first_conversion` with a list; scalar-string regression; and forwarding through both
  `model.save_pretrained_gguf` and `push_to_hub_gguf`.

  The bug was reproduced on the parent commit before fixing — the new list test fails there with
  `TypeError: unhashable type: 'list'` at `utils.py:12488`.
  
  Across the 789-test MLX + GGUF suite the failing set is byte-identical to the base commit; the only
  delta is the twelve added tests. The existing GGUF assertions are untouched — the shared scaffold
  gained an accumulating `quantize_calls` list alongside the pre-existing single-call
  `quantize_kwargs`.

  ## CI

  Second commit adds the file to the behavioral-gate job. It was previously only collected, and as that
  step's own comment says, `pytest tests/ --collect-only` proves only that a file imports — the same
  gap that let `test_mlx_finetune_last_n_layers` stay broken from #669 until #739. The file is
  deterministic, CPU-pure, fakes llama.cpp end to end (no binary, no network, no weights) and runs in
  well under a second.

  I fixed one stale number while pulling this up — the Tests section said "ten added tests" from before
  the CUDA-rule fix added two more.

### DIFF
--- a/.github/workflows/consolidated-tests-ci.yml
+++ b/.github/workflows/consolidated-tests-ci.yml
@@ -112,11 +112,18 @@ jobs:
         # CPU-pure and monkeypatch every Hub entry point, so they cost under two
         # seconds. Left out of a gate they would have been collected but never
         # run: `pytest tests/ --collect-only` above proves only that they import.
+        #
+        # test_mlx_save_export_edge_cases pins the MLX GGUF export contract:
+        # which quant targets are produced, that the expensive merge/convert
+        # runs once per export, and when the intermediate is scratch versus a
+        # requested artifact. It fakes llama.cpp end to end (no binary, no
+        # weights) and the whole file runs in well under a second.
         run: |
           python -m pytest \
             tests/test_training_utils_use_cache.py \
             tests/test_mlx_finetune_last_n_layers.py \
             tests/test_mlx_double_quant_reject.py \
+            tests/test_mlx_save_export_edge_cases.py \
             tests/test_hf_xet_fallback.py \
             tests/test_get_transformers_model_type_empty.py \
             tests/test_train_on_responses_list_labels.py \
--- a/tests/test_mlx_save_export_edge_cases.py
+++ b/tests/test_mlx_save_export_edge_cases.py
@@ -1362,6 +1362,71 @@ def test_gguf_scalar_full_precision_keeps_its_pre_pr_behaviour(monkeypatch, tmp_
     assert (out / "EdgeModel.BF16.gguf").exists()
 
 
+# --- Group 11e: intermediate must not down-convert an f32 request -----------
+
+
+def test_gguf_f32_request_converts_directly_to_f32(monkeypatch, tmp_path):
+    """bf16 and f16 both embed exactly in f32, so an f32 intermediate is never a
+    precision regression for the other targets -- but a bf16 intermediate makes
+    the requested F32.gguf f32-shaped bf16 data."""
+    mutils, calls = _gguf_export_scaffold(monkeypatch, tmp_path)
+    out = tmp_path / "out"
+    _export(mutils, out, ["f32", "f16"])
+
+    assert calls["convert_kwargs"]["quantization_type"] == "f32"
+    assert [c["quant_type"] for c in calls["quantize_calls"]] == ["f16"]
+    assert (out / "EdgeModel.F32.gguf").exists()
+    assert (out / "EdgeModel.F16.gguf").exists()
+
+
+def test_gguf_f32_request_alongside_a_kquant_still_converts_to_f32(
+    monkeypatch, tmp_path
+):
+    mutils, calls = _gguf_export_scaffold(monkeypatch, tmp_path)
+    out = tmp_path / "out"
+    _export(mutils, out, ["f32", "q4_k_m"])
+
+    assert calls["convert_kwargs"]["quantization_type"] == "f32"
+    assert {c["input_gguf"] for c in calls["quantize_calls"]} == {
+        str(out / "EdgeModel.F32.gguf")
+    }
+    assert (out / "EdgeModel.F32.gguf").exists()
+    assert (out / "EdgeModel.Q4_K_M.gguf").exists()
+
+
+def test_gguf_without_f32_the_intermediate_stays_bf16(monkeypatch, tmp_path):
+    """Deliberate non-change. f16 and bf16 are not orderable -- bf16 has the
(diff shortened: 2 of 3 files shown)


BardiaKoopah:
@codex review

BardiaKoopah:
Both Codex points are right. The singleton-list one was already fixed in `214ea0f2` — `_normalize_gguf_quantization_methods` now returns a `requested_as_sequence` flag and the full-precision carve-out in step 6 only applies to an original scalar, so `["f16"]` with `first_conversion="bf16"` emits `EdgeModel.F16.gguf` (`test_gguf_singleton_list_full_precision_is_still_emitted`), while the plain string `"f16"` keeps its pre-PR behaviour of no quantize pass at all.

The sharded intermediate one is fixed in `7f052fd6`. `save_pretrained_gguf` was building the llama-quantize input as `f"{output_base}.{first_conversion.upper()}.gguf"` and throwing away `convert_to_gguf`'s return, so once the intermediate crossed `--split-max-size` the converter wrote `{base}-00001-of-000NN.gguf` shards, that plain name never existed and the quantize step got a path that isn't there — and the cleanup only unlinked the same missing name, so the shards leaked too. `_resolve_gguf_intermediate` now takes the paths `convert_to_gguf` returned (falling back to a scan of the output dir, since the existing stubs and the pre-PR contract are the files on disk), quantizes from the first shard and removes the whole set when the intermediate is scratch. Passing the first shard is enough: llama-quantize hands an empty splits vector to `llama_model_loader`, which reads `split.count` and generates the rest via `llama_get_list_splits` (`src/llama-model-loader.cpp:600`), and the output is unsplit because we don't pass `--keep-split`. Three tests cover it — quantizing off shards, the scratch shards being removed whole, and the disk fallback.

Not verified on real weights: reproducing the split needs a ~13B model at f32, over 50GB on disk and well past this machine's 11.84GB Metal working-set ceiling for the merge, so the shard behaviour is exercised through the converter stub and checked against llama.cpp's loader source rather than an actual sharded GGUF.

BardiaKoopah:
@codex review

danielhanchen:
Confirmed the MLX path still keys quant_map on the raw argument at unsloth_zoo/mlx/utils.py:14654, so the documented list form of quantization_method never worked there. Will get this reviewed.

danielhanchen:
Confirmed main still resolves quantization_method through the scalar lookup in unsloth_zoo/mlx/utils.py, so the list form the CUDA path documents does not work on MLX, and reusing the single merge and convert for each target looks right. Will get this reviewed.

danielhanchen:
Confirmed the MLX path still keys the quant map on the whole value in unsloth_zoo/mlx/utils.py, so a list still fails there, and reusing one merge and one convert for several targets is the right shape. Main has moved on quite a bit since this branch (the imatrix handling around that same block), so could you rebase and confirm the list form and the imatrix guards agree?

PR #977, opened 2026-08-04
Feat(mlx): Honor adam_epsilon in MLXTrainer

BardiaKoopah:
`MLXTrainingConfig` mirrors SFTConfig / TrainingArguments field names and already carried
  `adam_beta1` / `adam_beta2` — but not `adam_epsilon`, which did not exist anywhere under
  `unsloth_zoo/mlx`.

  ## The bug
  
  The CUDA path forwards it alongside the betas (`unsloth/trainer.py:349-350`):

  ```python
  betas = (self.args.adam_beta1, self.args.adam_beta2),
  eps   = self.args.adam_epsilon,
  ```

  Because the MLX config validates its kwargs and rejects unknown ones, this was a hard failure
  rather than a silent drop:

  ```
  TypeError: MLXTrainingConfig got unexpected arguments: adam_epsilon
  ```

  So a script ported from CUDA died before the first step, rather than training with the wrong
  epsilon.

  ## What changed
  
  Add the field and forward it, scoped to `adam` / `adamw` exactly as CUDA scopes it. SGD, Muon
  and Lion accept no epsilon, and MLX's Adafactor `eps` is a 2-tuple whose meaning differs from
  HF's scalar, so it is deliberately not routed there.

  `None` means "not requested" and forwards nothing, leaving MLX's own default (1e-8, which is
  also the HF default) in force. The unset path is therefore bitwise unchanged, and only callers
  who set an explicit epsilon see any difference.

  The field is appended last and registered in `_MLX_CONFIG_OPTIONAL_COPY_FIELDS`, keeping those
  an exact suffix of the field list: the config binds positional arguments by field order, so
  inserting mid-list would shift every later field's slot.

  Consistent with HF and with the sibling betas, the value is not range-validated.

  ## Tests

  Eleven cases in a new `tests/test_mlx_adam_epsilon.py` (CPU-pure via the MLX simulation shim,
  no weights, ~0.1s for the module):

  - the config accepts `adam_epsilon` at all — the regression itself
  - the value reaches Adam and AdamW
  - unset forwards nothing, so the optimizer default stands
  - it composes with explicit betas
  - it is **not** forwarded to SGD / Muon / Lion, where it would be a `TypeError`
  - it is not forwarded to Adafactor, whose `eps` is a different shape
  - `_MLX_CONFIG_OPTIONAL_COPY_FIELDS` remains an exact suffix of the field list

  The bug was reproduced on the parent commit before fixing. Across the MLX suite the failing
  set is byte-identical to a clean checkout of `main`; the only delta is the eleven added tests.

  ## CI

  Second commit adds the new file to the behavioral-gate job. Outside those gate jobs `tests/`
  is collect-only, and as that step's own comment notes, collection proves solely that a file
  imports — a new test file left out of the gate would never actually run.

### DIFF
--- a/.github/workflows/consolidated-tests-ci.yml
+++ b/.github/workflows/consolidated-tests-ci.yml
@@ -112,11 +112,17 @@ jobs:
         # CPU-pure and monkeypatch every Hub entry point, so they cost under two
         # seconds. Left out of a gate they would have been collected but never
         # run: `pytest tests/ --collect-only` above proves only that they import.
+        #
+        # test_mlx_adam_epsilon pins that HF's adam_epsilon reaches Adam/AdamW
+        # and reaches nothing else, plus the appended-field ordering the config's
+        # positional binding depends on. CPU-pure via the MLX simulation shim,
+        # no weights, and the whole file runs in about a tenth of a second.
         run: |
           python -m pytest \
             tests/test_training_utils_use_cache.py \
             tests/test_mlx_finetune_last_n_layers.py \
             tests/test_mlx_double_quant_reject.py \
+            tests/test_mlx_adam_epsilon.py \
             tests/test_hf_xet_fallback.py \
             tests/test_get_transformers_model_type_empty.py \
             tests/test_train_on_responses_list_labels.py \
--- a/tests/test_mlx_adam_epsilon.py
+++ b/tests/test_mlx_adam_epsilon.py
@@ -1,0 +1,114 @@
+# SPDX-License-Identifier: LGPL-3.0-or-later
+# Copyright 2026-present the Unsloth AI Inc. team. All rights reserved.
+"""adam_epsilon on the MLX trainer (HF TrainingArguments parity).
+
+HF exposes ``adam_epsilon`` and the CUDA path forwards it to the optimizer
+(``unsloth/trainer.py:350``, alongside the betas). MLX carried ``adam_beta1`` /
+``adam_beta2`` but not the epsilon, and because ``MLXTrainingConfig`` rejects
+unknown kwargs, passing it raised ``TypeError`` before training began rather
+than being quietly ignored.
+
+CPU-pure: builds optimizers through the MLX simulation shim, no weights and no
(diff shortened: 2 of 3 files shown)


BardiaKoopah:
@codex review

BardiaKoopah:
Right, the guard ran before `_normalize_mlx_optimizer_name` so it fired for every optimizer, and `optim="sgd", adam_epsilon=-1.0` failed here while the same config trains on CUDA. Fixed in `c94f878b` — `adam_kwargs` (betas included, same scope) is now built after the name is resolved and gated on the Adam family, so SGD/Muon/Lion/Adafactor ignore the value entirely.

One wrinkle worth naming: it can't just move below the normalization, it has to go below the Adafactor rank>2 fallback, since that fallback demotes to AdamW and the demoted branch does take the epsilon. Added tests for both sides of that — the fallback still carries and validates the epsilon, and `sgd`/`muon`/`lion`/`adafactor` accept `-1.0`, NaN and `"not-a-number"` without forwarding anything.

BardiaKoopah:
@codex review

BardiaKoopah:
@oobabooga this as well pls

danielhanchen:
Confirmed unsloth_zoo/mlx/trainer.py carries adam_beta1 and adam_beta2 but no adam_epsilon, so a ported config fails in MLXTrainingConfig before the first step, and the resolution now sits after the Adafactor rank demotion so the AdamW fallback keeps it. One small thing: the description still says the value is not range-validated, but _resolve_adam_epsilon does reject negatives and NaN, so could you update the body to match?

danielhanchen:
Checked that adam_epsilon is still absent from unsloth_zoo/mlx/trainer.py on main and that the new forwarding sits after the Adafactor rank>2 demotion so the fallback keeps the epsilon. Please rebase, the branch is showing conflicts.

danielhanchen:
Confirmed unsloth_zoo/mlx/trainer.py only forwards the betas today, so adam_epsilon has nowhere to land, and the follow-up commit correctly moves the scoping below the Adafactor rank above 2 fallback. Will get this reviewed.

PR #978, opened 2026-08-04
Fix Qwen3.5 fp16 training dtype mismatches under UNSLOTH_FORCE_FLOAT32

chakshu-dhannawat:
Adds targeted dtype-normalization patches for Qwen3.5 when UNSLOTH_FORCE_FLOAT32=1.

Problem
- unsloth#7506 reports fp16 fine-tuning of Qwen3.5 failing with RuntimeError: expected mat1 and mat2 to have the same dtype, but got: c10::BFloat16 != c10::Half.
- On non-bf16 GPUs, unsloth loads Qwen3.5 in bf16 and then down-casts weights to fp16, but the trainer/autocast path may still present bf16 activations to those fp16 weights.

Fix
- Boundary-style wrappers for Qwen3_5GatedDeltaNet, Qwen3_5Attention, Qwen3_5MLP, and Qwen3_5ForCausalLM.
- Each wrapper casts inputs (and RoPE position embeddings for attention) to the actual submodule weight dtype before calling the original forward, then restores the caller dtype for the residual stream.
- Patches are gated on UNSLOTH_FORCE_FLOAT32 == "1", so bf16 / fp32 training and systems that do not need the fallback are untouched.

Validation
- Added tests/test_qwen3_5_float32.py with tiny Qwen3.5 models; it reproduces the mismatch on CPU (fp16 weights + bf16 activations) and verifies the patched forwards succeed.
- tests/test_temporary_patches_imports.py updated so the new submodule is covered by the import smoke suite.
- pytest tests/test_temporary_patches_imports.py tests/test_temporary_patches_exhaustive.py tests/test_qwen3_5_float32.py passes locally (143 passed, 12 skipped).

### DIFF
--- a/tests/test_qwen3_5_float32.py
+++ b/tests/test_qwen3_5_float32.py
@@ -29,6 +29,7 @@ import transformers.models.qwen3_5.modeling_qwen3_5 as qwen  # noqa: E402
 
 from unsloth_zoo.temporary_patches.common import TEMPORARY_PATCHES  # noqa: E402
 from unsloth_zoo.temporary_patches.qwen3_5_float32 import (  # noqa: E402
+    _unsloth_fused_loss_kwargs,
     _unsloth_is_default_causal_lm_loss,
 )
 
@@ -185,3 +186,17 @@ def test_custom_loss_rejected():
     """Custom losses must fall back to the logits + loss_function branch."""
     fake_loss = type("_FakeLoss", (), {"__name__": "CustomFocalLoss"})()
     assert _unsloth_is_default_causal_lm_loss(fake_loss) is False
+
+
+def test_fused_loss_kwargs_filter():
+    """Model-only kwargs must not leak into the fused CE kernel."""
+    kwargs = {
+        "num_items_in_batch": 8,
+        "shift_labels": False,
+        "output_attentions": True,
+        "output_hidden_states": True,
+        "return_dict": True,
+        "use_cache": True,
+    }
+    filtered = _unsloth_fused_loss_kwargs(kwargs)
+    assert filtered == {"num_items_in_batch": 8, "shift_labels": False}
--- a/tests/test_qwen3_5_float32.py
+++ b/tests/test_qwen3_5_float32.py
@@ -193,10 +193,17 @@ def test_fused_loss_kwargs_filter():
     kwargs = {
         "num_items_in_batch": 8,
         "shift_labels": False,
+        "ignore_index": -100,
+        "label_smoothing": 0.1,
         "output_attentions": True,
         "output_hidden_states": True,
         "return_dict": True,
         "use_cache": True,
     }
     filtered = _unsloth_fused_loss_kwargs(kwargs)
-    assert filtered == {"num_items_in_batch": 8, "shift_labels": False}
+    assert filtered == {
+        "num_items_in_batch": 8,
+        "shift_labels": False,
+        "ignore_index": -100,
+        "label_smoothing": 0.1,
+    }
--- a/tests/test_qwen3_5_float32.py
+++ b/tests/test_qwen3_5_float32.py
@@ -28,6 +28,9 @@ from transformers.models.qwen3_5.configuration_qwen3_5 import Qwen3_5TextConfig 
 import transformers.models.qwen3_5.modeling_qwen3_5 as qwen  # noqa: E402
 
 from unsloth_zoo.temporary_patches.common import TEMPORARY_PATCHES  # noqa: E402
+from unsloth_zoo.temporary_patches.qwen3_5_float32 import (  # noqa: E402
+    _unsloth_is_default_causal_lm_loss,
+)
 
 
 def _apply_temporary_patches():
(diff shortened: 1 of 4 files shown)


chakshu-dhannawat:
Thanks for the review. Pushed updates:

- Qwen3_5ForCausalLM: now pops return_dict/output_attentions/output_hidden_states from kwargs and forces return_dict=True for the internal self.model call, so outputs.last_hidden_state never fails when the caller requests tuple outputs.
- Added Qwen3_5ForConditionalGeneration wrapper to cover Qwen3.5 vision/multimodal models that hit the same bf16/fp16 lm_head mismatch.
- Updated tests/test_qwen3_5_float32.py to pytest.importorskip the specific transformers.models.qwen3_5.* modules so older supported Transformers jobs skip cleanly.

Re-ran pytest tests/test_temporary_patches_imports.py tests/test_temporary_patches_exhaustive.py tests/test_qwen3_5_float32.py locally: 143 passed, 12 skipped.

chakshu-dhannawat:
Addressed the latest Codex points in 4f71efe2:

- Tuple-output contract: both Qwen3_5ForCausalLM and Qwen3_5ForConditionalGeneration wrappers are now decorated with @can_return_tuple, so return_dict=False returns a tuple just like upstream.
- Fused-loss path: on the training branch the wrapper calls unsloth_fused_lm_head_loss directly instead of materialising full-vocabulary logits, keeping the memory-saving fused CE path.
- Vision tower: added dtype guards for Qwen3_5VisionAttention, Qwen3_5VisionMLP and Qwen3_5VisionPatchMerger so multimodal UNSLOTH_FORCE_FLOAT32 runs do not fail in vision linears.
- Expanded the weight-dtype detector to cover qkv / linear_fc1 and added a Qwen3_5ForCausalLM regression test that verifies both ModelOutput and tuple outputs pass.

Local run: pytest tests/test_temporary_patches_imports.py tests/test_temporary_patches_exhaustive.py tests/test_qwen3_5_float32.py -> 144 passed, 12 skipped.

chakshu-dhannawat:
Addressed the third Codex pass in b1bc5353:

- Config-driven output flags: no longer pop output_attentions / output_hidden_states; the wrapper sets return_dict=True and forwards the remaining kwargs, so config.output_hidden_states = True still works without the kwarg.
- UNSLOTH_RETURN_LOGITS=1: in the fused-loss branch we now materialise logits as well as calling the fused CE loss.
- UNSLOTH_RETURN_HIDDEN_STATES=1: returns the sliced hidden states in the logits field, matching the GRPO helper pattern used in qwen3_moe.py.
- Added a regression test confirming config.output_hidden_states = True propagates through the patched forward.

Local run remains 144 passed, 12 skipped.

chakshu-dhannawat:
Addressed the fourth Codex pass in 6a51a14f:

- Quantized lm_head guard: added _unsloth_weight_dtype which skips weights that have quant_state, are non-floating, or are fp8, so the wrapper never casts activations to a packed quantized storage dtype.
- Custom loss guard: the fused-CE shortcut now only runs when self.loss_function.__name__ is ForCausalLMLoss; any custom loss falls back to materialised logits + the configured loss function.
Both changes apply to Qwen3_5ForCausalLM and Qwen3_5ForConditionalGeneration.

Local run: pytest tests/test_temporary_patches_imports.py tests/test_temporary_patches_exhaustive.py tests/test_qwen3_5_float32.py -> 145 passed, 12 skipped.

danielhanchen:
Confirmed the gap: qwen3_5 is in FORCE_FLOAT32 in model_lists.py but there is no qwen3_5 float32 patch alongside gemma4_float32.py and qwen3_moe_float32.py, so fp16 loads hit the mismatch. The direction is right, but this also reimplements the causal-LM and conditional-generation forwards with fused loss and vision-tower handling. Could you narrow it to the dtype normalization following the qwen3_moe_float32.py shape and leave the loss path alone?

danielhanchen:
Rechecked and the missing qwen3_5 dtype normalization is still a real gap next to gemma4_float32.py and qwen3_moe_float32.py, but the branch still carries the causal-LM and conditional-generation forwards with fused loss and vision-tower handling. Could you push the narrowed version that only does the dtype casts in the qwen3_moe_float32.py shape and leaves the loss path untouched?

chakshu-dhannawat:
Hi Daniel, the branch at bbfbe5bc has been narrowed to component-level dtype wrappers only (Qwen3_5GatedDeltaNet, Qwen3_5Attention, Qwen3_5MLP). The causal-LM / conditional-generation forwards, fused-loss helpers, and vision-tower wrappers were removed. The PR body is also updated to match. Let me know if you'd like any further changes.

danielhanchen:
Checked the narrowed branch: it is now just the dtype wrappers for the gated delta net, attention and MLP, matching the qwen3_moe_float32.py shape, with the loss and vision paths dropped as asked. Will get this reviewed.

danielhanchen:
Confirmed the gap is still there, with qwen3_5 listed in model_lists.py but no float32 patch next to gemma4_float32.py and qwen3_moe_float32.py, and the branch is now narrowed to the component dtype wrappers as asked. Will get this reviewed.

chakshu-dhannawat:
Hi @danielhanchen, just a quick ping for a review when you have time. Thanks.

PR #979, opened 2026-08-04
Print the MLX 8-bit optimizer routing line once, on one line

danielhanchen:
Follow-up to #962. Trims the routing line that PR prints when `optim="adamw_8bit"` or `"adam_8bit"` resolves on MLX.

## Why

`_build_optimizer` runs once per training run and once per rank, and `adamw_8bit` is the default in every Unsloth README and notebook example. So the banner was not an unusual-configuration notice, it was a paragraph everyone saw on every run. Captured from `MLXTrainer._build_optimizer` on `dca7439d`:

the routing banner before this change: five wrapped lines per occurrence, twenty across four calls

Five wrapped lines at a 100 column terminal per occurrence. The second panel is four `_build_optimizer` calls, which is a four rank job or a re-run notebook cell: twenty lines before training starts.

Most of that length was mine. I appended the dtype and peak-memory caveats during the #962 review so the 35.8% figure would not travel alone, which was the right correction in the wrong place. Those belong in the module docstring, and they are already there.

## What is left

The banner existed because the old path rewrote `adamw_8bit` to `adamw` silently. That is now handled by the routing being correct. The one thing a user still cannot infer from the name is that this quantizes a single moment, where bitsandbytes' `adamw_8bit` quantizes both. That is worth one line:

```
Unsloth: adamw_8bit on MLX quantizes Adam's first moment to 8-bit (group_size=64); the second moment is unquantized.
```

116 characters, one line, printed at most once per process and only on the main process. `announce_quantized_optimizer` owns that; `describe_quantized_optimizer` still returns the string so it stays unit testable.

This matches the narrower convention `_build_optimizer` already follows a few lines up, where the Adafactor fallback prints only when behaviour deviates from what was asked for.

## Verified

Four `_build_optimizer` calls through the real trainer now emit one line instead of twenty:

```
optimizer: QuantizedMomentAdamW
lines printed across 4 _build_optimizer calls: 1
Unsloth: adamw_8bit on MLX quantizes Adam's first moment to 8-bit (group_size=64); the second moment is unquantized.
```

Tests: `test_mlx_quantized_optimizer_routing.py` 15 passed (the banner assertions now pin the one-line contract, plus print-once and rank-silence cases), `test_mlx_quantized_optimizer_state.py` 29 passed on `mlx-cpu`, and `test_mlx_trainer_internals.py` + `test_mlx_module_exports.py` + `test_mlx_double_quant_reject.py` + `test_mlx_finetune_last_n_layers.py` 260 passed.

No behaviour change to the optimizer itself.

### DIFF
--- a/tests/test_mlx_quantized_optimizer_routing.py
+++ b/tests/test_mlx_quantized_optimizer_routing.py
@@ -72,15 +72,38 @@ def test_8bit_names_are_advertised_as_supported():
     assert "adam_8bit" in SUPPORTED_MLX_OPTIMIZERS
 
 
-def test_description_states_what_is_and_is_not_quantized():
-    """The message must name the actual shape of the saving."""
+def test_description_names_the_one_thing_the_name_does_not():
+    """bitsandbytes' adamw_8bit quantizes both moments; this quantizes one. That
+    is the whole reason to print anything, so it must fit on one line."""
     from unsloth_zoo.mlx.optimizers_quantized import describe_quantized_optimizer
     message = describe_quantized_optimizer("adamw_8bit")
 
     assert "adamw_8bit" in message
-    assert "FIRST moment" in message
+    assert "first moment" in message
     assert "group_size=64" in message
-    assert "second moment is not quantized" in message
-    assert "multiple of 64" in message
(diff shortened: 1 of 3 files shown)


danielhanchen:
@codex review

PR #995, opened 2026-08-04
Feat(mlx): Support group_by_length for text training

BardiaKoopah:
# Description

  HF's `group_by_length` reshuffles which rows are batched together each epoch
  while keeping every batch length-homogeneous. MLX had no equivalent — the
  existing orders each give up one half of that:

  | `dataset_order` | padding | batch composition |
  | --- | --- | --- |
  | `"default"` | grouped (length-sorted) | computed **once** — every epoch trains
  on the same batches |
  | `"torch_randperm"` | pads to whatever lands together | fresh each epoch |

  To be precise about the default: the grouping itself never changes. The
  unlabeled path does permute the *order* those fixed batches are visited in each
  epoch (`visit_policy="epoch_permute"`), and the labeled
  `train_on_responses_only` path does not even do that. What neither offers is
  HF's behaviour — re-drawing *which rows are batched together* each epoch while
  keeping every batch length-homogeneous.

  ## What changed
  
  A `"length_grouped"` order that does both, exposed through the HF knob:

  ```python
  MLXTrainingConfig(group_by_length=True)
  ```

  The ordering helper is a step-for-step port of transformers'
  `get_length_grouped_indices`: seeded `randperm`, cut into mega-batches of
  `mega_batch_mult * batch_size`, sort each mega-batch by descending length, then
  swap the globally longest row into first position so an OOM surfaces immediately
  rather than minutes in.

  `_torch_randperm_order` already seeds a `torch.Generator` the way HF seeds its
  sampler generator, so **for a given seed the order matches CUDA exactly**.
  `test_matches_hf_reference_bit_for_bit` transcribes the HF function into the
  test file and asserts equality over 200 randomized (dataset size, batch size,
  seed) combinations, so the port is pinned to the real algorithm rather than to
  my reading of it.

  The order is reseeded per epoch like `torch_randperm`, so each pass gets a fresh
  mega-batch shuffle instead of repeating one grouping forever.

  ## Order resolution is now in one place

  `group_by_length`, `preserve_dataset_order` and `dataset_order` were previously
  combined inline at every call site. That is how the labeled and unlabeled text
  paths ended up accepting *different* `dataset_order` vocabularies —
  `_create_labeled_batches` handles `"default"` by length-sorting, while
  `_create_ordered_text_plan` raises `ValueError` on it. They only avoid
  disagreeing today because the caller happens to gate which one runs, and the
  comment asserting they must agree cites line numbers that are thousands of lines
  stale.

  `_resolve_text_dataset_order` now owns that decision, and
  `test_labeled_and_unlabeled_paths_produce_the_same_order` pins the two paths to
  the same stream. They have to agree or `train_on_responses_only` masks a
  different stream than it trains on.

  With `group_by_length` unset the resolver returns exactly what the old inline
  expressions returned, for all eight legacy combinations of the other two knobs.

  ## Scope

  - **Text only.** VLM raises: a row's token length does not account for image or
  audio tokens, so grouping on it would not bound padding the way it does for
  text.
  - **Sized datasets only.** `streaming=True` raises: the mega-batch permutation
  needs the whole dataset up front. `streaming_text_length_window_batches` is the
  lazy analogue.
  - **Training only**, matching HF, which installs the length-grouped sampler on
  the train dataloader alone. Eval keeps dataset order so metrics stay comparable
  run to run instead of being regrouped and per-epoch reshuffled underneath them.
  - Combining it with `preserve_dataset_order` or an explicit `dataset_order`
  raises rather than silently letting one win — both pick a concrete, conflicting
  order.

  The rejection helper is a no-op for every pre-existing order, so VLM and
  streaming runs that never set `group_by_length` are untouched.

  ## Tests
  `tests/test_mlx_group_by_length.py` (31 cases, CPU-pure index math plus torch's
  RNG, ~0.2s): bit-exact agreement with the HF reference; permutation/determinism
  properties; longest-row-first; measured padding reduction; empty and single-row
  datasets; every conflicting-knob rejection; the eval contract; the
  labeled/unlabeled agreement invariant; the VLM and streaming rejections plus
  proof the guard is a no-op for every pre-existing order; DDP behaviour (every
  rank derives an identical global order, rank slices are disjoint and cover the
  global batch, and grouping survives sharding); and the appended-field ordering
  the config's positional binding depends on.

  `tests/test_mlx_group_by_length_metal.py` (7 cases): end to end through the real
  builders — full row coverage, per-epoch reshuffle, seed determinism, padded
  text.
  - **Sized datasets only.** `streaming=True` raises: the mega-batch permutation
  needs the whole dataset up front. `streaming_text_length_window_batches` is the
  lazy analogue.
  - **Training only**, matching HF, which installs the length-grouped sampler on
  the train dataloader alone. Eval keeps dataset order so metrics stay comparable
  run to run instead of being regrouped and per-epoch reshuffled underneath them.
  - Combining it with `preserve_dataset_order` or an explicit `dataset_order`
  raises rather than silently letting one win — both pick a concrete, conflicting
  order.

  The rejection helper is a no-op for every pre-existing order, so VLM and
  streaming runs that never set `group_by_length` are untouched.

  ## Tests

  `tests/test_mlx_group_by_length.py` (31 cases, CPU-pure index math plus torch's
  RNG, ~0.2s): bit-exact agreement with the HF reference; permutation/determinism
  properties; longest-row-first; measured padding reduction; empty and single-row
  datasets; every conflicting-knob rejection; the eval contract; the
  labeled/unlabeled agreement invariant; the VLM and streaming rejections plus
  proof the guard is a no-op for every pre-existing order; DDP behaviour (every
  rank derives an identical global order, rank slices are disjoint and cover the
  global batch, and grouping survives sharding); and the appended-field ordering
  the config's positional binding depends on.
  
  `tests/test_mlx_group_by_length_metal.py` (7 cases): end to end through the real
  builders — full row coverage, per-epoch reshuffle, seed determinism, padded
  cells dropping versus `torch_randperm`, equal per-rank batch counts under DDP
  (unequal counts desynchronise collectives), disjoint rank shards, and the
  labeled `train_on_responses_only` builder. All 7 were executed on Apple Silicon
  (M2, mlx 0.31.2 / mlx-lm 0.31.3) and pass. This half needs a real MLX runtime,
  since the text plan builder cannot tokenize under the simulation shim (the same
  reason `tests/test_mlx_batching_and_decay.py` skips itself there), so it is
  wired into the Apple Silicon workflow rather than the CPU gate.

  Across the MLX suite the failing set is byte-identical to a clean checkout of
  `main`; the only delta is the added tests. On real MLX,
  `test_mlx_batching_and_decay` (56 passed), `test_mlx_trainer_internals` (223
  passed), `test_mlx_training_e2e_metal` (28 passed) and `test_mlx_ddp_metal` are
  unchanged versus main.

  ## CI

  The CPU half joins the behavioral gate — outside those jobs `tests/` is
  collect-only, which as that step's own comment notes proves only that a file
  imports. The end-to-end half joins `mlx-ci.yml`, where real MLX is available.

### DIFF
--- a/.github/workflows/consolidated-tests-ci.yml
+++ b/.github/workflows/consolidated-tests-ci.yml
@@ -129,11 +129,18 @@ jobs:
         # so a 3.10-only `match` and a module-level `str | Path` both reached main and
         # made the package un-importable on 3.9. It reads the floor from pyproject, so
         # raising requires-python relaxes it rather than making it lie.
+        #
+        # test_mlx_group_by_length pins the HF-exact ordering algorithm (checked
+        # against a transcription of transformers' get_length_grouped_indices),
+        # the conflicting-knob rejections, and that the labeled and unlabeled
+        # text paths order rows identically. CPU-pure index math plus torch's
+        # RNG; the real-MLX end-to-end half lives in the _metal companion.
         run: |
           python -m pytest \
             tests/test_training_utils_use_cache.py \
             tests/test_mlx_finetune_last_n_layers.py \
             tests/test_mlx_double_quant_reject.py \
+            tests/test_mlx_group_by_length.py \
             tests/test_hf_xet_fallback.py \
             tests/test_get_transformers_model_type_empty.py \
             tests/test_train_on_responses_list_labels.py \
--- a/.github/workflows/mlx-ci.yml
+++ b/.github/workflows/mlx-ci.yml
@@ -111,6 +111,13 @@ jobs:
         # CPT partition, target normalization and full-module save routing.
         run: python -m pytest tests/test_mlx_cpt_partition_metal.py -v -rs
 
+      - name: tests/test_mlx_group_by_length_metal.py
+        # group_by_length through the real text plan builder: full row coverage,
+        # per-epoch reshuffle, seed determinism, and that padded cells actually
+        # drop versus torch_randperm. Needs real MLX because the plan builder
+        # cannot tokenize under the simulation shim.
+        run: python -m pytest tests/test_mlx_group_by_length_metal.py -v -rs
+
       - name: tests/test_mlx_pr684_full_validation_metal.py
         # Deep behavior checks: resume determinism, completion-only training,
         # epoch step counts, SGD coupled decay, real Qwen2-VL LoRA training.
--- a/tests/test_mlx_group_by_length.py
+++ b/tests/test_mlx_group_by_length.py
@@ -19,6 +19,7 @@ CPU-pure: pure index math plus torch's RNG. No MLX, no weights, no Metal.
 from __future__ import annotations
 
 import random
+import sys
 
 import pytest
 import torch
@@ -127,6 +128,37 @@ def test_empty_and_single_row_datasets():
     order = _helper()
     assert order([], 4, 1) == []
     assert order([5], 4, 1) == [0]
+
+
+def test_ordering_works_without_torch(monkeypatch):
+    """pyproject excludes torch on darwin/arm64, so a native MLX install has
+    none. group_by_length is a stock HF knob and must not be the one option
+    that needs it: the permutation falls back to the stdlib RNG, and only the
+    shuffle changes -- the grouping still has to hold."""
+    from unsloth_zoo.mlx.utils import _length_grouped_order
+
+    monkeypatch.setitem(sys.modules, "torch", None)  # makes `import torch` fail
+
+    rng = random.Random(12)
+    lengths = [rng.randint(1, 1024) for _ in range(512)]
+    batch_size = 8
+    order = _length_grouped_order(lengths, batch_size, 77)
+
+    assert sorted(order) == list(range(len(lengths)))
+    assert order == _length_grouped_order(lengths, batch_size, 77)
+    assert order != _length_grouped_order(lengths, batch_size, 78)
+
+    def padding_cost(indices):
+        total = 0
+        for i in range(0, len(indices), batch_size):
+            batch = indices[i : i + batch_size]
+            widest = max(lengths[j] for j in batch)
+            total += sum(widest - lengths[j] for j in batch)
+        return total
+
+    plain = list(range(len(lengths)))
+    random.Random(77).shuffle(plain)
+    assert padding_cost(order) < padding_cost(plain) / 4
 
 
 # --- config resolution ---
--- a/tests/test_mlx_group_by_length.py
+++ b/tests/test_mlx_group_by_length.py
@@ -363,9 +363,11 @@ def test_order_reshuffles_across_epochs():
 
 
 def test_eval_batches_are_never_length_grouped():
-    """HF installs the length-grouped sampler on the train dataloader only.
-    Regrouping eval as well (and reshuffling it per epoch) would move metrics
-    around underneath the user for a knob that is about training throughput.
+    """A deliberate divergence: transformers length-groups eval too
+    (``Trainer._get_eval_sampler``). `_evaluate` accumulates ``loss * ntoks``
+    and divides by the total, so eval metrics do not depend on how rows are
+    grouped -- regrouping eval would only save padding on the eval pass, and
+    this knob is about training throughput.
     """
     from unsloth_zoo.mlx.trainer import _resolve_text_dataset_order
 
--- a/tests/test_mlx_group_by_length.py
+++ b/tests/test_mlx_group_by_length.py
@@ -442,3 +442,43 @@ def test_length_grouping_survives_rank_sharding():
     assert local_padding(order) < local_padding(
         _torch_randperm_order(len(lengths), 21)
     ) / 4
+
+
+def test_grouping_is_cut_at_the_global_batch_not_the_local_one():
+    """The plan builders consume the order in chunks of the GLOBAL micro-batch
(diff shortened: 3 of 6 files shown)


BardiaKoopah:
@chatgpt-codex-connector all three are right, fixed in `d4cabe1`, `66752bd` and `eb3715e`.

The torch one is the worst of them — `_length_grouped_order` went through `_torch_randperm_order`, which exists to mirror CUDA's sampler and raises ImportError without torch, and that is fine for the opt-in `dataset_order="torch_randperm"` but not for a stock HF knob on a platform where pyproject deliberately drops torch. The permutation now falls back to the stdlib RNG when torch is not importable; only the shuffle changes, the mega-batch grouping and the padding it saves are identical, and with torch present the CUDA-exact order is untouched. Verified on an M2 Pro by blocking torch at the import hook: `unsloth_zoo.mlx.utils` imports and `_length_grouped_order` returns a valid grouped order with `torch` never in `sys.modules`. I did not build a genuinely torch-free venv to prove the full plan builder end to end — transformers imports torch when the distribution is present even if the import hook blocks it, so that particular proof needs a clean `pip install .[mlx]` rather than a hook.

On the DDP grouping you're right that the mega-batches have to be cut at the batch the schedule consumes, and both text paths now pass `_distributed_global_batch_size(batch_size, comm_group)`. At world size 4 with a local batch of 2 the multiplier saturates at the 50x cap and gives 100-row windows against a consumed global batch of 8, so 100 % 8 = 4 and every straddling chunk pairs the shortest rows of one window with the longest of the next; over 1600 rows the per-rank padding waste after the rank slice goes from 62616 tokens to 8040. Note the common single-rank case is unaffected since the two sizes are equal there.

The response-only `max_steps` point holds for `torch_randperm` too, not just `length_grouped` — same expression, and the unlabeled plan already reseeds for both — so I fixed both rather than only the new order. `_create_labeled_batches` now builds epoch blocks until the `num_batches` horizon is covered instead of a fixed `num_epochs or 1`. Sequential, `preserve_dataset_order` and the legacy length-sorted default are deliberately left cycling one block: nothing in them reseeds the composition, and expanding the default would change the batch order of every existing step-based `train_on_responses_only` run for no correctness gain.

BardiaKoopah:
@codex review

BardiaKoopah:
Both correct, fixed. The grouping window now folds gradient_accumulation_steps in through a new `_length_grouped_window` in `utils.py` that both plan builders call, so the default grad_accum=4 config groups the same window transformers does — `test_window_matches_hf_sampler_construction` checks the resulting order against a transcription of HF's `get_length_grouped_indices` at `train_batch_size * gradient_accumulation_steps` over 100 randomized cases, and the labeled and unlabeled builders are pinned to the same stream at the builder level, not just the helper. While wiring it I found the unlabeled call site only passed grad_accum when num_epochs was set, so a max_steps run would have grouped differently from the same config run by epochs; it passes unconditionally now (`afd30039`).

On the preference trainers I rejected rather than routed it (`82a63743`): a preference row has a chosen and a rejected length, so there's no single length to group on, and CUDA rejects it too — TRL 0.23.1 leaves `Trainer._get_train_sampler` in place, and its `LengthGroupedSampler` raises "Can only automatically infer lengths for datasets whose items are dictionaries with an 'input_ids' key" on a prompt/chosen/rejected dataset, which I confirmed against the installed trl. `MLXDPOConfig(group_by_length=True)` now fails in `_prepare_data` with a message naming DPO/ORPO instead of silently ignoring the knob.

One thing I had wrong that came out of checking your `_get_train_sampler` reference: the docstring on `_resolve_text_dataset_order` claimed HF length-groups the train dataloader only, and it doesn't — `Trainer._get_eval_sampler` builds one too, at eval_batch_size and without the accumulation factor. MLX still evaluates in dataset order, since `_evaluate` accumulates `loss * ntoks` and divides by the total, so the metric is grouping-invariant and grouping eval would only save padding on the eval pass. Note corrected in `e73a8ecd`; if a maintainer wants literal parity there it's a small follow-up.

BardiaKoopah:
@codex review

BardiaKoopah:
@oobabooga can you check this please. Looks good on my end and codex is coming back clean

danielhanchen:
Confirmed there is no length-grouped order in unsloth_zoo/mlx today and the new one is pinned against the transformers reference, but this branch also changes existing torch_randperm behaviour by making grad_accum unconditional and by expanding _create_labeled_batches past num_epochs. Could you split those two out so the new order lands on its own?

danielhanchen:
Checked that the MLX path has no length-grouped order today and that the new resolver keeps the existing eight knob combinations unchanged, which is the part I care about most. The direction is right, though this branch also reworks the labeled train_on_responses_only epoch-block expansion for torch_randperm; could you split that out so the new order lands on its own?

danielhanchen:
Checked that nothing in unsloth_zoo/mlx handles group_by_length today, and the HF-reference parity test is the right way to pin the ordering, so this is a useful addition. The branch also reworks order resolution across both text paths and changes `_create_labeled_batches` epoch blocks for `torch_randperm`; could you split that second change out so the new order lands on its own?

danielhanchen:
Checked that main's MLX trainer has no length-grouped ordering, and this brings the HF group_by_length behaviour to text training behind the existing config knob. Will get this reviewed once the conflicts with main are resolved.

PR #997, opened 2026-08-05
Do not let run-to-run noise stand in for the audio signal

danielhanchen:
## The problem

Stage 4 of `tests/gemma4_audio_version_probe.py` proves audio reaches the objective by requiring two different tones to give two different losses:

```python
if abs(losses[0] - losses[1]) < 1e-6:
    raise AssertionError("two different tones gave the same loss ...")
```

That is only a proof on a machine where the same input twice gives the same loss, and on Apple Silicon this model does not. Five identical uncached forwards of Gemma 4 E2B gave five different logit sums, spread 28466618, with mlx-vlm and mlx only and no unsloth in the picture. A layerwise bisect puts the first differing decoder layer at 0, fifteen layers before `first_kv_shared_layer_idx=15`, with a max difference of about 1.0 on a single layer's output.

Against jitter that large a bare inequality is satisfied by the drift alone, so the stage goes green whether or not the audio is connected. The 0.6.2, 0.6.3 and 0.6.4 macOS cells all passed this stage and those three passes establish nothing.

The gate entry itself is not in question. The version boundary comes from which mlx-vlm versions load the checkpoint at all, which is a hard binary, and stage 3 compares integer counts. Neither can be moved by numeric drift. What is in question is this one stage's ability to detect a disconnected audio path on the platform it was written for.

## The fix

Measure the floor before trusting the gap. Run one tone three times and take the spread, then require the two-tone difference to clear it by a margin. Three rather than two because a single pair is one sample of a spread and can agree by luck, which would put the floor back at zero.

Where the model is reproducible the floor is zero and this reduces to the original test.

Two supporting changes:

- The decision is now `two_tone_verdict()`, a module level function. The branch that made this necessary only fires on hardware where the model drifts, and a test that needs a Mac and a 4 GB checkpoint to run is a test nobody runs.
- `UNSLOTH_ALLOW_CPU` and `UNSLOTH_IS_PRESENT` moved from import time into `main()`, so importing the module from a test no longer rewrites the environment for the rest of the pytest session. Both are still set before anything imports `unsloth_zoo`.

## Evidence

- New `tests/test_gemma4_probe_two_tone_gate.py`, 8 cases covering both reproducible and drifting platforms, connected and disconnected audio: **8 passed**.
- Mutation check: deleting the noise branch fails exactly the 3 drifting cases and leaves the other 5 green, so the tests exercise the new behaviour rather than passing alongside it.
- Two cases pin the reproducible-platform behaviour unchanged, so this tightens rather than loosens.
- 322 MLX tests green: `test_mlx_vlm_label_masks.py` 226, `test_mlx_shape_guard.py` 32, `test_mlx_batching_and_decay.py` 56, new suite 8.
- Linux CPU end to end, all 5 probe stages pass, stage 4 reporting `440Hz=19.3557 1760Hz=19.6828 signal=0.327 noise=0`. The zero floor is the point: Linux is bit exact, so the tightened stage returns the same numbers as before and only bites where the platform drifts.

## Note on the drift itself

Not a zoo bug and not fixed here. It reproduces with mlx-vlm and mlx alone. The KV-sharing explanation is ruled out by the bisect above, and an op level probe finds `embed_tokens`, `q_proj`, the attention block, the whole layer 0 decoder, bare `mx.quantized_matmul` and an unquantized control all bit identical, so the cause is still open. It matters here only as a reason no test should assert exact loss values on Metal.

### DIFF
--- a/tests/gemma4_audio_version_probe.py
+++ b/tests/gemma4_audio_version_probe.py
@@ -68,10 +68,25 @@ RATE = 16000
 
 results = {}
 
-# How far clear of the platform's own jitter the two-tone gap has to land.
-# Deliberately generous: stage 4 asks whether audio moves the loss at all, so
-# a connected model should dwarf the jitter, and a gap that merely edges past
-# it is worth a human look rather than a recorded pass.
+# How far clear of the platform's own spread the two-tone gap has to land.
+#
+# Measured on macos-14, mlx-vlm 0.6.3, mlx 0.32.0, three separate runs of this
+# probe against mlx-community/gemma-4-e2b-it-4bit:
+#
+#   run   signal (440 vs 1760)   noise (spread of three identical 440s)
+#     1   0.812                  2.27
+#     2   1.30                   0.827
+#     3   0.719                  1.25
+#
+# The spread of three identical inputs exceeds the gap between two different
+# tones. So on Apple Silicon this stage cannot separate connected audio from
+# disconnected audio at any margin, and no choice of this constant rescues it;
+# the honest report there is "could not measure", which is what Inconclusive
+# is for. On Linux the same probe gives noise=0 and signal=0.327, and the
+# stage decides normally.
+#
+# The margin only has to be large enough that jitter cannot masquerade as
+# signal on a platform where the two are separable at all.
 TWO_TONE_MARGIN = 4.0
 
 TWO_TONE_REASONS = {
@@ -122,6 +137,21 @@ class Inconclusive(Exception):
     machine it ran on. It still does not qualify the version -- nothing was
     demonstrated -- so it is not a pass either.
     """
+
+
+def gating_stages(stages):
+    """The stages whose result decides this version's exit code.
+
+    Stage 0 and its subchecks are diagnostic and never gated. Neither does a
+    stage that could not measure anything: this matrix exists to name the
+    mlx-vlm version that breaks Gemma 4 audio, and on Apple Silicon stage 4
+    cannot answer at all (see TWO_TONE_MARGIN). Failing the job over that
+    would report every version as broken on the one platform the feature
+    ships to. The measured verdict still appears in PROBE_RESULT, and stages
+    1 to 3, which are what actually decide the version boundary, still gate.
+    """
+    return {k: v for k, v in stages.items()
+            if not k.startswith("0_") and not v.get("inconclusive")}
 
 
 def stage(name):
@@ -408,9 +438,7 @@ def main():
     print("PROBE_RESULT " + json.dumps(
         {"mlx_vlm": version, "transformers": transformers.__version__,
          "model": args.model, "stages": results}), flush=True)
-    # Stage 0 and its subchecks are diagnostic and never gate the exit code.
-    verdict = {k: v for k, v in results.items() if not k.startswith("0_")}
-    sys.exit(0 if all(r["ok"] for r in verdict.values()) else 1)
+    sys.exit(0 if all(r["ok"] for r in gating_stages(results).values()) else 1)
 
 
 if __name__ == "__main__":
--- a/tests/gemma4_audio_version_probe.py
+++ b/tests/gemma4_audio_version_probe.py
@@ -47,7 +47,10 @@ without a single early failure hiding everything after it:
                                      the audio encoder emits, or the merge has
                                      nothing to put behind the surplus
   4. audio reaches the loss       -- distinct audio must give distinct losses,
-                                     or the model is training on text alone
+                                     by more than this platform's own
+                                     run-to-run spread, or the model is
+                                     training on text alone and a drifting
+                                     machine is supplying the difference
 
 Run: python tests/gemma4_audio_version_probe.py [--model REPO]
 Exit code 0 only if every stage that counts passed.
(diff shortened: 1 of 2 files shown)


danielhanchen:
Second commit `ba716765` acts on an independent review of the first. Recording what changed and, more usefully, the one thing that is not yet demonstrated.

### Not yet demonstrated: the margin

The gate requires the two tone gap to exceed 4x the spread of three repeats. That 4 has never been evaluated against Apple Silicon, which is the platform it was written for. The arithmetic is tighter than the prose implied: the only measured signal is 0.3271 nats on Linux, so the gate flips to `below_noise` the moment the Metal spread exceeds 0.0818 nats. And `4 x (max - min of 3)` is itself a noisy threshold, so the false red rate across the seven version matrix is nonlinear in the true spread.

My own test fixture models Apple Silicon as a spread of 0.51 nats, which is six times past that break even. If that were the real figure, a healthy connected model would be classified `below_noise`. I made that number up; nobody has measured the loss scale spread on Metal. The logit sum spread of 28466618 does not bound it, because cross entropy is invariant to a per position constant logit shift.

A staging job is running the probe three times on macos-14 to read the `noise=` figure the new code prints. I will post it here. If it lands near 0.02 nats the margin or the statistic has to change; if it is in the 1e-3 range the current value has ample headroom and I will record why.

### Changed in `ba716765`

- `below_noise` was an `AssertionError`, so it read as a stage failure. In a matrix whose purpose is that a red cell names the version that breaks Gemma 4 audio, that blames a version for a property of the machine. It has its own `Inconclusive` type and an `[INCONCLUSIVE]` line now. It still does not qualify the version, so it is not a pass either.
- The recorded detail carried `repeats[0]` and the spread but not the repeats. Those cannot tell jitter from a drift accumulating across sequential forwards, and the two want opposite responses. All three are recorded now, and the reported figure is their mean.
- Stale stage description in the header block.
- Tests: a dead `module_from_spec` call, and an environment assertion that could not hold. It checked `UNSLOTH_ALLOW_CPU`, which `conftest` sets for every session; `UNSLOTH_IS_PRESENT` is no better, since `unsloth/__init__.py:17` sets it on import. Neither can be asserted absent, so it clears each and re-executes, which is the property that matters. The margin boundary case is now built from the gate's own arithmetic rather than a hardcoded constant that was really testing float rounding.
- The new test file was missing the AGPL header.

### Worth flagging about the gate itself

`prepush_gate.py` cleared this branch before the commit and failed it after. The header check enumerates added files with `git diff --diff-filter=A`, so a staged but uncommitted new file reports as zero new files and the gate passes vacuously. That is how a headerless file reached the first push despite the gate being run.

### Verification

10 gate tests green, each new behaviour mutation checked: removing the noise branch fails 3, reverting the mean fails 2, dropping the repeats from the detail fails 1. 314 MLX tests green.

danielhanchen:
## The measurement, and it goes against the margin

Three runs of this probe on macos-14, mlx-vlm 0.6.3, mlx 0.32.0, against `mlx-community/gemma-4-e2b-it-4bit`:

| run | signal (440 vs 1760) | noise (spread of three identical 440s) |
|---|---|---|
| 1 | 0.812 | **2.27** |
| 2 | 1.30 | **0.827** |
| 3 | 0.719 | **1.25** |

The spread of three identical inputs is larger than the gap between two different tones. This is not a threshold to tune. At this sample size the stage cannot separate connected audio from disconnected audio on Metal at any margin, and no value of `TWO_TONE_MARGIN` rescues it. My earlier fixture standing in for Apple Silicon used a spread of 0.51, which I invented; the real figure is worse than that, and the fixture now uses run 2, whose repeats reproduce the measured 0.827 exactly.

That makes the labelling change in `ba716765` load bearing rather than cosmetic. An inconclusive stage no longer gates the exit code: it is not the version's fault, and gating would paint every mlx-vlm version red on the only platform this feature ships to. Stages 1 to 3 are what decide the version boundary and they still gate. The verdict is still recorded in `PROBE_RESULT`. The filter is `gating_stages()` now, tested both directions.

**What this PR is worth after the measurement.** The old assertion did not just risk passing on noise, it was passing on noise: every macOS green on stage 4 to date came from a machine where the same input twice disagrees by more than the tones do. Those greens should never have been recorded. The version boundary is unaffected, since it comes from which versions load the checkpoint at all plus stage 3's integer counts. Linux keeps deciding stage 4 normally at `noise=0, signal=0.327`.

## The more useful finding

The first stage 4 forward is **24.8986 in all three runs**, in three separate processes, while repeats two and three drift away from it. The layerwise bisect shows the same shape: round 0 was `-31402986.0` in both runs and only later rounds differed.

So the Metal behaviour is not random, it is history dependent. The same forward at the same point in the same sequence is reproducible; a forward at a different point in the sequence is not. That is accumulating state, not kernel jitter, and it is a better lead than the op level probe produced, which found `embed_tokens`, `q_proj`, the attention block, the whole layer 0 decoder, bare `mx.quantized_matmul` and an unquantized control all bit identical.

Still not a zoo bug and still not fixed here.

## CI

Every check on this PR reports `fail` at exactly 35s, and the job conclusion is `cancelled`, not `failure`. That is the org queue purge, not this branch. Verification is 326 tests green locally, all three MLX suites plus 12 gate tests, each new behaviour mutation checked.

PR #1004, opened 2026-08-06
MLX: name the config key when a mirrored config.json is incomplete

BardiaKoopah:
## What's wrong

mlx-lm builds `ModelArgs` through `from_dict`, which filters the config down to fields the dataclass declares. A required field with no default that is absent from `config.json` therefore fails in `__init__` with a bare `TypeError` naming neither the model nor the file:

```
TypeError: ModelArgs.__init__() missing 1 required positional argument: 'block_ff_dim'
```

`_load_mlx_lm_with_strict_fallback` catches only `ValueError`, so this propagates uncaught with no Unsloth context. The result reads as missing Apple Silicon support. In unslothai/unsloth#7306 it was diagnosed that way for two weeks — through an `xformers` red herring and a "does MLX support this model?" question — before the real cause surfaced: `unsloth/LFM2.5-230M`'s mirrored `config.json` lost upstream's `"block_ff_dim": 2560`.

## What this does

Catches `TypeError` at both mlx-lm `load_model` call sites (single-device and distributed) and re-raises naming the model, the missing key(s), and `config.json` as the cause:

```
Unsloth: unsloth/LFM2.5-230M's config.json is missing the key 'block_ff_dim',
which mlx-lm's 'lfm2' architecture requires and cannot default. This is an
incomplete config.json in the model repo - MLX and Apple Silicon support are
not the problem. Compare the config against the upstream repo this model was
mirrored from and add the missing key.
```

The original `TypeError` is preserved as `__cause__`.

## Two deliberate choices

**Anchored on `__init__`, not on the "missing N required positional arguments" wording alone.** This loader deliberately tolerates mlx-lm changing `load_model`/`_download`'s signature between releases (see the bypass comment in `_load_mlx_lm_with_strict_fallback`). Those failures carry identical wording, and reporting one as "config.json is missing 'model_path'" would point a reader at the wrong repo entirely. Anything that isn't a constructor failure keeps its original traceback. Covered by `test_mlx_lm_signature_drift_is_not_a_config_problem`.

**No backfill.** Deriving the value would be wrong. `block_ff_dim` is not `intermediate_size` — `LFM2-350M`, `LFM2-700M` and `LFM2-1.2B` omit `intermediate_size` entirely and set `block_auto_adjust_ff_dim: true`. They happen to coincide for LFM2.5-230M, which makes a backfill look safe when it isn't.

This is diagnostics only and does not repair the config. The fix for #7306 itself is adding the key back to the model repo on the Hub.

## Testing

New `tests/test_mlx_incomplete_config.py` (5 tests, follows the `test_mlx_qk_norm_version_gap.py` pattern with the `mlx_simulation` shim).

| Check | Result |
|---|---|
| `tests/test_mlx_incomplete_config.py` | 5 passed |
| `tests/test_mlx_qk_norm_version_gap.py` (sibling guard on the same `try`) | 7 passed |
| `tests/test_mlx_distributed_loader.py` (other patched function) | 7 passed |
| End-to-end: real loader, real `unsloth/LFM2.5-230M`, Apple Silicon M2, mlx-lm 0.31.3 / mlx 0.31.2 | guard fires, `__cause__` preserved |

Refs unslothai/unsloth#7306.

### DIFF
--- a/tests/test_mlx_incomplete_config.py
+++ b/tests/test_mlx_incomplete_config.py
@@ -1,0 +1,122 @@
+# Unsloth Zoo - Utilities for Unsloth
+# Copyright 2023-present Daniel Han-Chen, Michael Han-Chen & the Unsloth team. All rights reserved.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published
+# by the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+# A required mlx-lm ModelArgs field absent from config.json raises a bare
+# TypeError naming neither the model nor the file, which reads as missing Apple
+# Silicon support; the guard must name the config as the cause instead.
+
+from __future__ import annotations
+
+import pytest
+
+
+@pytest.fixture(autouse=True, scope="module")
+def _install_mlx_shim():
+    from mlx_simulation import simulate_mlx_on_torch
+
+    simulate_mlx_on_torch()
+
+
+# Verbatim from mlx-lm 0.31.3 loading unsloth/LFM2.5-230M, whose mirrored
+# config.json lost upstream's "block_ff_dim": 2560 (unslothai/unsloth#7306).
+_LFM2_MSG = (
+    "ModelArgs.__init__() missing 1 required positional argument: 'block_ff_dim'"
+)
+
+
+def test_missing_config_key_raises_actionable_error():
+    from unsloth_zoo.mlx.loader import _raise_if_incomplete_mlx_config
+
+    with pytest.raises(ValueError) as exc:
+        _raise_if_incomplete_mlx_config(
+            "unsloth/LFM2.5-230M", "lfm2", _LFM2_MSG, TypeError(_LFM2_MSG)
+        )
+    msg = str(exc.value)
+    assert "unsloth/LFM2.5-230M" in msg
+    assert "block_ff_dim" in msg
+    assert "config.json" in msg
+    assert "lfm2" in msg
+    # The whole point: stop readers concluding MLX is unsupported.
+    assert "Apple Silicon" in msg
+
+
+def test_multiple_missing_keys_all_named():
+    from unsloth_zoo.mlx.loader import _raise_if_incomplete_mlx_config
+
+    message = (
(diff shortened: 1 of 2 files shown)


BardiaKoopah:
Follow-up: this isn't a one-off. I audited every `unsloth/*` safetensors mirror against the mlx-lm arch it would load under, and **five live mirrors currently fail this way**.

Method: 1,426 `unsloth/*` repos → 390 whose `model_type` maps to an mlx-lm arch with no-default `ModelArgs` fields (excluding GGUF and bnb-4bit, which don't take this path) → fetched each `config.json` → ran mlx-lm's real `_get_classes` + `ModelArgs.from_dict` against it, and against the upstream repo it was mirrored from. Metadata only, no weights.

| mirror | missing key | downloads/mo | upstream |
|---|---|---:|---|
| `unsloth/LFM2.5-230M` | `block_ff_dim` | 8,802 | OK |
| `unsloth/Qwen3-Coder-Next` | `rope_theta` | 989 | OK |
| `unsloth/ERNIE-4.5-0.3B-PT` | `tie_word_embeddings` | 406 | OK |
| `unsloth/ERNIE-4.5-21B-A3B-Thinking` | `tie_word_embeddings` | 111 | OK |
| `unsloth/ERNIE-4.5-21B-A3B-PT` | `tie_word_embeddings` | 19 | OK |

Every one of these fails at `ModelArgs.__init__` with the bare `TypeError` this PR annotates, and in every case the upstream config builds fine — so the key was lost in mirroring, not missing at source. Three distinct architectures (`lfm2`, `qwen3_next`, `ernie4_5`/`ernie4_5_moe`) and three different keys, which is what makes a generic guard worth having over a per-arch fix.

Two things the audit ruled out, for completeness:

- `unsloth/Qwen2-VL-*` (5 repos) looked broken on a static field comparison but build fine — `_get_classes` remaps them. Static analysis alone gives false positives here; only running `from_dict` is decisive.
- `unsloth/NVIDIA-Nemotron-3-Ultra-550B-A55B` fails on `num_hidden_layers`, but `nvidia/...-BF16` fails identically, so that's an mlx-lm/`nemotron_h` gap rather than a mirroring problem. Out of scope here.

This PR doesn't fix any of the five — the fix is restoring the keys on the Hub. It just makes the failure say so instead of looking like missing Apple Silicon support.

BardiaKoopah:
@codex review

danielhanchen:
Confirmed both mlx-lm load_model call sites in unsloth_zoo/mlx/loader.py catch only ValueError, so a missing ModelArgs key does surface as a bare TypeError with no model or file named. Will get this reviewed.

danielhanchen:
Confirmed the TypeError from ModelArgs.__init__ still escapes the ValueError-only handler in unsloth_zoo/mlx/loader.py, and the __init__ anchoring keeps mlx-lm signature drift out of it. Will get this reviewed.

danielhanchen:
Confirmed the TypeError from ModelArgs.__init__ still escapes _load_mlx_lm_with_strict_fallback in unsloth_zoo/mlx/loader.py with no Unsloth context, and the __init__ anchor keeps mlx-lm signature drift out of it. Will get this reviewed.

PR #1013, opened 2026-08-08
MLX VLM: name the config key when a mirrored config.json is incomplete

BardiaKoopah:
> **Stacks on #1004.** That PR adds the diagnostic and the helper this one extends; the diff here shows #1004's commit too until it merges. Review/merge #1004 first, after which this collapses to the VLM-only change.

### The problem

mlx-vlm's `BaseModelConfig.from_dict` (`mlx_vlm/models/base.py`) filters the config on `inspect.signature(cls).parameters` before constructing — exactly as mlx-lm's `from_dict` does. So a required field missing from a mirrored `config.json` fails in `ModelConfig.__init__` with a bare TypeError that names neither the model nor the file:

```
TypeError: ModelConfig.__init__() missing 1 required positional argument: 'text_config'
```

Reproduced against mlx-vlm 0.6.3. That reads as "MLX doesn't support this model", which sends people to the wrong repo — the same failure mode #1004 fixes for the text side (see unslothai/unsloth#7306, where a mirrored LFM2.5 config had lost `block_ff_dim`).

This is not a corner case: **43 of mlx-vlm 0.6.3's architectures declare at least one required no-default `ModelConfig` field**, most commonly `text_config`, `vision_config`, `projector_config` and `model_type`.

### The change

All three VLM load entry points now surface the actionable error:

| Site | Note |
|---|---|
| `_load_mlx_vlm_with_extra_weight_filter` | First load only. The retry is deliberately unguarded — it runs only after a `ValueError`, which means the config already constructed. |
| `_load_mlx_vlm_distributed` | Already owned a `TypeError` handler for `sharded_load` signature drift. The config check runs **first** because it is the precise, `__init__`-anchored test; the drift branch stays reachable since that pattern cannot match `sharded_load() got an unexpected keyword argument 'tensor_group'`. |
| the runtime-quant branch | Calls `vlm_load` directly and so bypasses the extra-weight filter. The QK-norm guard already had to be repeated there for the same reason. |

`_raise_if_incomplete_mlx_config` grows a `library` argument so the message names mlx-vlm rather than mlx-lm. It defaults to `"mlx-lm"`, leaving #1004's text-side call sites byte-identical.

### Tests

`tests/test_mlx_incomplete_config_vlm.py`, 10 tests. **6 fail on the parent commit**; the 4 that pass there assert pre-existing behaviour (message-shape recognition, the mlx-lm default, non-matching TypeErrors passing through, successful loads untouched).

Beyond the behavioural tests, two assertions read `loader.py`'s source so a **fourth** VLM load path added later without the diagnostic fails CI rather than shipping the raw TypeError, and one pins the distributed handler's ordering.

Both incomplete-config test files are registered in the CPU gate — #1004's was not registered either. CPU-pure via the MLX simulation shim, no weights; the two together run in ~0.2s.

### Verification

- `verify_env.py` exit **0** (mlx 0.31.2 / mlx-lm 0.31.3 / mlx-vlm 0.6.3, `unsloth_zoo` resolving from the worktree)
- `tests/test_mlx_incomplete_config_vlm.py` — 10 passed
- `tests/test_mlx_incomplete_config.py` — 5 passed (no regression from the `library` parameter)
- `tests/test_mlx_qk_norm_version_gap.py` — 7 passed
- `tests/test_mlx_distributed_loader.py` — 7 passed
- narrow ruff gate (`--select E9,F63,F7,F82`): 13 findings, **identical to the parent baseline**; YAML parses; `compileall` clean

One invocation per file, per the repo's MLX suite convention.

### DIFF
--- a/.github/workflows/consolidated-tests-ci.yml
+++ b/.github/workflows/consolidated-tests-ci.yml
@@ -129,11 +129,22 @@ jobs:
         # so a 3.10-only `match` and a module-level `str | Path` both reached main and
         # made the package un-importable on 3.9. It reads the floor from pyproject, so
         # raising requires-python relaxes it rather than making it lie.
+        #
+        # The two test_mlx_incomplete_config files pin the diagnostic that turns a
+        # bare "ModelArgs/ModelConfig.__init__() missing N required positional
+        # arguments" into a message naming the model, the missing keys and
+        # config.json. Several of their assertions read loader.py's source to prove
+        # every mlx-lm / mlx-vlm load entry point is guarded, so a new load path
+        # added without the guard fails here rather than shipping the raw
+        # TypeError. CPU-pure via the MLX simulation shim, no weights, both files
+        # together run in about two tenths of a second.
         run: |
           python -m pytest \
             tests/test_training_utils_use_cache.py \
             tests/test_mlx_finetune_last_n_layers.py \
             tests/test_mlx_double_quant_reject.py \
+            tests/test_mlx_incomplete_config.py \
+            tests/test_mlx_incomplete_config_vlm.py \
             tests/test_hf_xet_fallback.py \
             tests/test_get_transformers_model_type_empty.py \
             tests/test_train_on_responses_list_labels.py \
--- a/tests/test_mlx_incomplete_config.py
+++ b/tests/test_mlx_incomplete_config.py
@@ -1,0 +1,122 @@
+# Unsloth Zoo - Utilities for Unsloth
+# Copyright 2023-present Daniel Han-Chen, Michael Han-Chen & the Unsloth team. All rights reserved.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published
+# by the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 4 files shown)


BardiaKoopah:
@codex review

danielhanchen:
Confirmed this hits the three mlx-vlm load paths in unsloth_zoo/mlx/loader.py where a required ModelConfig field missing from a mirrored config.json still surfaces as a bare TypeError, and the guard itself looks right. Could you drop the assertions that read loader.py's source and count call sites, and keep the tests behavioural, since those will break on any unrelated edit to that file?

danielhanchen:
Confirmed the VLM load paths in unsloth_zoo/mlx/loader.py still surface the bare ModelConfig.__init__ TypeError, and this names the model and the missing key instead; since it stacks on #1004, could you rebase once that lands so the diff is VLM-only, and trim the tests that assert on inspect.getsource of loader.py to behavioural checks?

danielhanchen:
Confirmed the three mlx-vlm load paths in unsloth_zoo/mlx/loader.py still surface the bare ModelConfig.__init__ TypeError today, so the diagnostic is worth having; could you drop the assertions in tests/test_mlx_incomplete_config_vlm.py that inspect loader.py's source text and rebase once #1004 lands so this is VLM-only?

danielhanchen:
Checked that main's mlx-vlm load paths in unsloth_zoo/mlx/loader.py still surface the bare ModelConfig TypeError, so this is useful once #1004 lands. Will review after that one and the conflicts are resolved.

PR #1019, opened 2026-08-08
Dispatch on width in the GRPO PrefixGrouper forward too

danielhanchen:
## What

`grpo_accumulated_loss` picks between the lm_head matmul helper
(`chunked_hidden_states_selective_log_softmax`) and the raw-logits one
(`chunked_selective_log_softmax`) on the width of what the forward returned, at
both of its own call sites. The PrefixGrouper forward, `_pg_grad_forward`,
did not. It handed `.logits` straight to
`GroupLayout.extract_logps`, which always calls the helper it is given as
`(hidden, lm_head, ids, chunks, ...)`, so raw `[T, vocab]` logits reached the
lm_head matmul:

```
a and b must have same reduction dim, but got [((s47*s87 + 255)//256), s33] X [1536, 151936]
```

`.logits` carries hidden states only when the model's forward is the generated
one that honours `UNSLOTH_RETURN_HIDDEN_STATES`; when generation falls back to
trl's own class it is a real `[T, vocab]` tensor. That is the same case
`grpo_accumulated_loss` already dispatches for.

## Why it was invisible

The handler below the call swallowed it. It untrusted the layout signature, set
`UNSLOTH_RETURN_HIDDEN_STATES=1` (a forward that already ignores that flag keeps
ignoring it) and fell back to the packed or padded result, so the step still
produced the right numbers. But the signature is untrusted again on the next
step, so a full extra forward was run and thrown away on every step for the rest
of the run, and PrefixGrouper never engaged at all for these models.

## Fix

Pass a raw-logits helper carrying `extract_logps`' own signature when the width
does not match, so the matmul and the scale / softcap the forward already
applied are skipped, exactly as the packed branch does. Nothing changes on the
hidden-states path, and `extract_logps` itself is untouched, so this works
against current and older `unsloth` releases.

## Note on the reshape in `chunked_hidden_states_selective_log_softmax`

The helper reshapes on its own last dim on purpose, and this PR leaves that
alone. Checked directly on torch 2.9.1:

- Compiling the helper with `dynamic = True, fullgraph = True` against a real
  `nn.Parameter` lm_head and correct-width hidden states succeeds across several
  `(B, T)` shapes. The hidden-size symbol is backed, so `Eq(s, 1536)` is decided
  by the real value and the guard specialises. There is no guard pathology to
  fix; the message above only appears when the tensor genuinely is the wrong
  width, and the `151936` in it is the vocab size.
- Reshaping against `lm_head`'s width instead turns that clear message into
  `shape '[-1, 6]' is invalid for input of size s33*s87` when the vocab is not
  divisible by the hidden size, and into a broadcast mismatch several lines
  further down when it is. Neither names the operands.
- A message-carrying `torch._check` is still rejected under `fullgraph = True`
  with `Unsupported: Failed to convert args/kwargs to proxy`
  (pytorch/pytorch#149997, pytorch/pytorch#163668), and a bare one reports only
  `Expected cond to be True`.

The comment there is extended to name this call site as well.

## Tests

`tests/test_grpo_prefix_grouper_raw_logits.py`, 6 tests. The block is exec'd
from its own source against a model that returns raw logits and the result is
checked against per-row logprobs; the hidden-states variant covers the unchanged
path. Two parametrised tests pin the property the reshape comment protects, one
with a vocab divisible by the hidden size and one without, since the divisible
case is where reshaping on `lm_head`'s width would silently rewrite the row
count.

Reintroducing the bug fails 2 of the 6; swapping the reshape for
`lm_head.shape[1]` fails the other 2.

Full suite: 4380 passed, 162 skipped, 0 failed.
`tests/test_upstream_pinned_symbols_trl_vllm.py` is excluded, it aborts
collection with a `PermissionError` on a path outside this checkout.

### DIFF
--- a/tests/test_grpo_prefix_grouper_raw_logits.py
+++ b/tests/test_grpo_prefix_grouper_raw_logits.py
@@ -1,0 +1,198 @@
+# Unsloth Zoo - Utilities for Unsloth
+# Copyright 2023-present Daniel Han-Chen, Michael Han-Chen & the Unsloth team. All rights reserved.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published
+# by the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+"""The PrefixGrouper forward needs the same width dispatch as the packed path.
+
+`.logits` carries hidden states only when the forward is the Unsloth generated
+one honouring UNSLOTH_RETURN_HIDDEN_STATES. The packed call site and its
+verifier both dispatch on width; `_pg_grad_forward` handed `.logits` straight to
+`extract_logps`, which always feeds its helper the lm_head matmul signature. A
+forward returning real [T, vocab] logits therefore raised the reduction-dim
+mismatch inside PrefixGrouper, the outer handler untrusted the signature, and
+the whole verify forward was repeated and thrown away on every later step.
+
+The block is exec'd from its own source so the dispatch is exercised, not read.
+
+The second half pins the property the reshape comment in
+chunked_hidden_states_selective_log_softmax protects: a wrong-width caller must
+still fail with a message naming BOTH operands, never be silently reshaped.
+"""
+
+import contextlib
+import inspect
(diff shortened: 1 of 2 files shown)


danielhanchen:
Corrected a count in the description above.

It said `grpo_accumulated_loss` dispatches on width at three of four call sites. An AST pass over pre-PR `origin/main` finds five direct calls to `chunked_hidden_states_selective_log_softmax`:

```
line 1180  grpo_accumulated_loss    width dispatch: yes
line 1219  grpo_accumulated_loss    width dispatch: yes
line 1448  forward (autograd)       width dispatch: no
line 1473  backward (autograd)      width dispatch: no
line 1501  efficient_log_softmax    width dispatch: no
```

So it is two of its own two, not three of four, and the undispatched sites are the PrefixGrouper machinery rather than packing calls. The fix itself is unchanged; only the framing was wrong.

danielhanchen:
@codex review

PR #1021, opened 2026-08-09
Pick the compile switch's position instead of inheriting the runner's

danielhanchen:
## What is red

Two tests in `tests/test_generated_fullgraph_fallback.py` fail whenever `UNSLOTH_COMPILE_DISABLE` is set:

```
FAILED tests/test_generated_fullgraph_fallback.py::test_maybe_compile_routes_fullgraph_through_the_fallback
    AssertionError: assert False
FAILED tests/test_generated_fullgraph_fallback.py::test_the_alias_routes_a_fullgraph_compile_through_the_fallback
    AssertionError: assert None is True
```

unsloth's `.github/workflows/consolidated-tests-ci.yml` pins `UNSLOTH_COMPILE_DISABLE=1` at job level, clones this repo at `main` and runs `pytest tests/` from it. So these two have been failing in all three matrix cells of the `Core (HF=...)` check on every open PR in `unslothai/unsloth`, and that job is the only thing that runs this suite.

## Root cause

The test, not the code. `temporary_patches/common.py` reads the flag into module constants at import:

```python
UNSLOTH_COMPILE_DISABLE = os.environ.get("UNSLOTH_COMPILE_DISABLE", "0") in ("1", "partial",)
```

and both entry points short-circuit on it by design. With the flag set `_maybe_compile` returns `lambda fn: fn` and `torch_compile` is `noop`, so nothing is compiled and nothing reaches `torch_compile_with_fallback`. That is the escape hatch working exactly as intended.

The two checks asserted the routing unconditionally, so with the flag set they read a correct no-op as a missing fallback. They went in with the file in #1001 and have never been run in that position here, since this repo has no CI of its own on `main`.

## The fix

The flag is read at import, so no in-process fixture can move it. Probe it in a subprocess, the same shape `tests/test_rl_replacements_compile_disable.py` and `tests/test_compile_disable_partial.py` already use, and assert both halves of the contract instead of only the one the runner happens to be in.

- `test_every_fullgraph_entry_point_carries_the_fallback` -- with compilation on, `torch_compile`, `_torch_compile` and `_maybe_compile` all reach the wrapper under `fullgraph`, in both the decorator and the positional spelling that `gemma.py` and `gpt_oss.py` use.
- `test_a_non_fullgraph_compile_is_left_alone` -- nothing else routes, `_raw_torch_compile` included, since its one caller applies the wrapper itself.
- `test_the_escape_hatch_still_turns_compilation_off[1|partial]` -- with the flag set nothing routes, the aliases hand back a Dynamo-disabled callable and `_maybe_compile` hands back the function itself. The routing must not become a way back in. This is the position the unsloth job runs in.

`test_maybe_compile_routes_fullgraph_through_the_fallback` keeps its source-level half under a name that says what is left (`test_maybe_compile_asks_for_the_fallback_only_under_fullgraph`); its runtime half moved into the probe.

The non-fullgraph direction needed a second observable. `_unsloth_fallback_state` is stamped on only under `fullgraph`, so it cannot tell a plain compile that wrongly went through the helper from one that did not: `torch_compile_with_fallback(fullgraph = False)` hands back a plain compiled object either way. The probe records the calls as well, with a positive control first, so a recorder that stops recording fails rather than passing everything for free.

No production code changed.

## Mutation results

Each mutation breaks the production behaviour one assertion guards, then the named test runs. All nine caught, all restored green afterwards.

| # | Mutation in `temporary_patches/common.py` | Caught by |
|---|---|---|
| 1 | `_compile_or_fall_back` never routes | `test_every_fullgraph_entry_point_carries_the_fallback` |
| 2 | `torch_compile` partials `torch.compile` directly (the original regression) | same, plus `test_the_alias_sites_exist_and_are_covered_by_the_alias_itself` |
| 3 | `_torch_compile` partials `torch.compile` directly | same two |
| 4 | `_maybe_compile` drops the fullgraph routing | same, plus `test_maybe_compile_asks_for_the_fallback_only_under_fullgraph` |
| 5 | `_compile_or_fall_back` routes non-fullgraph compiles too | `test_a_non_fullgraph_compile_is_left_alone` |
| 6 | `_maybe_compile` ignores the escape hatch | `test_the_escape_hatch_still_turns_compilation_off` |
| 7 | the aliases ignore the escape hatch | `test_the_escape_hatch_still_turns_compilation_off` |
| 8 | `_raw_torch_compile` double-wrapped through the fallback | `test_a_non_fullgraph_compile_is_left_alone` |
| 9 | `_maybe_compile` routes non-fullgraph compiles too | `test_a_non_fullgraph_compile_is_left_alone` |

Mutation 5 was missed by the first draft, which is what prompted the call recorder and the positive control.

## Suite baseline

Whole `tests/` suite, CPU, `UNSLOTH_COMPILE_DISABLE=1` to match the unsloth job, with the two CUDA-host tests that job deselects deselected:

| | result |
|---|---|
| before (`40963429`) | 16 failed, 4238 passed, 291 skipped |
| after | 14 failed, 4242 passed, 291 skipped |

The two removed are the two above. The other 14 are pre-existing, untouched by this PR and unrelated to it:

- `test_file_uri_resolution.py` (4)
- `test_gpt_oss_attention_mask.py` (1)
- `test_moe_lora_grouped_mm_capability.py` (1)
- `test_pretokenized_bypass_collator.py` (4)
- `test_rl_replacements_compile_disable.py` (4)

The last four are an artefact of the machine they were measured on rather than of this repo: that file's probe sets `PYTHONPATH = str(ROOT)`, replacing it, and the environment here needs an inherited path entry for the subprocess to import at all. The probe added in this PR prepends instead, for that reason. Left alone as out of scope.

The three `Core (HF=...)` cells on unsloth should go green with this merged, since they clone zoo `main` at job time.

### DIFF
--- a/tests/test_generated_fullgraph_fallback.py
+++ b/tests/test_generated_fullgraph_fallback.py
@@ -31,7 +31,13 @@ Lowering the limit makes the failure reachable on any GPU -- a T4 gets there
 unaided, a B200 at the default never does.
 """
 
+import json
+import os
 import re
+import subprocess
+import sys
+import textwrap
+from functools import lru_cache
 from pathlib import Path
 
 import pytest
@@ -81,22 +87,16 @@ def test_the_grpo_loss_regions_are_wrapped():
         "accumulate_chunk is still compiled bare"
 
 
-def test_maybe_compile_routes_fullgraph_through_the_fallback():
+def test_maybe_compile_asks_for_the_fallback_only_under_fullgraph():
     """Three more fullgraph regions go through this one helper, so wiring it
-    covers them without touching each decorator."""
-    from unsloth_zoo.temporary_patches import common
+    covers them without touching each decorator. What it does at runtime is in
+    `test_every_fullgraph_entry_point_carries_the_fallback` below, which has to
+    pick the compile switch's position rather than inherit the runner's."""
     src = (ZOO / "temporary_patches" / "common.py").read_text()
     body = src.split("def _maybe_compile(", 1)[1].split("\ndef ", 1)[0]
     assert "torch_compile_with_fallback" in body
     # Without fullgraph Dynamo already falls back by itself, so leave it alone.
     assert 'if not kwargs.get("fullgraph"):' in body
-
-    def f(x):
-        return x
-    wrapped = common._maybe_compile(fullgraph = True, dynamic = True)(f)
-    assert hasattr(wrapped, "_unsloth_fallback_state")
-    plain = common._maybe_compile(dynamic = True)(f)
-    assert not hasattr(plain, "_unsloth_fallback_state")
 
 
 def test_the_generated_preamble_imports_the_helper():
@@ -235,26 +235,12 @@ def test_the_alias_sites_exist_and_are_covered_by_the_alias_itself():
     assert raw, "the raw alias compile_with_eager_fallback needs is gone"
 
 
-def test_the_alias_routes_a_fullgraph_compile_through_the_fallback(monkeypatch):
-    from unsloth_zoo.temporary_patches import common as C
-    seen = {}
-
-    def _fake(**kwargs):
-        seen.update(kwargs)
-        return lambda fn: fn
-
-    monkeypatch.setattr("unsloth_zoo.temporary_patches.utils."
-                        "torch_compile_with_fallback", _fake)
-
-    @C.torch_compile(dynamic = True, fullgraph = True)
-    def _f(x): return x
-
-    assert seen.get("fullgraph") is True
-
-
 def test_the_alias_still_accepts_a_function_positionally():
     """`prepare = torch_compile(prepare, fullgraph = True)` is how gemma.py and
-    gpt_oss.py call it, and a kwargs-only helper would TypeError there."""
+    gpt_oss.py call it, and a kwargs-only helper would TypeError there.
+
+    True in either position of the compile switch: with it off the alias is
+    `noop`, which takes a function positionally too."""
     from unsloth_zoo.temporary_patches import common as C
 
     def _f(x): return x * 2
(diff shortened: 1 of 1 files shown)


danielhanchen:
Closing: superseded by #1022, which landed the same fix first.

This PR and #1022 were two answers to one red. #1022 merged as 4eadaa0 and is on `main`, so the two failures quoted
at the top of this description no longer happen. On `main` at 94b8860, `tests/test_generated_fullgraph_fallback.py`
is green in every position of the switch:

```
unset     124 passed, 2 skipped   EXIT=0
=1        EXIT=1, and the only failure is test_rmsnorm_recompile_guards::test_the_kernels_fall_back_to_eager_instead_of_aborting
=partial  EXIT=1, same single failure
```

(that remaining one is a different file, fixed in #1018.)

Rebasing this onto `main` conflicts in exactly the four hunks where #1022 put its own version of the fix, so
resolving it would mean choosing between two implementations of a fix that is already in the tree. Not worth it for
what is left.

Two assertions here were not redundant, and both were of the kind that passes on a broken tree, so they have been
carried into #1018 rather than dropped:

- `_maybe_compile`'s non-fullgraph case asserted the absence of `_unsloth_fallback_state`. The helper stamps that
  marker only under `fullgraph`, so it is absent whether or not a plain compile wrongly went through it. The
  call-recorder from this PR replaces it. Replacing the guard's body with `pass`, leaving its source text so the
  scan above still passes, went unnoticed unset, at `=1` and at `=partial` before; it is red in all three now.
- `test_a_non_fullgraph_compile_is_left_alone` ran in process, where the alias is `noop` whenever the switch is off
  and the recorder can never fire however the routing is written. That is the position unsloth's Core job runs the
  suite in. It now runs in the compile-enabled interpreter with a positive control, and covers `_torch_compile`,
  the positional spelling and `_raw_torch_compile` as this PR did.

The whole-file subprocess probe is not carried over. #1018 establishes that `_maybe_compile` reads both switches
inside its own body on every call, so a monkeypatch on the module attributes is exact there and a fresh interpreter
is only needed for the three import-bound aliases.

PR #1026, opened 2026-08-09
Fix MLX ragged VLM image batches

Lyxot:
## Summary

- preserve every `pixel_values` entry when aspect-ratio-dependent VLM preprocessing produces tensors that cannot be stacked
- keep the existing dense stack path for equal-shaped images
- add regression coverage for ragged ordering, reverse ordering, and dense-row data preservation

## Root cause

Gemma 4 can return one processed image tensor per example when a batch contains different aspect ratios. `_to_mx_vlm_batch` attempted to stack those tensors, caught the shape error, and then retained only `value[0]`. The vision tower consequently received a single 3-D image instead of the full batch and failed while unpacking `B, C, H, W`.

## Impact

Mixed-aspect-ratio MLX VLM batches now retain every processed image in order so model-specific eager vision paths can handle the ragged list. Equal-shaped batches remain dense.

## Validation

- `python -m pytest tests/test_mlx_vlm_label_masks.py -q` — 228 passed
- mutation checks independently verified the ragged fallback and dense-row data assertions
- validated with the first two `unsloth/LaTeX_OCR` rows and `mlx-community/gemma-4-E2B-it-qat-4bit`: both `(3, 384, 1584)` and `(3, 480, 1200)` tensors reached the real Gemma 4 input-embedding path

### DIFF
--- a/tests/test_mlx_vlm_label_masks.py
+++ b/tests/test_mlx_vlm_label_masks.py
@@ -4004,6 +4004,41 @@ def test_clips_of_unequal_duration_all_reach_the_model():
     assert tuple(stacked.shape) == (2, 80, 300)
 
 
+def test_mixed_aspect_ratio_images_all_reach_the_model():
+    """Gemma 4 keeps differently shaped processed images as a ragged list.
+
+    The generic conversion used to catch the failed stack and retain only the
+    first tensor, so a batch of two OCR examples reached the vision tower as one
+    unbatched 3-D image. Preserve every image and its row order instead.
+    """
+    from unsloth_zoo.mlx.utils import _to_mx_vlm_batch
+
+    source = (
+        np.full((3, 4, 7), 11, dtype=np.float32),
+        np.full((3, 6, 5), 29, dtype=np.float32),
+    )
+    for images in (source, source[::-1]):
+        pixels = _to_mx_vlm_batch({"pixel_values": list(images)})["pixel_values"]
+
+        assert isinstance(pixels, list)
+        assert [tuple(image.shape) for image in pixels] == [tuple(image.shape) for image in images]
+        assert [float(np.asarray(image)[1, -1, -1]) for image in pixels] == [
+            float(image[1, -1, -1]) for image in images
+        ]
+
+
+def test_equal_shape_images_stack_without_data_loss():
+    from unsloth_zoo.mlx.utils import _to_mx_vlm_batch
+
+    dense_source = [
+        np.full((3, 4, 7), 41, dtype=np.float32),
+        np.full((3, 4, 7), 73, dtype=np.float32),
+    ]
+    dense = _to_mx_vlm_batch({"pixel_values": dense_source})["pixel_values"]
+    assert tuple(dense.shape) == (2, *dense_source[0].shape)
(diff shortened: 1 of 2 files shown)


danielhanchen:
Verified this on real Apple Silicon and on Linux. No changes needed, it is correct as written.

### The bug is real and reproduces both ways

On `main`, two images with different aspect ratios collapse to the first one:

```
main:     [(3, 384, 1584), (3, 480, 1200)]  ->  array (3, 384, 1584)   second image dropped
this PR:  [(3, 384, 1584), (3, 480, 1200)]  ->  list of 2, both preserved
```

Equal shaped batches still take the dense path and keep every row: `(2, 3, 384, 384)` with row 1 intact. Confirmed on a macos-14 runner with real mlx on `Device(gpu, 0)`, and again with the real `SmolVLM-256M` processor, where `main` collapsed `(5, 3, 512, 512)` and `(9, 3, 512, 512)` to the first and this branch keeps both.

### The ragged list is what the consumer wants

`mlx_vlm/models/gemma4/processing_gemma4.py:252-257` is the producer: it stacks when all shapes agree and returns a plain list otherwise. `mlx_vlm/models/gemma4/vision.py:458-475` `VisionModel.__call__` opens with `if isinstance(pixel_values, list):` and concatenates per image. The chain from `_vlm_cce_forward` through `gemma4.py:129` `_scatter` reaches that branch intact, and Gemma 4 emits no companion `image_position_ids` for images, so there is no desynchronisation risk. Before this PR a mixed aspect ratio Gemma 4 batch silently trained on image 0 only.

### I checked whether the list can crash a consumer, and it cannot

The obvious worry is `compile.py:3155` and its siblings at 3309, 3663 and 3713 doing `pixel_values.astype(dtype)`, which a Python list has no method for. It is unreachable:

- Qwen's collator makes one batched processor call and gets back a single concatenated 2-D array, `(1024, 1176)` for two different sized images, delimited by `image_grid_thw`, not a per example list. Verified through `_processor_vlm_inputs` in both worktrees. So `_to_mx_vlm_batch` takes the `hasattr(value, "shape")` branch and the new code never runs.
- Swept all 94 image processors in transformers 5.15 with `return_tensors="np"`: 78 return an array, 0 return a list. The would be ragged ones raise inside `BatchFeature` rather than handing back a list.

So the only list producer in practice is gemma4, whose tower handles lists.

One side effect worth noting: this also fixes DeepSeek-OCR. Its contract is a two element `[patches, global_images]` pair with different shapes, so pre-PR the stack failed and `pixel_values[1]` read a row of patches.

Only follow up I would consider, and it is latent rather than a problem today: `image_grid_thw`, `image_sizes`, `spatial_shapes` and `pixel_values_videos` are not in `_VLM_RAGGED_MEDIA_PAYLOAD_KEYS`, so if a future processor ever returns per row lists for those they would still collapse to `value[0]`. Gemma 4 emits none of them, and `_VLM_PER_ROW_MEDIA_KEYS` already covers `audio_bounds`, `image_bound` and `tgt_sizes`.

Full suites, one pytest process per file: `test_mlx_vlm_label_masks.py` 228 passed, `test_mlx_save_export_regressions.py` 49, `test_mlx_shape_guard.py` 32. Apple Silicon CI green.

PR #1029, opened 2026-08-10
Apply logit scale and softcap on the compiled inference path

danielhanchen:
## Problem

The forward templates in `compiler.py` have an `elif labels is None:` arm that returns logits straight to the caller. It ends at

```python
logits = self.lm_head(hidden_states...)
```

with none of the logit transforms that the training arms immediately below it apply:

```python
if (scale_multiply) != ():
    logits = logits * (scale_multiply)
if (scale_divide) != ():
    logits = logits / (scale_divide)
if (softcapping) not in (None, (),):
    logits = logits / (softcapping)
    logits = torch.tanh(logits)
    logits = logits * (softcapping)
```

So for any model that configures `final_logit_softcapping` or a logit scale, inference returned untransformed logits while training used transformed ones. The two paths disagreed.

## Reproduction

`unsloth/gemma-2-2b-it` sets `final_logit_softcapping = 30.0`. Prompt "The capital of France is", compiled inference path:

| | absmax logit |
|---|---|
| before | 34.500, above the cap of 30.0 |
| after | 24.500 |

A logit of 34.5 against a cap of 30.0 is direct evidence the cap was not applied.

## Scope, stated precisely

This is narrower than it first looks. `tanh` is monotonic, so **top-1 is unchanged and greedy decoding is unaffected**. What changes is the shape of the distribution. On that prompt the last-position top-1 probability moves from 0.4154 to 0.3153 and entropy from 2.0891 to 2.7062 nats, with `KL(correct || previous) = 0.0562` nats.

So the paths affected are temperature and top-p sampling, returned logprobs, and anything scoring with the logits.

Models that leave these unset are unaffected: the emitted guards collapse to `if () != ():`, exactly as they already do in the training arms today.

## Change

Apply the same three transforms, in the same order, in the inference arm. Three template sites.

## Test

`tests/test_compiler_inference_logit_transforms.py` is a source guard, because reaching the generated code needs a real checkpoint and a GPU. It walks every `elif labels is None:` arm and asserts the softcap is applied before the arm ends, and separately pins the transform order shared with the training arms.

Against the current templates it fails with

```
AssertionError: an inference arm returns logits without applying the softcap;
the training arms below it do, so the two disagree
```

and passes with this change.

### DIFF
--- a/tests/test_compiler_inference_logit_transforms.py
+++ b/tests/test_compiler_inference_logit_transforms.py
@@ -1,0 +1,67 @@
+# Unsloth Zoo - Utilities for Unsloth
+# Copyright 2023-present Daniel Han-Chen, Michael Han-Chen & the Unsloth team. All rights reserved.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+"""The compiled inference branch must apply the logit transforms.
+
+`compiler.py` emits several forward templates. Each has an `elif labels is
+None:` arm that returns logits straight to the caller, and training arms that
+apply the captured logit scale and softcap. When the inference arm skips those
+transforms the returned distribution is wrong for any model configuring them:
+greedy decoding is unaffected because tanh is monotonic, but sampling,
+logprobs and logit-based scoring are not.
+
+This is a source guard rather than a runtime test, because reaching the
+generated code needs a real checkpoint and a GPU.
+"""
+import re
+
+import unsloth_zoo.compiler as compiler_module
+
+
+def _template_source() -> str:
+    import inspect
+    return inspect.getsource(compiler_module)
+
+
+def test_every_inference_branch_applies_the_softcap():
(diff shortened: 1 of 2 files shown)


danielhanchen:
@codex review

danielhanchen:
@codex review

danielhanchen:
Extra evidence that this is an omission rather than a deliberate fast path, and that it cannot double-apply.

**It is an omission.** `git log` puts the inference arm before the transforms existed. It entered as a bare shortcut:

```
1101ee09 Update compiler.py (#37)
-if not self.training and labels is None:
+if labels is None:
     logits = self.lm_head(hidden_states)
```

The `$LOGITSCALINGMULTIPLY$` / `$LOGITSCALINGDIVISION$` / `$LOGITSOFTCAPPING$` capture groups only arrived later, in e98791ab (Gemma 3), and were wired into the training arms. The shortcut was never revisited.

**Unsloth's other forward already does this.** `unsloth/models/llama.py`, `CausalLM_fast_forward`, on the same `labels is None` path:

```python
if logit_scaling != 0:
    logits *= logit_scaling
if logit_softcapping != 0:
    logits *= 1.0 / logit_softcapping
    logits.tanh_()
    logits *= logit_softcapping
```

So the hand-written forward and the generated one disagreed about the same model. This brings the generated path in line.

**No double application.** The find pattern consumes the model's own softcap lines (they are the `$LOGITSOFTCAPPING$` capture), so the emitted arm is the only place they can be applied. Verified by running `apply_fused_lm_head` over the real `Gemma2ForCausalLM.forward` from transformers and executing the emitted arm:

| build | softcap | emitted arm absmax | HF reference absmax | matches |
| main | 30.0 | 22.6906 | 19.1677 | no |
| this PR | 30.0 | 19.1677 | 19.1677 | yes, atol 0 |
| main | None | 22.6906 | 22.6906 | yes |
| this PR | None | 22.6906 | 22.6906 | yes |

A model that leaves the fields unset is byte for byte unchanged, and all three replacement templates carry the same block, so no arm is left behind.

PR #1031, opened 2026-08-10
Pick a module sizing backend that can carry the fp32 loader exceptions

danielhanchen:
Follow-up to #1028. The cross-platform runs on that PR's final head went red on ubuntu-latest and windows-latest with

```
FAILED tests/test_device_map_planner.py::test_fp32_loader_exceptions_apply_without_a_quantizer - assert 256 == (256 * 2)
```

and the failure is real, not a runner artefact.

## What happens

`_compute_module_sizes` walks two implementations in preference order:

* `transformers.integrations.accelerate.compute_module_sizes(model, hf_quantizer = ...)`
* `accelerate.utils.modeling.compute_module_sizes(model, dtype = ..., special_dtypes = ...)`

The transformers one is preferred because it asks the quantiser for each parameter's element size, which is exact. But it takes **only** `hf_quantizer`. On an unquantised model the information we need to pass is `special_dtypes`, which names the modules `from_pretrained` loads in float32 because the model class declares `_keep_in_fp32_modules` (the T5 family keeps `wo` that way). That implementation cannot express it, so the filtered kwargs came out empty, the call succeeded with no arguments, and the result was returned as if it were the answer. The fp32 exceptions were silently dropped and an fp16 checkpoint was charged half for those modules, which is exactly the undercount the check was added to prevent.

It passed locally because this transformers does not ship `transformers.integrations.accelerate`, so the walk fell through to accelerate's implementation, which does accept `special_dtypes`. Only a runner with a newer transformers picks the other branch.

## The fix

Signature support now decides the **order**, not just the arguments. An implementation that cannot carry everything `_quantized_size_kwargs` asked for is kept as a fallback and the walk continues to one that can; the fallback is used only if nothing better answers. The quantised path is unchanged, since `hf_quantizer` conveys everything there and still returns immediately.

## Test

`test_fp32_exceptions_survive_a_transformers_that_only_takes_a_quantizer` injects a `transformers.integrations.accelerate.compute_module_sizes` with the quantiser-only signature, so the preference branch is taken on any environment. It fails on current main with the same `256 == 512` assertion the runners hit, and passes here. The existing test that only exercises the accelerate path is kept.

41 tests pass locally.

### DIFF
--- a/tests/test_device_map_planner.py
+++ b/tests/test_device_map_planner.py
@@ -968,3 +968,48 @@ def test_fp32_loader_exceptions_apply_without_a_quantizer():
     assert upcast["norm"] == plain["norm"] * 2
     assert upcast["layers"] == plain["layers"]
     assert upcast[""] == plain[""] + plain["norm"]
+
+
+def test_fp32_exceptions_survive_a_transformers_that_only_takes_a_quantizer(monkeypatch):
+    """Signature support has to decide the ORDER, not just the arguments.
+
+    `transformers.integrations.accelerate.compute_module_sizes` is preferred and
+    takes only `hf_quantizer`, so on an unquantised model it cannot express the
+    fp32 loader exceptions at all. Returning its answer anyway dropped them and
+    charged an fp16 T5's `wo` half; the local environment happened not to ship
+    that implementation, so only the cross-platform runners caught it."""
+    import types
+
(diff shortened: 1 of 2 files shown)


danielhanchen:
@codex review

danielhanchen:
@codex review

PR #1033, opened 2026-08-10
fix(gguf): disable missing MTP export

Lyxot:
## Summary

- pass `--no-mtp` to llama.cpp text-model conversion when existing reconciliation finds that MTP is declared but the checkpoint has no MTP tensors
- preserve normal MTP metadata and export when MTP tensors are present
- keep `--no-mtp` off the mmproj command during VLM dual conversion
- omit `--no-mtp` for legacy converter entrypoints that do not declare the option
- add focused command-routing regression coverage to the existing MTP reconciliation tests

## Root cause

This was triggered by an upstream llama.cpp update.

The Qwen3.5 base checkpoint declares MTP, but the MLX merged checkpoint exports only the main model layers and contains no MTP tensors. Zoo reconciles that mismatch by stripping `mtp_num_hidden_layers` before conversion.

Previously, llama.cpp treated absent MTP metadata as effectively meaning “no MTP,” so stripping the key was sufficient and conversion succeeded implicitly.

On August 3, llama.cpp commit [`4ed2b13f`](https://github.com/ggml-org/llama.cpp/commit/4ed2b13f758e) added Qwen3-Next MTP inference. With the updated converter, absent MTP metadata can instead mean “infer the MTP count from tensors.” For Zoo's MLX merged checkpoint, inference finds no MTP tensors, leaves the count at zero, and reaches the converter assertion.

Zoo therefore needs to explicitly pass `--no-mtp` to disambiguate “MTP intentionally omitted” from “MTP metadata missing; infer it from tensors.”

## Impact

Qwen3.5 MLX merged checkpoints that retain an MTP declaration but contain only the main model layers can once again export to GGUF. Checkpoints that retain MTP tensors continue through the normal MTP export path.

## Validation

- `python -m pytest tests/test_convert_to_gguf_mtp_reconcile.py -q` (`14 passed`)
- mutation-checked the missing-MTP text, VLM routing, and legacy-converter assertions
- loaded the Studio Qwen3.5 0.8B adapter checkpoint, performed the real MLX merge, and exported BF16 text and mmproj GGUFs
- loaded both exported files with the bundled `llama-server`; `/health` returned `{"status":"ok"}`

### DIFF
--- a/tests/test_convert_to_gguf_mtp_reconcile.py
+++ b/tests/test_convert_to_gguf_mtp_reconcile.py
@@ -166,8 +166,10 @@ def test_convert_to_gguf_reconciles_mtp_config_to_index(llama_cpp, tmp_path, has
     updated = json.loads(config_path.read_text(encoding = "utf-8"))
     assert "unsloth_fixed_mtp" not in updated
     assert "unsloth_fixed_mtp" not in updated["text_config"]
-    assert ("mtp_num_hidden_layers" in updated) is has_mtp
-    assert ("mtp_num_hidden_layers" in updated["text_config"]) is has_mtp
+    # The declaration survives either way: kept as-is with MTP tensors, and kept
+    # because `--no-mtp` carries the intent without them.
+    assert "mtp_num_hidden_layers" in updated
+    assert "mtp_num_hidden_layers" in updated["text_config"]
     command, = _read_converter_commands(tmp_path)
     assert ("--no-mtp" in command) is not has_mtp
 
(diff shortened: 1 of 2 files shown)


danielhanchen:
Tested this one end to end on Linux and pushed two fixes to the branch.

### Confirmed the bug and the fix

Built a checkpoint in exactly the shape you describe (`mtp_num_hidden_layers` declared, zero MTP tensors) from `trl-internal-testing/tiny-Qwen3_5ForConditionalGeneration` and ran the real patched converter against a current llama.cpp:

- base: `assert self.opt_num_mtp_layers != 0` in `conversion/qwen.py:297`, no output
- this PR: `--no-mtp` in argv, `out_pr_text.gguf` written, `general.architecture = qwen35`, `qwen35.block_count = 2`, no `nextn_predict_layers` key

VLM dual conversion also checked: `--no-mtp` lands on the text command and never on the mmproj command. For a model with no MTP declared, argv is byte identical to base, so nothing changes on the normal path.

One thing worth adding to the description, because it makes the legacy branch provably safe rather than just defensive: `--no-mtp` arrived in llama.cpp `25558268` (2026-05-16), which predates the assertion added by `4ed2b13f` (2026-08-03). Every converter that can hit the assertion therefore also declares the flag.

### Fix 1: the flag is architecture gated, not just version gated

`_mtp_declared` is a config.json key check with no architecture awareness, but the converter hard rejects the flag for architectures with no MTP block:

```
ERROR:hf-to-gguf:--mtp / --no-nextn are not supported for LlamaForCausalLM
```

I reproduced this through `convert_to_gguf`: base succeeds, this PR raises `RuntimeError`, and it fails after the merge has already run, citing a flag the caller never passed. The reachable case is Qwen3.5 weights republished under `Qwen3ForCausalLM`; of 790 HF configs carrying `mtp_num_hidden_layers`, about 5 route to a non MTP architecture.

Probing the converter source cannot rule this out, because `25558268` added the flag and its architecture allowlist in the same commit. So `_converter_supports_no_mtp() == True` is exactly co-extensive with "the fatal gate exists" and there is no safe older window.

Fixed by retrying once without the flag when the converter rejects it, in the existing retry loop next to the dependency repair and OOM retries. Dropping the flag is precisely the pre-PR behaviour, which is the correct one for an architecture with no MTP block to strip.

### Fix 2: the probe only caught `OSError`

`_converter_supports_no_mtp(None)` raised `TypeError` out of `convert_to_gguf` instead of answering False. Widened to `(OSError, TypeError, ValueError)` so an unusable location degrades to "cannot prove support".

### Checks that found nothing

Worth recording so they are not re-run: CRLF line endings, UTF-8 BOM, CRLF plus BOM, lone CR, UTF-16, invalid UTF-8 bytes, embedded NULs, an 11.2 MB file, a 10 MB pathological file (0.286s, no backtracking), symlinks, dangling symlinks, directories, unreadable files, and a 345 character path. All behave correctly. `[^)]*` already matches `\r` and `\n`, so the multi-line `parser.add_argument(...)` in the real converter matches with or without `re.DOTALL`. The patched converter that `_download_convert_hf_to_gguf` generates also keeps the argparse block intact, so there is no false negative on the production path.

Tests: 14 to 21, including a converter fake that declares `--no-mtp` and then refuses it. The new tests fail on the pre-fix code with the exact `RuntimeError` above.

PR #1039, opened 2026-08-11
Pin the transformers kernel layout the two FLA opt-out tests assume

danielhanchen:
`Core (HF=latest + TRL=latest)` is red on unsloth main. Two of its four failures are these:

```
FAILED tests/test_fla_hopper_tile.py::test_opt_out_forces_pure_torch
FAILED tests/test_fla_hopper_tile.py::test_opt_out_outranks_the_source_preference_flags
  AssertionError: UNSLOTH_DISABLE_VENDORED_FLA defeated the opt-out for qwen3_5
```

## What changed upstream

transformers #47630, between 5.14.1 and 5.15.0, replaced the module-global kernel selection with a decorator:

```python
@use_kernel_func_from_hub_with_fallback("chunk_gated_delta_rule", "fla")
```

`transformers/integrations/hub_kernels.py` resolves it **at decoration time** and freezes the result into a closure:

```python
def decorator(torch_function):
    implementation = None
    try:
        module = importlib.import_module(package)
        implementation = resolve_internal_import(module, full_path)
    except Exception:
        ...
```

So on 5.15 there is no availability probe to answer False and no module global to unbind.

## Stale test probes, not a new bug

`fla_vendor.py` already knows about this. `_transformers_uses_availability_probe()` detects the layout, and on the new one `patch_vendor_fla()` deliberately does not return: it warns and falls through to the protection path, so the fla the decorator resolved is still patched. That contract has five dedicated tests and all of them pass on 5.15.

These two were simply the only opt-out tests that never pinned the layout. They assert the old-layout observables (probe False, global `None`), so on 4.57.6 they pass and on 5.15.0 they silently start exercising the new-layout branch and fail. The failure tracks `hasattr(hub_kernels, "use_kernel_func_from_hub_with_fallback")`, not any zoo change.

## The change

Test-only, one file, +62/-9.

- `test_opt_out_forces_pure_torch` pins `_transformers_uses_availability_probe` to True, with a docstring saying why pure torch only exists on that layout and where the new-layout contract is covered.
- `test_opt_out_outranks_the_source_preference_flags` is parametrized over both layouts. Old layout keeps the original assertion; new layout asserts neither source-preference flag short-circuits the protection path. That is strictly more coverage - the new-layout `UNSLOTH_FORCE_VENDORED_FLA` case was untested.

Grepped the tree for `_transformers_uses_availability_probe`: it exists only in `fla_vendor.py` (definition plus one call site) and this test file. No library source touched.

## Worth flagging

On transformers >= 5.15, `UNSLOTH_DISABLE_HOPPER_FLA_BWD=1` no longer yields the pure-PyTorch path, only the fixed-kernel path. Users are not left unprotected - the resolved fla still carries the BK-tile fix, so gradients are correct - and the library says so loudly at runtime. But it is a real narrowing of what that flag promises, and it is upstream's design, not something this PR introduces. Restoring the literal promise would mean making `import fla` fail before any linear-attn modeling module is imported, which breaks fla process-wide for every other consumer and does nothing once the module is already imported. That is a design decision rather than a CI fix, so it is left alone here.

## Verification

CPU-only, no GPU.

| | before | after |
|---|---|---|
| transformers 5.15.0 / trl 1.9.2 / torch 2.10.0+cpu | 2 failed, 16 passed, 17 skipped | **19 passed, 17 skipped** |
| transformers 4.57.6 / trl 0.22.2 | 18 passed, 17 skipped | **19 passed, 17 skipped** |

The two pre-fix failures match CI exactly; 18 to 19 is the added parametrization. Re-confirmed independently: the pre-change file still reproduces both failures on 5.15 in the same venv. Full zoo suite on the 5.15 venv shows no failure in `test_fla_hopper_tile.py` or `test_vendor_fla.py`, so no ordering regression.

### DIFF
--- a/tests/test_fla_hopper_tile.py
+++ b/tests/test_fla_hopper_tile.py
@@ -645,7 +645,19 @@ def test_in_place_patch_is_thread_safe(monkeypatch):
 def test_opt_out_forces_pure_torch(monkeypatch, fake_gated_delta_modeling):
     """UNSLOTH_DISABLE_HOPPER_FLA_BWD=1 must make transformers take its pure-torch
     gated-delta path, which needs both a False availability probe and the module
-    global unbound (the layer reads ``chunk_gated_delta_rule or torch_...``)."""
+    global unbound (the layer reads ``chunk_gated_delta_rule or torch_...``).
+
+    Pinned to the pre-#47630 layout, because that "or torch_..." read is what the
+    pure-torch outcome is made of: after #47630 the modeling files resolve the
+    kernel through ``use_kernel_func_from_hub_with_fallback``, which freezes the
+    implementation into a closure at import time, so there is no probe to answer
+    False and no global to unbind and pure torch is simply not reachable from
+    here. Without the pin this test silently swaps which branch of
+    ``patch_vendor_fla`` it covers as soon as the installed Transformers crosses
+    that release, and fails. The post-#47630 contract -- fall through to the
+    protection path instead of returning -- is
+    ``test_optout_falls_through_to_protection_on_the_new_layout``.
+    """
     import transformers.utils.import_utils as iu
 
     from unsloth_zoo.temporary_patches import fla_vendor as fv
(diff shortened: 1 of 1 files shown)


danielhanchen:
@codex review

PR #1043, opened 2026-08-11
Fix Qwen3.5 GGUF export when MTP config is nested under text_config

Souravrajvi0:
## Summary

Fixes unslothai/unsloth#8443

Qwen3.5 GGUF export fails with `assert self.opt_num_mtp_layers != 0` when `num_hidden_layers` and `mtp_num_hidden_layers` live only under `text_config`.

llama.cpp's Qwen MTP converter merges `text_config` in `index_tensors`, but its `__init__` reads those hyperparameters from the top level only. Unsloth already reconciles MTP keys against the weight index, but did not hoist the nested values the converter needs.

## Changes

- `unsloth_zoo/llama_cpp.py`: hoist `num_hidden_layers` and `mtp_num_hidden_layers` from `text_config` to the top level before writing `config.json` for conversion
- `tests/test_convert_to_gguf_mtp_reconcile.py`: cover nested-only configs and helper edge cases

## Test plan

- [x] `pytest tests/test_convert_to_gguf_mtp_reconcile.py`

### DIFF
--- a/tests/test_convert_to_gguf_mtp_reconcile.py
+++ b/tests/test_convert_to_gguf_mtp_reconcile.py
@@ -223,3 +223,61 @@ def test_convert_to_gguf_rejects_malformed_layer_count_before_rewrite(llama_cpp,
         )
 
     assert json.loads(config_path.read_text(encoding = "utf-8")) == config
+
+
+@pytest.mark.parametrize("has_mtp", (False, True))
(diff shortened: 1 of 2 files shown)


Souravrajvi0:
Closing — no maintainer review after waiting. Happy to reopen if the team wants to pick this up.

PR #1045, opened 2026-08-11
Spell relative imports absolutely where transformers cannot resolve them

danielhanchen:
`Paddle_OCR_(1B)_Vision` dies at the trainer's first checkpoint save:

```
FileNotFoundError: [Errno 2] No such file or directory:
'/usr/local/lib/python3.12/dist-packages/unsloth_zoo/mlx.runtime.py'
```

Nothing is missing from the wheel. That path is built by transformers, in `dynamic_module_utils.get_relative_imports`, which pastes whatever its regex captures after the directory and `.py`. It does not translate dots into separators, does not strip further leading dots, and does not consider that the target might be a package. Three shapes break:

| written | transformers opens |
|---|---|
| `from .mlx.runtime import x` | `unsloth_zoo/mlx.runtime.py` |
| `from .temporary_patches import x` | `unsloth_zoo/temporary_patches.py` (a directory) |
| `from .. import __version__` | `unsloth_zoo/mlx/..py` |

It reaches unsloth_zoo because a remote-code model takes the `custom_object_save` path on save, which walks relative imports recursively. One unresolvable import anywhere in that graph is a crash at checkpoint time, on a code path with nothing to do with the import.

36 sites across 18 files are now absolute. `from . import x` is untouched: the regex needs at least one non-space character after the dot, so it never matches it, and `from .sibling import x` resolves to a file that exists.

### Tests

`tests/test_relative_imports_resolve.py` asserts resolvability rather than dottedness, and runs `get_relative_import_files` itself against the package entry points. That end-to-end check is what found the 24 sites a dotted-only pass had missed.

Suite: 4726 passed. The 8 Gemma3n failures and 3 order-dependent rmsnorm recompile-budget failures are present on main and unaffected by this change.

### DIFF
--- a/tests/test_relative_imports_resolve.py
+++ b/tests/test_relative_imports_resolve.py
@@ -3,45 +3,36 @@
 
 """A relative import transformers cannot resolve breaks every remote-code save.
 
-`Paddle_OCR_(1B)_Vision` died at the trainer's first checkpoint save:
+`Paddle_OCR_(1B)_Vision` died at its first checkpoint save with
+`FileNotFoundError: .../unsloth_zoo/mlx.runtime.py`. Nothing was missing from
+the wheel, and `from .mlx.runtime import x` is valid Python: the dotted path is
+built by TRANSFORMERS, in `dynamic_module_utils`, which finds imports with a
+regex and joins the raw capture onto the directory:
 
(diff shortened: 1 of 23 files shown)


danielhanchen:
@codex review

PR #1053, opened 2026-08-12
Avoid persistent CUDA context creation during Zoo imports

oobabooga:
## Context

Unsloth Studio uses `unsloth_zoo.hf_xet_fallback` for Xet download health, tuning, watchdog, and HTTP fallback behavior. Importing that submodule first executes Zoo's package initializer, which imports `temporary_patches.gpt_oss`.

`gpt_oss.py` only needs total device capacity to configure combo kernels, but used `mem_get_info()`. That call creates a CUDA context which remains allocated for the lifetime of Studio's backend, even though Xet itself does not use the GPU.

## Change

Read total capacity from device properties instead. This returns the same value and preserves combo-kernel selection without creating the persistent context.

## Live measurements

Measured with Studio's Python environment, PyTorch 2.10.0+cu130, Zoo 2026.8.10, and an NVIDIA RTX 6000 Ada:

| Probe | `nvidia-smi` usage | Torch allocated / reserved |
| --- | ---: | ---: |
| Import Torch | No process | 0 / 0 |
| `torch.cuda.get_device_properties(0)` | No process | 0 / 0 |
| `torch.cuda.mem_get_info(0)` | 426 MiB | 0 / 0 |
| Import `unsloth_zoo.hf_xet_fallback` on `main` | 426 MiB | 0 / 0 |
| Import `unsloth_zoo.hf_xet_fallback` with this change | No process | 0 / 0 |

Both capacity APIs returned exactly `50,865,307,648` bytes.

## Testing

- Added an import regression that forbids `mem_get_info()` and verifies combo-kernel selection is preserved.
- Temporary-patches suite: 39 passed.
- Ruff and compile checks passed.

### DIFF
--- a/tests/test_temporary_patches_imports.py
+++ b/tests/test_temporary_patches_imports.py
@@ -127,7 +127,7 @@ def test_temporary_patches_submodule_list_is_complete():
 
 def test_gpt_oss_imports_without_visible_gpus():
     """gpt_oss.py computes device_memory at import; with UNSLOTH_ALLOW_CPU=1
-    DEVICE_TYPE stays "cuda" on GPU-less hosts, so mem_get_info must be
+    DEVICE_TYPE stays "cuda" on GPU-less hosts, so the capacity lookup must be
     guarded. Subprocess so conftest's mem_get_info stub cannot mask it."""
     import os
     import subprocess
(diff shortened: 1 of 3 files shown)


danielhanchen:
Confirmed independently on an NVIDIA B200, and the effect is larger on this card than on yours. Importing `unsloth_zoo.hf_xet_fallback` in a fresh process:

| | main | this PR |
| --- | ---: | ---: |
| nvidia-smi, this PID | 612 MiB | 0 MiB |

The mechanism is exactly as you describe: `mem_get_info()` needs a current context, `get_device_properties()` does not, and both return the same total. Worth adding that the cost is not fixed per machine, it is a bare CUDA context, so it scales with the card. I measured 712 MiB for `torch.zeros(1, device="cuda")` on this B200 against your 426 MiB, and none of it comes back from `empty_cache()` while the process lives.

I pushed `2f5af899` closing the same probe in `flex_attention/utils.py:54`:

```python
vram_of_gpu = min(torch.cuda.memory.mem_get_info(i)[-1]/1024/1024/1024 for i in range(torch.cuda.device_count()))
```

Same shape as the gpt_oss one: module scope, and `[-1]` is the total, so device properties answer it exactly. It matters here because `gpt_oss` imports `flex_attention`, so anyone reaching that path puts back the context this PR removes. Measured importing `unsloth_zoo.flex_attention.utils` alone: 612 MiB before, 0 MiB after, with `vram_of_gpu` identical at `178.3533935546875` from both APIs. Added a regression test in the same shape as yours; the temporary-patches import suite is 40 passed.

To check there was nothing else, I walked the AST of every module in the package for import-time calls that need a context. Those two files were the only ones, and both are now closed:

| file | call | still there |
| --- | --- | --- |
| `temporary_patches/gpt_oss.py:1394,1396` | `xpu/cuda.memory.mem_get_info()` | fixed by this PR |
| `flex_attention/utils.py:54` | `cuda.memory.mem_get_info()` | fixed by `2f5af899` |

Everything else at module scope is `is_available`, `device_count`, `get_device_properties`, `get_device_capability` or `get_device_name`, none of which create one.

For context on why this is worth caring about beyond the number: on the Studio side the same class of problem means one document ingested through RAG left an idle backend holding 866 MiB, because the embedder loads in-process and the context outlives every unload. That is unslothai/unsloth#8654.

danielhanchen:
@codex review

danielhanchen:
@codex review

PR #1054, opened 2026-08-12
Pad the GRPO log-softmax chunks so torch 2.12 can still compile them

danielhanchen:
`chunked_hidden_states_selective_log_softmax` and `chunked_selective_log_softmax` both split their rows with `torch.chunk(x, chunks = N)`, which leaves a ragged last chunk of `n - (N - 1)*ceil(n/N)`. Inside one dynamic `fullgraph` region that tail is a symbolic expression in `n`, and Inductor has to prove a vocab-wide node over it splits evenly.

torch 2.12 cannot, so a GRPO step dies at compile time:

```
InductorError: CantSplit: 202048*s47*s87 - 3434816*(((s47*s87 + 17)//18))
not divisible by s47*s87 - 17*(((s47*s87 + 17)//18))
```

That is the whole step, not a slow path. It is also not a real indivisibility: `202048*tail` over `tail` is `202048` by inspection.

## Measured, not inferred

Feeding that exact expression to `SizeVarAllocator.statically_known_multiple_of`:

| chunk size | 2.9.1 | 2.10.0 | 2.11.0 | 2.12.1 |
|---|---|---|---|---|
| ragged tail `n - 17*ceil(n/18)` | True | True | True | **False** |
| full chunk `ceil(n/18)` | True | True | True | True |

Confirmed on a real Tesla T4 running 2.12.1, which is where it was reported, and the traceback's `simd.py:847` matches 2.12's source.

So equal chunks avoid the question rather than working around the answer. Padding the row count up to a multiple of `chunks` removes the ragged tail; the padding repeats the last row and is sliced back off.

## Both call sites

The second function is the raw-logits sibling, reached whenever the model returns logits instead of hidden states. It chunks the same way over the same vocab width inside the same kind of region, so it carries the same exposure. Found by searching every `torch.chunk` in the package rather than stopping at the traceback. The remaining call sites, the GRPO gradient chunker and the fused cross-entropy loop, chunk outside the compiled region or over non-vocab widths and are not exposed.

## Verification

- Results are unchanged: the tests assert `torch.equal`, not closeness, against the pre-fix body across ragged splits, exact multiples, single rows, one chunk, and row counts below the chunk count where `torch.chunk` returns fewer chunks than asked.
- One test asserts the reference really is ragged, so the suite cannot pass vacuously.
- Both functions compile under `fullgraph = True, dynamic = True` on 2.12.1 with no graph breaks across four different pad values. This matters because `pad_rows` is a SymInt: the code deliberately avoids a Python `if` on it, which would either specialise the graph on `n % chunks` or graph-break under `fullgraph`.
- 111 tests pass across the chunked log-softmax, device, reduction-dim, compile-disable and raw-logits-dispatch suites.

## Scope note

The fix stays inside each function on purpose. `unsloth/models/rl.py` inlines both into generated trainers with `inspect.getsource`, dumping only the one function, so a new module-level helper would `NameError` at import there. That ruled out the alternative of hoisting the chunk loop out of the compiled region.

Cost is at most `chunks - 1` extra rows through a matmul that already runs over thousands.

A live Muse Glimmer GRPO run on Kaggle 2xT4 is in flight against this branch; I will post the result here.

### DIFF
--- a/tests/test_grpo_chunk_padding.py
+++ b/tests/test_grpo_chunk_padding.py
@@ -1,0 +1,144 @@
+# SPDX-License-Identifier: AGPL-3.0-only
+# Copyright 2026-present the Unsloth AI Inc. team. All rights reserved.
+
+"""Equal-chunk padding in ``chunked_hidden_states_selective_log_softmax``.
+
+The GRPO log-softmax chunks its rows with ``torch.chunk(x, chunks = N)``, which
+leaves a ragged last chunk of ``n - (N - 1)*ceil(n/N)`` rows. Inside one dynamic
+graph that tail is a symbolic expression in ``n``, and torch 2.12's Inductor
+cannot prove a vocab-wide node over it splits evenly, so the GRPO step dies with
+
+    InductorError: CantSplit: 202048*s47*s87 - 3434816*(((s47*s87 + 17)//18))
+    not divisible by s47*s87 - 17*(((s47*s87 + 17)//18))
+
+Padding the row count up to a multiple of ``chunks`` removes the ragged tail.
+These tests pin the two things that has to be true: the padding never changes a
+returned value, and the chunk sizes really are all equal.
+
+The function is lifted out with ``ast`` rather than imported, so the tests stay
+CPU-only and do not need the rest of unsloth_zoo. Decorators are stripped: the
+change is shape bookkeeping, and eager exercises it exactly.
+"""
+
+from __future__ import annotations
+
+import ast
+import os
+from pathlib import Path
+
+import pytest
+import torch
+
+
+SOURCE_PATH = Path(__file__).resolve().parents[1] / "unsloth_zoo" / "rl_replacements.py"
+FUNCTION_NAME = "chunked_hidden_states_selective_log_softmax"
+
+
+def _load_chunked_log_softmax():
+    tree = ast.parse(SOURCE_PATH.read_text(encoding = "utf-8"), filename = str(SOURCE_PATH))
+    for node in tree.body:
+        if isinstance(node, ast.FunctionDef) and node.name == FUNCTION_NAME:
+            node.decorator_list = []
+            namespace: dict = {"torch": torch, "os": os}
+            exec(
+                compile(ast.Module(body = [node], type_ignores = []), str(SOURCE_PATH), "exec"),
+                namespace,
+            )
+            return namespace[FUNCTION_NAME]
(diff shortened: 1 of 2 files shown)


danielhanchen:
Live result from Kaggle 2xT4, and it does not fix Muse Glimmer. Reporting it here rather than leaving the claim open.

The run installed `unsloth_zoo` at `f528321a909f` off this branch, which does contain the padding, and the GRPO step still died at the same cell with the same expression:

```
InductorError: CantSplit: 202048*s47*s87 - 3434816*(((s47*s87 + 17)//18))
not divisible by s47*s87 - 17*(((s47*s87 + 17)//18))
```

Byte for byte the same as before the change, which is the clue. The padding does make the compute uniform: all 18 chunks now run on `q = ceil(n/18)` rows. But the trailing `[:n_rows]` trim cuts the concatenated result back to `n`, and the slice of that trim belonging to the last chunk has length exactly `n - 17q`. Inductor fuses the trim back into that chunk's producer, the `logsumexp` over the vocab-wide logits, so the node it has to split is `202048 * (n - 17q)` and we are back where we started. The ragged expression moved from the input side to the output side and fusion carried it straight back in.

So equal chunks are not sufficient on their own. Any formulation that returns exactly `n` rows from an 18-way split reintroduces `n - 17*ceil(n/18)` somewhere the scheduler can fuse it into the vocab-wide reduction, which makes this hard to solve by reshaping alone inside a single `fullgraph` region on torch 2.12.

Holding this PR until the approach is settled rather than merging a change that costs padding work and fixes nothing. The measurements in the description still stand and are worth keeping: the predicate flips between 2.11 and 2.12 for that exact expression, and the whole thing is a torch 2.12 regression against arithmetic that is true by inspection.

danielhanchen:
Closing this. The padding approach does not fix the problem, and I would rather say so than leave a draft implying it might.

Making the chunks uniform does make the compute uniform at `q = ceil(n/18)`, but the trailing `[:n_rows]` trim still has to remove exactly `n - 17q` rows, and Inductor fuses that trim back into the vocabulary-wide `logsumexp`, regenerating the identical expression. A run on Kaggle 2xT4 with this branch failed at the same cell with the same `CantSplit`.

The general form of the problem: any formulation that returns exactly `n` rows from an N-way split reintroduces `n - (N-1)*ceil(n/N)` somewhere the backend can fuse into the reduction.

Superseded by #1055, which treats the codegen refusal as a fallback condition rather than trying to reshape the arithmetic around it, and which is verified passing on the hardware where the failure actually lives.

PR #1057, opened 2026-08-13
Stop a broken sibling package being reported as upstream drift

danielhanchen:
Eight tests on main announce upstream drift for something upstream did not do:

    DRIFT DETECTED: gemma3n.py:53 helper expects Gemma3nRMSNorm but it is
    missing on transformers 4.57.6

transformers 4.57.6 ships `Gemma3nRMSNorm` perfectly well. What actually happens is that `configuration_gemma3n` does

    from timm.data import ImageNetInfo, infer_imagenet_subset

and a newer `timm` no longer exports `ImageNetInfo`, so the whole gemma3n module raises `ImportError` and every class on it looks deleted. The message names transformers, which is the one component not at fault, and it names a version, which invites someone to go bisect transformers over a timm problem.

## Three places conflated two different findings

| where | before | after |
|---|---|---|
| `_try_get_class`, `test_temporary_patches_exhaustive.py` | swallowed every exception, returned `None`; 15 of its 23 callers turn `None` into a DRIFT failure | `None` only for a `ModuleNotFoundError` naming the module asked for; skip with the real exception otherwise |
| four gemma3n tests, `test_upstream_signatures.py` | inline `from ... import X`, so an ImportError surfaced as signature drift | routed through a new `_import_or_skip` with the same rule |
| `_load_modeling` and `_compile_and_get_cache`, `test_compiler_dynamic_exec.py` | caught only `ModuleNotFoundError`, so a raising module failed the parseable-cache test | skip on `ImportError` too, saying which |

**The drift signal is deliberately preserved.** A module that has genuinely gone away upstream is still reported, because that really is drift. Only a module that exists and raises on the way in is skipped, since it says nothing about upstream at all.

## Proving the distinction rather than assuming it

The new test builds both cases by hand rather than relying on the environment to supply one, so it stays meaningful after timm is fixed: an absent module still comes back as `None` so callers can fail on it, and a module that raises on a dependency is skipped.

One trap worth recording, since it silently produced a green result at first: pytest's `Skipped` derives from `BaseException`, so `pytest.raises(Exception)` lets it through and skips the asserting test itself. The test caught 25 skips instead of a pass until it asserted on `_pytest.outcomes.Skipped` directly.

## Measured

    test_temporary_patches_exhaustive.py
    + test_upstream_signatures.py
    + test_compiler_dynamic_exec.py      236 passed, 31 skipped

Before this change, 8 of those failed. No production code changes; this is test diagnostics only.

Sits on top of #1056, which fixes a different way this suite failed to report honestly (a collection error that ran zero tests while looking like a pass).

### DIFF
--- a/.github/workflows/consolidated-tests-ci.yml
+++ b/.github/workflows/consolidated-tests-ci.yml
@@ -372,6 +372,7 @@ jobs:
             tests/test_compiler_rewriter_exhaustive.py \
             tests/test_compiler_dynamic_exec.py \
             tests/test_temporary_patches_exhaustive.py \
+            tests/test_missing_parent_package_is_drift.py \
             tests/test_unsloth_zoo_lora_merge.py \
             tests/test_peft_paramwrapper_layout_drift.py \
             tests/test_transformers_moe_structure_drift.py \
--- a/tests/test_compiler_dynamic_exec.py
+++ b/tests/test_compiler_dynamic_exec.py
@@ -115,11 +115,9 @@ def _load_modeling(model_type: str):
             f"transformers, can't drive compiler"
         )
     except ImportError as exc:
-        # Present, but raising on the way in. gemma3n's config does
-        # `from timm.data import ImageNetInfo`, which a newer timm no longer
-        # exports, so the module blows up for a reason that has nothing to do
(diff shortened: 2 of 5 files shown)


danielhanchen:
@codex review

PR #1061, opened 2026-08-13
Reach a sibling test by path, not through the `tests.` package

danielhanchen:
Every open unsloth pull request is currently red on a zoo collection error.

`tests/` in this repo carries no `__init__.py`, so `tests.x` does not name anything here. It names whatever `tests` package happens to be importable, and under unsloth's Core CI there is one: the zoo checkout lands in `$RUNNER_TEMP/unsloth-zoo` while unsloth's own repo root is on `sys.path`, and unsloth **does** ship `tests/__init__.py`. So the two imports added in #1057 resolved against that project's package:

```
ModuleNotFoundError: No module named 'tests.test_temporary_patches_exhaustive'
ERROR ../../_temp/unsloth-zoo/tests/test_missing_parent_package_is_drift.py
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
5129 tests collected, 1 error in 11.87s
##[error]Process completed with exit code 2.
```

One module failing to import ends the whole job, so 5129 collected tests run zero of them and the pull request reads as broken by whatever it happened to change.

### The fix

The imports go by module name with this file's own directory on `sys.path`. Same file either way, and it no longer depends on which `tests` package wins the name.

### Why the grep

Running these files from this repo cannot catch it. `tests/test_missing_parent_package_is_drift.py` collects and passes here today, on the unfixed code, because pytest puts the rootdir first and `tests.x` resolves locally. It only misbehaves under a `sys.path` this suite does not control.

Reproduced by putting an empty `tests/__init__.py` package ahead of this repo on `PYTHONPATH`, which is the shape unsloth's CI produces:

- before: `ModuleNotFoundError: No module named 'tests.test_temporary_patches_exhaustive'`, no tests collected, 1 error
- after: 19 passed

The new `test_no_test_file_imports_a_sibling_through_the_tests_package` is red at `3f72f667` naming both offending lines, and green here. It also tests its own predicate, so a grep that silently stops matching does not pass by accident. This is the only file in the suite that had the pattern.

### DIFF
--- a/.github/workflows/consolidated-tests-ci.yml
+++ b/.github/workflows/consolidated-tests-ci.yml
@@ -134,6 +134,16 @@ jobs:
         # raising requires-python relaxes it rather than making it lie.
         # test_wheel_top_level_packages pins that discovery ships unsloth_zoo and
         # nothing else; collect-only would let the packaging config regress.
+        #
+        # The last two are here because `--collect-only` is exactly what missed
+        # the bug they cover. `tests/` is not a package, so `from tests.x import
+        # y` binds to whatever `tests` package is importable; unsloth's CI runs
+        # these files with its own repo root on sys.path and it DOES ship
+        # `tests/__init__.py`, so the import resolved there and took the module
+        # out at collection, ending that job at exit 2 on every unsloth pull
+        # request. Collecting the drift file here proves it still imports under
+        # a plain root, and the sweep is a stdlib-only grep that refuses the
+        # pattern anywhere in the suite.
         run: |
           python -m pytest \
             tests/test_training_utils_use_cache.py \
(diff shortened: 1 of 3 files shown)


danielhanchen:
End-to-end confirmation, running unsloth's failing CI step verbatim from an unsloth checkout with zoo at each revision. The step is `python -m pytest --collect-only -q "$RUNNER_TEMP/unsloth-zoo/tests/"` in `consolidated-tests-ci.yml`.

zoo at `main` (`3f72f667`):

```
ERROR ../temp/zooci_main/tests/test_missing_parent_package_is_drift.py
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
5144 tests collected, 1 error in 6.59s
EXIT=2
```

zoo at this branch:

```
5165 tests collected in 6.19s
EXIT=0
```

Same interpreter, same unsloth checkout, same command. The 21 extra collected tests are the two files this PR touches.

danielhanchen:
@codex review

danielhanchen:
@codex review

PR #1064, opened 2026-08-14
Finish the job #1063 started on a bitsandbytes that is not importable

danielhanchen:
#1063 widened three module-scope `bitsandbytes` guards from `ImportError` to `Exception`, because a wheel built against a different torch fails its own import with

```
bitsandbytes/backends/cuda/ops.py:69
_get_raw_stream = torch._C._cuda_getCurrentRawStream
AttributeError: module 'torch._C' has no attribute '_cuda_getCurrentRawStream'
```

Those three were not the whole chain. Two more module-scope sites reach `bitsandbytes` without a guard that catches this, and `import unsloth` still died on both.

### The two sites

**1. `unsloth_zoo/patching_utils.py:733`** imported `transformers.integrations.bitsandbytes` with no guard at all, and on transformers 4.x that module does a bare module-scope `import bitsandbytes as bnb` of its own.

```
File "unsloth_zoo/patching_utils.py", line 733, in <module>
  import transformers.integrations.bitsandbytes
File "transformers/integrations/bitsandbytes.py", line 20, in <module>
  import bitsandbytes as bnb
AttributeError: module 'torch._C' has no attribute '_cuda_getCurrentRawStream'
```

Guarding only the import turns the crash into a `NameError`. The `import transformers.integrations.bitsandbytes` statement is what binds `transformers` in this module, and the transformers 5.x block below reads the same module twice more. So all four reference sites move together onto one alias that is `None` when the dep is unusable.

**2. `unsloth_zoo/vllm_utils.py:420`** gated its whole `bitsandbytes` block on `importlib.util.find_spec("bitsandbytes") is not None`. A mismatched wheel is on disk, so `find_spec` says yes and the `import bitsandbytes.functional` behind the gate raises:

```
File "unsloth_zoo/vllm_utils.py", line 421, in <module>
  import bitsandbytes.functional
AttributeError: module 'torch._C' has no attribute '_cuda_getCurrentRawStream'
```

The gate now imports the package to decide, which is the only answer `find_spec` cannot give.

### Neither site invents a new fallback

`patching_utils` skips a dynamic-4bit patch that needs `inspect.getsource` of a function it cannot reach. `vllm_utils` takes the `else:` branch of its own gate, the one that already defines the no-op `patch_bitsandbytes_quant_state` / `patch_bitsandbytes_compute_dtype` / `unpatch_bitsandbytes_compute_dtype` stand-ins for a host with no `bitsandbytes`.

The transformers 5.x `should_convert_module` branch does **not** get skipped. It is gated on `quantizers_utils.should_convert_module`, which is the marker that actually separates 4.x from 5.x and needs no `bitsandbytes` to read, so a broken wheel still gets that patch.

### The healthy path is byte-for-byte unchanged

With the real wheel installed (`bitsandbytes` 0.50.0, transformers 4.57.6), before and after this branch:

| | parent 848fd15f | this branch |
|---|---|---|
| patched source sha256 | `759f6abb...` | `759f6abb...` |
| bytecode sha256 | `6c0a7043...` | `6c0a7043...` |
| `tib._replace_with_bnb_linear.__name__` | `_unsloth_replace_with_bnb_linear` | same |
| `mark_parent_error` | `False` | `False` |
| names exec'd into globals | `Conv1D, bnb, init_empty_weights, nn, replace_with_bnb_linear, signature` | same |
| `vllm_utils.Linear4bit` base | `bitsandbytes.nn.modules.Linear4bit` | same |

The 5.x `should_convert_module` branch, which this host's transformers 4.57 never reaches, was exercised by simulating 5.x (drop `_replace_with_bnb_linear`, add `quantizers_utils.should_convert_module`) and lands identically on both.

### Tests

`tests/test_broken_bitsandbytes_import.py` fakes the dep with a one-line stub module on `sys.path`, in a subprocess. No network, no real broken wheel, and nothing leaks into the rest of the suite.

- `test_patching_utils_imports_when_bitsandbytes_cannot` and `test_vllm_utils_takes_the_no_bnb_path_when_bitsandbytes_cannot_import` fail at 848fd15f and pass here.
- `test_the_stub_really_breaks_the_import` keeps the other cases from passing vacuously on a host that quietly resolved the real wheel.
- `test_a_healthy_bitsandbytes_still_gets_the_dynamic_4bit_patch` and `test_a_healthy_bitsandbytes_still_opens_the_vllm_utils_gate` pass on both, so a guard cannot buy the fix by turning 4bit off for everybody.

Related suites, all green:

```
tests/test_broken_bitsandbytes_import.py tests/test_temporary_patches_imports.py
tests/test_allow_cpu_import_driverless.py tests/test_compiler_rewriter_exhaustive.py
tests/test_extended_dep_api_pins.py tests/test_weak_dictionary_writes_not_compiled.py
tests/test_zoo_source_upstream_refs.py tests/test_upstream_signatures.py
tests/test_upstream_pinned_symbols_transformers.py tests/test_zoo_history_regressions.py
tests/test_zoo_history_regressions_deep.py tests/test_gguf_bitsandbytes_guard.py
tests/test_vllm_to_hf_conversion.py tests/test_fp8_dense_merge.py

521 passed, 17 skipped in 113.86s
```

### One thing this cannot fix from here

`peft/helpers.py` has the same `except ImportError` on `bitsandbytes` and sits right behind these two, reached from `unsloth_zoo/training_utils.py:22` (`from transformers import Trainer`) and `unsloth_zoo/vllm_utils.py:3100`. So a genuinely broken wheel still needs that fixed upstream before `import unsloth` completes. Simulating the upstream one-word fix locally, the import completes and reports the expected `bitsandbytes is not installed - 4bit QLoRA unallowed, but 16bit and full finetuning works!`.

### Sweep

I AST-scanned every module-scope import in `unsloth_zoo/` for the same class. Everything demonstrably broken is fixed above. Not fixed, because I could not make them fail here, listed so they are on the record:

- `vllm_utils.py:176` `import vllm.model_executor.layers.quantization.bitsandbytes` behind `find_spec("vllm")`, unguarded. Does not import `bitsandbytes` at module scope with the installed vLLM, so a broken wheel passes straight through it.
- `loss_utils.py:24` `from triton import __version__` and `compiler.py:51` `import triton`, both fully unguarded and both on the `import unsloth` path. A triton mismatched with torch fails the same way. Left alone: triton is not optional on the CUDA path, and the MLX and skip-GPU-init hosts already get `stubs/triton_stub.py`.
- `vllm_lora_worker_manager.py` (8 imports), `vllm_lora_request.py:10`, `vllm_utils.py:112,115`: `except ImportError` around `vllm.*`. Importing any `vllm.*` runs the whole vLLM init, which raises `RuntimeError` / `OSError` on a torch or CUDA-lib mismatch more often than `ImportError`.
- `mlx/cce/__init__.py:22` `import mlx.core` under `except ImportError`; an mlx wheel on the wrong arch fails with a dylib `OSError`. Note the ~13 other module-scope `mlx` imports in the package are unguarded, so the guard here is already inconsistent rather than load-bearing.
- `saving_utils.py:35,39` `except (ImportError, ModuleNotFoundError)` on `transformers.integrations.mxfp4`. Safe against the pinned transformers, which has no module-scope `triton_kernels` import there, but narrow if a newer one adds it.
- `patching_utils.py:821` `import transformers.quantizers.quantizers_utils`, unguarded. Verified that module has no module-scope `bitsandbytes` or `triton` import.

`patching_utils.py` is the only module-scope importer of `transformers.integrations.bitsandbytes` in the package, and `integrations/bitsandbytes.py` is the only installed transformers module with an unguarded module-scope `import bitsandbytes`.

### DIFF
--- a/tests/test_broken_bitsandbytes_import.py
+++ b/tests/test_broken_bitsandbytes_import.py
@@ -1,0 +1,215 @@
+# SPDX-License-Identifier: AGPL-3.0-only
+# Copyright 2023-present the Unsloth team. All rights reserved.
+"""A bitsandbytes that fails its own import with something other than ImportError
+must not take the whole unsloth_zoo import down with it.
+
+A bitsandbytes wheel built against a different torch raises out of its own module
+scope, and not as an ImportError:
+
+  bitsandbytes/backends/cuda/ops.py:69
+  _get_raw_stream = torch._C._cuda_getCurrentRawStream
+  AttributeError: module 'torch._C' has no attribute '_cuda_getCurrentRawStream'
+
+Two module-scope sites let that through. `patching_utils` reaches bitsandbytes second
+hand: its dynamic-4bit patch imports `transformers.integrations.bitsandbytes`, and that
+transformers module does a bare module-scope `import bitsandbytes as bnb` of its own.
+`vllm_utils` reaches it directly, behind an `importlib.util.find_spec` gate that only
+proves the wheel is on disk. Either one put the AttributeError on the `import unsloth`
+path, breaking 16bit and full finetuning on a host whose only real problem was a 4bit dep.
+
+Both have a no-bnb path already, and that is where a broken wheel now goes. The healthy
+path is pinned here too, so a guard cannot buy the fix by turning 4bit off for everybody.
+
+The dep is faked with a one-line module on `sys.path`, never a real wheel and never the
+network, and every case runs in a fresh interpreter so the poisoned `bitsandbytes` cannot
+leak into the rest of the suite.
+"""
+
+import os
+import pathlib
+import subprocess
+import sys
+
+import pytest
+
+
+_ROOT = pathlib.Path(__file__).resolve().parents[1]
+
+# The message the mismatched wheel actually dies with, reproduced verbatim so a
+# guard narrowed back to ImportError fails here rather than passing vacuously.
+_BREAKAGE = "module 'torch._C' has no attribute '_cuda_getCurrentRawStream'"
+
+_STUB = f'raise AttributeError("{_BREAKAGE}")\n'
+
+
+def _run(code, *, poisoned):
+    """Run `code` in a fresh interpreter, optionally with the broken dep first
+    on the path. The repo root goes on PYTHONPATH so the child imports the tree
+    under test rather than an installed unsloth_zoo."""
+    path = [str(_ROOT)]
+    if poisoned:
+        path.insert(0, str(poisoned))
+    if os.environ.get("PYTHONPATH"):
+        path.append(os.environ["PYTHONPATH"])
+    return subprocess.run(
+        [sys.executable, "-c", code],
+        capture_output = True,
+        text = True,
+        env = dict(
+            os.environ,
+            PYTHONPATH = os.pathsep.join(path),
+            UNSLOTH_ALLOW_CPU = "1",
+        ),
+        timeout = 900,
+    )
+
+
+@pytest.fixture
+def poisoned_bitsandbytes(tmp_path):
+    (tmp_path / "bitsandbytes.py").write_text(_STUB)
+    return tmp_path
+
+
+def test_the_stub_really_breaks_the_import(poisoned_bitsandbytes):
+    """Without this, every other case in the file could pass on a host that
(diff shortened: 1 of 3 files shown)


danielhanchen:
@codex review

PR #1068, opened 2026-08-16
fix: stub torch.float8_e8m0fnu before transformers import

Souravrajvi0:
## Summary

Fixes unslothai/unsloth#8933.

On torch builds before 2.7, `import unsloth` fails during Studio training with:

```
module 'torch' has no attribute 'float8_e8m0fnu'
```

Pinned transformers releases bind `torch.float8_e8m0fnu` at import time in `integrations/finegrained_fp8.py`, even when the run will never touch UE8M0 FP8 checkpoints.

This PR stubs the missing dtype to `torch.float8_e4m3fn` before transformers is imported, so normal LoRA training can start on older torch. A clearer fallback error is added when no FP8 dtypes exist at all.

## Test plan

- [x] `pytest tests/test_torch_float8_e8m0fnu_stub.py` (6 passed)
- [x] **torch 2.6 integration** (`scripts/verify_e8m0_import.py`, 4/4 passed):
  - Studio `finegrained_fp8` bind fails without stub (reproduces #8933)
  - Same bind succeeds after stub
  - `unsloth_zoo.temporary_patches.utils` imports on real torch 2.6.0+cpu
  - UE8M0 path checked on torch 2.12 venv (no regression)
- [x] `import unsloth` reaches zoo init on torch 2.6 (blocked later only by CPU-only torch CUDA ops; Studio users have GPUs)

Fixes unslothai/unsloth#8933

### DIFF
--- a/scripts/verify_e8m0_import.py
+++ b/scripts/verify_e8m0_import.py
@@ -1,0 +1,207 @@
+#!/usr/bin/env python3
+"""Manual integration verification for unslothai/unsloth#8933."""
+
+from __future__ import annotations
+
+import os
+import sys
+import traceback
+
+ROOT = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
+STUDIO_TRANSFORMERS_ROOT = os.path.expanduser(
+    "~/.unsloth/studio/.venv_t5_510"
+)
+
+
+def _banner(title: str) -> None:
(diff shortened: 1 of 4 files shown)


Souravrajvi0:
Closing — no maintainer review after waiting. Happy to reopen if the team wants to pick this up.

PR #1072, opened 2026-08-18
fix: forward enable_thinking=False in vision collator for Qwen3.8

### LINKED ISSUE #1070: qwen3.8 disable system prompt
tokenizer of new qwen3 model (qwen/qwen3.8) adds system_promt `<|im_start|>system
Reasoning effort is set to xhigh. Please think carefully through the task, validate key assumptions, consider plausible alternatives, and prioritize correctness, consistency, and clarity in the final answer.` if `enable_thinking=False` flag is not specified. please add support forwarding args to https://github.com/unslothai/unsloth-zoo/blob/main/unsloth_zoo/vision_utils.py#L951 disable adding system_prompt.

Souravrajvi0:
## Summary

Fixes #1070

Qwen3.8 chat templates inject a default system prompt such as:

> Reasoning effort is set to xhigh. Please think carefully...

unless `enable_thinking=False` is passed to `apply_chat_template`. `UnslothVisionDataCollator` was calling the template without that kwarg, polluting training examples.

## Changes

- Auto-detect templates that mention `enable_thinking` and default to `enable_thinking=False` for training
- Add optional `chat_template_kwargs` collator argument for explicit overrides
- Forward per-example `chat_template_kwargs` from dataset rows (matching the text collator path)
- Route all collator `apply_chat_template` calls through one helper

## Test plan

- [x] `pytest tests/test_vision_collator_chat_template_kwargs.py` (6 passed)
- [x] Real `Qwen/Qwen3.8-27B` tokenizer: reasoning system prompt injected without flag, absent with `enable_thinking=False`

### DIFF
--- a/tests/test_vision_collator_chat_template_kwargs.py
+++ b/tests/test_vision_collator_chat_template_kwargs.py
@@ -6,21 +6,17 @@
 # the Free Software Foundation, either version 3 of the License, or
 # (at your option) any later version.
 
-"""Qwen3.8-style chat_template_kwargs forwarding in UnslothVisionDataCollator."""
+"""chat_template_kwargs forwarding in UnslothVisionDataCollator."""
 
 from __future__ import annotations
 
-from unittest.mock import MagicMock
-
-import pytest
-
 from unsloth_zoo.vision_utils import (
     UnslothVisionDataCollator,
-    _QWEN38_REASONING_SYSTEM_PROMPT,
-    _default_chat_template_kwargs,
     _merge_chat_template_kwargs,
-    _template_supports_enable_thinking,
 )
+
+
+_QWEN38_REASONING_SYSTEM_PROMPT = "Reasoning effort is set to"
 
 
 class _FakeTokenizer:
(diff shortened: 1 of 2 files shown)


danielhanchen:
Confirmed the vision collator at unsloth_zoo/vision_utils.py:951 calls apply_chat_template with no kwargs, so per-example chat_template_kwargs never reach it the way they do in the text path. The forwarding part is the right fix, but could you drop the auto-default that flips enable_thinking to False whenever the template string mentions it, and keep this to explicit collator and per-example kwargs?

Souravrajvi0:
Dropped the auto-default that set `enable_thinking=False` whenever the template mentioned it.

The vision collator still forwards:
- constructor `chat_template_kwargs`
- per-example `chat_template_kwargs` from dataset rows

Qwen3.8 reasoning prompts stay off only when those kwargs are passed explicitly, e.g. `chat_template_kwargs={"enable_thinking": False}`.

danielhanchen:
Thanks for dropping the auto-default, the collator now only honours explicit constructor and per-example chat_template_kwargs, which matches how the text path already merges them in unsloth_zoo/dataset_utils.py:2569. Confirmed main's vision collator forwards neither. Will get this reviewed.

danielhanchen:
Thanks for dropping the auto-default, the collator and per-example chat_template_kwargs forwarding is now the shape I wanted and default rendering is untouched. Will get this reviewed.

danielhanchen:
Thanks for dropping the auto-default, the explicit collator and per-example chat_template_kwargs forwarding is what I wanted and it lines up with #1070 where the vision path never passed enable_thinking through. Will get this reviewed.

PR #1080, opened 2026-08-20
Stop the Mac lane dying on an mlx teardown segfault

danielhanchen:
## What breaks

`tests/test_qwen35_vjp_metal.py` passes all 4 tests, then the process segfaults during interpreter shutdown:

```
4 passed, 1 warning in 6.61s
Segmentation fault: 11   python -m pytest tests/test_qwen35_vjp_metal.py
##[error]Process completed with exit code 139
```

Nothing is failing. The job dies on a green suite.

This has been invisible until now because it is the step immediately after `tests/test_mlx_switch_lora.py`, which has been red since 2026-08-18. The job stopped there and skipped everything below, this step included. Fixing that earlier failure in #1078 is what exposed this one.

## It is mlx 0.32.1

Measured on an Apple Silicon runner, same tree, flipped both directions inside one job:

| mlx | result |
| --- | --- |
| 0.32.1 | 4 passed, Segfault 11, exit 139 |
| 0.32.0 | 4 passed, **exit 0** |
| back to 0.32.1 | 4 passed, Segfault 11, exit 139 |

The last green scheduled run of this workflow was 2026-08-17 on 0.32.0; 0.32.1 shipped on the 18th. The same before-and-after shows up independently in the scheduled runs on the staging mirrors. This is the second thing 0.32.1 broke here, the other being saving over a loaded file, which #1078 fixes in our own code.

## It is not ours to fix from Python

The crash needs a fused Metal custom-kernel forward to be the last operation the process touched. Only `test_disable_fused_mrope_fixes_rotary_grad` arms it; the other three tests in the file are clean when run alone. Doing a `value_and_grad` after that forward is also clean, which is why the file only fails as a whole.

Using the arming shape directly, three samples each:

```
build model + fused rotary forward            139 139 139
  + del refs, gc.collect(), mx.synchronize()  139 139 139
  + gc.collect() registered at exit           139 139 139
  + os._exit() after the work                   0   0   0
```

Dropping every reference we hold, collecting, and flushing the stream still crashes, so there is nothing left for us to release. The fault is in mlx's own static destruction, after our cleanup has run. Clearing the cache at exit does not help either. Only skipping interpreter finalization avoids it.

## The change

That one file now runs through a small wrapper that calls `pytest.main` and then `os._exit` with pytest's own return code.

This hides nothing. The exit code is pytest's, verified locally:

| case | exit |
| --- | --- |
| passing file | 0 |
| failing test | 1 |
| collection error | 4 |

No test is skipped, relaxed or marked xfail, and mlx stays unpinned so the next release is still exercised on this lane. The wrapper is removed once mlx fixes the teardown crash.

## Verification

This alone does not turn the lane green, because on `main` the job still dies earlier at `test_mlx_switch_lora.py`. Both this and #1078 are needed. Validated on a real Apple Silicon runner with both applied, all 14 test steps green including the seven that have been skipped since 2026-08-18:

```
success  tests/test_mlx_torch_shim_smoke.py
success  tests/test_mlx_switch_lora.py
success  tests/test_mlx_save_over_source.py
success  tests/test_qwen35_vjp_metal.py
success  tests/test_mlx_training_e2e_metal.py
success  tests/test_mlx_generate.py
success  tests/test_mlx_generate_metal.py
success  tests/test_mlx_cpt_partition_metal.py
success  tests/test_mlx_pr684_full_validation_metal.py
success  tests/test_mlx_quantized_optimizer_state.py
success  tests/test_mlx_shape_guard.py
success  tests/test_mlx_vlm_label_masks.py
success  tests/test_mlx_batching_and_decay.py
```

Worth flagging that those seven steps had not run for three days, so this is also the first confirmation that nothing else regressed behind them.

This is worth reporting upstream to ml-explore/mlx with the reproducer above.

### DIFF
--- a/.github/workflows/mlx-ci.yml
+++ b/.github/workflows/mlx-ci.yml
@@ -90,7 +90,13 @@ jobs:
 
       - name: tests/test_qwen35_vjp_metal.py
         # Metal VJP regressions for the qwen3_5 training patches (issue 6002).
-        run: python -m pytest tests/test_qwen35_vjp_metal.py -v -rs
+        #
+        # Run through the wrapper, not `python -m pytest`: on mlx 0.32.1 a fused
+        # Metal kernel left live at interpreter shutdown segfaults, so this file
+        # exits 139 with every test passing. The wrapper returns pytest's own exit
+        # code, so a real failure still fails the job. See its docstring for the
+        # measurements. Revert to plain pytest once mlx fixes the teardown crash.
+        run: python tests/_run_then_exit_hard.py tests/test_qwen35_vjp_metal.py -v -rs
 
       - name: tests/test_mlx_training_e2e_metal.py
         # Real LoRA training smoke: tiny 4-bit model through FastMLXModel +
--- a/.github/workflows/mlx-ci.yml
+++ b/.github/workflows/mlx-ci.yml
@@ -90,12 +90,9 @@ jobs:
 
       - name: tests/test_qwen35_vjp_metal.py
         # Metal VJP regressions for the qwen3_5 training patches (issue 6002).
-        #
-        # Run through the wrapper, not `python -m pytest`: on mlx 0.32.1 a fused
-        # Metal kernel left live at interpreter shutdown segfaults, so this file
-        # exits 139 with every test passing. The wrapper returns pytest's own exit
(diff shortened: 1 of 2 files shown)


danielhanchen:
@codex review

danielhanchen:
@codex review

PR #1081, opened 2026-08-21
fix(mlx): rewind the PRNG by reseeding instead of assigning to mx.random.state

Lyxot:
## What breaks

mlx 0.32.1 replaced `mx.random.state` — previously a plain list — with a `mlx.core.random._RandomState` sentinel exposing `__len__`, `__getitem__` and `__iter__` but no `__setitem__` ([ml-explore/mlx#3828](https://github.com/ml-explore/mlx/pull/3828), on PyPI 2026-08-18). Reads still work; every assignment into it now raises `TypeError: 'mlx.core.random._RandomState' object does not support item assignment`.

Three places captured and restored the global PRNG key by assigning into `mx.random.state`, and they fail in two different ways:

- The DDP local-gradient compile fallback assigned unconditionally, so it raises *while recovering from a compile failure* — replacing the compile error an operator needs to see with an unrelated `TypeError`.
- The single-process compile fallback and `_preserved_preprocessing_rng` were guarded on `isinstance(state, list)`. That guard exists for the torch simulation shim, whose `mx.random.state` is a callable. On 0.32.1 the sentinel lands in the same branch, so both silently stop restoring anything: a run that falls back to eager now diverges from an eager run of the same seed, with nothing reported.

Measured on 0.32.1, the old mechanism captures nothing and restores nothing:

```text
OLD captured anything?  False
OLD restored correctly? False
NEW captured anything?  True
NEW restored correctly? True
```

## What changed

A capture/restore pair in `unsloth_zoo/mlx/utils.py`, used at all three sites:

- `_mlx_rng_key()` returns the current key as its two 32-bit words, or `None` when the state cannot be read.
- `_restore_mlx_rng_key(words)` rewinds via `mx.random.seed((words[0] << 32) | words[1])`.

`mx.random.key(seed)` builds `array([seed >> 32, seed & 0xFFFFFFFF], uint32)`, so key packing is exactly invertible and reseeding with a captured key's own words restores that key. This is one code path with no version branching, and `mx.random.seed` is already the seeding idiom in `unsloth_zoo/mlx/loader.py`.

The `isinstance(..., list)` guards are gone. Capture now decides once whether there is a key to rewind to, which keeps the simulation shim a no-op without also swallowing an upstream API change — that conflation is what let this ship silently.

## Why reseeding is equivalent, not approximate

Verified against both mlx versions:

- On 0.32.0 the seed round-trip and the previous `state[0] = key` assignment produce **bit-identical** subsequent draws.
- `mx.random.seed` accepts and exactly round-trips the **full unsigned 64-bit range**, including `0x8000000000000000` and `0xFFFFFFFFFFFFFFFF`. Keys with the high bit set are half of all real keys, so this was checked rather than assumed; a signed packing anywhere would have broken half of all runs.

Because the reconstruction has no reachable failure mode, `_restore_mlx_rng_key` is unguarded. A blanket `except` there would reintroduce exactly the defect this fixes: a failure indistinguishable from an intentional no-op.

## Risk and compatibility

- Behaviour on mlx < 0.32.1 is unchanged, measured bit-exact rather than argued.
- The torch simulation shim still no-ops, now because capture reports no key rather than because of a type check.
- This does not depend on the companion Studio change in `unsloth/unsloth` and can merge in either order.

## Validation

```bash
python -m pytest tests/test_mlx_training_e2e_metal.py tests/test_mlx_trainer_internals.py -q
```

229 passed with mlx 0.32.0 installed, and again with mlx 0.32.1.

Every added test was mutation-checked — reverting the production change fails exactly the test that covers it, and nothing else:

- restoring by item assignment fails `test_restore_works_where_the_state_refuses_item_assignment`, on either mlx version, because the stand-in state has no `__setitem__` just as the real one does not;
- deleting either compile-fallback restore fails `test_every_compile_fallback_rewinds_the_rng_before_retrying_eagerly`;
- deleting the single-process restore also fails `test_a_runtime_compile_failure_retries_eagerly_from_the_captured_key`, which drives `MLXTrainer._train_inner` on real MLX with `mx.compile` failing once at runtime after consuming the stream.

The DDP fallback needs a distributed group no unit test can raise, so it is covered by the structure test only; that limitation is stated in the test's own docstring rather than left implied.

### DIFF
--- a/.github/workflows/consolidated-tests-ci.yml
+++ b/.github/workflows/consolidated-tests-ci.yml
@@ -421,6 +421,20 @@ jobs:
             echo "real mlx is installed but the numeric tests skipped"; exit 1
           fi
 
+      - name: tests/test_mlx_rng_rewind.py (HARD GATE)
+        # PRNG capture/rewind across the compile fallbacks. Needs mlx core only,
+        # no Metal, so it belongs on this lane rather than the opt-in Apple
+        # Silicon one -- which is usually skipped, and a rewind guarded only
+        # there is a rewind nothing guards on a pull request.
+        run: .lanes/venv-numerics/bin/python -m pytest tests/test_mlx_rng_rewind.py -v -rs
+
+      - name: Fail if the RNG rewind tests skipped
+        run: |
+          .lanes/venv-numerics/bin/python -m pytest tests/test_mlx_rng_rewind.py -q -rs | tee rng.txt
+          if grep -qiE '[0-9]+ skipped' rng.txt; then
+            echo "real mlx is installed but the RNG rewind tests skipped"; exit 1
+          fi
+
       # One pytest process per file: tests/mlx_simulation/ replaces
       # sys.modules["mlx.core"] process-wide, disarming whatever runs after it.
       - name: tests/test_mlx_shape_guard.py
--- a/.github/workflows/consolidated-tests-ci.yml
+++ b/.github/workflows/consolidated-tests-ci.yml
@@ -422,10 +422,9 @@ jobs:
           fi
 
       - name: tests/test_mlx_rng_rewind.py (HARD GATE)
-        # PRNG capture/rewind across the compile fallbacks. Needs mlx core only,
-        # no Metal, so it belongs on this lane rather than the opt-in Apple
-        # Silicon one -- which is usually skipped, and a rewind guarded only
-        # there is a rewind nothing guards on a pull request.
+        # PRNG capture/rewind across the compile fallbacks. Needs mlx core, not
+        # Metal, so it gates here; the Apple Silicon lane is opt-in and usually
+        # skipped, which guards nothing on a pull request.
         run: .lanes/venv-numerics/bin/python -m pytest tests/test_mlx_rng_rewind.py -v -rs
 
       - name: Fail if the RNG rewind tests skipped
--- a/tests/test_mlx_rng_rewind.py
+++ b/tests/test_mlx_rng_rewind.py
@@ -101,9 +101,7 @@ def _draw_after_seed(seed):
     return _draw()
 
 
-# ---------------------------------------------------------------------------
-# Key arithmetic
-# ---------------------------------------------------------------------------
+# --- Key arithmetic ---
 
 @pytest.mark.parametrize("seed", [
     0, 1, 2**31 - 1, 2**31, 2**32 - 1, 2**32, 2**63 - 1, 2**63, 2**64 - 1,
@@ -147,8 +145,7 @@ def test_restore_works_where_the_state_refuses_item_assignment():
 
 
 def test_capture_reads_nothing_where_the_state_is_not_indexable():
-    # The torch simulation shim exposes a callable, which must stay a no-op
-    # rather than raise or reseed.
+    # The torch simulation shim's state is a callable: a no-op, not a raise.
     with _shadowing_random_state(lambda: {"counter": 0}):
         assert _mlx_rng_key() is None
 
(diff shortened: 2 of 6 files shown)


Lyxot:
@codex review

danielhanchen:
I went through this against the mlx versions the project actually supports rather than only the two the description covers, and pushed three changes to the branch. The core of the PR is right: `mx.random.key` is built upstream as `{seed >> 32, (uint32) seed}` (`python/src/random.cpp`), so reseeding with a captured key's own words restores it exactly, and I confirmed that over 50k seeds, every single-bit value and the whole high half of the unsigned 64-bit range, on mlx 0.21.1, 0.25.2, 0.29.4, 0.30.0 (CUDA backend), 0.30.6, 0.31.2, 0.32.0 and 0.32.1. Where the old item assignment still works, the reseed is bit-identical to it across ten distributions.

Three things needed fixing.

**1. The test helper deletes `mx.random.state` on mlx < 0.31.2.**

`_shadowing_random_state` unshadows with `delattr`. That is only correct where `state` is served by a module `__getattr__`. Upstream binds it as a real module attribute (`m.attr("state") = default_key().state()`) through 0.31.0, and only from 0.31.2 is it lazily thread-local. On the earlier range the `finally` deletes the PRNG state itself, every later `mx.random.state` read raises `AttributeError`, `_mlx_rng_key` swallows that into `None`, and unrelated tests quietly stop rewinding. Measured on mlx 0.29.4, the helper as submitted turns 30 passing tests into 16 failures. pyproject declares `mlx>=0.22.0`, so both eras are in support. The helper is now version-aware and there is a regression test for it.

**2. `_restore_mlx_rng_key` can raise, in the two places where a raise does the most damage.**

The argument for leaving it unguarded is sound, but it holds only if the words cannot put `mx.random.seed` out of range. Capture no longer type-checks the state, so an int32-typed or otherwise non-uint32 state comes back as negative words and `mx.random.seed` rejects anything outside `[0, 2**64)` with a `TypeError`. That lands either in `_preserved_preprocessing_rng`'s bare `finally`, replacing whatever the block raised, or in the DDP fallback, replacing the compile error being recovered from. Masking both words to 32 bits is a no-op for real keys and makes "no reachable failure mode" true rather than contingent. It also fixes `(0, 2**32)`, which previously restored the wrong key.

**3. A key that is not two words returns a bare `None`.**

That is the same silent divergence the `isinstance(state, list)` guard caused, moved from the type check to the length check. If mlx ever widens the key, every compile fallback stops restoring and nothing is reported. It now warns once and declines.

**On coverage.** All five behavioural tests went into `tests/test_mlx_training_e2e_metal.py` behind `@metal_only`. The Apple Silicon job reported SKIPPED on this PR, so none of them ever ran; the only new test that executed was the source-grep structure test, which cannot catch this regression. Nothing in them needs Metal, and I verified all five pass unchanged on a Linux CPU mlx build. They now live in `tests/test_mlx_rng_rewind.py` on the Linux CPU lane as a hard gate, with the same skip check the numeric suite has.

The structure test now asserts on the parse tree rather than counting substrings and slicing `source[announced - 400:]`. It pins that both restores sit on a recovery branch and precede the eager retry there, without pinning which construct that branch is, since the single-process fallback recovers inside `except` and the DDP one inside an `if` after the try. Mutation-checked: deleting either restore fails it, and so does moving one below its retry, which the substring version did not catch.

**Verification.** 823 assertions across seven isolated venvs (five mlx versions on the CPU backend, one on the CUDA backend, one against the installed package), all green. Full `tests/test_mlx_*.py` sweep on mlx 0.32.1 is identical to the merge base apart from the new file. On mlx 0.32.1 the pre-PR code still measurably fails the way you described: the guarded form captures nothing and restores nothing, the unguarded DDP form raises `TypeError` while handling the compile error. Reproduced, not taken on trust.

One thing outside this PR: `unsloth/studio/backend/core/inference/mlx_inference.py` has a byte-for-byte copy of these two helpers from unslothai/unsloth#9478, so points 2 and 3 apply there too and will need a follow-up.

Issue #849, opened 2026-07-02
MoE grouped_mm backend silently trains on transposed expert weights when `2*moe_intermediate == hidden_size` (bf16 and bnb-4bit)


pjordanandrsn:
`temporary_patches/moe_utils.py::preprocess_weight` converts transformers-v5 F.linear expert
layout (`gate_up (E, 2I, H)`, `down (E, H, I)`) to grouped_mm layout by shape inspection:

```python
    if proj_type == "gate_up":
        # Want (E, hidden_dim, 2*intermediate).
        if weight.shape[1] == hidden_dim:
            return weight
        else:
            return weight.transpose(-2, -1)
```

When `2*moe_intermediate_size == hidden_size` the check is ambiguous — `weight.shape[1]` equals
`hidden_dim` for the F.linear layout too — and the weight comes back **untransposed**, so the
gate_up matmul runs on transposed weights in every expert and every MoE layer (the analogous
`down` branch has the same ambiguity when `moe_intermediate_size == hidden_size`). Loss stays
finite and training appears to run normally, but outputs are numerically wrong. The
`_WEIGHT_PREPROCESSORS` registry that could disambiguate per-arch has no registrations
(2026.6.7 wheel and main as of 2026-07-02).

## Repro

unsloth 2026.6.9 / unsloth-zoo 2026.6.7 / transformers 5.5.0, torch 2.11.0+cu128 (outside
unsloth's `<2.11` pin — the layout logic is version-independent by code reading), any
grouped_mm-capable GPU:

```python
import unsloth, torch, torch.nn.functional as F
from transformers import Qwen3MoeConfig, Qwen3MoeForCausalLM

cfg = Qwen3MoeConfig(vocab_size=1024, hidden_size=64, intermediate_size=128,
                     moe_intermediate_size=32,   # 2*32 == hidden_size -> square gate_up
                     num_hidden_layers=1, num_attention_heads=4, num_key_value_heads=2,
                     head_dim=16, num_experts=8, num_experts_per_tok=2,
                     decoder_sparse_step=1, initializer_range=0.2)
torch.manual_seed(0)
model = Qwen3MoeForCausalLM(cfg).to("cuda", torch.bfloat16).eval()
experts = model.model.layers[0].mlp.experts

T, K = 64, 2
hs = torch.randn(T, cfg.hidden_size, device="cuda", dtype=torch.bfloat16)
idx = torch.randint(0, 8, (T, K), device="cuda")
[...]

PR #1082, opened 2026-08-21
Resolve the MoE weight-preprocessor registry through the package so the compiled-cache copy sees it

pjordanandrsn:
The piece @danielhanchen asked me to split out of #915.

`moe_utils.py` copies itself into `unsloth_compiled_cache`, and `get_forward_moe_backend()` hands back the forward from that copy, which is its own module object (`unsloth_cached_moe_utils`, plus the bare `moe_utils` the compiled modules import). Each copy had its own `_WEIGHT_PREPROCESSORS`, so registering through the package never reached the forward that actually runs. `preprocess_weight` over there just fell back to shape inference, which for a square weight with no readable sibling is the #849 guess. Nothing in-tree registers one yet btw, OLMoE in #915 would be the first.

Fix is small: the registry functions go through `_weight_preprocessor_registry()`, which hands back the package module's dict if it's loaded (a `sys.modules` lookup, no import) and the local one if not. Package path doesn't change.

Test loads a real cache copy via `_load_cached_moe_utils_module()` from a scratch `UNSLOTH_COMPILE_LOCATION`, checks the resolved backend really is the copy's forward, and registers both ways. 3 of 4 fail on main. Also put it and `tests/test_moe_preprocess_weight_transpose.py` in the hard gate, since it turns out nothing was running the #849 test. CPU-only, no weights.

Either order with #915 is fine. That one works on its own since #913, this just makes its registration stick on the cache path.

### DIFF
--- a/.github/workflows/consolidated-tests-ci.yml
+++ b/.github/workflows/consolidated-tests-ci.yml
@@ -215,8 +215,8 @@ jobs:
         #
         # The two MoE registry files pin preprocess_weight's layout resolution (#849)
         # and that the weight-preprocessor registry is one dict across the package
-        # module and its unsloth_compiled_cache copy (#915). CPU-pure torch, no
-        # weights; neither was executed by any job before.
+        # module and its unsloth_compiled_cache copies. CPU-pure torch, no weights;
+        # neither was executed by any job before.
         #
         # The seven base-model-source files are here for the same reason. They pin
         # that an unreachable Hub is never reported as a missing or unquantized
--- a/.github/workflows/consolidated-tests-ci.yml
+++ b/.github/workflows/consolidated-tests-ci.yml
@@ -213,6 +213,11 @@ jobs:
         # mapping (no GPU, no vLLM needed). test_sft_prepare_dataset_double_bos
         # pins add_special_tokens for text that already carries BOS.
         #
+        # The two MoE registry files pin preprocess_weight's layout resolution (#849)
+        # and that the weight-preprocessor registry is one dict across the package
+        # module and its unsloth_compiled_cache copy (#915). CPU-pure torch, no
+        # weights; neither was executed by any job before.
+        #
         # The seven base-model-source files are here for the same reason. They pin
         # that an unreachable Hub is never reported as a missing or unquantized
         # model, which is what stops save_pretrained_merged from writing nothing
@@ -269,6 +274,8 @@ jobs:
             tests/test_find_common_token_ids_no_match.py \
             tests/test_vllm_direct_lora_loading.py \
             tests/test_sft_prepare_dataset_double_bos.py \
+            tests/test_moe_preprocess_weight_transpose.py \
+            tests/test_moe_weight_preprocessor_registry_shared.py \
             tests/test_python39_compatibility.py \
             tests/test_checkpoint_modes.py \
             tests/test_wheel_top_level_packages.py \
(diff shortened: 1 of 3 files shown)


danielhanchen:
Went through this in depth and pushed two commits to the branch.

**`adc5947` Guard the registry resolve on `model_type`.** `preprocess_weight` is not weight-prep-time only, it runs once per projection per MoE forward from `_gate_up_provider` / `_down_provider`, again from the recompute backward, and twice per forward in `moe_utils_fp8`. Resolving the registry unconditionally cost about 100ns on every one of those calls (290 -> 392 ns/call over 200k iterations). Moving the resolve inside the existing `model_type` check drops that to 6ns, which is noise. Same behaviour, cost only on the registered path.

**`3104b96` Cover the bare `moe_utils` copy.** The tests pinned `unsloth_cached_moe_utils`, but that is not the copy the compiled model modules import. `compiler.py:191` puts `UNSLOTH_COMPILE_LOCATION` on `sys.path` and `compiler.py:1206` emits a bare `from moe_utils import ...`, so there is a **third** module object under that name, and it was unpinned. New test fails on main, passes here. The file now gates 5 tests, 4 of which fail on main.

Also condensed the comment blocks a bit.

### Before and after, measured

Real B200 through `torch._grouped_mm`, forward and backward, dumping outputs plus all three gradients for `model_type` in `{None, deepseek_v3, gemma4_moe, olmoe}` crossed with square and non-square `gate_up`. Worth noting `deepseek_v3` and `gemma4_moe` really do set `_unsloth_model_type`, so the truthy path is a shipped path, not a hypothetical one.

| Comparison | Result |
|---|---|
| main vs this branch, nothing registered | bit-identical, all 8 cases |
| main, registered vs not registered | bit-identical, so the registration is silently dropped. This is the bug |
| this branch, registered vs not registered | differs, so the registration reaches the forward that runs. This is the fix |
| this branch, a correct registration vs inference | bit-identical, so registering does not perturb numerics |

So it is a real bug, not a theoretical one, and since nothing in-tree registers a preprocessor yet the change is a no-op for every shipped model today.

### Compatibility

- **Python 3.9 through 3.14, plus both free-threaded builds.** 10 invariants each, 8/8 interpreters green. `gil_enabled=False` included, since the registry is now shared mutable state across module copies.
- **Two full stacks.** py3.13 / torch 2.9.1 / transformers 4.57 and, in an isolated venv, py3.12 / torch 2.11 / transformers 5.15. The CI hard gate passes on both (1061 and 1060 passed).
- **Platform.** Every changed function is walked at AST level to confirm none of them touch `os`, `shutil`, `pathlib`, `sys.platform` or the path modules, so there is no OS-dependent surface. Behaviour is also identical under `sys.platform` faked to win32, darwin, linux and cygwin, and the cache copy loads from paths with spaces, non-ASCII, long components and trailing separators.
- **Old installs.** A stale pre-PR `unsloth_compiled_cache/moe_utils.py`, for instance where the compile location is read-only and `install_to_cache` could not overwrite, still loads and runs with pre-PR behaviour rather than an `AttributeError`. The reverse skew, a new cache copy against an older package module, binds correctly. No cache dir at all falls back to the package's own forward.
- **Robustness.** Package module absent from `sys.modules`, package present but partially initialised, registry attribute of the wrong type, an unrelated top-level `moe_utils` shadowing the name, `importlib.reload`, 200 concurrent registrations across 16 threads, and fork / spawn / forkserver children. All handled.

440 simulation cases plus 288 bitwise parity cases, all green, no defects found beyond the two things I pushed.

One thing I did not change but is worth knowing: only `_WEIGHT_PREPROCESSORS` is unified. Other module state stays per-copy, notably `_WARNED_AMBIGUOUS_LAYOUTS`, so the #849 warning can still print once per loaded copy. Fine for this PR's scope.

CI has still never run on this one, `statusCheckRollup` is empty. I will get the workflows approved.

danielhanchen:
@codex review

danielhanchen:
@codex review

PR #1090, opened 2026-08-23
feat(mlx): evaluate ORPO and DPO runs, not just sample from them

Lyxot:
Stacked on #1011 — please review that one first. Both target `main`, so until #1011 merges GitHub shows its commit here too; the work to review in this PR is the **last two commits**, and the diff to read is `feat/mlx-preference-generate..HEAD`.

ORPO and DPO runs on MLX reported no eval loss and no reward metrics. An eval dataset fed the sampling pass #1011 adds, and nothing else: there was no number to select a best model on, no early-stopping signal, and nothing to compare against the same run on CUDA. This adds the evaluation itself, reporting the metric set TRL logs for these objectives so a run reads the same on either backend.

```
eval_loss  eval_rewards/chosen  eval_rewards/rejected  eval_rewards/accuracies
eval_rewards/margins  eval_logps/chosen  eval_logps/rejected
eval_logits/chosen  eval_logits/rejected
```

ORPO additionally reports `eval_nll_loss`, `eval_log_odds_ratio`, and `eval_log_odds_chosen`, the three terms its loss decomposes into.

## What changed

**A separate eval scorer, not the training loss.** `make_preference_eval_fn` scores one batch and returns `(loss, pairs, stats)`. The training loss is compiled and normalizes across a gradient-accumulation window that evaluation has no equivalent of, so reusing it would report a number that depends on the accumulation setting rather than on the held-out set.

**Every metric is divided by the total it is a mean over.** The scorer reports sums rather than per-batch means, so the eval set's value is one division at the end. Rewards, logps and ORPO's log-odds terms divide by the pair count; the two logit metrics and ORPO's `nll_loss` are means over tokens, so each reports a token count alongside its sum. That distinction only shows on a ragged final batch, where batches hold different numbers of tokens — TRL averages its per-batch means unweighted there, and weighting them by pairs is no closer, because neither can recover a mean over tokens once the counts differ.

**`eval_loss` is the pair-weighted mean of each batch's own objective value.** The scorer returns the batch's pair count as its weight rather than 1. TRL takes a mean of per-batch means; pair weighting agrees with it on every full batch and is more exact on a ragged final one, where weighting batches equally would give a one-pair tail the same say as a full batch of pairs.

**Eval batches are sequential and single-pass.** The held-out set is scored in the order it was declared, every evaluation scores the same batches, and `grad_accum=1` leaves each batch's normalizers describing only itself — which is what the pair weighting assumes.

**No `eval_perplexity`.** A preference loss is not a per-token likelihood, so there is nothing to exponentiate. `add_eval_callback` receives `perplexity=None`, and `report_to` omits the key rather than logging `None` — the W&B call logs eval loss and perplexity together inside one `try/except`, so a rejected `None` would have taken the eval loss down with it and charted nothing at all.

**`generate_during_eval` no longer short-circuits the evaluation.** Sampling is an addition to scoring, and runs after it. Its prompts come from the pass the scoring plan already makes over each split rather than from a second pass over the eval dataset, so evaluation as a whole reads that dataset once — measured, not assumed: a full run iterates the eval dataset exactly one time. The formatting hand-off that makes this work is applied inside the preference branch only, so no VLM or text SFT split is ever wrapped. Each row is formatted once and rendered once, and the prompt that gets sampled is the same text that gets scored — which matters because neither a `formatting_func` nor a chat template is required to return the same thing twice. Because the sampler keeps text and token ids rather than the row, no row is copied and nothing is assumed about what a row is; a dataset that rewrites one buffer per iteration is safe.

**`load_best_model_at_end` and `early_stopping_patience` are accepted** for these objectives now that an evaluation produces a metric to select on. `metric_for_best_model` can name any reported metric, so selecting on `rewards/accuracies` works. Both still require an eval dataset, which must report a length and be non-empty. The length is what stands in for termination — the plan materializes every row and nothing can prove an arbitrary iterable ends — so a bare generator is refused even though a single traversal would now suffice for one. Beyond that the dataset is never indexed and is read exactly once per run.

**`max_eval_batches` stays rejected.** It caps an unbounded stream by batch count. The preference plan is finite and epoch-driven, so honoring a batch cap would wrap into a second epoch and score part of the held-out set twice — silently, as a slightly wrong number.

## Validation

Cross-checked term by term against TRL 0.23.1's `DPOTrainer.dpo_loss` and `ORPOTrainer.odds_ratio_loss` on shared synthetic logits — eight checks, all passing: reference-free DPO, referenced DPO, DPO with label smoothing, and ORPO all agree, including the reward, logp, and log-odds terms and ORPO's NLL over the chosen sequence.

The two logit metrics are checked separately because TRL is not consistent between its own trainers: `DPOTrainer` averages the logit metric over the completion positions only, `ORPOTrainer` over the whole sequence. Each is followed rather than reconciled, so either metric compares directly against the run it mirrors. The DPO one is checked again under per-row response lengths, the fixture shape that separates a token-weighted mean from a row-weighted one, since TRL's is token-weighted across whatever shares the batch.

Every added test was mutation-checked — the production behaviour each one names was reverted or weakened, and exactly that test failed.

**The eval-mode gate is proven by a real Metal run, not by reading.** Evaluation runs the model in eval mode and restores training mode afterwards, which is what keeps NEFTune noise out of a reported eval number. The gate is the embedding's own training flag, so only a real `mlx.nn` module tree exercises it: the simulation shim's `Module.eval()` is a no-op and the class has no `training` attribute at all, so a test there would pass whether the gate worked or not. `test_neftune_noise_is_gated_out_of_the_preference_eval_forward` first proves the gate is live — the embedding is noised while training, redraws every forward, and is bit-identical to its base class while evaluating — then records the flag at every embedding call during a NEFTune-enabled ORPO run with an eval dataset. Removing `self.model.eval()` from `_evaluate` fails it with "NEFTune noise leaked into the eval forward"; removing the paired `self.model.train()` fails it with "_evaluate left the model in eval mode". A companion test covers the path that motivated putting the restoration in a `finally`: an evaluation that raises must still leave the model in training mode, or every step after it trains without NEFTune noise or dropout. This is the failure mode worth a real-model test: it moves an eval number without breaking anything visibly.

```
pytest tests/test_mlx_preference.py -q
102 passed

pytest tests/test_mlx_training_e2e_metal.py -q -k "preference or neftune or failed_evaluation"
3 passed
```

## Not in this PR

Train-side per-step preference metrics. The stats contract added here is reusable for them, but wiring them into the training log is separate work. The remaining CUDA-parity gaps — TRL's config defaults, BOS/EOS tokenization differences, and the CPO/KTO/GRPO trainers — are being handled separately as well.

### DIFF
--- a/tests/test_mlx_preference.py
+++ b/tests/test_mlx_preference.py
@@ -1323,3 +1323,29 @@ def test_generation_prompt_rejects_a_budget_with_no_room():
             {"prompt": "p", "chosen": "", "rejected": ""},
             max_seq_length=8, max_new_tokens=8,
         )
+
+
+def test_eval_loss_weights_every_pair_once_across_a_ragged_tail():
+    """Three rows at batch_size 2 leave a one-pair tail. Weighting each batch
+    equally would give that tail the same say as the two pairs before it."""
+    from unsloth_zoo.mlx.preference import (
+        create_preference_batch_plan, make_preference_eval_fn,
+    )
+
+    options = dict(max_seq_length=64, num_epochs=1, grad_accum=1,
+                   preserve_dataset_order=True)
+    split = create_preference_batch_plan(
+        rows(3), Tokenizer(), batch_size=2, **options)
+    whole = create_preference_batch_plan(
+        rows(3), Tokenizer(), batch_size=3, **options)
+    assert len(split) == 2 and len(whole) == 1
+    model = TinyModel()
+    eval_fn = make_preference_eval_fn("dpo", beta=0.1, reference_free=True)
+    total, weight = 0.0, 0
+    for index in range(len(split)):
+        loss, pairs, _stats = eval_fn(model, *split[index])
+        total += float(loss) * int(pairs)
+        weight += int(pairs)
+    assert weight == 3, "every pair is weighted once"
+    assert math.isclose(total / weight, float(eval_fn(model, *whole[0])[0]),
+                        rel_tol=2e-5, abs_tol=2e-5)
--- a/tests/test_mlx_preference.py
+++ b/tests/test_mlx_preference.py
@@ -820,26 +820,6 @@ def test_final_loss_averages_logical_updates_with_ragged_epoch_tail(
     assert result["train_loss"] == pytest.approx(expected, rel=1e-6)
 
 
-@pytest.mark.parametrize("eval_dataset,eval_steps", [
-    (rows(1), 1), (rows(1), 0), (None, 1),
-])
-def test_preference_capabilities_fail_before_model_setup(eval_dataset, eval_steps):
-    import mlx.nn as nn
-    from unsloth_zoo.mlx.trainer import MLXORPOConfig, MLXORPOTrainer
-
-    class CapabilityModel(nn.Module):
-        def __init__(self):
-            super().__init__()
-            self._config = {"model_type": "tiny"}
-
-    trainer = MLXORPOTrainer(
-        CapabilityModel(), Tokenizer(), rows(1), eval_dataset=eval_dataset,
-        args=MLXORPOConfig(eval_steps=eval_steps),
-    )
-    with pytest.raises(ValueError, match="evaluation, best-model loading"):
-        trainer._prepare_data(False)
-
-
 def test_trainer_applies_preference_formatter_once_per_row():
     import mlx.nn as nn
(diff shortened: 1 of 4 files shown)


Lyxot:
Folded into #1011 — the evaluation commits now sit on that branch, and its description covers both the scoring pass and the sampling pass. Closing this one so there is a single PR to review.

PR #1093, opened 2026-08-24
Make the concurrent CI lanes record the failure that actually happened

danielhanchen:
## The bug

Each fused CI lane runs in the background and writes its own exit status, which is the only thing the job ever learns about it. The capture looks like this:

```bash
{ cmd1; cmd2; ...; cmdN } > "$log" 2>&1 || rc=$?
echo "$rc" > ".lanes/$id.install"
```

A brace group reports the status of its **last** command only, and there is no `set -e` inside it, so execution continues past a failure and `rc` ends up being whatever `cmdN` returned. Every failure except the last one was recorded as `rc=0`. The guard immediately below it:

```bash
[ "$rc" = "0" ] || { echo "::error::$id environment could not be built (status $rc)"; exit 1; }
```

could therefore never fire, and the job continued with a half-built venv.

This is what it looked like in practice on #1092. The suites lane's pip install died with:

```
ERROR: Could not install packages due to an OSError: ('Connection broken:
IncompleteRead(323825 bytes read, 889 more expected)', ...)
```

The lane then ran `pip install pytest==9.0.3`, which succeeded, so `rc` was 0 and "Build both environments (concurrently)" reported **success**. The run went red two steps later in "Assert the real MLX runtime is what got installed" with `ModuleNotFoundError: No module named 'mlx'`, a step that had nothing to do with the actual failure.

## Why not `set -e`

Adding `set -e` inside the group does not work, and it is worth writing down because it is the obvious fix. The group sits on the left of `|| rc=$?`, and POSIX specifies that errexit is ignored for a compound command that is part of an AND-OR list. Verified:

```
{ false; true; }              -> rc=0    # today: failure masked
( set -e; false; true; )      -> rc=0    # still masked, on the left of ||
{ false && true; }            -> rc=1    # propagates
```

`&&`-chaining is what actually works: the first failure both stops the lane and becomes the recorded status. A deliberately optional step keeps its `|| true` by wrapping it, `{ cmd || true; } &&`, so it stays tolerated without breaking the chain.

## Scope

All three fused lanes had it:

| job | step | introduced in |
| --- | --- | --- |
| `interpreter-collect` | Install and collect on every interpreter | #1074 |
| `mlx-cpu-linux` | Build both environments | #1075 |
| `core-upstream-matrix` | Run all three upstream pins | #1075 |

The `core-upstream-matrix` lane was the worst of the three: its last command is `pip show transformers trl peft torch`, so a failed install was reported as `pip show`'s status. The `interpreter-collect` lane then gates its collect on `rc -eq 0`, and a collect failure is reported without failing the job, so a broken install there could go fully green.

Worth being clear that #1074 and #1075 were right in design. Recording a per-lane status and failing loudly on a missing one is the correct shape, and the comment in #1074 says so explicitly ("A lane that dies without recording a status is a failure, not a pass. Backgrounding is only safe if a missing result is loud."). Only the capture itself was inert.

## The transient

The trigger was a genuine network flake, so the lanes now wrap their network-bound installs in a small `retry()` (3 attempts, 15s apart). Note that bumping pip's own `--retries` would not have helped here: per `pip install --help` it is the "Maximum attempts to establish a new HTTP connection", so a response body truncated mid-download is not covered by it at all.

## Tests

`tests/test_ci_lane_scripts_parse.py` gains three checks:

- every statement in a status-capture group is `&&`-chained to the next
- every network-bound install is wrapped in `retry()`
- the shell semantics the first check relies on, pinned by executing them, so the reasoning above is verified rather than asserted

The first check found the `core-upstream-matrix` lane, which I had not spotted by reading. Both new checks were confirmed against injected regressions: removing one `&&` fails the chaining check, removing one `retry` fails the retry check, and the restored file passes.

```
31 passed
```

### DIFF
--- a/.github/workflows/consolidated-tests-ci.yml
+++ b/.github/workflows/consolidated-tests-ci.yml
@@ -92,6 +92,29 @@ jobs:
         run: |
           mkdir -p .lanes
 
+          # A lane records its own exit status, so every command in it has to be
+          # able to set that status. A brace group reports only its LAST command,
+          # and `set -e` inside it would be ignored because the group sits on the
+          # left of `|| rc=$?` (POSIX: errexit does not apply there). So the steps
+          # are `&&`-chained: the first failure both stops the lane and becomes rc.
+          #
+          # retry(): pip's own --retries only covers ESTABLISHING a connection, so
+          # a response body truncated mid-download (ProtocolError / IncompleteRead)
+          # is never retried and the install just dies. Wrap the network-bound
+          # installs so a truncated wheel costs a retry instead of a red run.
+          retry() {
+            attempt=1
+            until "$@"; do
+              if [ "$attempt" -ge 3 ]; then
+                echo "retry: giving up after $attempt attempts: $*"
+                return 1
+              fi
+              echo "retry: attempt $attempt failed, retrying in 15s: $*"
+              attempt=$((attempt + 1))
+              sleep 15
+            done
+          }
+
           # CPU index avoids the multi-GB CUDA wheel set. `--no-deps unsloth`
           # satisfies the find_spec("unsloth") guard at unsloth_zoo/__init__.py:128.
           lane() {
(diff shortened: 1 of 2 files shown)


danielhanchen:
@codex review

danielhanchen:
@codex review

PR #1094, opened 2026-08-24
fix(gguf): Retry GGUF conversion with --no-mtp when the converter asserts on an absent MTP head

Lyxot:
## The bug

GGUF export aborts for Qwen MTP models. Exporting a merged `Qwen/Qwen3.5-4B` save fails with:

```text
  File "conversion/qwen.py", line 303, in __init__
    assert self.opt_num_mtp_layers != 0
AssertionError
```

The export produces nothing at all — this is a hard failure, not a degraded GGUF.

## Root cause

llama.cpp's shared `_QwenMtpMixin` sizes the speculative block from `mtp_num_hidden_layers`, and when that key is absent falls back to counting `mtp.layers.<i>` tensors. It has no way to express "this architecture can carry an MTP head and this checkpoint does not have one", so when both signals come up empty it asserts instead of exporting a headless model.

A merged MLX save lands in exactly that state, and legitimately so: the `mlx-lm` and `mlx-vlm` sanitizers discard every `mtp.*` weight at load time, so the merged weights genuinely contain no head, and the staged config's speculative-layer sync drops the now-false declaration to match. Both halves are correct on their own; the converter simply cannot represent the result.

This is not specific to unsloth. A pristine Hugging Face checkpoint shaped like `Qwen/Qwen3-Coder-Next` — registered as `Qwen3NextForCausalLM`, no `mtp_num_hidden_layers`, and no `mtp.*` tensors — hits the same assertion under the stock `convert_hf_to_gguf.py` with no unsloth involvement.

The reason this surfaced now is upstream `4ed2b13f` ("model: MTP support for Qwen3-Next"), which removed Qwen3.5's own mixin — it did `block_count += hparams.get("mtp_num_hidden_layers", 0)` and so tolerated a missing key — and folded Qwen3.5 onto the shared class, which treats `0` as "infer or assert".

## What changed

One case is added to the converter self-heal loop in `convert_to_gguf`, symmetric to the existing case that drops `--no-mtp` when an architecture refuses the flag: when the converter fails on that assertion, retry once passing `--no-mtp` to state the head's absence explicitly.

Reacting to the converter's own failure, rather than predicting the situation from the config, is deliberate. Predicting it would need either a hardcoded list of the architectures llama.cpp routes to a speculative-block class — duplicating a registry upstream has already reorganised once — or a scan of every export's checkpoint for tensors that almost no model has. It also covers the checkpoints whose config never declares a head at all, which no config-derived signal can see.

The retry only widens the export's capability; it never narrows it. A checkpoint that does have a head either declares one or has recognisable `mtp.*` tensors, and in both cases the converter succeeds on the first attempt and this code is never reached.

## Risks and tradeoffs

The retry silently changes what the GGUF contains — it is the only self-heal case that alters the output rather than the process — so it prints a notice saying the resulting GGUF has no MTP head. The first run's traceback is still discarded along with the failed attempt, as with the other retries.

Matching is narrow by design. The pattern requires the assertion and its `!= 0` comparison on one line, so a run that logs an inferred count and then fails for an unrelated reason, or a same-line assertion over the same counter with a different invariant, keeps its head and surfaces its real error. An unterminated stderr stream concatenated onto stdout cannot splice the two words together into a false match either.

This is a workaround for a defect that belongs upstream. If `_QwenMtpMixin` gains a way to represent a headless checkpoint, this case becomes dead code and should be deleted rather than maintained.

## Validation

Focused suite, plus mutation checks — five mutations of the production change, each caught by a distinct test:

```bash
python -m pytest tests/test_convert_to_gguf_mtp_reconcile.py -q
```

End-to-end against the real converter with `Qwen/Qwen3.5-0.8B`, tracing the converter's argv:

```text
1: flags=['--outfile', '--outtype', '--split-max-size']
2: flags=['--outfile', '--outtype', '--split-max-size', '--no-mtp']
3: flags=['--outfile', '--outtype', '--mmproj', '--split-max-size']
```

The retry fires exactly once, the separate mmproj pass is unaffected, and the resulting file reports `qwen35.block_count = 24` with 320 tensors and a maximum block of `blk.23` — the text stack intact, no speculative head.

### DIFF
--- a/tests/test_convert_to_gguf_mtp_reconcile.py
+++ b/tests/test_convert_to_gguf_mtp_reconcile.py
@@ -472,10 +472,9 @@ def test_convert_to_gguf_still_strips_the_declaration_for_legacy_converters(llam
 
 
 def _write_asserting_converter(path: Path) -> Path:
-    """A stub that fails with llama.cpp's MTP assertion until `--no-mtp` is passed.
+    """Fails with llama.cpp's MTP assertion until `--no-mtp` is passed.
 
-    It replays the traceback rather than re-deriving it, so it pins the retry
-    wiring, not the upstream condition.
+    Replays the traceback, so it pins the retry wiring, not the upstream condition.
     """
     converter = path / "asserting_convert.py"
     command_log = path / "converter_commands.jsonl"
@@ -542,10 +541,10 @@ def test_convert_to_gguf_retries_with_no_mtp_after_the_inference_assertion(llama
     first, second = _read_converter_commands(tmp_path)
     assert "--no-mtp" not in first
     assert "--no-mtp" in second
-    # The model directory stays the trailing positional argument.
+    # The model directory stays the trailing positional.
     assert second[-1] == str(model_dir)
     assert (tmp_path / "output.gguf").exists()
-    # Dropping the head is the one thing this retry does that a user cannot see.
+    # Dropping the head is the one effect a user cannot otherwise see.
     assert "--no-mtp" in capsys.readouterr().out
 
 
@@ -555,8 +554,8 @@ def test_convert_to_gguf_retries_with_no_mtp_after_the_inference_assertion(llama
         ("  File \"qwen.py\", line 303\n    assert self.opt_num_mtp_layers != 0\nAssertionError\n", True),
         ("AssertionError: something else entirely\n", False),
         ("INFO:hf-to-gguf:opt_num_mtp_layers resolved to 1\n", False),
-        # Both words, different lines: a checkpoint whose head was recognised and
-        # which then failed for another reason keeps it.
+        # Both words, different lines: a recognised head that then failed for
+        # another reason keeps it.
         (
             "INFO:hf-to-gguf:opt_num_mtp_layers resolved to 1\n"
             "AssertionError: tensor shape mismatch\n",
@@ -614,14 +613,13 @@ def test_add_no_mtp_refuses_to_loop(llama_cpp):
 
 
 def test_convert_to_gguf_refuses_to_retry_when_the_checkpoint_has_mtp_tensors(llama_cpp, tmp_path):
-    """The assertion proves the converter recognised zero canonical
-    `mtp.layers.<i>`, not that there is nothing to keep. `--no-mtp` makes it
-    discard every `mtp.*`, so a checkpoint that does carry a head must surface
-    the disagreement instead of exporting a quietly reduced GGUF."""
+    """The assertion proves zero *recognised* `mtp.layers.<i>`, not a headless
+    checkpoint. `--no-mtp` discards every `mtp.*`, so a checkpoint that does
+    carry a head must surface the disagreement, not export a reduced GGUF."""
     model_dir = tmp_path / "model"
     model_dir.mkdir()
-    # No declaration -- exactly the shape that reaches the assertion -- but the
-    # head is there, indexed past the trunk.
+    # No declaration, the shape that reaches the assertion, but the head is
+    # there, indexed past the trunk.
     (model_dir / "config.json").write_text(
         json.dumps(
             {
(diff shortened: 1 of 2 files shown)


danielhanchen:
Reviewed this end to end against the real converter and pushed a follow-up commit to the branch.

## The bug is real, and this fixes it

Built tiny Qwen3-Next checkpoints and ran the actual `convert_hf_to_gguf.py` (llama.cpp `c1d0e7a`, which carries the assertion at `conversion/qwen.py:303`) from base main and from this branch:

| Fixture | base main | this branch |
|---|---|---|
| no declaration, no head | fails on `assert self.opt_num_mtp_layers != 0` | OK, one retry, `block_count=4`, no `nextn_predict_layers` |
| head present, no declaration | OK, `block_count=5`, `nextn_predict_layers=1` | identical, no retry |
| head present, declared | OK, `block_count=5`, `nextn=1` | identical |
| declared, no head (#1033) | OK, `block_count=4` | identical |

Also confirmed upstream chronology: `--no-mtp` landed in `25558268` (2026-05-16) and is an ancestor of the assertion commit `4ed2b13f` (2026-08-03), so no stock upstream revision has the assertion without the flag. Against a pre-assertion converter (`2b63e061`) all four fixtures behave exactly as on base, so older llama.cpp installs are unaffected.

## What the follow-up commit changes

The assertion proves the converter recognised zero canonical `mtp.layers.<i>` tensors, not that the checkpoint is headless, and `--no-mtp` makes `filter_tensors` discard every `mtp.*`. A checkpoint carrying the shared `mtp.fc`/`mtp.norm` block but no countable `mtp.layers.<i>` reproduces it: base fails loudly, this branch silently discarded the tensors and reported success. It now keeps the assertion and explains the disagreement.

Four smaller ones:

- The retry is limited to required (text) runs. `--no-mtp` never belonged on the mmproj command, but the branch sits in the loop both passes share, so a projector failure quoting the assertion added it there.
- `captured` is joined on a newline instead of concatenated, and the pattern uses horizontal whitespace only around `!= 0`. stderr ending `assert self.opt_num_mtp_layers !=` plus stdout starting ` 0` completed the pattern and retried an unrelated failure. The join alone was not enough, since `\s` spans newlines. (The existing splice test uses an `AssertionError` boundary, which `\bassert\b` cannot match anyway, so it did not cover this.)
- The failed attempt's output is removed before the retry, as the OOM branch already does. Callers scan the output directory for `*.gguf`, so a truncated `--outfile` or a leftover shard would be shipped as valid.
- The first attempt's failure is kept in the final error, so a retry that fails for its own reason does not discard the assertion that caused it.

## Validation

- 40 tests in `test_convert_to_gguf_mtp_reconcile.py`, plus `test_convert_to_gguf_self_heal.py` and `test_convert_hf_to_gguf_patcher.py`; 476 passed across the gguf/convert/save suites with no regressions.
- Mutation check: the original change had 2 of 5 mutations caught (dropping `\bassert\b` and widening `[^\r\n]*` to `.*` both survived). All 9 mutations of the current change are now caught.
- Python 3.9 through 3.14 in isolated venvs: identical `convert_to_gguf` signature to base, identical behaviour, all suites pass on every version.
- Platform coverage is by construction rather than by CI: nothing in the changed path branches on `IS_WINDOWS`/`sys.platform`/`platform.system()`, and nothing imports torch, so NVIDIA/AMD/CPU and Linux/macOS/Windows/WSL are the same code. Simulated the parts that do differ - CRLF and CR line endings, cp1252/latin-1/utf-16 console output through `errors="replace"`, and Windows drive-letter, UNC, extended-length, spaced and non-ASCII model paths through `_add_no_mtp` under `shell=False`.
- Retry state machine: nine scripted converter sequences over the cartesian product of dep-error, flag refusal, assertion and success, asserting invocation counts, termination, and that the model directory stays the trailing positional in every invocation.

danielhanchen:
@codex review

Issue #969, opened 2026-07-31
merge_and_overwrite_lora writes a text-only config against full VLM weights when text_only=True


Zelys-DFKH:
### Summary

`merge_and_overwrite_lora` writes a text-only `config.json` next to full VLM weights when the model was loaded with `text_only = True`. The saved checkpoint reloads with every weight randomly initialized, and nothing raises.

### Reproduction

```python
from unsloth import FastLanguageModel

model, tokenizer = FastLanguageModel.from_pretrained(
    model_name = "unsloth/gemma-3-4b-it-unsloth-bnb-4bit",
    max_seq_length = 512,
    load_in_4bit = True,
    text_only = True,
)
model = FastLanguageModel.get_peft_model(
    model, r = 8, lora_alpha = 8,
    target_modules = ["q_proj", "k_proj", "v_proj", "o_proj"],
)
model.save_pretrained_merged("out", tokenizer, save_method = "merged_16bit")
```

Then inspect what was written:

```python
import json, collections
cfg = json.load(open("out/config.json"))
keys = list(json.load(open("out/model.safetensors.index.json"))["weight_map"])
print(cfg.get("architectures"), cfg.get("model_type"))
print(collections.Counter(k.split(".")[0] for k in keys))
```

Output:

```
None gemma3_text
Counter({'language_model': 444, 'vision_tower': 437, 'multi_modal_projector': 2})
```

The config says `gemma3_text` and carries no `architectures` key. The weights are the full VLM under `language_model.*`, `vision_tower.*` and `multi_modal_projector.*`. There are zero `model.layers.*` tensors, which is what a text-only `Gemma3ForCausalLM` looks for.

### Consequence

Reloading the 8.6GB checkpoint silently discards all of it:

```python
from transformers import AutoModelForCausalLM
m = AutoModelForCausalLM.from_pretrained("out", dtype="bfloat16")
```

Every parameter is reported missing and newly initialized, including `model.embed_tokens.weight`, all 34 `model.layers.*`, `model.norm.weight` and `lm_head.weight`. No exception is raised, so a finetune can be saved, reloaded and served as an untrained model without any signal that something went wrong.

The export also writes the whole vision tower that `text_only = True` was asked to skip, so the file is roughly twice the size it needs to be.

### The default path is fine

Running the identical flow without `text_only = True` produces a consistent checkpoint:

```
architectures: ['Gemma3ForConditionalGeneration']   model_type: gemma3
[...]

PR #1097, opened 2026-08-25
Drop the vision and audio tensors from a text_only merged_16bit export

Zelys-DFKH:
Follows #1073, which fixed the first half of #969. Please read the second paragraph before the rest: this changes what a `text_only = True` merge writes, on a path Studio is on.

#969 had two directions. #1073 took the first one: the merge writes VLM weights, so save a config that describes them. That made the export loadable. This takes the second one, for the `text_only = True` case only: bring the weights to the request instead. The vision and audio tensors go, the text tensors are renamed into the text-only namespace, and the export becomes the causal LM that `text_only = True` asked for.

So for a `text_only` merge the config that lands is the text config again, not the composite one #1073 saves. That is a deliberate reversal of #1073 on this one path. #1073 stays load-bearing: whenever the text weights cannot be located in the base checkpoint, this stands down with a warning and the export is exactly what #1073 writes today. Since Studio passes `text_only = True` for any non-image dataset, that fallback is the part I would most like reviewed.

One correction to my own framing on the #1073 thread, where I said this halves the export. It does not, or not reliably: dropping 439 of `unsloth/gemma-3-4b-it`'s 883 tensors takes it from 8.60 GB to 7.76 GB, because SigLIP is small next to a 3.9B decoder. Tensor count is not bytes. The saving scales with how much of the checkpoint is not the text model, so an omni checkpoint carrying audio, speech and talker stacks gives back far more. The reason to do this is that the export is the model that was asked for. What it costs on disk is a side effect, and how big a one depends on the checkpoint.

## Finding the text weights

The hard part is that there is no one place a VLM keeps them. The layout is decided when the checkpoint is written, not by the transformers doing the reading, so two checkpoints from the same family disagree. Five arrangements turned up in real files:

| where the text weights live | head | seen on |
|---|---|---|
| `language_model.model.*` | tied, no tensor | `unsloth/gemma-3-4b-it` as published |
| `model.text_model.*` | `lm_head.weight` | `HuggingFaceM4/Idefics3-8B-Llama3` |
| `thinker.model.*` | `thinker.lm_head.weight` | `Qwen/Qwen2.5-Omni-3B` |
| `model.language_model.*` | `lm_head.weight` | `Gemma3ForConditionalGeneration` in memory |
| `model.language_model.model.*` | tied, no tensor | gemma3 written by `save_pretrained` |

`_checkpoint_conversion_mapping` is `None` on both Gemma3 classes, so there is nothing upstream to borrow. Hardcoding any one of these silently mismaps the others, which is the same class of bug #969 reported.

All five are the text key with one prefix swapped for another, so `_text_only_key_map` derives that pair from the two key sets rather than assuming it. The expected text keys come from meta-instantiating `AutoModelForCausalLM.from_config(config)`, which is the reload contract itself: `config` is what gets written, so whatever a reloader builds from it is exactly the set of tensors that must be on disk. For each text key the base does not already hold, it offers every (text prefix, base prefix) pair that lands it on a real tensor, keeps only the pairs every such key agrees on, and requires the survivors to describe one mapping. Anything else raises, and the caller falls back.

A tied head is handled explicitly rather than by accident: `unsloth/gemma-3-4b-it` ships 883 tensors and not one is an `lm_head`, because the reload reties it from the embeddings.

## After the map

Post-pass after the Step 5 merge loop rather than inside `_merge_and_overwrite_lora`, which is shared with the mxfp4 and fp8 paths. One shard is read at a time, so the working set is what `split_safetensors_to_shards` already uses on this path. Shards that were pure vision keep nothing and are deleted, `renumber_safetensor_files` closes the gaps, and the index is rewritten to the surviving shards or removed when one file is left. A stale index is not cosmetic here: it still names shards that no longer exist.

The config is flipped to the text config only after the drop succeeds, so a failed remap leaves nothing half-written. `architectures` is set from the class the expected key set was built from, since #969 saw `architectures: null` beside the text config.

One combination stands down and warns: `low_disk_space_usage` together with `push_to_hub` uploads and deletes each shard inside the merge loop, so by the time this post-pass runs there is nothing left on disk to read the text weights out of. The warning says that, and the export is the correct #1073 one, just larger than asked for. `low_disk_space_usage` on its own still drops, and there is a test pinning that, because an over-broad guard here would quietly cost everyone saving locally.

One family maps successfully and still deserves a flag. `MllamaForCausalLM` built from an `MllamaTextConfig` carries `model.layers.N.cross_attn.*`, because on mllama the cross-attention into the vision tower lives in the text stack. So a `text_only` mllama export keeps those weights: the checkpoint is self-consistent and reloads, and mllama skips cross-attention when there are no vision states, but the tensors are along for the ride. I have not run this on real mllama weights, only confirmed the shape of the expected key set. If you would rather that family stand down too, it is a `model_type` check in `_is_text_only_export`.

## Files changed

- `unsloth_zoo/saving_utils.py`: `TextOnlyRemapError`, `_is_text_only_export`, `_text_only_key_map`, `_text_only_expected_keys`, `_shard_keys_on_disk`, `_rewrite_shards_text_only`, `_write_text_only_index`, `_export_text_only_config`, and the post-pass wired in before Step 7.
- `tests/test_text_only_key_map.py`: new. The five layouts above plus five negatives, each one a case where the mapping is not uniquely determined and the code has to raise instead of picking. Key sets only, no weights, no network.
- `tests/test_text_only_shard_rewrite.py`: new. `unsloth/gemma-3-4b-it` keeps text weights in both its shards, so an end-to-end run never deletes a shard, never renumbers, and never has to decide whether the index should still exist. These drive those three branches directly on tiny real safetensors, because a checkpoint whose index names a file that is not there fails at load with nothing useful to say.
- `tests/test_merge_e2e_text_only_config.py`: extended to the new contract. `_reload_info` now resolves the class from the saved `architectures` instead of assuming the VLM, so it reads whichever checkpoint is in front of it.

## Test results

Full suite on Linux, Python 3.12.3, torch 2.11.0+cu130, transformers 5.5.0, peft 0.20.0, run twice in two separate clones so neither run could touch the other's tree:

| | passed | failed | skipped | time |
|---|---|---|---|---|
| `c1dca718`, the base this branches from | 5175 | 12 | 193 | 12:43 |
| this branch | 5193 | 12 | 193 | 13:04 |

The two failure lists are the same twelve node IDs, `diff` clean: five in `test_allow_cpu_import_driverless`, four in `test_pretokenized_bypass_collator`, and one each in `test_fla_hopper_tile`, `test_temporary_patches_imports`, `test_zoo_source_upstream_refs`. None of them import `saving_utils`, and they fail the same way on the base commit, so this change neither causes nor fixes them. I am listing them rather than describing the suite as green.

The 18 extra passes are 19 tests added and one removed, which is the next section.

Key sets can agree with each other and still be wrong about a real file, so there is also a run on real weights: `unsloth/gemma-3-4b-it` merged with `text_only = True` drops 439 tensors and keeps 444 as `Gemma3ForCausalLM`, config `gemma3_text`, two shards, every `weight_map` entry naming a file that exists and holding the key it claims, and the reload reports 0 missing, 0 unexpected, 0 mismatched.

## One test from #1073 is deleted

`test_text_only_export_config_matches_written_weights` asserts `model_type != "gemma3_text"` for a `text_only` export. That is precisely the contract this PR inverts, so it cannot survive as written, and I would rather delete it in the open than weaken it until it passes. `test_text_only_export_drops_the_vision_tower` replaces it: same fixture, same shape, the opposite config assertion, and the reload check carried over unchanged. If you would rather keep the #1073 behavior reachable than replace it, that is a reasonable call and I will rework this behind a flag.

## Credit

#969's second direction and its framing are kgt392's, from the #1073 thread. The `get_text_config` catch and the wider family sweep are danielhanchen's.

### DIFF
--- a/tests/test_merge_e2e_text_only_config.py
+++ b/tests/test_merge_e2e_text_only_config.py
@@ -153,7 +153,9 @@ def test_text_only_export_drops_the_vision_tower(tmp_path):
 
 
 def test_text_only_export_is_smaller_than_the_base(tmp_path):
-    """Halving the checkpoint is the point of the drop, so measure it rather than assume it."""
+    """The saving is the point, so measure it. How large it is depends on the model: the
+    tower is 10% of gemma-3-4b-it and most of an omni checkpoint, so only pin the direction.
+    """
     H.set_offline_cpu_env()
     base_dir, cfg = _write_vlm_base(tmp_path)
     out_dir = os.path.join(str(tmp_path), "merged")
--- a/tests/test_merge_e2e_text_only_config.py
+++ b/tests/test_merge_e2e_text_only_config.py
@@ -192,6 +192,26 @@ def test_index_and_shards_agree_after_the_drop(tmp_path):
         "the index and the shards disagree about which tensors exist")
 
 
+def test_low_disk_space_still_drops_without_a_push(tmp_path):
+    """Only the upload half of low_disk_space_usage blocks the drop, so keep the guard narrow.
+
+    That combination uploads and removes each shard inside the merge loop under a name chosen
+    before the drop changes how many shards there are, which is why it stands down. Saving to
+    a local directory removes nothing early, so there is no reason to give up the drop there.
+    """
+    H.set_offline_cpu_env()
+    base_dir, cfg = _write_vlm_base(tmp_path)
+    out_dir = os.path.join(str(tmp_path), "merged")
+
+    pm = _attach_lora(_text_only_model(cfg, base_dir))
+    H.run_merge(pm, base_dir, out_dir, save_dtype=torch.float32, low_disk_space_usage=True)
+
+    written = list(H.read_safetensors_dir(out_dir))
+    assert written, "the export wrote nothing"
+    assert not [k for k in written if "vision" in k], "low_disk_space_usage skipped the drop"
+    _assert_clean_reload(out_dir)
+
+
 def test_remap_failure_keeps_the_whole_checkpoint(tmp_path, monkeypatch):
     """When the text weights cannot be placed, fall back to #1073 rather than guess.
 
--- a/tests/test_merge_e2e_text_only_config.py
+++ b/tests/test_merge_e2e_text_only_config.py
@@ -14,12 +14,17 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-"""The merged_16bit config must describe the weights the merge actually writes (#969).
+"""The merged_16bit config and weights must describe the same model (#969).
 
 `merge_and_overwrite_lora` takes its weights from the resolved base checkpoint but used to
 save `model.config`, which `text_only = True` has already replaced with the nested text
 config, so a `gemma3_text` config landed next to full VLM weights. Nothing raised: every
 tensor was silently re-initialized on reload, so a finetune could be served untrained.
+
+There are two ways to make the pair agree, and this suite covers both. A `text_only` merge
+drops the vision and audio tensors and exports the text model that was asked for. When the
+text weights cannot be located in the base checkpoint the drop stands down, and the export
+falls back to the whole VLM under a config that describes it -- correct, just larger.
 
 The fixture is the real shape -- a text-only decoder whose `_name_or_path` points at a VLM
 checkpoint. Passing is not "no exception" (the bug threw none) but the saved architecture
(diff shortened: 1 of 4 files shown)


danielhanchen:
Confirmed main already writes a loadable export via the base config in saving_utils.py, and this goes the other way by actually dropping the vision and audio tensors a text_only merge asked to skip, with a warned fallback when the text weights cannot be located. Since this inverts the config contract merged in #1073 on that one path, I want to look closely at the fallback before this goes in.

Zelys-DFKH:
The `try` is read-only. It reads key names off the shards, builds the expected set on the meta device, and compares the two. Nothing on disk changes until `_text_only_key_plan is not None`. By then the directory already holds the full #1073 export, so the fallback has nothing to undo. It just does not run, and the warning says which export you got.

That is why the `except` is broad: any failure lands on a checkpoint that is already correct.

The rewrite half is stricter. The keys on disk are checked against the mapped set, and a mismatch raises instead of warning. That raise leaves a half-rewritten directory, and the message tells you to delete it and merge again. Writing to a sibling directory and swapping at the end would make it atomic, if you prefer that.

`test_remap_failure_keeps_the_whole_checkpoint` covers the fallback end to end.

danielhanchen:
Checked this against main, where the merged_16bit branch now saves the base config, so this is the follow-up that actually makes a text_only export the causal LM that was requested, and the stand-down path when the text weights cannot be located looks right. Will get this reviewed.

danielhanchen:
Checked this against the current merged_16bit path in saving_utils.py, and it deliberately reverses what #1073 writes for a text_only merge, so anyone reloading that export as the composite model gets a different checkpoint than before; could you confirm the fallback path when the text weights cannot be located, and say whether you would rather put the new behaviour behind a flag?

danielhanchen:
This reverses what #1073 writes for a `text_only` merge, which is the path Studio takes for any non-image dataset, and the shards are rewritten in place before the verification raise in saving_utils.py. Could you confirm the fallback and the mllama cross-attention case on real weights, and say whether you would rather keep the #1073 behaviour reachable behind a flag?

Zelys-DFKH:
@danielhanchen thank you for going through this twice, and sorry it sat long enough to need the second look. You were right about the order: the shards were rewritten in place before the check that could refuse them. I've rebased onto main and reworked that part, and the description is updated to match. Here are your three questions:

**Fallback.** Nothing is written over an original shard until the whole plan has been staged and checked. The text tensors go to hidden files next to each shard, and the staged key set has to match the plan exactly before anything touches the originals. If a key can't be placed, if the config class is one `AutoModelForCausalLM` doesn't know, if the disk fills up, or if the check fails, the staged files are deleted, one warning is printed, and the directory stays exactly the #1073 export, untouched. If something fails after that point, while the files are being moved into place, it cleans up and raises an error that says the export is incomplete, rather than leaving a half-converted checkpoint that looks finished. The tests break each of these on purpose.

**mllama on real weights.** I ran it, and my earlier description was wrong. It said mllama maps and keeps its `cross_attn` tensors, but that came from the key set, not the weights. In practice unsloth's loader declines `text_only` for mllama (`resolve_model_class` comes back `None` for its text config), so the full model loads and this code never runs. On `unsloth/Llama-3.2-11B-Vision-Instruct` the export is the #1073 checkpoint: 906 tensors, 64 of them `cross_attn`, reloads as `MllamaForConditionalGeneration` with nothing missing, and generates. If something did hand it an `MllamaTextConfig` directly, the expected keys can't be built, so it falls back. `gemma-3-4b-it` still drops 439 and keeps 444, reloads clean, and says Paris.

**Flag.** I'd lean against one, and I'll add it if you'd still like it. `text_only` is decided at load, so the export should describe the model that was actually loaded. A save-side switch would let the loaded model and the export disagree, and that disagreement is exactly how #969 happened. Whenever the drop can't be done safely, you still get the #1073 export.

I've also been following your loader-side `text_only` work (#1350, unslothai/unsloth#11584, unslothai/unsloth#11684, unslothai/unsloth#11692), and it's good to see that path getting this much attention. If any of it changes which checkpoints reach this path, I'm happy to adjust.

PR #1098, opened 2026-08-25
MLX: int8 W8A8 prefill for quantized models on Apple M5

danielhanchen:
## What this does

MLX computes quantized matmuls in 16-bit regardless of how the weights are stored, so a 4-bit model prefills no faster than an 8-bit one. On Apple M5, whose neural accelerators run int8 at roughly twice the fp16 rate, routing prefill-sized matmuls through an `int8 x int8 -> int32` GEMM recovers that. Decode is memory-bound and keeps MLX's kernels untouched.

Adapted from [JetBrains' int8 NAX prefill patch for mlx-vlm](https://github.com/JetBrains/mlx-vlm/tree/feature/int8-prefill) (MIT, attribution retained in the kernel file), which measured **+29.5% at 6.4k tokens and +47.3% at 11.8k** on Qwen3.6-27B-4bit for Junie Local.

Worth noting the gap is not MLX-specific. llama.cpp has had int8 W8A8 on CUDA for years (`ggml-cuda/mmq.cuh`, with `mma.cuh` emitting `mma.sync...s32.s8.s8.s32`), but its **Metal** backend dequantizes the weight tile to 16-bit and runs `matmul2d` with a float accumulator (`ggml-metal/kernels/mul_mm.metal`), exactly as MLX does. So this is unfilled on Apple silicon generally, not just here.

## Usage

Opt-in, and a hard no-op everywhere it is not supported.

```python
from unsloth_zoo.mlx import int8_prefill

if int8_prefill.enable():          # False, and a no-op, off M5
    int8_prefill.warmup(model)     # required: builds the allow-list
    ok, detail = int8_prefill.self_test()
```

`warmup()` is not optional. Nothing is accelerated until a weight is registered, and registration is also where every `mx.eval` happens, which is what leaves the hot path safe inside `mx.compile`.

## Differences from the reference implementation

- **Patches `mx.quantized_matmul` at op level** rather than `nn.QuantizedLinear.__call__`. Tied lm_heads reach the op through `QuantizedEmbedding.as_linear` (68 model files in mlx-lm), MLA through `QuantizedMultiLinear`, sharded models through the distributed linears. A layer-level patch needs five separate class patches to cover the same ground. Verified that nothing in mlx, mlx-lm or `unsloth_zoo.mlx` imports the symbol by value, so the `sys.modules` sweep is insurance rather than a requirement.
- **Dispatch is gated on a warmup-built registry**, not per-call shape heuristics. This keeps the quantized-KV attention path (`mlx_lm/models/base.py`, where the "weight" is a per-step cache tensor) out by construction, moves every `mx.eval` into warmup so the hot path survives `mx.compile`, and makes scope policy a registration decision, since warmup knows the module path and the op does not.
- **The fast path is an `mx.custom_function`** whose vjp delegates to the stock 4-bit op. `mx.fast.metal_kernel` outputs carry no vjp, so without this a LoRA backward through an intercepted prefill would fail inside the trainer.
- **Group sizes 32/64/128**, and the nibble arithmetic is parameterized on bits rather than assuming 4.
- **The MetalPerformancePrimitives include is isolated to the GEMM**, so the activation-quant and weight-requant kernels compile on any Metal device and can be validated in CI.
- **A self-test at `enable()`** runs registered shapes, including a ragged `M % 128 != 0`, against MLX's own op on the real device and disables permanently on mismatch.
- Idempotent, with `disable()` restoring `mx.quantized_matmul` exactly.

## Affine 8-bit is implemented but off by default

This is arithmetic, not caution. Requantizing affine-8 to symmetric int8 has no headroom: the requant step is `max_g range_g / 254` against a native `range_g / 255`, so any channel whose inter-group range ratio exceeds about 1.2 comes out **worse than the weights it started from**. At 4 bits the same requant has roughly 16x of cushion, which is the only reason the technique works at all. Enable with `UNSLOTH_MLX_INT8_ALLOW_8BIT=1` only alongside a KLD check.

## What is verified, and what is not

**The speedup is unverified.** No GitHub runner tier offers an M5. The macOS leg measured that `mpp::tensor_ops::matmul2d` with int8 operands does not even load on M1 (`[metal::Device] Unable to load kernel`) while a trivial Metal kernel JITs fine on the same runner, so the GEMM is M5-only by nature rather than by choice.

| | Verified | How |
| --- | --- | --- |
| Dispatch: what is intercepted, what falls through bit-identically | yes | 70 tests, any MLX backend |
| W8A8 arithmetic and its error against MLX's 4-bit op | yes | portable backend, machine-independent |
| `mx.compile` safety, LoRA backward, idempotency, no-op path | yes | 70 tests |
| Packed-weight unpacking on real Metal | yes | macOS CI, all 3 group sizes, within one int8 LSB |
| GEMM tile conventions | no | needs M5 |
| Any performance claim | no | needs M5 |

The one int8 LSB tolerance is Metal's `rint` rounding halves to even where `mx.round` does not, so ties can differ by one and nothing else may.

Mitigations for the untested GEMM: the `enable()` self-test above, and `UNSLOTH_MLX_INT8_VERIFY=1` shadow mode which logs max relative error per intercepted call.

If you have an M5, I have a bench harness that reproduces JetBrains' context points (256 / 2k / 6.4k / 11.8k / 20k) for a direct comparison. Happy to add it here if wanted.

## CI

Added to the existing `mlx-smoke` job rather than a new workflow: macOS concurrency is capped at 5 jobs account-wide and that runner already has the MLX stack provisioned, so a second workflow would spend a slot reinstalling what is already there.

The most important assertion is inverted from what you would expect: the capability probe **must return False** on the M1 runner, and enabling the module must leave a model forward bit-identical. That is the configuration every current Mac user is in, so it is the check that protects them.

## Not covered

MoE gains nothing yet. Expert FFNs go through `mx.gather_qmm` (`mlx_lm/models/switch_layers.py`), not `mx.quantized_matmul`, and on a Qwen3-MoE-class model those are the dominant prefill FLOPs. A grouped GEMM over ragged per-expert row counts is a separate kernel. Registry entries already carry an optional expert dimension so it stays a GEMM rather than a rework.

## Side finding

MLX 0.32.1's CUDA backend raises `cuGraphAddKernelNode "invalid argument"` for `mx.quantized_matmul` at `group_size=128`. Reproducible with none of this code loaded, and absent on Metal. One numerics test skips on Linux for that reason and the macOS leg asserts the same call works there, so it is covered rather than dropped. Worth reporting upstream to ml-explore separately.

### DIFF
--- a/.github/workflows/consolidated-tests-ci.yml
+++ b/.github/workflows/consolidated-tests-ci.yml
@@ -464,6 +464,39 @@ jobs:
             echo "real mlx is installed but the numeric tests skipped"; exit 1
           fi
 
+      - name: int8 prefill suites (HARD GATE)
+        # The Apple Silicon lane is opt-in behind an `mlx` label, so on a normal pull
+        # request this is the only place the int8 prefill code is exercised at all.
+        # Everything here is backend-independent: dispatch, eligibility, the W8A8
+        # arithmetic via the portable backend, mx.compile safety and the LoRA vjp. The
+        # Metal kernels and the M5 capability probe are not reachable from Linux and are
+        # not claimed to be.
+        env:
+          PYTHONPATH: ${{ github.workspace }}/tests
+        run: |
+          .lanes/venv-numerics/bin/python -m pytest -v -rs \
+            tests/test_mlx_int8_prefill_dispatch.py \
+            tests/test_mlx_int8_prefill_numerics.py \
+            tests/test_mlx_int8_prefill_safety.py
+
+      - name: Fail if the int8 prefill suites collapsed to skips
+        # One skip is expected and understood: mx.quantized_matmul at group_size 128
+        # cannot be evaluated on some backends, so the comparison against it is skipped
+        # on the reference rather than on a backend guess. Anything beyond that means
+        # the suites stopped testing and quietly passed.
+        env:
+          PYTHONPATH: ${{ github.workspace }}/tests
+        run: |
+          .lanes/venv-numerics/bin/python -m pytest -q -rs \
+            tests/test_mlx_int8_prefill_dispatch.py \
+            tests/test_mlx_int8_prefill_numerics.py \
+            tests/test_mlx_int8_prefill_safety.py | tee int8.txt
+          SKIPPED=$(grep -oE '[0-9]+ skipped' int8.txt | grep -oE '[0-9]+' || echo 0)
+          if [ "${SKIPPED:-0}" -gt 1 ]; then
+            echo "int8 prefill suites skipped ${SKIPPED} tests, expected at most 1"
+            exit 1
+          fi
+
       - name: tests/test_mlx_rng_rewind.py (HARD GATE)
         # PRNG capture/rewind across the compile fallbacks. Needs mlx core, not
         # Metal, so it gates here; the Apple Silicon lane is opt-in and usually
--- a/.github/workflows/mlx-ci.yml
+++ b/.github/workflows/mlx-ci.yml
@@ -199,3 +199,82 @@ jobs:
         run: |
           python -m pytest tests/test_mlx_batching_and_decay.py -v -rs \
             --deselect tests/test_mlx_batching_and_decay.py::test_compiler_review_guards_are_present
+
+      # int8 W8A8 prefill. Added to this job rather than to a new workflow: macOS
+      # concurrency is capped at 5 jobs account-wide, and this runner already has the
+      # MLX stack provisioned, so a second workflow would spend a slot to reinstall
+      # what is already here.
+      #
+      # What this leg can and cannot show. The int8 path needs the M5 neural
+      # accelerators, and no GitHub runner tier offers one, so nothing here measures
+      # performance. What it does establish, and the Linux leg cannot:
+      #
+      #   1. The capability probe declines on a real Metal device that lacks M5 int8
+      #      tensor ops, and enabling the module then changes a model forward not at
+      #      all. That is the configuration every current Mac user is in, so it is the
+      #      assertion that matters most.
+      #   2. The activation-quant and weight-requant Metal kernels compile and run for
+      #      real. They deliberately carry no MetalPerformancePrimitives include, so
+      #      they execute here even though the GEMM cannot, and their packed-nibble and
+      #      word-boundary arithmetic is the likeliest source of a silent wrong-answer
+      #      bug. Linux cannot check it at all.
+      #
+      # Measured on macos-15/26: mpp::tensor_ops::matmul2d with int8 operands does not
+      # load on M1 ("[metal::Device] Unable to load kernel") while a trivial Metal
+      # kernel JITs fine on the same runner, so the GEMM stays M5-only by nature rather
+      # than by choice.
+      - name: tests/test_mlx_int8_prefill_dispatch.py
+        env:
+          PYTHONPATH: ${{ github.workspace }}/tests
+        run: python -m pytest tests/test_mlx_int8_prefill_dispatch.py -v -rs
+
+      - name: tests/test_mlx_int8_prefill_numerics.py
+        env:
+          PYTHONPATH: ${{ github.workspace }}/tests
+        run: python -m pytest tests/test_mlx_int8_prefill_numerics.py -v -rs
+
+      - name: tests/test_mlx_int8_prefill_safety.py
+        env:
+          PYTHONPATH: ${{ github.workspace }}/tests
+        run: python -m pytest tests/test_mlx_int8_prefill_safety.py -v -rs
+
+      - name: int8 prefill Metal kernels vs reference
+        # The two MetalPerformancePrimitives-free kernels, checked against the plain-MLX
+        # implementations of the same arithmetic. A tolerance of one int8 LSB, because
+        # Metal's rint rounds halves to even while mx.round does not, so ties can differ
+        # by one and nothing else may.
+        run: |
+          python -c "
+          import mlx.core as mx, mlx.nn as nn
+          from unsloth_zoo.mlx.int8_prefill import scales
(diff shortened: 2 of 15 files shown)


Lyxot:
Tested this on an M3 Max (macOS 26.5.1, MLX 0.32.0) and on GitHub's macOS runners. Two useful results: most of the "needs M5" table is verifiable today, and the capability gate has a hole.

## The GEMM is testable without an M5

On M3 Max the MPP int8 kernel compiles, loads, and is **bit-exact against a numpy int32 reference on 96/96 configs** — M ∈ {1, 127, 128, 129, 512, 549, 1000, 2048}, N ∈ {128, 256, 2048}, K ∈ {32, 256, 1536, 4096}. That covers the transposed-right extents, the cooperative-tensor index ordering, and the ragged-M epilogue guard.

Full path vs `mx.quantized_matmul`: rel err 1.1e-2–1.7e-2 across 3 shapes × group sizes 32/64/128 × rows 512/549/2000; Metal vs portable agree to ≤7.4e-3; requant within the documented 1 LSB. `mx.grad` and `mx.compile` both work through the Metal backend. `N % 128 != 0` does corrupt output (N=200 → 36% of elements wrong), so that rule is load-bearing.

## The gate admits M3, where the path is slower

Runner results (`macos-15` and `macos-26`, same workflow):

| Runner | Chip | macOS | GEMM | `is_supported()` |
| macos-15 | M1 (Virtual) | 15.7.7 | `MetalPerformancePrimitives.h` not in SDK | False — *"macOS 15 below 26"* |
| macos-26 | M1 (Virtual) | 26.5.2 | `Unable to load kernel` | False — probe failed |
| local | **M3 Max** | 26.5.1 | runs, bit-exact | **True** |

So the probe does discriminate on M1 — but the current CI assertion runs on `macos-15`, where `_MACOS_MIN = 26` short-circuits before the probe ever executes, so it passes without testing the discriminator. (Caveat: the macOS 26 runner is an `Apple Paravirtual device`, not bare-metal M1.)

On M3 Max the probe passes, `enable()` returns True, `self_test()` passes, and the forward changes by 1.7e-2 — while prefill gets slower. Interleaved A/B, min-of-80, half-split convergence check:

- raw int8 `matmul2d` vs bf16 `mx.matmul`: **0.92×** (2048×11008×4096) and **0.93×** (6400×4096×4096), both converged to <2%. No int8 rate advantage on this silicon.
- end to end vs `mx.quantized_matmul`: **0.90×** at 11008×4096 / 6400 rows, **0.85×** at 2048×2048 / 6400 rows. Across ~30 paired measurements the ratio never exceeded 1.0.

Suggestion: gate on measured throughput instead of on whether the kernel loads — at `warmup()`, time `int8_gemm_raw` against the stock op on one registered shape and enable only past a margin. "Does it load" separates M1 from M3; "is it faster" separates M3 from M5, and only the second is the property the module trades quality for. Device-string gating is fragile — the macOS 26 runner reports `air64_v27`, not an `applegpu_*` family.

## `MAX_OUT` misses in both directions

`MAX_OUT` is `n > 32768`, so every vocab ≤ 32768 slips through. On TinyLlama quantized 4-bit g64:

```
registered=112 skipped=44
REGISTERED head/embedding-like modules: ['lm_head', 'model.embed_tokens']
lm_head: N=32000 K=2048 -> eligible=(True, None)   MAX_OUT=32768
```

Same for Llama-2/CodeLlama/Mistral v0.1–v0.2 (32000) and Mistral-7B-Instruct-v0.3 (exactly 32768). This is the case the comment says the constant exists to prevent, and prefill needs one row of those logits.

It also fires as a false positive the other way: `Molmo-7B-D-0924-4bit` has `intermediate_size=37888`, so a genuine MLP up-projection is rejected as `N=37888 above 32768 (probably an lm_head)`.

## `self_test` uses data unlike real activations

`mx.random.normal` has max|x|/rms ≈ **5.3**. Real `down_proj` input (post-SwiGLU) on TinyLlama is **64–69**, and that is the one projection where real activations are worse than the synthetic ones:

| projection | max\|x\|/rms | real-act err | gaussian err |
| --- | --- | --- | --- |
| layers.5.mlp.**down_proj** | 63.8 | 4.55e-2 | 1.67e-2 (**2.7×**) |
| layers.15.mlp.**down_proj** | 68.8 | 6.49e-2 | 3.72e-2 (**1.7×**) |
| layers.15.mlp.gate_proj | 39.8 | 1.27e-2 | 1.37e-2 (0.9×) |
| layers.5.self_attn.q_proj | 44.6 | 1.62e-2 | 1.73e-2 (0.9×) |

Activation outliers aren't mentioned in the error analysis, which covers weight-side headroom thoroughly. End to end on 1536 real tokens, top-1 argmax agreement vs stock: **95.90%** at the default warmup, 96.35% minus lm_head, 96.48% minus down_proj, **97.85%** minus both. Perplexity moves by noise either way (10.8775 → 10.8759), so it's the wrong metric here.

Note `scope="mlp"` points the wrong way — `down_proj` lives in the MLP, so MLP-only is the worst subset, not the conservative one.

## Two smaller things

`enable()` does not call `self_test()`, despite "A self-test at `enable()`" in the description — and it can't, since `enable()` precedes `warmup()` and the registry is empty. `warmup()` doesn't call it either, so the stated mitigation for the untested GEMM only runs if the user types it.

Coverage, across 49 distinct affine-quantized configs: gate/up eligible 24/49, down 27/49, and 22/49 get nothing. Main reasons are `bits != 4` (8-bit off by default; 2/3/5/6-bit unsupported) and non-affine mxfp4/mxfp8/nvfp4 — plus `N % 128` biting real models (`N=10944` on DeepSeek-V2-Lite, `N=2112` on gemma-4-26b-a4b).

Happy to open the bit-exact GEMM sweep as a test file if useful — it runs anywhere the probe passes.

PR #1103, opened 2026-08-25
Do not let a heterogeneous config kill the dtype walk

danielhanchen:
## The bug

`unsloth/gemma-4-E2B-it` does not load at all on transformers 5.15.1. Measured on a Tesla T4, before a single weight is touched:

```
File "unsloth_zoo/patching_utils.py", line 467, in __fix_dtype
    __fix_dtype(getattr(config, key, None))
File "transformers/integrations/heterogeneity/configuration_utils.py", line 298
    raise AmbiguousGlobalPerLayerAttributeError(
AmbiguousGlobalPerLayerAttributeError: 'head_dim' is a per-layer attribute and
may vary across layers.
```

`__fix_dtype` walks every key `config.to_dict()` reports, looking for nested configs that carry a dtype. From transformers 5.15 a MatFormer or other heterogeneous config raises straight out of `PretrainedConfig.__getattribute__` for any attribute that varies per layer, so a key `to_dict()` itself just listed cannot be read back.

The code looks defensive and is not. `getattr(config, key, None)` has a default, and a default only suppresses `AttributeError`. `AmbiguousGlobalPerLayerAttributeError` is not one, so it propagates through the entire load.

## The fix

Guard the read and skip the key. That is safe because of what the walk is for: it descends looking for nested config objects that might carry a dtype, and a per-layer scalar like `head_dim` is never one, so an unreadable key has nothing to contribute either way.

The `try` wraps the read only, with the recursion outside it. A `try` around both would swallow errors raised deeper in the walk and turn a real failure into a silent skip, which is the same bug in a better disguise.

## Tests

Five, fully synthetic and CPU-only, since the failure is a property of attribute access and needs neither gemma-4 nor a GPU nor transformers 5.15.

- The first reproduces the crash against the **unfixed** walk, so the fix is measured against a failure that genuinely happens rather than against nothing.
- One asserts the dtype is still corrected. Surviving while silently skipping the dtype would be the same class of bug: the load succeeds and the model carries bfloat16 on a card that has none.
- One asserts a nested config is still reached, which is what a too-wide fix would break.
- One reads the shipped source and asserts the `try` covers the read and not the recursion.

Both mutations caught: reverting the fix, and widening the `try` to cover the recursion.

## Provenance

Found by the Kaggle T4 CI work, which tried to add a `gemma-4-E2B-it` training leg and could not get past the model load.

### DIFF
--- a/tests/test_heterogeneous_config_dtype_walk.py
+++ b/tests/test_heterogeneous_config_dtype_walk.py
@@ -1,0 +1,166 @@
+# SPDX-License-Identifier: AGPL-3.0-only
+# Copyright 2026-present the Unsloth AI Inc. team. All rights reserved.
+
+"""A heterogeneous config must not kill the dtype walk in patch_model_and_tokenizer.
+
+Observed on real hardware, not constructed here: `unsloth/gemma-4-E2B-it` on
+transformers 5.15.1 with unsloth 2026.8.20 fails to load AT ALL, on a Tesla T4,
+before a single weight is touched:
+
+    File "unsloth_zoo/patching_utils.py", line 467, in __fix_dtype
+        __fix_dtype(getattr(config, key, None))
+    File "transformers/integrations/heterogeneity/configuration_utils.py", line 298
+        raise AmbiguousGlobalPerLayerAttributeError(
+    AmbiguousGlobalPerLayerAttributeError: 'head_dim' is a per-layer attribute
+    and may vary across layers.
+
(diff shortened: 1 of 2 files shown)


danielhanchen:
Confirmed patching_utils.py line 467 on main still reads keys back with a bare getattr default, which does not cover the per-layer attribute error, so the whole load dies before any weight is touched. Will get this reviewed.

PR #1105, opened 2026-08-26
Restrict builtins and imports for model generated RL code

danielhanchen:
`create_locked_down_function` is what executes the Python our GRPO and RL game notebooks get back from the model. It is the one place where a reward hacking or otherwise bad completion turns into code running on the user's Colab or Kaggle box, and right now it does not actually restrict anything.

## The problem

Three separate things line up badly:

**1. The policy default is permissive.** `create_locked_down_function` calls `load_single_function(function)` with no keyword arguments, so it takes the default `builtins_policy = "full"`, which maps to the real `builtins.__dict__`. Generated code can reach `__import__`, `open`, `eval`, `exec`, `compile` and `getattr`.

**2. The lock down line is worse than a no-op.**

```python
f = types.FunctionType(f.__code__, {})
```

CPython injects the real builtins module into any globals mapping that has no `__builtins__` key, so binding to an empty dict hands everything straight back. This also means the `allowlist` branch already present in the file could never have worked even if something selected it, and nothing ever did.

**3. `check_python_modules` is not a security check.** It exists to stop reward hacking via numpy and torch. It accepts `os`, `subprocess`, `socket`, `shutil` and `ctypes` because those are all stdlib, and it only walks `Import` and `ImportFrom` nodes, so `__import__("os")` is invisible to it. Its `ok` return value is also computed and then discarded at every call site in the notebooks.

The result is that a completion like this passes every check we have and runs:

```python
def matmul(A, B):
    import os
    os.system("...")
```

I confirmed this against the current code before changing anything. `import os` inside a generated function returned a live pid, and `__import__("socket").gethostname()` returned the hostname.

## The fix

- **`_safe_import`**, a replacement for `__import__` that resolves only the 23 modules in `_SAFE_IMPORT_MODULES`. This part matters for compatibility. Simply dropping `__import__` would break the 2048 notebook's own documented sample, which does `import math` and `from typing import Callable`. Those still work; `os` and `subprocess` do not.
- **Extended allowlist builtins**, covering what the notebook prompts actually ask the model for (list and numeric work, comprehensions, sorting, `try`/`except`) while leaving out every capability primitive.
- **`_reject_dunder_access`.** Restricting builtins on its own does not stop `().__class__.__bases__[0].__subclasses__()`, which walks from any literal to `subprocess.Popen` using nothing but attribute access. Generated matmul and strategy code has no reason to touch dunders.
- **Globals rebound onto the allowlist** instead of `{}`, so the restriction survives the call.
- **`create_locked_down_function` now defaults to the allowlist policy.** Pass `builtins_policy = "full"` to restore the previous behaviour. Existing positional calls keep working.

## Compatibility

No user facing behaviour changes. The two in tree importers of `create_locked_down_function`, `unsloth/_gpu_init.py` and `unsloth/models/_utils.py`, are re-exports rather than call sites, so nothing internal depended on the permissive default.

Legitimate generated code is unaffected: the notebook prompts ask for self contained functions using builtins plus at most math and typing, and all of that still runs. Code that reaches for numpy or torch now fails with `ImportError`, which the reward functions already catch and score, rather than crashing training.

## Verification

Checked on CPU, no GPU and no training run required.

Blocked:

| Case | Result |
| --- | --- |
| `import os` then `os.system` | `ImportError: Import of 'os' is not allowed` |
| `__import__("os").listdir(".")` | `RuntimeError: Name '__import__' is not allowed` |
| `(1).__class__.__bases__[0].__subclasses__()` | `RuntimeError: Attribute '__subclasses__' is not allowed` |
| `open("/etc/passwd")` | `NameError: name 'open' is not defined` |
| `eval("1+1")` | `NameError: name 'eval' is not defined` |
| `import subprocess` | `ImportError` |
| `import socket` | `ImportError` |
| `getattr(A, "__class__")` | `NameError: name 'getattr' is not defined` |
| `import numpy` | `ImportError`, scored by the reward function as before |

Still working:

| Case | Result |
| --- | --- |
| 2048 sample, `import math` and `from typing import Callable` | returns `"W"` |
| The matmul sample embedded in the gpt-oss GRPO notebook | returns `[[19, 22], [43, 50]]` |
| Sudoku sample, pure builtins | returns `(0, 0, 1)` |
| `sorted` / `set` / `try`-`except` strategy | returns `3` |

## Scope

This is defence in depth, not a real sandbox, and the docstring now says so plainly. It does not bound CPU, memory or recursion. Genuinely untrusted code still belongs in a container.

The companion notebooks change is unslothai/notebooks, which switches the four gpt-oss matmul GRPO notebooks off their own inline copy of this helper and onto this one. The 2048 and Sudoku notebooks already import from here, so they pick this up with no notebook change at all.

### DIFF
--- a/tests/test_rl_locked_down_function.py
+++ b/tests/test_rl_locked_down_function.py
@@ -29,6 +29,13 @@ BLOCKED = [
     # The dunder is a string constant here, so the AST check does not see it.
     # It fails anyway because getattr is not in the allowlist at all.
     ("getattr", 'def matmul(A, B):\n    return getattr(A, "__class__")\n', NameError),
+    # Allowlisted modules re-export unsafe ones: random keeps a private handle
+    # on os, and typing keeps sys. The facade has to deny both.
+    ("random._os", 'def matmul(A, B):\n    import random\n    return random._os.system("echo pwned")\n', AttributeError),
+    ("typing.sys", 'def matmul(A, B):\n    import typing\n    return typing.sys.modules["os"].getpid()\n', AttributeError),
+    # attrgetter takes a dotted string, so it walks dunders without the AST
+    # check ever seeing them. operator is therefore not allowlisted.
+    ("operator", 'def matmul(A, B):\n    import operator\n    return operator.attrgetter("__class__")(1)\n', ImportError),
 ]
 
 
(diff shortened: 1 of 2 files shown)


danielhanchen:
Confirmed this hits rl_environments.py:563, where binding generated code to an empty globals dict lets CPython hand the real builtins back, so the lock down never applied and import os still ran. Will get this reviewed.

PR #1109, opened 2026-08-26
Stop every pull request writing a pip cache nobody can read

danielhanchen:
Stacked on #1107 (base is `cache-janitor-20gib-ceiling`, not `main`). The diff here is only this change; merge #1107 first.

## The problem

`actions/setup-python`'s `cache: 'pip'` is read-write, and it saves from its own post-step on **whatever ref the job ran on**. A cache written on a `pull_request` ref is restorable only by re-runs of that same pull request, because caches created on the base branch are available to the topic branch and not the other way round.

So every PR writes a private copy that nobody else can ever read, and that copy competes for the repo's 20 GiB budget against main's copy, which every PR **can** read.

Nothing about it fails. Over quota GitHub deletes whole entries by last access date, so main's copy goes, the next PR misses, downloads, and writes its own copy. CI gets slower and that reads as the cost of CI. Merged PR #1084 alone held **6.6 GiB** this way, against a repo sitting at 19.97 of 20 GiB.

The closed-PR sweep added in #1107 deletes those entries after the fact. This is the source.

## The change

Restore on every ref, save on `refs/heads/main` only. A PR still restores from main's entry, it just stops writing one. Ported from `unslothai/unsloth`'s `pip-cache-restore` / `pip-cache-save` pair, with two deliberate differences.

**`name` is part of the key, and required.** In unsloth every job hashes its own dependency files into `pip-<os>-<arch>-py<ver>-<hash>`, so five concurrently-live caches share one prefix and no janitor can rank generations without deleting a live one. That is exactly why ~14 GiB of superseded pip entries accumulated there with nothing able to touch them. Naming the job makes each family its own prefix, so `cache-janitor.yml` can prune it the way it prunes everything else.

**The restore takes a prefix fallback.** unsloth's does not, and for its `frontend-dist` cache that is correct: a dist cache is a build output, so the wrong generation is wrong. This is pip's HTTP cache, addressed by URL and content hash, where the previous generation is the current one minus whatever moved. A dependency bump then costs the changed wheels instead of all of them, and a stale entry cannot serve wrong content. It is the same argument that already lets the key carry `3.12` rather than `3.12.14`.

For the same reason most jobs key on `pyproject.toml` alone even where inline pins exist. Adding the workflow file would roll the cache on every edit to it, comments included, and buy nothing: the hash exists to roll the entry forward so stale wheels do not accumulate, not to protect correctness. The three jobs whose deps are named nowhere else (`lint`, `pip-scan`, `studio-export-fix`) do key on their own file, and their entries are small.

## Call sites

Nine, across seven workflows:

| workflow | job | `name` | key-files |
| --- | --- | --- | --- |
| consolidated-tests-ci | python-version-collect | `collect` | pyproject.toml |
| consolidated-tests-ci | repo-tests-cpu | `repo-tests` | pyproject.toml |
| consolidated-tests-ci | mlx-cpu-linux | `mlx-lanes` | pyproject.toml |
| consolidated-tests-ci | core-upstream-matrix | `core-drift` | pyproject.toml |
| gemma4-audio-probe | probe | `gemma4-audio` | pyproject.toml |
| mlx-ci | mlx-smoke | `mlx` | pyproject.toml |
| lint-ci | source-lint | `lint` | (its own file) |
| security-audit | pip-scan-packages | `pip-scan` | (its own file) |
| studio-export-fix-ci | studio-export-fix | `studio-export-fix` | (its own file) |

Each save sits **directly after the step that installs**, not at the end of the job. A composite step with no `if:` is skipped once an earlier step has failed, so the save action's own `always()` only covers what sits between the pair. Putting it at the end would silently discard a completed install whenever a later test failed.

## Test

`tests/test_pip_cache_actions.py` (39 cases) pins the contract: no workflow uses the built-in cache, every restore has exactly one save after it, at most one step sits between them, `name` is well-formed and unique across all jobs, every `key-files` entry resolves, and the save stays gated to main with `always()`.

That last one is not tidiness. The grep for `cache: 'pip'` missed `studio-export-fix-ci.yml`, which spells it unquoted as `cache: pip`. The YAML-aware assertion is what found it.

### DIFF
--- a/.github/actions/pip-cache-restore/action.yml
+++ b/.github/actions/pip-cache-restore/action.yml
@@ -1,0 +1,137 @@
+# SPDX-License-Identifier: AGPL-3.0-only
+# Copyright 2026-present the Unsloth AI Inc. team. All rights reserved.
+
+# Restore half of the pip cache, replacing `actions/setup-python`'s built-in
+# `cache: 'pip'`.
+#
+# The built-in cache is read-write and saves from its own post-step on WHATEVER
+# REF the job ran on. A cache written on a pull_request ref can only be restored
+# by re-runs of that same pull request -- caches created on the base branch are
+# available to the topic branch, but not the other way round -- so every PR
+# writes a copy nobody else can ever read, and that copy competes for the 20 GiB
+# budget against main's copy, which every PR CAN read.
+#
+# Measured on this repo 2026-08-26: 19.97 of 20 GiB, of which merged PR #1084
+# alone held 6.6 GiB on refs/pull/1084/merge. Nothing about that is visible as a
+# failure. Over quota GitHub deletes whole entries by last access date, so main's
+# copy goes, the next PR misses, downloads, and writes its own copy. CI just gets
+# slower and everyone assumes that is the cost.
+#
+# Deleting those entries after the fact -- which cache-janitor.yml now does --
+# treats the symptom. This is the source: restore on every ref, save on main.
+#
+# Adapted from unslothai/unsloth's action of the same name, with one change:
+# `name` is part of the key. There, every job hashes its own dependency files
+# into `pip-<os>-<arch>-py<ver>-<hash>`, so five concurrently-live caches share
+# one prefix and no janitor can rank generations without deleting a live one.
+# Naming the job makes each family a distinct prefix, hence prunable.
+#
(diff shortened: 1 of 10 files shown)


danielhanchen:
@codex review

PR #1112, opened 2026-08-27
Keep the device constants reachable under UNSLOTH_ZOO_DISABLE_GPU_INIT

danielhanchen:
## The lane was not just red, it was missing 43 tests

`Security audit` on main fails, and the failure is a **collection error**, not an assertion:

```
ERROR collecting tests/security/test_compile_model_type_sink_guard.py
unsloth_zoo/compiler.py:58: in <module>
    from . import DEVICE_TYPE
E   ImportError: cannot import name 'DEVICE_TYPE' from 'unsloth_zoo'
```

pytest reports that as one red. What it actually means is that a whole security test module never ran. Those 43 tests are the ones guarding `unsloth_compile_transformers` and `create_new_function` against a `model_type` string reaching an import path or a file write outside the compiled cache, added in #1108. They have not executed on CI since.

## Why the name is missing

`security-audit.yml` sets `UNSLOTH_ZOO_DISABLE_GPU_INIT=1` (added in #1089) so `import unsloth_zoo` skips the device init that ends in `find_spec("unsloth") is None -> ImportError`, which is the only reason a supply-chain-hygiene job would need `unsloth` installed. That reasoning is still correct and the flag stays.

The problem is what the skip branch leaves behind. `__init__.py` has three paths:

| path | `DEVICE_TYPE` and friends |
| --- | --- |
| MLX | set eagerly, all four |
| normal | imported from `.device_type`, all four |
| `UNSLOTH_ZOO_DISABLE_GPU_INIT=1` | **none of them** |

So the skip left the package half-defined. It was meant to defer work, not to remove names. The flag's own comment promises the stubbed imports resolve to "a loud no-op instead of a hard ImportError", and that promise was kept for triton and bitsandbytes but not for the device constants.

Only four names are read off the package root by other modules (`DEVICE_TYPE`, `DEVICE_TYPE_TORCH`, `DEVICE_COUNT`, `ALLOW_PREQUANTIZED_MODELS`), so the fix is narrow.

## Two obvious fixes that are both wrong

**Import `.device_type` eagerly in the skip branch.** Measured on this tree:

```
import unsloth_zoo (skip path):   22 ms,  torch loaded: False
+ from .device_type import ...  1382 ms,  torch loaded: True
```

That is precisely the work the flag exists to skip for the download-only `hf_xet_fallback` child, which never reads one of these constants. So the resolver is lazy (PEP 562 module `__getattr__`), and import stays at ~20 ms for that child.

**Define `DEVICE_TYPE = "cpu"`.** `compiler.py` branches on `cuda`, `hip` and `xpu` and has no cpu arm, so a "cpu" value would fall through all three silently. This repo already has a convention for a driverless host: `UNSLOTH_ALLOW_CPU=1`, which `get_device_type` honours by returning the `"cuda"` sentinel, and which `tests/conftest.py` sets at module scope before collection. Nothing new is needed and the workflow is unchanged.

## The change

`unsloth_zoo/__init__.py` gains a module-level `__getattr__` that resolves those four names from `.device_type` on first read and caches into `globals()`. PEP 562 only consults it when normal lookup fails, so on the MLX and normal paths, where all four are real globals, it never runs and nothing about those paths changes. Anything not in the set still raises `AttributeError`, so it does not become a catch-all that hides typos.

## Tests

`tests/security/test_disable_gpu_init_device_constants.py`, 8 tests, all in subprocesses since both properties are about what happens during `import unsloth_zoo`.

The laziness pair is the point. `test_importing_the_package_does_not_pull_in_torch` would pass vacuously on its own, so it is paired with `test_reading_a_constant_is_what_pulls_torch_in`. Together they fail if anyone ever replaces the resolver with a top-level `from .device_type import ...`, which would satisfy every other test here while silently undoing the reason the flag exists.

## Evidence

Run driverless (`CUDA_VISIBLE_DEVICES=""`) with the flag set, as the runner does.

```
tests/security/test_compile_model_type_sink_guard.py    43 passed
tests/security/test_disable_gpu_init_device_constants.py 8 passed
tests/security (whole lane)                            103 passed in 366s
```

Mutation-tested rather than trusted. With the `__init__.py` change reverted from a byte snapshot:

```
ERROR tests/security/test_compile_model_type_sink_guard.py
Interrupted: 1 error during collection
6 failed, 2 passed
```

and restored (md5 verified identical to the snapshot) all 8 pass again. The 2 that pass under revert are the laziness check, vacuously true when nothing resolves at all, and the `AttributeError` check, which the default lookup satisfies. That is exactly the hole the paired positive test covers.

## Not in scope

The other main red, `Core drift (3 upstream pins)` on `Tests CI`, is unrelated (transformers 5.16.1 moved `mxfp4` symbols) and is being handled separately.

### DIFF
--- a/tests/security/test_disable_gpu_init_device_constants.py
+++ b/tests/security/test_disable_gpu_init_device_constants.py
@@ -46,6 +46,13 @@ def _public_upper_names(body) -> set:
 
     Descends into nested `if`s, because those still bind at module level when
     the branch is taken, but not into `def`/`class`, whose locals do not.
+
+    Annotated assignment counts. `device_type.py` declares these very constants
+    as `DEVICE_TYPE : str = get_device_type()`, so a fifth one arriving in the
+    MLX branch in that style is the likely shape, and reading only `ast.Assign`
+    would miss it, shrink the intersection back to four and let this file pass
+    while the skip path is short a name. Unpacked targets count for the same
+    reason. A bare `X: int` with no value only annotates and binds nothing.
     """
     found = set()
     stack = list(body)
@@ -53,11 +60,16 @@ def _public_upper_names(body) -> set:
         node = stack.pop()
         if isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef, ast.ClassDef)):
             continue
+        targets = ()
         if isinstance(node, ast.Assign):
-            for target in node.targets:
-                if isinstance(target, ast.Name) and target.id.isupper():
-                    found.add(target.id)
-        elif isinstance(node, ast.ImportFrom):
+            targets = node.targets
+        elif isinstance(node, ast.AnnAssign) and node.value is not None:
+            targets = (node.target,)
+        for target in targets:
+            for leaf in ast.walk(target):
+                if isinstance(leaf, ast.Name) and leaf.id.isupper():
+                    found.add(leaf.id)
+        if isinstance(node, ast.ImportFrom):
             for alias in node.names:
                 name = alias.asname or alias.name
                 if name.isupper():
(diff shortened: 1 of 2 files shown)


danielhanchen:
@codex review

danielhanchen:
@codex review

PR #1115, opened 2026-08-27
Drop stale line numbers from the find_spec("unsloth") guard comments

danielhanchen:
Four CI comments point at `unsloth_zoo/__init__.py:128` as the `find_spec("unsloth")` guard that a `--no-deps unsloth` install exists to satisfy.

The guard is at line **250**. Line 128 is unrelated: `UNSLOTH_ENABLE_LOGGING` / Triton env setup. The citations predate #1112 and #1114 and are wrong in all four places.

Rather than renumber them, this cites the file and drops the line number, so the comment cannot rot the next time the module shifts.

Comment-only. No workflow logic changes; both files still parse as YAML.

Sites fixed:
- `.github/workflows/consolidated-tests-ci.yml` (3 places: the drift lane installer, the repo-tests installer, and the `UNSLOTH_IS_PRESENT` handshake note)
- `.github/workflows/wheel-smoke.yml` (1 place: the import-smoke pivot note)

### DIFF
--- a/.github/workflows/consolidated-tests-ci.yml
+++ b/.github/workflows/consolidated-tests-ci.yml
@@ -131,7 +131,7 @@ jobs:
           }
 
           # CPU index avoids the multi-GB CUDA wheel set. `--no-deps unsloth`
-          # satisfies the find_spec("unsloth") guard at unsloth_zoo/__init__.py:128.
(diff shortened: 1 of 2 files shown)


danielhanchen:
@codex review

PR #1118, opened 2026-08-27
Close two mxfp4 detector holes, and fix the corruption they hid on transformers 5.0.0

danielhanchen:
Two commits, deliberately separable. `403df55b` is the detector work and is pin-independent; `2796ba13` is the 5.0.0 code fix. Dropping the second leaves a coherent tree.

## The detector holes (the part that matters most)

Both would let the **next** upstream arity change through silently, on versions squarely inside the supported set.

**1. The signature detector could not fail.** `test_mxfp4_dequantize_convertops_signature` used `_assert_params_superset(fn, required=["blocks", "scales"])`. A 3-arg upstream **is** a superset of a 2-arg requirement, so the exact property `can_safely_patch` enforces sailed straight through the test meant to police it. Demonstrated against real 5.0.0, not a synthetic: old assertion GREEN where the patch is in fact being refused.

**2. The layout probe crashed instead of measuring.** It hardcoded a 2-positional call, so on 5.0.0 it died capturing the golden with `TypeError: dequantize_convertops() missing 1 required positional argument`. Right colour, wrong reason, and it told you nothing about the layout. It now passes `len(inspect.signature(fn).parameters)` positionals:

```
old probe + old zoo -> TypeError (crashes before measuring)
new probe + old zoo -> FAIL: loader got (2, 4, 96), stock gives (2, 96, 4)
new probe + new zoo -> pass
```

**3. `test_mxfp4_shard_and_distribute_module_present` on transformers 5.10.0.** The test called `pytest.fail` unconditionally while the code it guards is version-gated, reddening a lane no change to zoo could ever green. That is how hole 1 survived. It is now `xfail` on exactly 5.10.0 and a hard failure everywhere else.

Not a skip, and not a widened gate. Aligning the gate with `gpt_oss.py`'s `< 5.16.0` does not work, because 5.10.0 sits inside that range. And this is not drift: upstream moved the module to `transformers/distributed/tensor_parallel.py`, deleted `shard_and_distribute_module` at that tag, then restored the `integrations` layout in 5.10.1 fourteen minutes later, and PyPI has yanked 5.10.0. A drift detector firing on a retracted artifact is a false positive. `xfail` keeps it in the report and flips to XPASS if the module returns; a skip would read "not applicable," which is false.

## The corruption on 5.0.0

The existing comment said 5.0.0's 3-arg `dequantize_convertops` was "excluded by pyproject." That reads as safe, and it is not. `hasattr` is true, so zoo attempts the 2-arg patch, `can_safely_patch` refuses it (`Parameter count mismatch: 3 vs 2`, logged only under `UNSLOTH_ENABLE_LOGGING`), and `convert_moe_packed_tensors` has already been swapped for the un-transposed version. Nothing restores `transpose(1, 2)`, so GPT-OSS loads with dims 1 and 2 swapped. The pin never protected the code path; it only stops a default resolve.

The fix dispatches on the **observed parameter names**, not on `transformers_version`. Arity is the exact property `can_safely_patch` enforces, so reading it directly cannot disagree with it, whereas a version gate is a proxy a backport or fork can falsify. An unrecognised signature falls through to the 2-arg arm and is rejected loudly.

The 5.0.0 arm mirrors upstream's own body (verified against the `v5.0.0` tag: `empty_cache` when `target_device == "cpu"`, result placed on `target_device`) with the transpose added back before placement.

## Measured, nine versions

Layout tuples from an arity-agnostic probe, golden captured pre-patch in-process on the same input.

| transformers | arity | stock | before | after | drift file (after) |
|---|---|---|---|---|---|
| 4.57.6 | absent | n/a | n/a | n/a | 101 passed, 18 skipped |
| **5.0.0** | **3** | (2, 96, 4) | **(2, 4, 96)** | **(2, 96, 4)** | 97 passed, 22 skipped |
| 5.1.0 | 2 | (2, 96, 4) | (2, 96, 4) | (2, 96, 4) | 97 passed, 22 skipped |
| 5.5.0 | 2 | (2, 96, 4) | (2, 96, 4) | (2, 96, 4) | 106 passed, 13 skipped |
| 5.9.0 | 2 | (2, 96, 4) | (2, 96, 4) | (2, 96, 4) | 106 passed, 13 skipped |
| 5.10.0 | 2 | (2, 96, 4) | (2, 96, 4) | (2, 96, 4) | 105 passed, 13 skipped, 1 xfailed |
| 5.10.1 | 2 | (2, 96, 4) | (2, 96, 4) | (2, 96, 4) | 106 passed, 13 skipped |
| 5.16.0 | 2 | (2, 96, 4) | (2, 96, 4) | (2, 96, 4) | 103 passed, 16 skipped |
| 5.16.1 | 2 | (2, 96, 4) | (2, 96, 4) | (2, 96, 4) | 103 passed, 16 skipped |

Signature census: 5.0.0 alone at 3 params, the other seven at 2. No collateral failures on any version.

**Mutation-verified.** The first commit alone is the mutation: detectors without the code fix, swept over all nine versions, green everywhere and on 5.0.0 both fire against genuine upstream (`arity=3 golden=(2, 96, 4) after=(2, 4, 96) dqc_patched=False verdict=CORRUPT_SHAPE` plus `FAILED test_mxfp4_dequantize_convertops_signature`). The tightened signature assertion also goes red if upstream renames the third parameter or grows a fourth, and stays green on both real signatures.

## One correction to an earlier claim

I previously described the 5.10.0 problem as test-only. It is not, though it is conditional. `return raise_error(...)` sits at line 466 **inside** `patch_gpt_oss` (114-577), so it abandons the rest of that function and loses two patches, `load_and_swizzle_mxfp4` and `replace_with_mxfp4_linear`. But `patch_gpt_oss` returns unconditionally at line 156 when `triton_kernels` is absent (traced: `last_line_executed=156`, `reached_tp_import=False`), so the loss only materialises on real GPT-OSS MXFP4 GPU setups.

That mid-function `return raise_error(...)` pattern, where one failed optional import silently abandons every later patch, is worth its own issue. It is not specific to mxfp4 and fixing it changes behaviour on many versions, so it is out of scope here.

## Pins

Unchanged, deliberately. Recording the verification only: `!=5.1.0` is unnecessary (5.1.0 measured clean), the `<=5.5.0` cap is stale across these nine (5.9.0 through 5.16.1 all clean), and `!=5.10.0` is not needed since the PyPI yank already stops resolvers. Note the constraint string is duplicated at `pyproject.toml` lines 55 and 100.

### DIFF
--- a/tests/test_mxfp4_load_path_layout.py
+++ b/tests/test_mxfp4_load_path_layout.py
@@ -21,9 +21,9 @@ Zoo's `convert_moe_packed_tensors` replacement returns the UN-transposed
 [E, G*B*2, D]:
 
   * 4.x             -> module level `mxfp4.dequantize`, called by quantizer_mxfp4
-  * 5.1.0 and newer -> `Mxfp4Dequantize` (a ConversionOps) -> `dequantize_convertops`
+  * 5.0.0 and newer -> `Mxfp4Dequantize` (a ConversionOps) -> `dequantize_convertops`
 
-Zoo only ever patched `dequantize`, so from 5.1.0 the transpose was dropped and
+Zoo only ever patched `dequantize`, so from 5.0.0 the transpose was dropped and
 GPT-OSS loaded with dims 1 and 2 swapped; 5.16.0 deleting the function is what
 finally turned the drift detector red. The existing mxfp4 tests missed it because
 they only exercise the SAVE path, which stayed self-consistent. So assert the
@@ -45,6 +45,7 @@ import pytest
 # against the installed transformers rather than a hardcoded, version-gated shape.
 _PROBE = textwrap.dedent(
     """
+    import inspect
     import torch
     import transformers
     import transformers.integrations.mxfp4 as m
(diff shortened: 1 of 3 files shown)


danielhanchen:
@codex review

PR #1127, opened 2026-08-28
Stop mlx-lm's ArraysCache.advance() stranding a Metal buffer per call in batched generation

Lyxot:
## The bug

`model.fast_generate` on an MLX model with linear-attention layers leaks one live Metal buffer per layer per decode token. A long enough batched generation terminates the process:

```text
RuntimeError: [metal::malloc] Resource limit (499000) exceeded.
```

Measured on `Qwen/Qwen3.5-0.8B` (24 layers: 18 GatedDeltaNet + 6 full attention) at batch 2, death arrives at decode token 29,316. A 48-layer `qwen3_next` (36 linear layers) would die near 14.3k tokens. The bytes are trivial — 4 B per call — but the buffer *count* is what trips Metal's resource limit.

The worst case is a GRPO-style rollout: N identical prompts in one `fast_generate` call.

## Root cause

`ArraysCache.advance` in mlx-lm 0.31.3 (`mlx_lm/models/cache.py`):

```python
def advance(self, N):
    if self.lengths is not None:
        self.lengths -= N
    if self.left_padding is not None:
        self.left_padding -= N
```

Each `-=` rebinds to a new **unevaluated** graph node. The Python-int operand is eagerly converted to an `mx.array` scalar that already owns a Metal buffer, and the dead node pins it. `mx.clear_cache()` cannot reclaim it: the buffer is active, not cached.

Nothing collapses that chain during decode. The metadata is absent from `ArraysCache.state`, and `GenerationBatch._step` evaluates only `inputs` and `_current_logprobs`. Only the cache that owns the shared SSM mask has `make_mask` rebuilt each forward, so exactly one cache per model stays evaluated and every other linear-attention layer accumulates. On Qwen3.5-0.8B that is 17 of 18 caches, or 68.3 B — precisely 17 int32 scalars — per decode step.

### Why it hides

The metadata is armed only on the batch path: `_merge_caches` → `ArraysCache.merge` sets `left_padding` when the merged caches are empty. `finalize()` clears it, but `prompt()` only calls `prepare()`/`finalize()` when `max_padding > 0`. So:

| path | affected |
|---|---|
| `model.generate` → `stream_generate` (single stream) | no — metadata never armed |
| batch, **unequal** prompt lengths | no — right padding triggers `finalize()`, which clears it |
| batch, **equal** prompt lengths (including batch-of-1) | **yes** |

That asymmetry is inverted from intuition: ragged batches are safe, identical batches leak. It survives casual testing and shows up in rollouts.

## The change

`generate_batch` installs a replacement `ArraysCache.advance` that accumulates the count in a Python int, plus properties over `lengths` and `left_padding` that fold that count into a single `value - pending` node the first time the field is read. Nothing is allocated per call, and nothing is evaluated inside the forward pass.

Evaluating eagerly in `advance` was the obvious alternative and is worse: it would add one host sync per linear-attention layer per decode token, inside the forward, to keep a 4-byte scalar tidy.

Because `lengths` and `left_padding` are read from outside `cache.py` as well as by `merge`/`extend`/`filter`/`make_mask` inside it, the fold has to happen on read, which is why these become properties rather than the accumulator being kept privately. A subclass would not work either — mlx-lm constructs these caches itself.

### Installation is deliberately conservative

Rewriting a third-party class at runtime is the risk this PR carries, so the install is gated:

- It runs only from `generate_batch`, once per process, under a lock, with the decision published only after it completes. A failed attempt — mlx-lm not importable, candidacy check raising — is retried rather than latched.
- A class is a candidate only if its `advance` matches `_stock_advance` (a reference function carrying the 0.31.3 body, never installed and never called) on **compiled body and signature**: `inspect.signature`, `co_code`, `co_consts`, `co_names`, `co_varnames`. Both are compiled by the running interpreter, so this is exact-operations identity, not a name-table heuristic and not a version string.
- Anything else keeps stock behaviour. The guard fails *toward* stock: an mlx-lm that still leaks but compiles differently is simply not patched, rather than having semantics silently overwritten.

Measured discrimination:

| candidate `advance` | replaced |
|---|---|
| mlx-lm 0.31.3 | yes |
| same body, reformatted, comments added | yes |
| `+=` instead of `-=` | no |
| clamped with `mx.maximum` | no |
| same decrements plus an unrelated side effect | no |
| `def advance(self, N=1)` | no |
| metadata reads already mediated by descriptors | no |

Deciding candidacy never constructs or invokes the candidate.

Also handled: a cache constructed before installation adopts its plain attributes rather than losing them to the new descriptors.

## Scope

- mlx-vlm ships its own `ArraysCache` that already defers advancement, so the vision batch path is untouched.
- `BatchKVCache.offset` looks like the same pattern but does not accumulate — RoPE consumes it each step, which collapses the chain. Confirmed by measurement, below.
- `model.generate` is unaffected and unchanged.

## Validation

mlx 0.32.2, mlx-lm 0.31.3, Apple silicon.

**Buffer-count stress**, one armed cache, `resource_limit` 499000:

```text
before:  !! DIED at advance #498999: RuntimeError: [metal::malloc] Resource limit (499000) exceeded.
after:   # survived 5000000 advances in 0.82s, left_padding=[-5000000]
```

**Real decode**, `Qwen/Qwen3.5-0.8B` through `mlx_lm.generate.BatchGenerator`, B=2, 600 steps, measuring what force-evaluating the metadata afterwards releases — i.e. the size of the pending chain:

```text
before:  released by forced eval = 41076 B (68.5 B/step)
after:   released by forced eval =   272 B (0.5 B/step, constant rather than per-step)
```

Sampled tokens and final `left_padding` are identical between the two runs.

**Public path**, `generate_batch` (the engine behind `model.fast_generate`) with four identical prompts, the rollout shape where `max_padding == 0` and `finalize()` never runs: generated token ids identical with and without the change.

**No other cache on this path accumulates.** Force-evaluating every array-valued metadata field across all 24 caches of a live batch (18 `ArraysCache` + 6 `BatchKVCache`) releases 228 B total after 400 steps — constant, not per-step.

**Equivalence.** A differential run over `left_padding`, `lengths`, `make_mask`, `filter`, `extend` (including two caches carrying different deferred counts), `merge`, `prepare` and `finalize` produces byte-identical output before and after; only memory differs. Dtypes are unchanged.

**Cost.** 20,000 isolated `advance` calls: 0.0996 s before, 0.0034 s after. The replacement is strictly less work — it allocates nothing.

**Tests.** `tests/test_mlx_generate_metal.py` gains coverage for the memory contract, the arithmetic across mlx-lm's own metadata plumbing, pre-patch cache adoption, candidacy in both directions, idempotence, retry after a failed attempt, and the requirement that installation happens before any decoding. Every added assertion was mutation-checked: thirteen mutations of the production change each fail the tests.

## Related

Same `[metal::malloc]` signature as unslothai/unsloth#6990, but a different site — that one was the training loop, this is batched generation.

### DIFF
--- a/.github/workflows/consolidated-tests-ci.yml
+++ b/.github/workflows/consolidated-tests-ci.yml
@@ -325,6 +325,7 @@ jobs:
             tests/test_train_on_responses_list_labels.py \
             tests/test_mlx_preference.py \
             tests/test_mlx_trainer_internals.py \
+            tests/test_mlx_arrays_cache_advance.py \
             tests/test_mlx_quantized_optimizer_routing.py \
             tests/test_check_hf_model_exists_transport_errors.py \
             tests/test_hub_transport_errors_single_segment.py \
--- a/tests/test_mlx_arrays_cache_advance.py
+++ b/tests/test_mlx_arrays_cache_advance.py
@@ -1,0 +1,335 @@
+"""The ArraysCache.advance patch, on any platform.
+
+Everything that decides *whether* to rewrite a third-party class -- the body
+matcher, the install latch, the retry policy, the deferred arithmetic and the
+adoption of caches built before installation -- is plain Python. None of it needs
+mlx, Metal or Apple Silicon, and the companion Metal suite is opt-in behind a
+label, so this is the coverage that runs on every pull request.
+
+The Metal-only assertions (live buffer counts, real decoding) stay in
+tests/test_mlx_generate_metal.py.
+"""
+
+import importlib.util
+import sys
+import types
+
+import pytest
+
+if importlib.util.find_spec("mlx") is None:
+    from mlx_simulation import simulate_mlx_on_torch
+    simulate_mlx_on_torch()
+
+import unsloth_zoo.mlx.generate as engine  # noqa: E402
+
+
+def stock_bodied_cache():
+    """A class carrying mlx-lm 0.31.3's advance, byte for byte."""
+
+    class ArraysCache:
+        lengths = left_padding = None
+
+        def advance(self, N):
+            if self.lengths is not None:
+                self.lengths -= N
+            if self.left_padding is not None:
+                self.left_padding -= N
+
+    return ArraysCache
+
+
+@pytest.fixture
+def fresh(monkeypatch):
+    """Un-latch the installer and let a test aim it at a class it owns."""
+
+    monkeypatch.setattr(engine, "_ARRAYS_CACHE_ADVANCE_RESOLVED", False)
+    monkeypatch.setattr(engine, "_VLM_ARRAYS_CACHE_ADVANCE_RESOLVED", False)
+
+    def aim(cls, module_name="mlx_lm.models.cache"):
+        module = types.ModuleType(module_name)
+        module.ArraysCache = cls
+        monkeypatch.setitem(sys.modules, module_name, module)
+        return cls
+
+    return aim
+
+
+# --------------------------------------------------------------------------
+# The matcher
+# --------------------------------------------------------------------------
+
+
+def test_matcher_accepts_the_body_it_reproduces():
+    assert engine._has_replaceable_advance(stock_bodied_cache())
+
+
+def test_matcher_rejects_the_opposite_sign():
+    class Incrementing:
+        lengths = left_padding = None
+
+        def advance(self, N):
+            if self.lengths is not None:
+                self.lengths += N
+            if self.left_padding is not None:
+                self.left_padding += N
+
+    assert not engine._has_replaceable_advance(Incrementing)
+
+
+def test_matcher_rejects_the_same_decrements_plus_a_side_effect():
+    class SideEffecting:
+        lengths = left_padding = None
+
+        def advance(self, N):
+            self.touched = True
+            if self.lengths is not None:
+                self.lengths -= N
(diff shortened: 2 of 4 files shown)


Lyxot:
@codex review

Lyxot:
Note for maintainers: the new tests in `tests/test_mlx_generate_metal.py` are `metal_only`, so they only execute on the Apple Silicon runner, and `MLX CI on Mac M1` is opt-in via the `mlx` label. I don't have permission to add labels here — if someone adds `mlx`, that job will pick up the new coverage.

Locally on mlx 0.32.2 / mlx-lm 0.31.3: `pytest tests/test_mlx_generate_metal.py tests/test_mlx_generate.py -q` gives 56 passed, plus 2 pre-existing FastVLM failures unrelated to this change (`ModuleNotFoundError: No module named 'timm'`).

danielhanchen:
Confirmed there is nothing handling ArraysCache.advance in unsloth_zoo/mlx/generate.py today, so batched decode on linear-attention MLX models keeps stranding a scalar buffer per layer per step until Metal refuses to allocate. I will sort out the mlx label so the new Metal-only tests actually run, and get this reviewed.

danielhanchen:
@codex review

PR #1132, opened 2026-08-29
Never hand a packed 4-bit weight to the Linear4bit F.linear fallback

lonexreb:
## What

Fixes unslothai/unsloth#9867. Training `unsloth/Qwen3.8-27B-unsloth-bnb-4bit` (hybrid `Qwen3_5ForConditionalGeneration`) crashes on the first forward: the GatedDeltaNet projections (`in_proj_z` etc.) reach the patched `Linear4bit.forward` with a weight that is still the **packed uint8 nf4 buffer** and no `quant_state` on the parameter, so the quant_state-None fallback runs `F.linear` on the packed buffer:

```
RuntimeError: mat1 and mat2 shapes cannot be multiplied (82x5120 and 1x15728640)
```

(`15728640 = 6144 × 5120 / 2` — exactly the packed shape of the (6144, 5120) weight, per the reporter's analysis.) The existing `shape[-1] == 1` probe does not cover every packed layout transformers 5 leaves behind during init, and a uint8 weight can never be a valid plain-linear operand regardless.

## Fix

In the quant_state-None branch, when the weight is uint8: call `fix_4bit_weight_quant_state_from_module` and re-read `quant_state` — the recovered state then routes through `matmul_4bit` exactly like the regular attention/MLP paths. If it is still missing, raise an error naming the layer's actual problem instead of the shape error. Genuinely unquantized (float) layers keep the `F.linear` fallback byte-for-byte.

## Tests

New `tests/test_linear4bit_packed_fallback.py`, with bitsandbytes faked so it runs on CPU-only hosts (house pattern: `test_broken_bitsandbytes_import.py`), using the reporter's exact projection shape:
- packed weight + recoverable state → routes through `matmul_4bit` (**fails on main**: F.linear shape error)
- packed weight + unrecoverable state → named RuntimeError, not the mat1/mat2 error (**fails on main**)
- float weight without quant_state → `F.linear` fallback preserved (passes on main and here)

## Caveat

I don't have an AMD gfx1201 box; the control flow is pinned by the tests above, but the end-to-end confirmation on the reporter's hardware is outstanding — they offered to run patched builds in unslothai/unsloth#9867, and I'll ask them to verify this branch.

### DIFF
--- a/tests/test_linear4bit_packed_fallback.py
+++ b/tests/test_linear4bit_packed_fallback.py
@@ -1,0 +1,122 @@
+# SPDX-License-Identifier: AGPL-3.0-only
+# Copyright 2023-present the Unsloth team. All rights reserved.
+"""The Linear4bit fallback never hands a packed 4-bit weight to F.linear.
+
+Qwen3.5's GatedDeltaNet projections can reach the patched forward with a
+weight that is still the packed uint8 nf4 buffer and no ``quant_state`` on the
+parameter (unsloth#9867). The old fallback ran ``F.linear`` on that buffer and
+died with ``mat1 and mat2 shapes cannot be multiplied (BxH and 1xNUMEL)``. The
+patched fallback must instead re-derive the quant state from the module, and
+when that is impossible, fail with an error naming the real problem.
+
+bitsandbytes is faked (house pattern: test_broken_bitsandbytes_import.py), so
+this runs on CPU-only hosts and pins the control flow, not bnb's kernels.
+"""
+
+import sys
+import types
+
+import pytest
+
+torch = pytest.importorskip("torch")
+
+
+def _install_fake_bitsandbytes(monkeypatch, *, fix_attaches_state):
+    calls = {"fix": 0, "matmul": 0}
+
+    class _QuantState:
+        pass
+
+    class _Params4bit(torch.nn.Parameter):
+        pass
+
+    def fix_4bit_weight_quant_state_from_module(module):
+        calls["fix"] += 1
+        if fix_attaches_state:
+            module.weight.quant_state = _QuantState()
+
+    def matmul_4bit(x, weight, bias = None, quant_state = None):
+        calls["matmul"] += 1
+        # Shape-checked stand-in: the real kernel dequantizes to (out, in).
+        return x.new_zeros(x.shape[0], 6144)
+
+    bnb = types.ModuleType("bitsandbytes")
+    bnb.matmul_4bit = matmul_4bit
+    bnb.nn = types.ModuleType("bitsandbytes.nn")
+    bnb.nn.modules = types.ModuleType("bitsandbytes.nn.modules")
+
+    class Linear4bit(torch.nn.Module):
+        # patch_function replaces an existing forward; signature must match.
+        def forward(self, x: torch.Tensor):
+            raise NotImplementedError
+
+    bnb.nn.modules.Linear4bit = Linear4bit
+    bnb.nn.Linear4bit = Linear4bit
+    bnb.nn.modules.Params4bit = _Params4bit
+    bnb.nn.modules.fix_4bit_weight_quant_state_from_module = (
+        fix_4bit_weight_quant_state_from_module
+    )
+    monkeypatch.setitem(sys.modules, "bitsandbytes", bnb)
+    monkeypatch.setitem(sys.modules, "bitsandbytes.nn", bnb.nn)
+    monkeypatch.setitem(sys.modules, "bitsandbytes.nn.modules", bnb.nn.modules)
+    return bnb, Linear4bit, calls
+
+
+def _patched_forward(monkeypatch, *, fix_attaches_state):
+    bnb, Linear4bit, calls = _install_fake_bitsandbytes(
+        monkeypatch, fix_attaches_state = fix_attaches_state
+    )
+    from unsloth_zoo.temporary_patches import bitsandbytes as patch_module
+
+    patch_module.patch_bitsandbytes_linear4bit_forward()
+    return Linear4bit, calls
+
+
+def _packed_module(Linear4bit):
+    # The reporter's in_proj_z: a (6144, 5120) nf4 weight packs to
+    # (6144 * 5120 / 2, 1) uint8 — F.linear on it is the mat1/mat2 error.
+    module = Linear4bit()
+    module.weight = torch.nn.Parameter(
+        torch.zeros(6144 * 5120 // 2, 1, dtype = torch.uint8), requires_grad = False
+    )
+    module.bias = None
+    module.compute_type_is_set = True
+    module.compute_dtype = torch.float16
+    return module
+
+
+def test_packed_weight_recovers_its_quant_state(monkeypatch):
+    Linear4bit, calls = _patched_forward(monkeypatch, fix_attaches_state = True)
+    module = _packed_module(Linear4bit)
+    # shape[-1] == 1 also triggers the init-time probe, so the packed branch is
+    # exercised with a wide packed layout too, which that probe does not cover.
+    module.weight.data = module.weight.data.reshape(1, -1)
+    x = torch.zeros(82, 5120, dtype = torch.float16)
+    out = module.forward(x)
+    assert calls["fix"] >= 1
+    assert calls["matmul"] == 1, "recovered state must route through matmul_4bit"
+    assert out.shape == (82, 6144)
+
+
+def test_unrecoverable_packed_weight_fails_with_a_named_error(monkeypatch):
+    Linear4bit, calls = _patched_forward(monkeypatch, fix_attaches_state = False)
+    module = _packed_module(Linear4bit)
+    module.weight.data = module.weight.data.reshape(1, -1)
+    x = torch.zeros(82, 5120, dtype = torch.float16)
+    with pytest.raises(RuntimeError) as excinfo:
+        module.forward(x)
+    message = str(excinfo.value)
+    assert "packed uint8 weight" in message
+    assert "quant_state" in message
+    assert "mat1" not in message
+
+
+def test_a_genuinely_unquantized_layer_still_takes_f_linear(monkeypatch):
+    Linear4bit, calls = _patched_forward(monkeypatch, fix_attaches_state = False)
+    module = Linear4bit()
+    module.weight = torch.nn.Parameter(torch.zeros(8, 4, dtype = torch.float32))
+    module.bias = None
+    x = torch.zeros(2, 4, dtype = torch.float32)
+    out = module.forward(x)
+    assert out.shape == (2, 8)
+    assert calls["matmul"] == 0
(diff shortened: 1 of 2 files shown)


danielhanchen:
Confirmed the fallback in unsloth_zoo/temporary_patches/bitsandbytes.py still reaches F.linear when a packed uint8 weight misses the shape[-1] == 1 probe, and the float path is left untouched. Will get this reviewed, and it would help to have the reporter confirm the end to end run on their box.

lonexreb:
Thanks for the confirm. Reporter verification is in motion: both @Awardin (original, ROCm gfx1201) and @rockerBOO (independent CUDA repro of the same signature on `Qwen3.8-27B` + LoRA, per unslothai/unsloth#9867) have just been pinged there with the exact `pip install git+…@fix/gated-deltanet-packed-4bit-fallback` line and their own original repro steps to re-run. Will relay results here as soon as either reports back.

rockerBOO:
Edit: Removed invalid log.  I had it set with `use_exact_model_name=True, ` so it never used the pre-quantized model data. 

When using `load_in_4bits=True` it loads `unsloth/Qwen3.8-27B-unsloth-bnb-4bit` which is where the quantized model problem is. If we load it from the model (using `use_exact_model_name=True, `) it quantizes it fine. So this specific issue might be fixed from updating this model to use the quant. 

Got the similar error Awardin got.

```
  Unsloth: a 4-bit layer holds a packed uint8 weight of shape (15728640, 1) with no quant_state, so it
  cannot be dequantized or matmul'd. Please report the model and layer at
  https://github.com/unslothai/unsloth/issues
```

Awardin:
Ran the end-to-end test on the real gfx1201 hardware as requested (@lonexreb). Result: **the patch engages, but the forward pass still fails** — now with the patch's own clean guard instead of the raw matmul crash.

**Environment**
- AMD Radeon AI PRO R9700 (gfx1201), Windows 11, native (no WSL)
- torch `2.11.0+rocm7.13.0`, HIP `7.13.99004`, bitsandbytes `0.50.2.dev0`
- unsloth-zoo installed from `fix/gated-deltanet-packed-4bit-fallback` (reports `2026.8.16`)
- Model: `unsloth/Qwen3.8-27B-unsloth-bnb-4bit` via `FastLanguageModel` + TRL `SFTTrainer`, 5 steps, seq 512

**Observed**
- Load completes: 20.57 GB VRAM after load (20.62 reserved), 79.7M trainable / 16,531M total params.
- Step 0 forward: two warnings, then the new RuntimeError from the patch:

```
[bitsandbytes.nn.modules|WARNING] FP4 quantization state not initialized. Please call .cuda() or .to(device) on the LinearFP4 layer first.
[bitsandbytes.nn.modules|WARNING] FP4 quantization state not initialized. Please call .cuda() or .to(device) on the LinearFP4 layer first.
RuntimeError: Unsloth: a 4-bit layer holds a packed uint8 weight of shape (15728640, 1) with no quant_state, so it cannot be dequantized or matmul'd.
```

- Failure site: `Qwen3_5GatedDeltaNet_forward` → `z = self.in_proj_z(hidden_states)` → `unsloth_zoo/temporary_patches/bitsandbytes.py:82`.
- VRAM peak at failure: 20.87 GB (no OOM involved).

**Reading:** the old `mat1 and mat2 shapes cannot be multiplied (…x… and 1x15728640)` is gone — the fallback now correctly detects the packed uint8 weight. But on this load path the GatedDeltaNet projections (`in_proj_z`, presumably its siblings too) end up with **no `quant_state` attached at all** (matching the two FP4 "state not initialized" warnings), so there is nothing for `matmul_4bit` to dequantize with. Looks like the quant_state for these modules is lost/never mapped during model load rather than at matmul time.

<details>
<summary>Full traceback</summary>

```
attn(
  File "torch\nn\modules\module.py", line 1779, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "torch\nn\modules\module.py", line 1790, in _call_impl
    return forward_call(*args, **kwargs)
  File "unsloth_compiled_cache\unsloth_compiled_module_qwen3_5.py", line 669, in forward
    return Qwen3_5GatedDeltaNet_forward(self, hidden_states=hidden_states, cache_params=cache_params, attention_mask=attention_mask)
  File "torch\_dynamo\external_utils.py", line 227, in nonrecursive_disable_wrapper
    return fn(*args, **kwargs)
  File "unsloth_compiled_cache\unsloth_compiled_module_qwen3_5.py", line 507, in Qwen3_5GatedDeltaNet_forward
    z = self.in_proj_z(hidden_states)
  File "torch\nn\modules\module.py", line 1779, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "torch\nn\modules\module.py", line 1790, in _call_impl
    return forward_call(*args, **kwargs)
  File "unsloth_zoo\temporary_patches\bitsandbytes.py", line 82, in forward
    raise RuntimeError(
RuntimeError: Unsloth: a 4-bit layer holds a packed uint8 weight of shape (15728640, 1) with no quant_state, so it cannot be dequantized or matmul'd.
```
</details>

Happy to re-run on this box once the quant_state mapping is addressed.

Awardin:
Mapping this onto your test matrix: our real-hardware run lands exactly in your "packed weight + **unrecoverable** state" branch — i.e. `fix_4bit_weight_quant_state_from_module` finds nothing to recover for the GatedDeltaNet projections on an actual gfx1201 load (consistent with the two `FP4 quantization state not initialized` warnings right before the raise). So the PR works as designed — clean, named error instead of the shape crash — but the quant_state going missing for these modules at load time looks like a separate, still-open issue in the loader path. Happy to re-test once that part is addressed.

Awardin:
Follow-up: I can confirm @rockerBOO's finding on the gfx1201 box — **the workaround trains successfully**, which pins the root cause on the pre-quantized checkpoint rather than the code path.

**Setup** (same environment as my earlier report): loaded the original `Qwen/Qwen3.8-27B` with `use_exact_model_name=True` + `load_in_4bit=True`, so bitsandbytes quantizes at load time instead of pulling `unsloth/Qwen3.8-27B-unsloth-bnb-4bit`.

**Result — QLoRA runs end-to-end:**
- Load + on-the-fly nf4 quantization from local disk: **81 s**
- Forward pass through the GatedDeltaNet projections: **no error** (this is exactly where both the raw matmul crash and the patch's named guard fired before)
- 5 SFT steps: loss **3.87 → 1.85**, monotonic, all finite
- VRAM: 16.83 GB after load, **17.99 GB peak** (32 GB card), ~60 s/step at seq 512, bs 1, grad-accum 4
- torch `2.11.0+rocm7.13.0`, HIP `7.13.99004`, bitsandbytes `0.50.2.dev0`, Windows native

**Reading:** when quant_state is created at load time it is present and the (unpatched, release `unsloth_zoo` 2026.8.15) path works fine — so the missing quant_state is a property of the published `unsloth/Qwen3.8-27B-unsloth-bnb-4bit` checkpoint (its GatedDeltaNet projections), not of the runtime. Regenerating that checkpoint would likely fix the default `load_in_4bit=True` path for everyone; this PR's named error remains valuable as a guard for any checkpoint with the same defect.

lonexreb:
Synthesis of the hardware verification (thanks @Awardin — two runs, both decisive):

1. **The patch behaves exactly as designed**: the raw `mat1 and mat2` crash is gone, the packed uint8 weight is detected, re-derivation is attempted, and when nothing can be recovered the named error fires at the true failure site (`in_proj_z`). The two `FP4 quantization state not initialized` warnings confirm the re-derivation had nothing to work with.
2. **The state genuinely does not exist on that path**: the published `unsloth/Qwen3.8-27B-unsloth-bnb-4bit` checkpoint's GatedDeltaNet projections load with no quant_state, while quantize-at-load from the base model trains end-to-end in the identical environment on *release* zoo. Filed the checkpoint regeneration as unslothai/unsloth#10276, since that fix is an artifact rebuild, not code.

So this PR's role settles as the guard: it converts a silent inscrutable crash into an error that names the layer and the cause, for this checkpoint and any future one with the same defect. Two follow-ups I can push here if wanted: (a) extend the error message to name the recovery that works today (`load_in_4bit = True` on the base model, per #10276) so the error is actionable for users, and (b) nothing else — the control-flow tests already pin both branches. Say the word on (a), @danielhanchen.

danielhanchen:
Confirmed the fallback in unsloth_zoo/temporary_patches/bitsandbytes.py still runs F.linear on a packed uint8 weight when the shape[-1] == 1 probe misses, so the recovery branch makes sense. Will get this reviewed.

danielhanchen:
Confirmed the fallback in unsloth_zoo/temporary_patches/bitsandbytes.py still hands a packed uint8 weight straight to F.linear when quant_state is missing, which matches the mat1/mat2 error in the report. Could you follow up with the reporter's end to end run on their hardware so we have that confirmation alongside the tests?

danielhanchen:
Confirmed main's fallback at unsloth_zoo/temporary_patches/bitsandbytes.py still sends a packed uint8 weight straight into F.linear when quant_state is missing, and the shape[-1] == 1 probe does not catch that layout. Could you get the reporter to confirm the branch end to end on their setup before I review?

rockerBOO:
Isn't the issue that the weights on huggingface just need to be remade, at least for this flagged weight? When we do it in process (not using the pre-quantized weights) it runs fine.

This PR tries to fix it but still fails and errors (correctly).

lonexreb:
@danielhanchen the end-to-end run you asked for is already in this thread, twice, from the original reporter on the reporting hardware (Aug 31, gfx1201 / ROCm 7.13 / Windows native):

1. **This branch, `unsloth/Qwen3.8-27B-unsloth-bnb-4bit`, `load_in_4bit = True`:** loads, then at step 0 the raw `mat1 and mat2 shapes cannot be multiplied` crash is gone and the branch's named error fires at the true site (`Qwen3_5GatedDeltaNet_forward → in_proj_z`), after two `FP4 quantization state not initialized` warnings showing re-derivation had nothing to recover. So the guard does exactly what it says.
2. **Release zoo, base `Qwen/Qwen3.8-27B` with `use_exact_model_name = True` + `load_in_4bit = True`:** quantizes at load and trains 5 SFT steps end to end (loss 3.87 → 1.85, 18 GB peak) through the same projections.

@rockerBOO reproduced the same pair on CUDA. Together they settle it: the state is genuinely absent from the published checkpoint's GatedDeltaNet projections, and no runtime patch can invent a quant_state that was never saved. rockerBOO, agreed on both counts: the fix for that model is a rebuild of the artifact (filed as unslothai/unsloth#10276, still unanswered), and this PR "fails correctly": its whole job is to turn the inscrutable shape error into one that names the layer and the cause.

Pushed b85e52db to make that error actionable rather than just named: it now says the checkpoint was likely saved without the state and points at the recovery that works today (load the base model with `load_in_4bit = True`), with the pin asserting that text. Nothing else changed; the float path and the recovery branch are as reviewed.

danielhanchen:
Main now covers this in unsloth_zoo/temporary_patches/bitsandbytes.py: the Linear4bit fallback raises a named error for any packed [N, 1] weight without quant_state instead of reaching F.linear, and it covers the float quant_storage layouts too, so I do not think this is needed any more. Sorry for the slow review, and thanks for chasing this down with the reporters on real hardware.

PR #1133, opened 2026-08-29
Do not report unsloth's own bitsandbytes stub as a test module leak

vineethsaivs:
Part of unslothai/unsloth#8138, though narrower than that issue: this is a false positive in the leak guard itself, not the shim-vs-real-mlx collision the issue is mainly about.

## What happens

`tests/conftest.py`'s `sys.modules` guard reports a leak against the **first test in a session that imports `unsloth_zoo`**, on any host where bitsandbytes is not installed. On an Apple Silicon checkout that is two tests, and the error tells them to fix something they did not do:

```
ERROR tests/test_mlx_batching_and_decay.py::test_compiler_review_guards_are_present
ERROR tests/test_extended_dep_api_pins.py::test_bnb_functional_dequantize_4bit_present

AssertionError: tests/test_mlx_batching_and_decay.py::test_compiler_review_guards_are_present
replaced ['bitsandbytes.nn'] in sys.modules and did not put it back, so every later test
in this process imports the substitute. Use monkeypatch.setitem(sys.modules, ...), which
restores on teardown.
```

Neither test mentions bitsandbytes. The first one is four `inspect.getsource` assertions; all it does is `import unsloth_zoo.compiler`.

## Why

`unsloth_zoo/__init__.py:191` calls `bitsandbytes_stub.inject_into_sys_modules()` when `real_bitsandbytes_available()` is false, and `_BnbFinder.find_spec` then builds each submodule on demand:

```
bitsandbytes             __file__=.../stubs/bitsandbytes_stub.py  origin=.../bitsandbytes_stub.py
bitsandbytes.nn          __file__=None                            origin=bitsandbytes_stub
bitsandbytes.functional  __file__=None                            origin=bitsandbytes_stub
```

`_is_module_stub` keys on `__file__`, so the parent escapes the check because the stub module is a real file on disk. Its lazily created submodules are bare `ModuleType` and do not, which is the only reason they were reported. The package installing its own stub is the same thing that happens in production on that host, not a test swapping a module out.

It matters beyond the noise: the guard raises in **teardown**, so the test itself passes and the run still reports an error, and whichever test happens to sort first wears it. Move a file and the error moves with it.

## The fix

Exempt the stub by the spec origin `_BnbFinder` sets explicitly, and only when nothing was bound before, so `was is not None` keeps the swap case reported.

This does not weaken what the guard was written for. Its own comment records the motivating case: `test_vllm_to_hf_conversion` left a `bitsandbytes.functional` carrying only `dequantize_4bit`, and every later `from bitsandbytes.functional import QuantState` failed with "(unknown location)". The unsloth stub has the opposite property, a module-level `__getattr__` that answers every name, so it cannot cause that.

## Test

`tests/test_module_leak_guard.py` drives the two hooks directly with explicit before/after `sys.modules` states, so no case depends on whether an earlier test already imported `unsloth_zoo`:

- the unsloth stub appearing where nothing was bound, for `bitsandbytes.nn` and `bitsandbytes.functional`, is not reported
- importing the real package where nothing was bound is not reported
- a partial substitute appearing where nothing was bound is still reported, which is the motivating case above
- a module that was already bound being replaced, **even by the unsloth stub**, is still reported

```
$ pytest tests/test_module_leak_guard.py -q
5 passed

# without the conftest change
FAILED tests/test_module_leak_guard.py::test_unsloth_bitsandbytes_stub_is_not_a_leak[bitsandbytes.nn]
FAILED tests/test_module_leak_guard.py::test_unsloth_bitsandbytes_stub_is_not_a_leak[bitsandbytes.functional]

$ pytest tests/test_mlx_batching_and_decay.py::test_compiler_review_guards_are_present \
         tests/test_extended_dep_api_pins.py::test_bnb_functional_dequantize_4bit_present -q
2 passed                     # before: 2 passed, 2 errors
```

Against the twelve test files that mention bitsandbytes, the failure set is identical to `main` apart from `test_bnb_functional_dequantize_4bit_present`, which stops erroring. Run on macOS arm64, Python 3.11, no bitsandbytes installed, at `52aff0d`.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -328,6 +328,22 @@ def _is_module_stub(mod):
     return mod is not None and getattr(mod, "__file__", None) is None
 
 
+def _is_unsloth_bnb_stub(mod):
+    """Whether this is the bitsandbytes stub `import unsloth_zoo` installs for itself.
+
+    `unsloth_zoo/__init__.py` calls `bitsandbytes_stub.inject_into_sys_modules()` when no
+    real bitsandbytes is installed, and `_BnbFinder` then builds each submodule on demand
+    with this spec origin. That is the package behaving as designed on a host without the
+    real one, not a test substituting a module, and unlike the partial substitute this
+    guard was written for it answers every attribute rather than a chosen few.
+
+    The parent `bitsandbytes` already escapes the check below because the stub module has
+    a `__file__` of its own; its submodules are bare ModuleType and do not, which is the
+    only reason the first test to import unsloth_zoo was reported as a leak.
+    """
+    return getattr(getattr(mod, "__spec__", None), "origin", None) == "bitsandbytes_stub"
+
+
 def pytest_runtest_setup(item):
     import sys as _sys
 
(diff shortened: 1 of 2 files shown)


danielhanchen:
main now exempts the zoo's own bitsandbytes stub in tests/conftest.py via the IS_UNSLOTH_STUB marker, so I do not think this is needed any more. Sorry for the slow review, and thanks for the very clear write up.

vineethsaivs:
Confirmed main exempts the zoo-owned stub through `IS_UNSLOTH_STUB`. Closing this as superseded.

PR #1134, opened 2026-08-30
Restore parent package attributes when the mlx shim fixtures put sys.modules back

danielhanchen:
#1131 was a net regression. It fixed the two failures it was aimed at and introduced six others, and unsloth's `Core` job reports three of them against zoo main today, one per matrix leg depending on which worker draws which file.

### What went wrong

`test_mlx_preference.py` and `test_mlx_trainer_internals.py` drop every `unsloth_zoo.mlx.*` module so their shim is what those modules bind against. #1131 made them save the originals and put them back, which is right as far as it goes and only half the state.

Dropping `unsloth_zoo.mlx` and letting it be re-imported makes the import machinery bind

```python
unsloth_zoo.mlx = <the module built under the shim>
```

as an attribute on the parent package, and that binding outlives the `sys.modules` entry that produced it. Restoring `sys.modules` therefore leaves two lookup paths disagreeing, because they are read by different things:

```python
import unsloth_zoo.mlx.trainer as t              # walks parent attributes
from unsloth_zoo.mlx.trainer import MLXTrainer   # resolves via sys.modules
```

Measured after the fixture runs:

```
import-as module id : 139681420121392
sys.modules entry   : 139700800574768  same=False
t.MLXTrainer        : 616397424
fresh MLXTrainer    : 618847840        SAME=False
```

Any file doing both is left holding two live copies of one module, and an `isinstance` against a class from one is False for an object the other built. That is exactly what `dataset_utils.train_on_responses_only` does to decide whether it has an `MLXTrainer`, so it silently took the Hugging Face branch and called a non-callable tokenizer wrapper.

### The fix

Both fixtures now go through `mlx_simulation.restore_modules`, which puts `sys.modules` and the parent attributes back together. It also drops the attribute outright for a submodule that was imported only while the shim was up, since leaving that bound is the same split in reverse: an attribute walk reaching a module no `sys.modules` lookup can.

### Confirmation

Each is a deterministic two-file repro, run as `pytest A B`:

| paired with `test_mlx_preference.py` | before #1131 | at #1131 | here |
| --- | --- | --- | --- |
| `test_pr684_review_fixes_a.py` | 97 passed | 1 failed | 97 passed |
| `test_mlx_text_path_contract.py` | 98 passed | 1 failed | 98 passed |
| `test_mlx_distributed_loader.py` | 87 passed | 4 failed | 87 passed |

What #1131 fixed still holds: `test_mlx_preference` + `test_mlx_gated_delta_vjp`, `test_mlx_gated_delta_vjp` + `test_mlx_neftune_quant_map`, and `test_mlx_trainer_internals` + `test_mlx_gated_delta_vjp` all pass, as do both polluters on their own.

Whole suite as `Core` runs it (`-n 4 --dist loadfile`, same ignores and deselects): **11 failures before, 5 after**, and the 5 are the local-environment ones this box produces with or without these changes and that CI does not report (`test_rl_replacements_compile_disable`, which fails here even in isolation, and `test_moe_lora_grouped_mm_capability`).

### Why nothing caught it

Nothing asserted that the two lookup paths agreed, which is how #1131 traded two failures for six without it showing up. `tests/test_mlx_shim_module_restore.py` now asserts it, and also pins both fixtures to the shared helper so a future edit that goes back to a bare `sys.modules.update` is a failing test rather than six failures in unrelated files one worker over. Removing the parent repair fails two of its three tests.

### DIFF
--- a/tests/mlx_simulation/__init__.py
+++ b/tests/mlx_simulation/__init__.py
@@ -143,3 +143,58 @@ def mlx_is_simulated() -> bool:
         return False
     origin = f"{getattr(module, '__name__', '')} {getattr(module, '__file__', '') or ''}"
     return "mlx_simulation" in origin
+
+
+def snapshot_modules(is_owned):
+    """Record every currently-imported module the caller is about to disturb."""
+    import sys
+
+    return {name: module for name, module in sys.modules.items() if is_owned(name)}
+
+
+def restore_modules(saved, is_owned):
+    """Undo a fixture's sys.modules surgery, PARENT PACKAGE ATTRIBUTES INCLUDED.
+
+    Putting the old entries back in sys.modules is only half of it. Dropping
+    `unsloth_zoo.mlx` and letting it be re-imported makes the import machinery
+    bind `unsloth_zoo.mlx = <the new module>` as an attribute on the parent
+    package, and that binding outlives the sys.modules entry that produced it.
+
+    The two then disagree, because they are read by different things:
+    `import unsloth_zoo.mlx.trainer as t` walks parent attributes, while a
+    `from unsloth_zoo.mlx.trainer import MLXTrainer` inside the code under test
+    resolves through sys.modules. A file that does both is left holding two live
+    copies of one module, and `isinstance(trainer, MLXTrainer)` is False for an
+    object the other copy built. That is not hypothetical: it is what
+    dataset_utils.train_on_responses_only does to decide whether it is looking at
+    an MLXTrainer, and it silently took the Hugging Face branch instead.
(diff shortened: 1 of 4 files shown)


mahiatlinux (approved):
reproduced through unsloth Core: the stale parent-package binding causes the current MLX failures. the shared restore helper fixes the split module state.

mahiatlinux:
Approved. This is blocking current unsloth Core: https://github.com/unslothai/unsloth/actions/runs/33293510781/job/99209169516. The failures match the stale MLX shim state fixed here.

PR #1137, opened 2026-08-31
Make test_mlx_gated_delta_vjp's real-mlx gate reject the torch shim

danielhanchen:
The five failures unsloth's `Core` job reports against zoo `main`, all in `tests/test_mlx_gated_delta_vjp.py`:

```
test_checkpointed_layer_detaches_integer_arguments        AttributeError: 'bool' object has no attribute 'item'
test_router_gradient_matches_detached_reference           mlx-shim: 'mlx.core.argpartition' is not implemented
test_index_derived_graphs_stay_differentiable[_sparse_mask_loss-shape1]
                                                          mlx-shim: 'mlx.core.put_along_axis' is not implemented
test_only_the_index_argument_is_detached                  only Tensors of floating point dtype can require gradients
test_unfused_projection_matches_the_fused_one             mlx-shim: 'mlx_vlm.models.glm5_next.config.TextConfig' is not implemented
```

## Why

The file gates on `importlib.util.find_spec("mlx") is not None`. `find_spec` consults `sys.modules` first, and several sibling modules call `simulate_mlx_on_torch()` while being IMPORTED. Collection imports every module in the session, so once it has reached one of those the gate reads the shim as a real mlx. The `requires_real_mlx` cases then run against a stub that has no `mx.argpartition`, no `put_along_axis` and no `mlx_vlm.models.glm5_next`, and that returns a Python `bool` where mlx returns an array. Those are the five messages above, one for one.

Whether an installer is imported first is alphabetical serially, where none is, and worker assignment under `-n 4 --dist loadfile`, which is how the `Core` job runs this suite. That is the whole reason the file passes locally and fails in CI.

Deterministic two-file repro, five failures before this change and none after:

```
pytest tests/test_mlx_arrays_cache_advance.py tests/test_mlx_gated_delta_vjp.py
```

This is the same gate, and the same reason, as `test_mlx_neftune_quant_map.py` in #1131, which added `mlx_simulation.mlx_is_simulated()` for exactly this. That change did not reach this file. Its note that the remaining five were local-environment failures CI does not report was wrong: they are these, and CI reports them on every run.

## Why skip rather than grow the shim

The alternative is implementing `argpartition`, `put_along_axis` and a `glm5_next` config in the stub. These cases assert mlx autodiff semantics, so checking them against a torch stub tests the stub; and the `glm5_next` case needs a real `mlx_vlm` module, which cannot be stubbed into a meaningful assertion. #1131 settled the same question the same way.

The tests themselves are right about what they check. #1120 added them with the intent that they need real MLX; only the gate is wrong.

## Coverage

None moves on Apple Silicon. Every `simulate_mlx_on_torch()` call in the suite is conditional on a real mlx being absent, verified by AST over all 36 callers, so where a real mlx exists the shim is never installed and the gate evaluates exactly as before. On Linux CI these five skip, which is what they already do when the file runs on its own.

## Standing check

`tests/test_mlx_real_mlx_gates_reject_the_shim.py` imports each of the three modules holding a real-mlx gate in a subprocess with the shim installed first, and fails if the gate comes back `True`. Verified non-vacuous: it fails on the pre-fix file and passes after. It also asserts each gate is still a module-level name, so a rename cannot leave it guarding nothing.

## Verification

Suite as the `Core` job runs it (`-n 4 --dist loadfile`, same ignores and deselects): `5 failed, 5383 passed`. The five above are gone. The five that remain are this box's pre-existing ones, unrelated to MLX: four in `test_rl_replacements_compile_disable.py` reproduce on an unmodified checkout, and `test_moe_lora_grouped_mm_capability.py::test_no_grad_decoding_retains_nothing` passes in isolation on both trees. CI reports none of them.

Two files, `test_mlx_distributed_loader.py` and `test_mlx_save_export_regressions.py`, were excluded from that local run: they spin at 100% CPU indefinitely on this host, on an unmodified checkout as well, and pass in CI.

### DIFF
--- a/tests/test_mlx_gated_delta_vjp.py
+++ b/tests/test_mlx_gated_delta_vjp.py
@@ -33,9 +33,22 @@ import types
 
 import pytest
 
-_HAS_REAL_MLX = importlib.util.find_spec("mlx") is not None
-if not _HAS_REAL_MLX:
-    from mlx_simulation import simulate_mlx_on_torch
+from mlx_simulation import mlx_is_simulated, simulate_mlx_on_torch  # noqa: E402
+
+# Not "the import worked": several test modules install the MLX-on-torch shim into
+# sys.modules while being IMPORTED, and collection imports every module, so `mlx`
+# can already BE that shim by the time this file is reached. Gating on find_spec
+# alone then reads the shim as real, and the cases marked requires_real_mlx run
+# against it -- they need mx.checkpoint, a vjp that differentiates integer arrays,
+# and mlx_vlm's glm5_next, none of which the shim implements. That made the file
+# pass or fail on collection order: stable serially, where alphabetical order puts
+# no installer first, and not under `-n N --dist loadfile`, where worker assignment
+# decides. Same gate, and same reason, as test_mlx_neftune_quant_map.py.
+_SHIM_ALREADY_INSTALLED = mlx_is_simulated()
+_HAS_REAL_MLX = (
+    importlib.util.find_spec("mlx") is not None and not _SHIM_ALREADY_INSTALLED
+)
+if not _HAS_REAL_MLX and not _SHIM_ALREADY_INSTALLED:
     simulate_mlx_on_torch()
 
 import mlx.core as mx  # noqa: E402  (real, or the torch shim on CI)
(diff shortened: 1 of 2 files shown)


danielhanchen:
Closing as a duplicate of #1136, which was opened first and carries the same one-line gate fix plus a broader guard (an AST rule over every test module, rather than a subprocess check of the three known gates).

Two things found while verifying this that are now on #1136:

- `find_spec("mlx_vlm.models.glm5_next")` at module level raises rather than returning None when mlx_vlm is absent, so on a real-MLX machine without mlx-vlm the whole file fails to collect. Fixed there in 99b44dfb.
- mlx publishes manylinux x86_64 wheels, and with `mlx[cpu]` + mlx-lm + mlx-vlm installed this file runs 22 passed / 12 skipped on Linux, the 12 being Metal-only. So the five tests need not stay skipped on CI; that is a follow-up on the workflow, not on this file.

PR #1141, opened 2026-09-01
Prune an open PR's caches once nothing has read them

danielhanchen:
`cache-janitor.yml` pass 1 frees a cache when its pull request is no longer open and keeps it otherwise. That treats PR state as a liveness signal, which it is not. Lookup is scoped by ref, so only that PR's own runs can ever restore the entry. If none of them has in days, it is holding budget for a reader that does not exist, and whether the PR is still open says nothing about that either way.

The zoo shows the failure concretely: 6.18 GiB on `refs/pull/929/merge`, 30.9% of its 20 GiB ceiling, three of the four entries never read once since they were written on 2026-08-26. PR #929 is open, so the sweep spares them and reported `pruned 0 caches, 0.0 GiB` against a 16.9 GiB total that had already tripped the 80% warning. Dependabot PRs are the usual shape, staying open for weeks or months in both repos.

## The janitor could not see idleness at all

The inventory selected `id, created_at, size_in_bytes, ref, version, key`. `last_accessed_at` is returned by the API and was discarded, and `created_at` was used purely as a sort key for generation ranking, never against a clock. So there was no way to express "nothing has read this".

It now also selects `last_accessed_at`, which is the field GitHub's own eviction policy runs on.

## The rule

Prune a cache on an **open** PR's ref once it has not been read for `open_pr_idle_days`, default 3.

By age, not by author. A stale human PR costs exactly the same, and a bot PR whose CI ran this morning is genuinely live, so an author check would add API calls and failure modes without improving the decision.

This only brings the reclaim forward rather than freeing anything GitHub would keep. An entry unread for 7 days is evicted regardless; the budget is what cannot wait, since a week of overlap times every ref that ran is larger than it.

## Three guards against freeing something in use

- **`last_accessed_at` is re-read immediately before acting.** The inventory is minutes old by the time the loop reaches an entry, and a run that restored it in between resets the clock. Deciding on the stale copy would delete a cache that is back in use, which is exactly what this rule claims not to do.
- **The re-check runs in report mode too.** Report mode exists to predict what delete mode does, and the documented way to adopt a change here is to read the report first. A preview listing candidates the real sweep would spare sends someone looking for bytes that were never going to be freed.
- **PR state is now `open`, `closed` or `unknown`.** It previously defaulted to `open` on a lookup failure, which was the same outcome as unknown while `open` only ever meant keep. Now that an open PR can also be pruned, the two have to differ. An API error, an unparseable timestamp, or a failed revalidation all still keep the cache.

`open_pr_idle_days` is validated with the same `[1-9]|[1-9][0-9]` whitelist as `keep`, and for the same reason recorded there: an unusable value must not silently become a cutoff of 0, which would treat every cache as idle.

## Verification

Both repos get the byte-identical change, applied by one patch script rather than hand-edited twice.

Both janitors' embedded scripts were extracted and run against stubbed inventories with a stub `gh`, in report and delete mode, covering:

| case | expected |
| --- | --- |
| open PR, idle 4d | pruned |
| open PR, idle 2d | kept |
| open PR, lookup fails | kept |
| closed PR, read 0h ago | pruned regardless of age |
| open PR idle 4d, read again between inventory and delete | kept |
| open PR idle 4d, revalidation itself fails | kept |
| idle 500h cache on `refs/heads/main` | untouched by the PR rule |
| three `pip-v2-*` generations | oldest pruned, newest two kept |
| open PR idle 4d, two generations | both pruned by pass 1, not shielded by `keep` |

18 assertions, all passing on both files, with report and delete agreeing on every case. The harness lives outside the repo; say the word if it should be checked in.

Roll out by running `workflow_dispatch` in **report** mode first and diffing the candidate list against `gh api /repos/<repo>/actions/caches` before letting the schedule delete.

## Note for the zoo

The #929 residue expires on its own at approximately 2026-09-02T09:36Z, and the wider `setup-python-*` family (10 entries, 13.07 GiB, 77.5% of zoo usage) within about 33 hours of that measurement, since nothing writes that key scheme any more. Run report mode before then if you want to see it in the candidate list. That timing is why this PR adds no special case for retired key schemes: it would reclaim those bytes for roughly a day and then be dead code, and the janitor header already documents leaving that family to the 7-day timer.

### DIFF
--- a/.github/workflows/cache-janitor.yml
+++ b/.github/workflows/cache-janitor.yml
@@ -88,6 +88,10 @@ on:
         description: 'Generations kept per prefix. 2 so an in-flight run cannot lose the cache it just resolved, and so a reverted dependency bump still hits.'
         type: string
         default: '2'
+      open_pr_idle_days:
+        description: 'Delete a cache on an OPEN pull request ref after this many days with no read. Nothing but that PR can reach it, so no read means no reader.'
+        type: string
+        default: '3'
 
 # Two sweeps would race on the same ids and spend their DELETEs on 404s.
 concurrency:
@@ -114,6 +118,7 @@ jobs:
           # `false` alike, so `inputs.mode == false` is true when unset.
           DELETE: ${{ (github.event_name == 'schedule' || inputs.mode == 'delete') && 'yes' || 'no' }}
           KEEP: ${{ inputs.keep || '2' }}
+          OPEN_PR_IDLE_DAYS: ${{ inputs.open_pr_idle_days || '3' }}
         run: |
           set -euo pipefail
           repo="$GITHUB_REPOSITORY"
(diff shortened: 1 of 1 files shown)


danielhanchen:
@codex review

PR #1142, opened 2026-09-01
MLX: fused Metal kernels for the gated-delta training backward

Lyxot:
## What this changes

Training on MLX with a gated-delta model (Qwen3.5, Qwen3-Next, Kimi Linear) currently takes an ops VJP through the recurrence. It is memory-correct but dispatch-bound: roughly `T * 12` lazy primitives per layer per step, which `mx.compile` cannot fuse. This adds fused Metal kernels for the training forward and backward and routes eligible calls to them.

Gated-delta forward+backward gets **2.3-2.9x** faster, which is worth **1.08-1.25x** on a whole training step. Numerics are unchanged to within accumulation-order noise, and peak memory is unchanged.

Nothing else about the path changes: shapes the kernels cannot place still take the ops VJP in both directions, inference is untouched, and `gated_delta_kernel_supported` is the single predicate deciding eligibility.

## How it works

Three kernels share one blocked-sequential thread layout. Where mlx-lm's kernel gives each state row its own simdgroup — so every contraction costs a full `simd_sum` and each of the `Dv` rows re-reads the same `k`/`q` from device memory — here `P = Dk/W` threads cover a row inside a single simdgroup, `DB` rows share a threadgroup, and the threadgroup stages a `TB`-step time block of `k`/`q`/`v`/`g`/`beta` once.

- **Forward** runs the recurrence in that layout.
- **Checkpoint** replays the chunk keeping only the state at each time-block boundary, plus every step's `k · state` contraction.
- **Backward** reverse-scans, replaying each block from its entry checkpoint, so no per-step state buffer is needed and the per-row state cotangent stays in registers.

The BPTT chunk length is derived per call rather than fixed. Longer chunks amortize the launch and the block replay — 1.9x on the backward between 64 and 512 steps — but hold more block checkpoints, so the length is chosen against a checkpoint memory budget and clamped to a 64-512 range.

The forward and backward want different layouts: the forward prefers wide segments (more rows per threadgroup, fewer redundant `k`/`q` reads), the backward prefers narrow ones because it carries three state fragments per thread and replays blocks. They are therefore selected independently. Forcing them to agree costs 1.9x on the backward.

The last commit fixes a placement gap found while benchmarking: at `Dk >= 256` the reduction scratch crowded out every time block at full threadgroup width, so no layout was found and those shapes silently fell back to the ops VJP. The layout search now retries at half threadgroup width, which halves the reduction scratch while costing the time block far less.

## Results

Apple M3 Max, batch 1, bfloat16, measured against this branch's merge base. Every figure is min-of-N wall time with the arms interleaved, and a duplicate baseline arm runs in every measurement so the spread between its two timings shows how much of a difference is the machine. All cells below are under 3% control spread.

Gated-delta layer, at the head configurations the Qwen3.5 checkpoints actually run:

| model | T | forward | forward+backward |
|---|---:|---:|---:|
| 0.8B / 2B | 512 | 1.61x | **2.37x** |
| 0.8B / 2B | 1024 | 1.49x | **2.35x** |
| 0.8B / 2B | 2048 | 1.53x | **2.62x** |
| 4B | 512 | 1.63x | **2.33x** |
| 4B | 1024 | 1.72x | **2.44x** |
| 4B | 2048 | 1.82x | **2.87x** |

Whole training step — gradient checkpointing, chunked cross-entropy and `mx.compile` all on, which is what the MLX trainer runs:

| model | T | full-parameter tok/s | LoRA tok/s | full-parameter | LoRA |
|---|---:|---:|---:|---:|---:|
| Qwen3.5-0.8B | 512 | 989 → 1,072 | 1,362 → 1,619 | 1.096x | 1.188x |
| Qwen3.5-0.8B | 1024 | 924 → 1,093 | 1,344 → 1,616 | 1.199x | 1.200x |
| Qwen3.5-0.8B | 2048 | 887 → 1,079 | 1,295 → 1,616 | 1.227x | 1.245x |
| Qwen3.5-2B | 512 | 510 → 546 | 739 → 820 | 1.079x | 1.110x |
| Qwen3.5-2B | 1024 | 503 → 543 | 763 → 822 | 1.088x | 1.082x |
| Qwen3.5-2B | 2048 | 517 → 596 | 765 → 870 | 1.165x | 1.152x |
| Qwen3.5-4B | 512 | 225 → 247 | 334 → 371 | 1.104x | 1.114x |
| Qwen3.5-4B | 1024 | 268 → 299 | 349 → 389 | 1.114x | 1.118x |
| Qwen3.5-4B | 2048 | 191 → 222 | 307 → 354 | 1.156x | 1.152x |

LoRA is r=16, alpha=32 on every linear in both mixer kinds and the MLP.

The end-to-end figure is much smaller than the layer figure because gated delta is a shrinking share of a step that also runs MLPs, full attention every fourth layer, and a cross-entropy over a 248k vocabulary. The gain rises with sequence length and falls with model width, and it is the same whether training full-parameter or with LoRA.

## Correctness

Losses on a real training step agree with the current path to at most **1.1e-4** relative across all eighteen model-level measurements. Gradients are checked in the test suite against autodiff over a step-by-step fp32 reference, elementwise rather than globally so a small misrouted entry cannot hide behind a large one.

Atomic accumulation in the backward leaves low-order gradient bits nondeterministic, as elsewhere on Metal. The forward and backward use different layouts, so the backward replays states differing from the forward's in the low bits: at most 2.4e-7 relative per gradient over a T=512 fp32 chunk, against up to 1.9e-7 between two runs of the same backward, which already disagree through the atomics.

## Risks and tradeoffs

- **No memory win.** Isolated, the fused backward cuts layer peak memory 8.6-11x. At model scale that saving is already taken by gradient checkpointing, which recomputes exactly the activations the fused backward avoids storing. Measured end to end, peak memory moves by at most 136 MB in either direction. This is a speed change, not a memory change.
- **Single-machine calibration.** The layout search and the chunk-length budget were tuned on one Apple GPU. `mx.device_info()` does not expose a core count, so the thresholds cannot scale themselves to the device. They are conservative rather than tight.
- **Absolute throughput is machine-state dependent.** The same configuration measured 415.8 ms and 517.6 ms on an idle versus a loaded machine. The ratios held across both; read the tok/s figures as relative, not as a specification.

## Validation

```bash
python -m pytest tests/test_qwen35_vjp_metal.py -q        # 49 passed
python -m pytest tests/test_mlx_gated_delta_vjp.py -q     # 36 passed
```

Run the two files separately. A pre-existing failure occurs when both run in one process, and it reproduces on the merge base.

The tests cover both gating shapes, all three dtypes, thread prefixes 1 to 32, segment widths 4 to 32, GQA ratios 1 to 4, sequence lengths that are not a multiple of the time block, and the chunk seam. A dispatch check confirms the fused kernels are actually reached at every benchmarked shape, with blocked-forward call counts matching the expected chunk counts, so the measured gains are attributable to the kernels rather than to a dispatch that quietly declined.

### DIFF
--- a/.github/workflows/consolidated-tests-ci.yml
+++ b/.github/workflows/consolidated-tests-ci.yml
@@ -566,6 +566,31 @@ jobs:
 
       - name: tests/test_mlx_batching_and_decay.py
         run: .lanes/venv-suites/bin/python -m pytest tests/test_mlx_batching_and_decay.py -v -rs
+
+      # The three gated-delta files ran in no workflow at all: this lane's list
+      # omitted them and the Apple Silicon lane, which is opt-in behind the `mlx`
+      # label, names only test_qwen35_vjp_metal.py. So the ops VJP, the batched
+      # gradient regression from #776 and the consumer-binding sweep shipped
+      # ungated. They need mlx core plus mlx-lm, not Metal, so they belong here;
+      # the Metal-only halves skip and stay the macOS lane's job.
+      - name: tests/test_mlx_gated_delta.py
+        run: .lanes/venv-suites/bin/python -m pytest tests/test_mlx_gated_delta.py -v -rs
+
+      - name: tests/test_mlx_gated_delta_vjp.py
+        run: .lanes/venv-suites/bin/python -m pytest tests/test_mlx_gated_delta_vjp.py -v -rs
+
+      - name: tests/test_mlx_gated_delta_batch_grad.py
+        # Gradient parity vs plain autodiff at B >= 2. Every test here needs the
+        # real runtime, so under the torch shim the file skips wholesale -- which
+        # is exactly what the gate below refuses to accept as a pass.
+        run: .lanes/venv-suites/bin/python -m pytest tests/test_mlx_gated_delta_batch_grad.py -v -rs
+
+      - name: Fail if the gated-delta gradient tests skipped
+        run: |
+          .lanes/venv-suites/bin/python -m pytest tests/test_mlx_gated_delta_batch_grad.py -q -rs | tee gd.txt
+          if ! grep -qiE '[1-9][0-9]* passed' gd.txt; then
+            echo "real mlx is installed but no gated-delta gradient test ran"; exit 1
+          fi
   # Core (HF/TRL/peft) drift matrix. Three cells: HF=4.57.6+TRL<1, HF=latest+TRL=latest,
   # and pyproject defaults.
   #
--- a/tests/test_mlx_gated_delta_vjp.py
+++ b/tests/test_mlx_gated_delta_vjp.py
@@ -406,6 +406,67 @@ def test_kernel_dispatch_guards_partial_threadgroup_rows():
     assert not gv.gated_delta_kernel_supported(q, g, None, bad_v, q)
 
 
+@pytest.mark.skipif(_HAS_METAL, reason="the off-Metal answer is what is pinned here")
+def test_kernel_support_declines_off_metal_without_touching_the_layout_search():
+    """Off Metal the predicate must answer False, not raise.
+
+    The layout search reads `dtype.size`, which only a real mx.Dtype has, so with
+    it ahead of the device guard a training call under the torch shim died with
+    AttributeError instead of falling back to the ops VJP.
+    """
+    import unsloth_zoo.gated_delta_vjp as gv
+
+    q = mx.zeros((1, 8, 2, 64))
+    k = mx.zeros((1, 8, 2, 64))
+    v = mx.zeros((1, 8, 2, 64))
+    for g in (mx.zeros((1, 8, 2)), mx.zeros((1, 8, 2, 64))):
+        assert gv.gated_delta_kernel_supported(q, g, None, v, k) is False
+
+
+@pytest.mark.skipif(_HAS_METAL, reason="the off-Metal answer is what is pinned here")
+def test_training_call_falls_back_to_ops_vjp_off_metal(monkeypatch):
+    """The whole dispatch, not just the predicate: an open training window off
+    Metal must reach gated_delta_ops_efficient."""
+    import types
+
+    import unsloth_zoo.gated_delta_vjp as gv
+
+    # patch_gated_delta does `from mlx_lm.models import gated_delta`, which reads
+    # the package ATTRIBUTE, so sys.modules alone is not enough to redirect it.
+    package = types.ModuleType("mlx_lm")
+    models = types.ModuleType("mlx_lm.models")
+    module = types.ModuleType("mlx_lm.models.gated_delta")
+    module.gated_delta_update = lambda *args, **kwargs: ("original", None)
+    module.gated_delta_kernel = lambda *args, **kwargs: ("kernel", None)
+    module.compute_g = lambda A_log, a, dt_bias: mx.zeros(a.shape)
+    models.gated_delta = module
+    package.models = models
+    monkeypatch.setitem(sys.modules, "mlx_lm", package)
+    monkeypatch.setitem(sys.modules, "mlx_lm.models", models)
+    monkeypatch.setitem(sys.modules, "mlx_lm.models.gated_delta", module)
+
+    reached = []
+    real_ops = gv.gated_delta_ops_efficient
+    monkeypatch.setattr(
+        gv, "gated_delta_ops_efficient",
+        lambda *args, **kwargs: (reached.append(True), real_ops(*args, **kwargs))[1],
+    )
+    gv.patch_gated_delta()
+
+    q = mx.zeros((1, 8, 2, 64))
+    v = mx.zeros((1, 8, 2, 64))
+    a = mx.zeros((1, 8, 2))
+    acquire_mlx_training_patches()
+    try:
+        module.gated_delta_update(
+            q, q, v, a, a, mx.zeros((2,)), mx.zeros((2,)),
+            state=None, mask=None, use_kernel=True,
+        )
+    finally:
+        release_mlx_training_patches()
+    assert reached, "training call off Metal did not reach the ops VJP"
+
+
 # -- training window and index detachment -------------------------------------
 
 @pytest.fixture
--- a/tests/test_mlx_gated_delta_vjp.py
+++ b/tests/test_mlx_gated_delta_vjp.py
@@ -402,8 +402,8 @@ def test_kernel_dispatch_guards_partial_threadgroup_rows():
     g = mx.zeros((1, 8, 2))
     ok_v = mx.zeros((1, 8, 2, 16))
     bad_v = mx.zeros((1, 8, 2, 30))
-    assert gv.gated_delta_kernel_supported(q, g, None, ok_v)
-    assert not gv.gated_delta_kernel_supported(q, g, None, bad_v)
+    assert gv.gated_delta_kernel_supported(q, g, None, ok_v, q)
+    assert not gv.gated_delta_kernel_supported(q, g, None, bad_v, q)
 
 
 # -- training window and index detachment -------------------------------------
--- a/tests/test_qwen35_vjp_metal.py
+++ b/tests/test_qwen35_vjp_metal.py
@@ -290,13 +290,11 @@ def test_end_to_end_training_step_all_patches():
     assert any(float(mx.abs(g).max()) > 0 for g in flat), "all grads zero"
 
 
-# Two tolerances per output. The global one bounds error against the tensor's
-# largest magnitude; the elementwise one bounds each output against its own
-# magnitude, so a single badly-computed small entry cannot hide behind a large
-# one. bf16 inputs quantize y to ~4e-3 either way; the fp32 state carries no
-# such rounding and is held to fp32 accumulation error.
-_Y_TOL = {mx.bfloat16: 5e-3, mx.float32: 5e-6}
-_Y_ELEM_TOL = {mx.bfloat16: 8e-3, mx.float32: 2e-5}
+# The global tolerance bounds error against the tensor's largest magnitude,
+# the elementwise one against each entry's own, so a small badly computed
+# entry cannot hide behind a large one.
+_Y_TOL = {mx.bfloat16: 5e-3, mx.float16: 5e-4, mx.float32: 5e-6}
+_Y_ELEM_TOL = {mx.bfloat16: 8e-3, mx.float16: 1e-3, mx.float32: 2e-5}
 _STATE_TOL = 1e-6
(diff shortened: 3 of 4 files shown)


Lyxot:
@codex review

danielhanchen:
This routes the default MLX gated-delta training path through new Metal kernels whose layout search and chunk budget were calibrated on one machine, so a slower or narrower Apple GPU could pick a layout that has not been measured. Could you confirm the fallback and timings on at least one other Apple Silicon device before I review?

Lyxot:
Confirmed on a second device — an 18-core M3 Pro / 36 GB (macOS 26.6.2, mlx 0.31.2) against the 40-core M3 Max / 128 GB it was calibrated on (macOS 26.5.1, mlx 0.32.2). 45% of the GPU cores, 28% of the memory.

**The layout search has no device-derived input.** I dumped every decision the branch makes — eligibility verdict, forward and backward `(P, W, DB, TB, threads)`, and the derived BPTT chunk length — over 378 combinations (9 `Dk` × 7 `Dv` × 3 dtypes × vector/scalar gating), then diffed the two machines:

```
1,2c1,2
< # device: Apple M3 Max  arch: applegpu_g15s
< # mlx: 0.32.2
---
> # device: Apple M3 Pro  arch: applegpu_g15s
> # mlx: 0.31.2
```

The only lines that differ are the two naming the machine. All 378 rows — 282 eligible, 96 falling back to the reference path — are identical tuple for tuple, across an mlx version change as well.

That is structural rather than lucky: the search reads shapes, dtypes and three compile-time constants (`_BLOCKED_TG_MEMORY = 32768`, `_BLOCKED_TG_THREADS = 256`, `_BLOCKED_PAD = 8`). 32 KB of threadgroup memory is the same on every Apple GPU, so there is no per-device quantity a narrower GPU could vary. Dispatch matches too: every Qwen3.5 model/length pair reaches the fused kernel and splits into the expected number of chunks on both machines.

**Timings are equal or better on the narrower GPU.** Isolated gated-delta layer, bfloat16, batch 1, min-of-5, interleaved arms with a duplicate `main` control (control spread ≤1.0% in every cell below):

| model | T | fwd+bwd, M3 Max | fwd+bwd, M3 Pro |
|---|---|---|---|
| 0.8B | 512 / 1024 / 2048 | 2.37x / 2.35x / 2.62x | 2.65x / 2.83x / 3.21x |
| 2B | 512 / 1024 / 2048 | 2.43x / 2.48x / 2.60x | 2.70x / 2.81x / 3.23x |
| 4B | 512 / 1024 / 2048 | 2.33x / 2.44x / 2.87x | 2.98x / 3.41x / 3.87x |

End-to-end training — real optimizer steps, `mx.compile` + chunked cross-entropy + gradient checkpointing on both arms, batch 1, tok/s `main` → branch:

| model | T | M3 Pro full-parameter | M3 Pro LoRA r=16 |
|---|---|---|---|
| 0.8B | 512 | 527 → 622 (1.180x) | 586 → 703 (1.200x) |
| 0.8B | 1024 | 535 → 640 (1.196x) | 574 → 700 (1.218x) |
| 0.8B | 2048 | 506 → 628 (1.242x) | 533 → 673 (1.263x) |
| 2B | 512 | 288 → 314 (1.090x) | 348 → 385 (1.107x) |
| 2B | 1024 | 297 → 327 (1.102x) | 348 → 391 (1.124x) |
| 2B | 2048 | 289 → 326 (1.128x) | 333 → 385 (1.156x) |
| 4B | 512 | 127 → 142 (1.120x) | 152 → 172 (1.135x) |
| 4B | 1024 | 126 → 143 (1.129x) | 147 → 169 (1.163x) |
| 4B | 2048 | 121 → 143 (1.177x) | 139 → 165 (1.188x) |

For comparison the M3 Max reads 1.08–1.23x full-parameter and 1.08–1.25x LoRA, so no cell regresses on either machine. Peak memory is neutral to slightly lower on the branch. The two `main` arms produced bit-identical losses in all 18 cells — the guard that each arm got its own compile trace — and branch losses agree with `main` to 1.4e-4 relative, consistent with fp32 accumulation order differing between the paths.

Two things worth flagging rather than burying: forward-only on 4B is a hair behind the M3 Max (1.74x vs 1.82x at T=2048), the one column where the narrower GPU does not lead; and full-parameter 4B at T=2048 peaks at 21.5 GB, which pages on a 36 GB machine unless it is otherwise idle — a first pass under load read 0.998x at a 9.4% control spread, and re-running it idle gave the 1.177x above at 0.67%.

danielhanchen:
Reviewed this end to end. I pushed two commits to the branch; everything else below is for you to take or leave.

## What I ran

Apple Silicon is the only place the kernels actually execute, so I ran the repo's own `MLX CI on Mac M1` lane against this head on a `macos-15` runner. All 16 files green, and `tests/test_qwen35_vjp_metal.py` reports **49 passed, 0 skipped**, matching your validation section. `test_mlx_training_e2e_metal.py` (56) and `test_mlx_pr684_full_validation_metal.py` (6) are green too, so real LoRA training through the new path works.

Everything that is not Metal I checked on Linux with real MLX (`mlx` + `mlx-cpu`, the same combination the `mlx-cpu-linux` lane installs), plus the `tests/mlx_simulation` torch shim:

- **Layout search, exhaustively.** 2160 combinations of `Dk` x `Dv` x dtype x gating rank x config. For all 2040 placements I recomputed the threadgroup byte count from the kernels' own `threadgroup` declarations rather than from `_blocked_layout`'s formula, and checked `Dk % P`, `W % 4`, `W <= max_segment`, `Dv % DB`, `DB % (32/P)`, `threads == P*DB`, `32 <= threads <= 1024` and `NSIMD >= 1`. Zero violations. The formula and the declarations agree, including the vectorized `red_g` and `g_s` rows, and `DB % (32/P) == 0` does guarantee `threads >= 32`, so `red_q[NSIMD][Dk]` is never zero length.
- **Fallback numerics, base versus head.** 96 cases over ragged `T` (1, 3, 63, 64, 65, 77, 128, 129), GQA ratios 1/2/4, fp32 and bf16, both gating ranks, batch 2. Forward plus all six gradients: **672 arrays, bit identical** to the merge base. The ops VJP is genuinely untouched.
- **Regression surface.** `pytest -k "mlx or gated_delta or pinned_symbols"` on head-with-fix versus merge base: same 82 failures on both (24 unique, all pre-existing), zero new, ten more passing.
- **Compatibility.** Every symbol a consumer binds is signature identical. `unsloth_zoo` still imports with no mlx and no unsloth, and does not pull in `gated_delta_vjp` as a side effect. `is_mlx_available()` is False across [Windows, Linux, WSL, macOS x86] x [NVIDIA, AMD, CPU], so nothing here is reachable off Darwin arm64.

## One bug, fixed in 88e8feb9

`gated_delta_kernel_supported` evaluated `_blocked_layouts(q, g, v)` before its two device clauses, and the layout search reads `dtype.size`, which only a real `mx.Dtype` has. Off Metal the predicate therefore raised instead of answering False:

```
AttributeError: 'torch.dtype' object has no attribute 'size'
  gated_delta_vjp.py:1144  and _blocked_layouts(q, g, v) is not None
  gated_delta_vjp.py:561   in_dtype.size * (dk_rows * (Dk + _BLOCKED_PAD)
```

It is reachable through the real dispatch, not just a direct call: with a training window open, which is what `MLXTrainer.train()` does, a training forward under the torch shim dies there. The merge base returns False and falls back to the ops VJP. I moved the two device checks to the front. They are the cheapest clauses and the ones that settle the question off Metal, and since every operand is side-effect free, reordering an all-`and` chain changes nothing on Apple Silicon. Two tests come with it, both skipped on Metal.

This does not affect users on Apple hardware, where `mx.Dtype.size` exists. It affects the repo's own Linux simulation lane.

## CI coverage, fixed in f0e90c2f

`tests/test_mlx_gated_delta.py`, `test_mlx_gated_delta_vjp.py` and `test_mlx_gated_delta_batch_grad.py` were run by **no workflow at all**. The `mlx-cpu-linux` lane drives a named-file list that omitted them, and `mlx-ci.yml` is opt-in behind the `mlx` label and names only `test_qwen35_vjp_metal.py`. That is 47 tests that pass today with nothing running them, including the batched-gradient regression from #776 and the consumer-binding sweep. I added them to the suites venv with the same did-anything-actually-run gate the other numeric files there have.

## Three things for you, no action required from me

1. **The framing in the description.** "Training on MLX with a gated-delta model currently takes an ops VJP through the recurrence" is not what the merge base does for eligible shapes. At `0f372d4b` an eligible training call already ran a fused path: mlx-lm's `gated_delta_kernel` forward plus `_make_gd_chunk_states_kernel` and `_make_gd_chunk_backward_kernel` for the backward, at a fixed 64-token chunk. The ops VJP was the fallback. Your measurements are against that merge base so the numbers stand, but as written it reads as introducing the first fused backward rather than replacing one. Worth a sentence, since the actual win, dropping the every-step state buffer for block checkpoints and letting the chunk grow, is a better story than the one currently told.

2. **Eligibility moves both ways.** Sweeping 2808 shapes with the device guards stubbed true: 310 gain the fused path (small `Dv` of 1, 2, 3, 5, 6, 7, 10, which base rejected via `Dv % 4`), and 266 lose it, all at `Dk` in {288, 640, 768, 1024}, the last two only with vector gating. `Dk = 288` is structurally unplaceable: `W = 288/P` has to be a multiple of 4 and at most 32, and no power-of-two `P` satisfies both. Every real head dim (64, 128, 256 against `Dv` 64, 128, 256, both gating kinds) stays eligible, so I do not think this blocks anything, but a lost shape is a silent fall back to the ops VJP rather than an error, and it is worth knowing it is intentional.

3. **The checkpoint budget is one-sided.** `_kernel_chunk_size` clamps up to `_KERNEL_CHUNK_MIN`, so when a single state already exceeds 128 MiB the floor wins and the budget is not honoured: at `B=16, Hv=64, Dk=Dv=256, TB=4` the checkpoints come to 3.75 GiB. That is still under half what the merge base would have allocated at the same shape, so it is not a regression, and the configuration is unrealistic. Separately, at `TB=4` with a 512-token chunk the boundary count reaches 127 against the base's 64, so in the small-state regime the head can hold about 2x the base's buffer while staying far below the budget in absolute terms. The comment says "longest chunk whose checkpoints fit the budget", which is true only when the budget binds.

Also worth noting for whoever merges: `gated_delta_kernel_supported` gained a `k` parameter. All three in-repo call sites were updated, but an external caller still passing four arguments now gets False rather than True, which is a silent loss of the fused path rather than a `TypeError`.

danielhanchen:
@codex review

danielhanchen:
@codex review

PR #1145, opened 2026-09-02
Guard the disabled-hook eager fallback against checkpointed regions

danielhanchen:
## What this changes

`_fall_back_to_eager_on_recompile_limit` in `unsloth_zoo/temporary_patches/utils.py` has four arms that can move a wrapper from compiled to eager. Three of them ask `_in_non_reentrant_checkpoint()` / `_PACKED_COMPILED_IN_CHECKPOINT` before changing mode, and either defer to the step boundary or end the step.

The fourth did not. The arm that catches Dynamo refusing to inline one of our own `torch.compiler.disable`d gradient checkpointing hooks was:

```python
if not _is_our_own_disabled_hook(e):
    raise
state["eager"] = True
_warn(...)
return eager_func(*args, **kwargs)
```

It flipped on the spot, with no checkpoint question asked, no deferral, and no entry in either `_LATCHED_EAGER_LABELS` or `_RECENT_EAGER_LABELS`.

## Why it matters

Dynamo only discovers a disabled hook while tracing. Under `@torch_compile_with_fallback(fullgraph = True, dynamic = True)` the first trace of a recompute-only path therefore happens during the backward, after the forward already packed compiled activations. Flipping to eager there desynchronises the pack and the recompute of a live checkpointed region.

The failure that comes out of it is a metadata error rather than a count error, which is worth spelling out because it looks confusing in a log. For an RMSNorm, measured on this branch:

| mode | tensors packed |
|---|---|
| eager | 6 |
| compiled (static or dynamic, identical) | 3 |

The forward packs 3, the eager recompute wants 6, and `early_stop` raises `_StopRecomputationError` at exactly 3. So `len(weak_holders) == recomp_counter`, torch falls through to the per-index comparison, and you get:

```
CheckpointError: Recomputed values for the following tensors have different
metadata than during the forward pass
```

## The fix

`_give_up_on_backend` already had the right contract for a refusal decided at compile time, so the two now share it. Its body becomes `_give_up_at_compile_time(e, args, kwargs, marker_before, warning)`, and the disabled-hook arm reaches it through `_give_up_on_disabled_hook`.

The arm now restores `marker_before`, records the latch in both label sets, and when `(_in_non_reentrant_checkpoint() or _PACKED_COMPILED_IN_CHECKPOINT) and label in _COMPILED_OK_LABELS` it sets `_RAISED_INSIDE_CHECKPOINT` and re-raises instead of running eager mid-region. Silent backward corruption becomes a well defined "end this step, everything is latched eager, the retry is consistent".

Two deliberate limits on the blast radius:

- A first-call refusal, where nothing compiled has been packed by this label, still falls back to eager exactly as before. The common case does not change.
- Only this label latches, not every borrower. Budget exhaustion is process wide and takes the other borrowers with it; a compile-time refusal is a property of one region's graph, so knocking unrelated regions eager would cost their compilation for nothing.

## Tests

`test_a_disabled_hook_does_not_flip_a_region_that_packed_compiled` in `tests/test_checkpoint_compile_mix.py` fails on `main` and passes here, so it discriminates rather than decorates. Verified by running the new file against the unmodified `main` source, not just against its own branch. Two further tests are non-regression guards and pass on both.

`tests/test_disabled_hook_graph_break.py` gets an autouse fixture that resets label and marker state, because every `_wrap` in that file reuses the label `"TestMod"` and this arm now records by label the way the codegen arm always has.

Across the 24 test files touching the compile and fallback machinery, torch 2.9.1+cu128 and transformers 5.5.0:

| tree | result |
|---|---|
| `main` | 805 passed, 19 skipped |
| this branch | 808 passed, 19 skipped |

Also green on transformers 4.57.6 with torch 2.9.1+cpu.

## What this does not do

**This does not close the Gemma 4 gradient checkpointing report, and should not be merged as though it does.** The state captured on that run was:

```
latched:         []
pending:         ["Gemma4RMSNorm.forward (scaled)"]
packed_compiled: true
```

with the warning "ran out of recompilation cache ... switching to eager at the next step". This arm never touches `_PENDING_EAGER_LABELS` and emits different text, so it cannot produce that state. That report is the budget path in `_retry_with_more_budget`, where the retry recompiles inside the recompute and produces a different variant than the one the forward packed. Tracked separately.

So this is a real inconsistency found while investigating that bug, fixed on its own merits, rather than a fix for it.

## Scope

One file changed under `unsloth_zoo/`, plus two test files. No new torch APIs: it reuses `_in_non_reentrant_checkpoint`, `_COMPILED_OK_LABELS` and `_RAISED_INSIDE_CHECKPOINT`, all already version guarded, so the portability profile is the same as `_give_up_on_backend`'s. Parses under the 3.9 grammar floor declared in pyproject.

Only torch 2.9.1+cu128 was available to test on, so the behaviour is unverified on 2.6 through 2.13 and on Mac and Windows.

### DIFF
--- a/.github/workflows/consolidated-tests-ci.yml
+++ b/.github/workflows/consolidated-tests-ci.yml
@@ -296,6 +296,11 @@ jobs:
         # test_wheel_top_level_packages pins that discovery ships unsloth_zoo and
         # nothing else; collect-only would let the packaging config regress.
         #
+        # test_disabled_hook_checkpoint_guard pins the disabled-hook arm of the
+        # eager fallback: that it still falls back for a first-call refusal, that
+        # it ends the step instead when a compiled pack is outstanding, and that
+        # only its own label latches. CPU-pure, no model weights.
+        #
         # test_loss_normalization_contract pins num_items_in_batch: the grad-accum
         # guard, and the packed / padding-free boundary count. The second one
         # shipped broken for the same reason this comment keeps repeating -- the
@@ -356,6 +361,7 @@ jobs:
             tests/test_missing_parent_package_is_drift.py \
             tests/test_sibling_imports_do_not_go_through_a_tests_package.py \
             tests/test_hip_bf16_offload_dtype.py \
+            tests/test_disabled_hook_checkpoint_guard.py \
             -v
 
       - name: pytest MLX adapter-filter gate (HARD GATE)
--- a/.github/workflows/consolidated-tests-ci.yml
+++ b/.github/workflows/consolidated-tests-ci.yml
@@ -297,9 +297,8 @@ jobs:
         # nothing else; collect-only would let the packaging config regress.
         #
         # test_disabled_hook_checkpoint_guard pins the disabled-hook arm of the
-        # eager fallback: that it still falls back for a first-call refusal, that
-        # it ends the step instead when a compiled pack is outstanding, and that
-        # only its own label latches. CPU-pure, no model weights.
+        # eager fallback: still falls back on a first-call refusal, ends the step
+        # when a compiled pack is outstanding, latches only its own label. CPU-pure.
         #
         # test_loss_normalization_contract pins num_items_in_batch: the grad-accum
         # guard, and the packed / padding-free boundary count. The second one
--- a/tests/test_checkpoint_compile_mix.py
+++ b/tests/test_checkpoint_compile_mix.py
@@ -443,3 +443,126 @@ def test_the_last_sequential_segment_is_outside_every_region():
     is packed while it executes and eager is safe there even when the earlier
     segments are non-reentrant."""
     assert _sequential_walk(2, False) == [True, True, False, False]
+
+
+# ---- the disabled-hook arm, which used to flip with no question asked -----
+
+def _disabled_hook_error():
+    """Dynamo's refusal to inline a `torch.compiler.disable`d callable."""
+    import torch._dynamo.exc as exc
+    cls = getattr(exc, "Unsupported", None)
+    if cls is None:
+        pytest.skip("this torch has no torch._dynamo.exc.Unsupported")
+    try:
+        return cls(
+            "Skip calling `torch.compiler.disable()`d function\n"
+            "  Explanation: Skip calling function `requires_grad_pre_hook` "
+            "since it was wrapped with `torch.compiler.disable`\n"
+            "  Hint: Remove the `torch.compiler.disable` call"
+        )
+    except Exception:
+        pytest.skip("Unsupported cannot be constructed on this torch")
+
+
+@pytest.fixture
+def hook_label():
+    label = "test_disabled_hook_under_checkpoint"
+    for s in (U._LATCHED_EAGER_LABELS, U._PENDING_EAGER_LABELS,
+              U._RECENT_EAGER_LABELS, U._COMPILED_OK_LABELS):
+        s.discard(label)
+    packed = U._PACKED_COMPILED_IN_CHECKPOINT
+    yield label
+    for s in (U._LATCHED_EAGER_LABELS, U._PENDING_EAGER_LABELS,
+              U._RECENT_EAGER_LABELS, U._COMPILED_OK_LABELS):
+        s.discard(label)
+    U._PACKED_COMPILED_IN_CHECKPOINT = packed
+    U._RAISED_INSIDE_CHECKPOINT = False
+    U._CHECKPOINT_SETTLE_ATTEMPTS = 0
+
+
+@pytest.mark.skipif(
+    U._in_non_reentrant_checkpoint() is None,
+    reason = "torch < 2.8 cannot report whether a checkpoint region is live",
+)
+def test_a_disabled_hook_does_not_flip_a_region_that_packed_compiled(hook_label):
+    """The Gemma4 abort.
+
+    Dynamo only discovers a disabled hook while TRACING, so this arm fires on a
+    recompile -- routinely the backward recompute of a region whose forward
+    already packed compiled. It used to set `state["eager"]` and run eager on
+    the spot, with none of the checkpoint question the other give-up arms ask,
+    and torch reported
+
+        CheckpointError: Recomputed values for the following tensors have
+        different metadata than during the forward pass
+
+    with `_LATCHED_EAGER_LABELS` still empty, because this arm recorded nothing.
+    """
+    calls = {"n": 0}
+
+    def compiled(*args, **kwargs):
+        calls["n"] += 1
+        if calls["n"] == 1:
+            return "compiled"
+        raise _disabled_hook_error()
+
+    fn = U._fall_back_to_eager_on_recompile_limit(
+        compiled, lambda *a, **k: "eager", hook_label)
+
+    def region(x):
+        # First call packs compiled, so the region owes a compiled recompute.
+        assert fn(x) == "compiled"
+        # A later trace in the same region refuses. Eager here is the abort.
+        return fn(x) * x
+
+    with pytest.raises(Exception) as ei:
+        checkpoint(region, torch.randn(4, 4, requires_grad = True),
+                   use_reentrant = False)
+    message = str(ei.value)
+    assert "torch.compiler.disable" in message, message[:200]
+
+    assert hook_label in U._LATCHED_EAGER_LABELS, "the flip must be recorded"
+    assert hook_label in U._RECENT_EAGER_LABELS, "and as this step's evidence"
+    assert U._RAISED_INSIDE_CHECKPOINT, "the abandoned region must be settled"
+    assert fn._unsloth_fallback_state["eager"], "the retry has to be eager"
+
+    # `ei` roots the traceback, which roots the abandoned generator and leaves
+    # its saved-tensor hooks installed for every later test in the process --
+    # the case `_settle_abandoned_checkpoint_generator` documents.
+    del ei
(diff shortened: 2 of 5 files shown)


danielhanchen:
Verification pass on this PR, since it changes shared machinery rather than one model's patch.

## Before and after

Before, the disabled-hook arm was `state["eager"] = True; return eager_func(...)`. It never raised, never wrote to the global label sets, and never restored the packed marker. After, it goes through `_give_up_at_compile_time`: it restores the marker, records the latch in `_LATCHED_EAGER_LABELS` and `_RECENT_EAGER_LABELS`, and raises instead of running eager when this label already compiled successfully and a region is live.

## Real issue or fake

Real, and narrower than the Gemma 4 report it was found under. The arm was the only one of four that changed compile mode with no checkpoint question asked. Flipping mid-region desynchronises a pack and its recompute, which is silent wrong gradients rather than an error, because `early_stop` truncates the recompute to the forward's tensor count so the counts agree and only the metadata differs.

## The behaviour change a merger should weigh

This converts some runs that previously continued into runs that stop. That is the point, since what they continued into was a desynchronised backward, but it is user visible and there is no automatic recovery today: `force_eager_fallback` has **no caller anywhere in unsloth**. `unsloth/models/_utils.py` only wires `apply_pending_eager_fallbacks` into `Trainer.training_step`. So a user hits a fatal `Unsupported` rather than a retried step. Making that retry automatic is an unsloth-side change, tracked with the rest in #1146.

The blast radius is bounded deliberately in two ways, both pinned by tests: a first-call refusal, where nothing compiled has been packed by this label, still falls back to eager exactly as before; and only this label latches, not every borrower, unlike budget exhaustion which is process wide.

One place it is deliberately pessimistic, now pinned by `test_conservative_when_another_region_packed_compiled_this_step`: `_PACKED_COMPILED_IN_CHECKPOINT` is global for the step and `_COMPILED_OK_LABELS` is sticky for the process, so the arm can raise when this particular wrapper packed nothing in this region. That is the same trade `_give_up_on_backend` has always made, not a new one.

## Simulation suite

13 tests written against the real wrapper, run on `main` and on this branch so every result is a pair. Weighted towards "does the old case still work" rather than "does the new case work", because that is where the risk is. Covering: a refusal outside any checkpoint, a first-call refusal inside a live region, both the pre-2.7 and post-2.7 refusal spellings, an unrelated graph break still raising (the over-catching hazard), the new raise, label recording, a rebuilt wrapper inheriting a latch, unrelated labels keeping their compilation, marker restore in both directions, the codegen arm still behaving, and an assertion that the shared give-up path contains no platform or device branch.

## Torch matrix

Each cell a throwaway `uv venv`, CPU wheels:

| torch | disabled-hook detection (new / old spelling) | budget config names | simulation suite |
|---|---|---|---|
| 2.6.0 | True / True | `cache_size_limit` only | 12 passed, 1 skipped |
| 2.7.1 | True / True | both | 12 passed, 1 skipped |
| 2.8.0 | True / True | both | 12 passed, 1 skipped |
| 2.9.1 | True / True | both | 12 passed, 1 skipped |
| 2.10.0 | True / True | both | 12 passed, 1 skipped |
| 2.11.0 | True / True | both | 12 passed, 1 skipped |
| 2.12.1 | True / True | both | 12 passed, 1 skipped |
| 2.13.0 | True / True | both | 12 passed, 1 skipped |

Detection matters because it is matched on Dynamo's message text, which was respelled at 2.7, and a miss means the exception escapes as a hard error. Both spellings are recognised on every version, and 2.6.0 correctly shows only the old budget config name.

## Regression checks

| suite | main | this branch |
|---|---|---|
| 24 zoo files touching compile and fallback machinery (torch 2.9.1+cu128, transformers 5.5.0) | 805 passed, 19 skipped | 808 passed, 19 skipped |
| same files, transformers 4.57.6 + torch 2.9.1 cpu | 530 passed, 27 skipped | 533 passed, 27 skipped |
| unsloth core: settle-eager-fallbacks, gradient-checkpointing-restore, pretrain-compile-reset, bnb-checkpoint-skip-modules | 23 passed | 23 passed |

The core suite matters because unsloth has 14 `fullgraph = True` sites that consume this wrapper.

While running the matrix, `test_checkpoint_modes.py` and `test_generated_fullgraph_fallback.py` each failed one test. That was a missing `datasets` in my minimal venv, confirmed by reproducing both failures identically on unmodified `main` in the same venv. Not this change.

## Platform and Studio

No browser or Studio surface. `unsloth-studio` has zero `fullgraph = True` sites and does not import `torch_compile_with_fallback`, `apply_pending_eager_fallbacks` or `force_eager_fallback`, so it never builds one of these wrappers. The shared give-up path contains no `sys.platform`, `platform.system`, `cuda`, `hip`, `mps`, `darwin` or `win32` reference, which is asserted rather than claimed. Parses under the 3.9 grammar floor.

Old installs are unaffected: core already guards the zoo import with `try/except` and treats a missing `apply_pending_eager_fallbacks` as "older unsloth_zoo has no deferred switch", so old core with new zoo and new core with old zoo both hold.

Only Linux with torch CPU wheels plus one B200 was available, so Mac and Windows behaviour is inferred from the absence of platform branches rather than executed.

danielhanchen:
CI green on `75998d88`: Tests CI, Lint CI and Wheel CI all pass, MLX skipped as expected for a change with no MLX surface. All four Tests CI jobs pass, including Core drift across the 3 upstream pins.

The new file executes in the hard gate rather than only being collected. From the log:

```
pytest behavioral gates (HARD GATE)
  test_disabled_hook_outside_any_checkpoint_still_falls_back_to_eager PASSED
  test_first_call_refusal_inside_checkpoint_still_falls_back          PASSED
  test_old_torch_message_signature_also_falls_back                    PASSED
  test_an_unrelated_graph_break_still_raises                          PASSED
  test_refusal_after_compiling_ok_inside_a_live_region                PASSED
  test_fix_records_the_latch_in_the_global_label_sets                 PASSED
  test_a_new_wrapper_with_a_latched_label_starts_eager                PASSED
  test_only_this_label_latches_not_every_borrower                     PASSED
  test_marker_is_restored_to_what_it_was                              PASSED
  test_marker_true_is_preserved_when_no_raise_happens                 PASSED
  test_backend_refusal_arm_still_behaves                              SKIPPED
  test_changed_path_has_no_platform_or_device_branch                  PASSED
  test_conservative_when_another_region_packed_compiled_this_step     PASSED
```

`test_backend_refusal_arm_still_behaves` skips on the CI runner because it needs a specific `BackendCompilerFailed` shape that this torch does not expose. It passes locally where that shape exists. Worth knowing so nobody reads the skip as a silent hole: the codegen arm's own behaviour is covered by `tests/test_inductor_codegen_fallback.py` in the same gate.

Two earlier attempts on this sha were cancelled rather than failed, both times with Tests CI and Wheel CI dying at the same instant while the shorter Lint job survived. No repo-wide sweep at those timestamps, so it looked like contention on the org queue rather than anything about this branch. Re-triggered and they pass.

Ready for review.

danielhanchen:
Correcting something I said above. I wrote that there was no repo-wide sweep at those timestamps. That was based on too narrow a window. Looking at the last 90 minutes across the whole repo, cancellations are landing on `main` as well:

| branch | runs cancelled |
|---|---|
| fix/disabled-hook-checkpoint-guard | 5 |
| refs/pull/1145/head | 2 |
| main | 4 |

So it is a repo-wide sweeper cancelling active runs, not queue contention and not anything about this branch. The conclusion is unchanged, but the evidence for it is better than what I first gave.

Where that leaves the checks. Everything substantive already passed on `75998d88` and those results stand: Tests CI with all four jobs, Lint CI, Wheel CI, with the new guard suite executing inside the behavioral gates hard gate rather than only being collected. The rollup reads UNSTABLE because of leftover cancelled entries, including `Analyze (python)` and `Analyze (actions)`, which are CodeQL default setup and cannot be re-run through the API.

I am going to stop re-triggering rather than keep feeding the sweeper. Anyone merging can re-run the checks once the sweeping stops. Nothing here is a test failure.

danielhanchen:
Filled in the two axes the torch matrix did not cover.

**transformers 5.x** (the torch matrix pinned 4.57.6, so the guard suite had never run against a 5.x line). py3.12, torch 2.9.1:

| transformers | `tests/test_disabled_hook_checkpoint_guard.py` | fallback machinery suite |
|---|---|---|
| 5.0.0 | 12 passed, 1 skipped | 521 passed, 28 skipped, 1 failed |
| 5.5.0 | 12 passed, 1 skipped | 521 passed, 28 skipped, 1 failed |
| 5.16.1 | 12 passed, 1 skipped | 521 passed, 28 skipped, 1 failed |

The one failure is `test_generated_fullgraph_fallback.py::test_the_grpo_loss_regions_are_wrapped`, and it is not from this branch. Ran that single test on a fresh clone of unmodified main (ff68556) and on this tree, on both pins:

```
transformers==4.57.6   main: 1 failed   this branch: 1 failed
transformers==5.16.1   main: 1 failed   this branch: 1 failed
```

Identical on both trees, so it is pre-existing and orthogonal. It passes when the whole suite runs in order and fails when run alone, which points at test ordering rather than at either tree.

**Python 3.10 through 3.13** (torch 2.9.1, transformers 4.57.6): the gated suite passes on all four.

Re-ran the runs that got cancelled earlier so the checks reflect the head commit rather than the sweep.

danielhanchen:
@codex review

danielhanchen:
@codex review

PR #1153, opened 2026-09-04
test: update getsource guard spelling to 3-tuple

anmolxlight:
Follow-up to #1149, which widened both getsource guards to `except (OSError, TypeError, tokenize.TokenError)` but left the test asserting the old 2-tuple spelling. This fixes `test_it_catches_what_getsource_actually_raises`, `test_the_nn_forward_patch_loop_is_guarded` and `test_both_getsource_guards_are_present` — currently failing on zoo main and breaking downstream unsloth Core CI (which installs zoo@main). Test-only change, 3 lines.


danielhanchen:
main now asserts the widened guard in tests/test_unreadable_modeling_source.py using the _catches helper rather than the literal tuple spelling, so I do not think this is needed any more. Sorry for the slow review, and thanks for putting it together.

Issue #1122, opened 2026-08-27
Xet transport errors (503 from xet-bridge CDN) bypass the HTTP fallback, then unsloth continues into an unguarded in-process load and hangs


olsh:
### Summary

A **transient Xet-side error** (HTTP 503 from the `xet-bridge` CDN, or a CAS decode error) is classified as a deterministic child error and re-raised, so the HTTP fallback never runs. `unsloth` then swallows that exception and continues with the **in-process** load — which still has Xet enabled and no stall watchdog — turning a fast, retryable CDN error into an unbounded silent hang.

Observed twice in a row on fresh cloud GPU pods while HF's Xet CDN was degraded: the base-model download froze at ~3 GB of ~8 GB with the NIC at **~42 KB/s** and the GPU idle, for 15+ minutes, until killed. A plain `curl` of the same shard from the same pod, in the same minute, ran at **6.6 MB/s**.

### Environment

| | |
|---|---|
| unsloth | 2026.8.21 |
| unsloth_zoo | 2026.8.15 |
| huggingface_hub | 1.28.0 |
| transformers | 5.5.0 |
| torch | 2.10.0+cu128, CUDA 12.8 |
| model | `unsloth/Qwen3.5-4B` (bf16, `load_in_16bit`) |
| hardware | RTX A4500 20 GB, Linux, RunPod community cloud |

### What happened

```
[unsloth_zoo.hf_xet_fallback|WARNING] Download stalled for 'unsloth/Qwen3.5-4B' -- retrying on Xet (attempt 2 of 2)
[unsloth_zoo.log|WARNING] Unsloth: Could not pre-download unsloth/Qwen3.5-4B
    (RuntimeError: HfHubHTTPError: Server error '503 Service Unavailable' for url
     'https://us.aws.cdn.hf.co/xet-bridge-us/<repo>/<hash>?***'); continuing with the normal load.
```

On another pod the same position produced `RuntimeError: Task error: File reconstruction error: CAS Client Error: Format error: I/O error: error decoding response body`.

After that line the process never progresses again. Measured on the wedged pod:

| probe | value |
|---|---|
| `du -sh $HF_HOME` over 20 s | 2.9 GB → 2.9 GB (unchanged) |
| `*.incomplete` blob mtime | frozen 15 min earlier |
| eth0 receive | **~42 KB/s** |
[...]

PR #1156, opened 2026-09-05
Reduce comment volume in tests/

danielhanchen:
Trims restating and boilerplate comments across 71 files in `tests/`, the first batch of a comment reduction pass over this repo. Net 771 comment lines removed.

## What was kept

The dominant comment register here is evidence, not description, so the rule applied was: the sentence that names the failure mode or the model survives, the sentence that restates the code goes. Explicitly preserved:

- named failure modes ("would silently train on nothing", "crashes on first forward", "rather than iterating its characters")
- model, version and platform names (Gemma-4, Falcon-H1, Qwen2-MoE, transformers 5.10-5.14, mlx-vlm 0.6.4, macos-14)
- measured numbers and literals (`atol=1e-4, rtol=1e-3`, the 2 GB xet-core default, `0600` / umask `0002`, `HIDDEN = 1152`, `2300 * 33.94 = 78k > 65504`)
- upstream references (`unslothai/unsloth-zoo#1122`, `unsloth#6273`, `vllm/v1/worker/utils.py request_memory`, `compiler.py:66-67`)
- aligned mini-tables and the legends that make bare tuples readable, for example the `(B, T, Hk, Hv, Dk, Dv, dtype, tol, vectorized)` key in `test_mlx_gated_delta_vjp.py`
- every functional comment: `# noqa`, `# pragma: no cover`, `# type: ignore`
- all 15 line licence headers, byte for byte

A review pass over the full diff restored 73 comments that the first pass cut too far, mostly cases where a comment had been truncated mid-sentence and left a fragment with no subject. Those are fixed here rather than in a follow-up.

## Verification against main

- **No code changes.** For all 71 files the non-comment token stream is byte-identical to main. This is stronger than an AST comparison: it also proves no `#` inside a string literal was altered, which matters because several tests use commented source as fixture data.
- **Comment-only diff**, via the in-repo `scripts/verify_comment_only_diff.py`: 71/71 OK.
- **Codegen unchanged**, 96 files, no rewriter in `compiler.py` changes its decision. This repo rewrites Python source text at runtime, and several finders anchor on consecutive lines, so a comment sitting between them suppresses a rewrite and deleting it turns the rewrite on. That change is invisible to an AST diff.
- **Comment sentinels intact**: the `vllm_utils.py` `# LM Head` slice anchor and the 800 character window after it, the PrefixGrouper anchor whose slice is `exec`d, the mlx fp16 phrase asserted by a hard-gated test, `_AITER_MARKER`, `_full_license_header`, and the versioning header whose line count is parsed positionally.
- **Licence headers** byte-identical in all 50 files that carry one.
- **Lint**: `ruff check` and `ruff format --check` produce identical results to main; the narrow CI gate (`E9,F63,F7,F82`) and `compileall` are clean.

## Tests

- mlx files, one process each since the shim replaces `sys.modules["mlx.core"]` process-wide: 222 passed, 49 skipped

`test_rl_replacements_compile_disable.py` reports 4 failures. That file is byte-identical to main and the same 4 failures reproduce on a clean main checkout, so they are unrelated to this change.

### DIFF
--- a/tests/test_check_hf_model_exists_transport_errors.py
+++ b/tests/test_check_hf_model_exists_transport_errors.py
@@ -203,7 +203,6 @@ def test_a_plain_absent_repo_does_not_warn_about_gating(monkeypatch):
     assert [w for w in caught if "gated" in str(w.message)] == []
 
 
-# The success paths are unchanged.
 
 def test_repo_with_safetensors_returns_true(monkeypatch):
     _patch_ls(monkeypatch, [
--- a/tests/test_checkpoint_compile_mix.py
+++ b/tests/test_checkpoint_compile_mix.py
@@ -41,17 +41,13 @@ from torch.utils.checkpoint import checkpoint
 from unsloth_zoo.temporary_patches import utils as U
 
 
-# aot_eager, not inductor: this file tests the compile-mode switch and what
-# autograd saves, both of which AOTAutograd already covers. Inductor adds a C++
-# codegen step that fails for reasons of its own once an earlier test touched
-# the inductor config -- why these passed alone and failed in a full run.
+# aot_eager, not inductor: inductor's C++ codegen fails once an earlier test touched its config.
 _BACKEND = "aot_eager"
 
 
 def _norm(w, x, k):
-    # `k` is a plain int, so Dynamo specialises on its value and every block
-    # forces a fresh compilation. A real vision run exhausts the cache the same
-    # way, one guard per shape rather than per constant.
+    # `k` is a plain int, so Dynamo specialises on it and every block recompiles, the
+    # way a real vision run exhausts the cache one guard a shape.
     return x * torch.rsqrt(x.pow(2).mean(-1, keepdim=True) + 1e-6) * w * (1.0 + 0.0 * k)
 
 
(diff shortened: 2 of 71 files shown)


danielhanchen:
Idempotency evidence, run on a staging copy of this branch so the main CI queue is not used for it.

Two properties matter for a comment-only change in this repo, because `compiler.py` rewrites source text and several of its finders anchor on consecutive lines:

**I1, rewriter decisions unchanged.** For every rewritable module, each source-to-source rewriter was run over the base text and the head text, comparing whether it fired, how many lines it inserted, and a hash of those lines. 480 rewriter checks, 12 firings, **0 decision changes**.

**I3, rewriters are fixed points.** `f(f(src)) == f(src)`. **0 non-fixpoint rewriters.**

Both checks were mutation-calibrated rather than just run, since a check that only ever passes proves nothing:

- Shortening the two-line comment in `temporary_patches/ministral.py` to one line, a purely cosmetic edit, moves the `value_states` cast inside `fix_attention_dtype_consistency`'s 200 character lookahead at `compiler.py:856` and flips it from inserting a 3-line cast to skipping. I1 caught it; firings dropped 12 to 11. The AST is unchanged, the tests still pass, and `scripts/verify_comment_only_diff.py` reports the file clean, so nothing else in the tree can see this.
- Making the marker emitted at `compiler.py:548` diverge from the `_AITER_MARKER` guard tested at `:507` makes `replace_sdpa_with_amd_aiter` re-wrap its own output, +49 lines on a second pass. I3 caught it.

One note on the second case. That rewriter returns at its first line unless `get_amd_attention_implementation() == "amd_aiter"`, so on any non-AMD runner it never fires and a fixpoint check over the repo's own files passes while executing nothing. It is exercised on a fixture with the device probe stubbed, and the check asserts the fixture was actually rewritten before concluding anything about stability.

I2 from the original plan, "re-running the trim is a fixpoint", is not reported here: this trim was done by judgement rather than by a script, so there is no trimmer to re-run and no honest way to test that property.

PR #1158, opened 2026-09-05
Reduce comment volume in the rest of tests/

danielhanchen:
Finishes the pass started in #1156. That one covered 71 of the 281 files in `tests/`; this trims 114 of the remaining 210. Net 1391 comment lines removed.

## What went, and what stayed

Removed: section banners and bare `# ----` / `# ====` rule lines, `Step 1` through `Step N` scaffolding, and comments that restate the assertion directly below them (`# Base weights stay frozen.` above the assert that checks exactly that).

Kept: named failure modes, measured values, version windows (`transformers 5.10-5.14`, `trl 0.20.0`, `bitsandbytes < 0.46.0`), upstream commit and issue references, model and platform names, aligned mini-tables and the legends that make bare tuples and subscripts readable, every "why this test greps source instead of running it" note, and all functional comments (`# noqa`, `# pragma: no cover`, `# type: ignore`).

Yield varies a lot by file, because it depends on whether a file was written comment-first or comment-last. Several files gave 30 percent or more; around thirty gave nothing and were left untouched rather than padded. For example `test_moe_gategrad_identity.py` kept all 33 of its comments: they carry the fp16 underflow failure mode, `eps 1e-12` against fp16's smallest normal of about 6e-5, and `rtol=5e-4 / atol=5e-5` with the accumulation-order reason.

## One non-comment change

Removing a standalone banner leaves three consecutive blank lines where it used to sit. 238 blank lines were collapsed back to the file's own convention, two before a top-level statement and one before an indented one. That is the only part of the diff that is not a comment, and it is whitespace only.

## Verification against main

- **No code changes.** For all 114 files the non-comment token stream is byte-identical to main. This is stronger than an AST comparison: it also proves nothing inside a string literal moved, which matters because several tests here embed fake source containing `#` lines and assert on it.
- **Comment-only diff** via the in-repo `scripts/verify_comment_only_diff.py`: 114/114 OK.
- **125 comment-dependent cross-file assertions still resolve.** Meta-tests in this suite read sibling test files' source, so a reworded comment in one file can break an assertion in another. Each of those 125 is a literal that exists only inside a comment in the file that owns it, and all are byte-frozen here.
- **Codegen unchanged.** No rewriter in `compiler.py` changes its decision, `create_new_function`'s import-selection substring test is stable over 2335 symbols, and all 8 `re.sub` replacement strings are byte-identical. That last one matters because an introduced backslash in a replacement string corrupts every substitution that uses it.
- **No orphaned or truncated comments introduced**, checked with a detector calibrated on both defect shapes.
- **Licence headers** byte-identical in all 87 files that carry one.
- **Lint**: narrow CI gate (`E9,F63,F7,F82`), `compileall`, and both dynamic-execution linters clean.

## Tests

Base and head measured on the same host and compared file by file, not just in aggregate:

- all 51 mlx files, one pytest process each, since `tests/mlx_simulation/` replaces `sys.modules["mlx.core"]` process-wide

Identical in every partition. The 4 failures in `test_rl_replacements_compile_disable.py` reproduce on a clean main checkout, and that file is byte-identical to main here.

Worth noting that a number of the mlx files are Metal gated and skip on Linux, so the macOS CI leg is the real signal for those paths rather than a local run.

## Review pass

A review of the full diff restored 67 comments that the first pass cut too far, before this PR opened rather than as a follow-up. Those fell into three groups: facts the surviving text no longer carried (the `torch <= 2.9.x` allocator boundary getting expandable plus roundup where 2.10 gets expandable only), legends for otherwise bare values (`dim = 16  # hidden == intermediate`, `# scale is 2-D (64, 1)`), and structure keyed to something outside the file (20 rule lines mapping onto the numbered coverage matrix in a module docstring). Every restoration was taken from the `origin/main` bytes rather than retyped, so a restore cannot quietly reword what it claims to restore.

### DIFF
--- a/tests/_merge_e2e_helpers.py
+++ b/tests/_merge_e2e_helpers.py
@@ -252,7 +252,6 @@ def assert_merge_correct_model_space(*, family, base_dir, out_dir,
         f"-{sorted(set(base_p) - set(merged_p))[:4]})"
     )
 
-    # map each adapted key onto its parameter name (usually identical)
     resolved: dict[str, _Adapted] = {}
     for ak, a in adapted.items():
         if ak in base_p:
@@ -301,7 +300,6 @@ def assert_merge_correct(*, family, base_tensors, out_dir, save_dtype,
     assert merged, f"[{family}] no safetensors written to {out_dir}"
     base_keys = set(base_tensors.keys())
 
-    # no adapter remnants leaked
     leaked = [k for k in merged if any(m in k for m in _LORA_REMNANT_MARKERS)]
     assert not leaked, f"[{family}] adapter remnants in merged output: {leaked[:5]}"
 
@@ -320,7 +318,6 @@ def assert_merge_correct(*, family, base_tensors, out_dir, save_dtype,
             adapted=adapted, save_dtype=save_dtype,
         )
 
-    # resolve adapted keys against real base keys (VLM prefix tolerance)
     resolved: dict[str, _Adapted] = {}
     for ak, a in adapted.items():
         rk = _resolve_key(ak, base_keys)
@@ -351,18 +348,15 @@ def assert_merge_correct(*, family, base_tensors, out_dir, save_dtype,
             _assert_equal(family, key, mt, base_tensors[key])
             n_passthru += 1
 
-    # every adapted target appeared and was checked
     if not allow_missing_adapted:
         missing = [k for k in resolved if k not in merged]
         assert not missing, f"[{family}] adapted targets missing from merged output: {missing}"
     assert n_adapted >= 1, f"[{family}] no adapted tensors were checked (LoRA not attached?)"
 
-    # MoE merge must not have fallen back
     assert _MOE_MERGE_STATE.get("fallback", 0) == 0, (
         f"[{family}] MoE merge fell back: {_MOE_MERGE_STATE}"
     )
 
-    # index consistency
     idx = read_index(out_dir)
     if idx is not None:
         wm = idx.get("weight_map", {})
@@ -412,7 +406,6 @@ class FamilySpec:
     fused_params: tuple = ()        # fused-expert target_parameters
 
 
-# distinct dims so fused-expert orientation is unambiguous (transpose bugs show)
 _H = 32
 _I = 64
 _MOE_I = 48      # distinct from _H and 2*_MOE_I
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -44,11 +44,6 @@ import sys
 import types
 
 
-# ---------------------------------------------------------------------------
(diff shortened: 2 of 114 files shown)


danielhanchen:
Idempotency evidence, run on a staging copy of this branch so the main CI queue is not used for it. All green there, including the full test suite.

This batch only touches `tests/`, so these are negative controls rather than discoveries. They are worth stating anyway, because the same harness is what will gate the batches that do touch `unsloth_zoo/`, and a control that passes is how you know the harness is wired up.

Four properties, all mutation calibrated rather than merely run:

**I1, rewriter decisions unchanged.** 480 rewriter checks over 96 files, 12 firings, 0 decision changes. Calibrated by shortening the two-line comment in `temporary_patches/ministral.py` to one line: that moves the `value_states` cast inside `fix_attention_dtype_consistency`'s 200 character lookahead and flips it from inserting a 3 line cast to skipping. Firings drop 12 to 11 and the check catches it. Nothing else in the tree can see that: the AST is unchanged, the tests pass, and `verify_comment_only_diff.py` reports the file clean.

**I3, rewriters are fixpoints.** `f(f(src)) == f(src)`, 0 non-fixpoint. Calibrated by making the marker emitted at `compiler.py:548` diverge from the `_AITER_MARKER` guard tested at `:507`: the rewriter then re-wraps its own output, +49 lines on a second pass.

**I4, import selection.** `create_new_function` picks which names to import with `x in new_source`, a plain substring test over text that includes comments. Deleting the last mention of a symbol drops its import; a reworded comment can add a spurious one. Stable over 2335 symbols. Calibrated by adding a comment naming a symbol absent from a file, which produces exactly that spurious import.

**I5, `re.sub` replacement strings.** All 8 sites byte-identical (`_cross_entropy_code`, `cross_entropy_replacement_1/2/3`, `COMPILED_LORA_FORWARD`, `_license_header`, `_disabled_sdpa_code`, `custom_gradient_checkpointing_replacements`). An introduced backslash in one of these corrupts every substitution that uses it. Calibrated by injecting one, 0 to 2 backslashes.

Two notes on the harness itself, since a check that quietly verifies nothing is worse than no check.

I5 first covered only 2 of the 8 sites while reporting a pass, because those 8 are five different AST node types (`Constant`, `Call`, `BinOp`, `JoinedStr`, `List`) and evaluating values only reached the plain constants. It now compares each assignment's source segment, which handles every node type and also catches a comment edited inside one of these literals. It fails loudly on any site it cannot locate rather than passing over it.

The aiter fixpoint check was vacuous at first for the opposite reason: `replace_sdpa_with_amd_aiter` returns at its first line unless the host is AMD, so it never fires here and `f(f(x)) == f(x)` held trivially. It now runs against a fixture with the device probe stubbed, and asserts the fixture was actually rewritten before drawing any conclusion about stability.

PR #1159, opened 2026-09-06
Reduce comment volume in unsloth_zoo pure-logic modules

danielhanchen:
Third batch of the comment reduction pass, after #1156 and #1158 (`tests/`) and #1157 (`unsloth_zoo/mlx/`). This is the first batch over source that `compiler.py` scans, so it is deliberately conservative: **182 comment lines removed across 14 files, deletion only, zero insertions.**

## How scope was set

A read-only safety triage ran over every candidate file before any edit, checking five things per comment: whether a test asserts on its text, whether it sits between two lines a `compiler.py` finder anchors on as consecutive, whether it is inside a string literal, whether it contains `def` above a `def`, and whether the file's source is read at runtime or by a test.

**Excluded outright**, about 7300 comments left untouched:
- `compiler.py`, `saving_utils.py`, `patching_utils.py`, `vllm_utils.py`, `llama_cpp.py`, `rl_replacements.py`
- everything under `temporary_patches/`, which holds every live rewriter firing and all the `find("def")` dedent sites
- `compile_cache.py`, `tiled_mlp.py`, `hf_xet_tuning.py`: triage found 3 to 6 safe lines each and judged the risk not worth it
- `dataset_utils.sft_prepare_dataset`: its source text is consumed by the `unsloth` repo's RL patcher, which is not in this tree. The function documents this itself at `_ccm = 'create_' + 'causal_mask_mapping'`, code deliberately split to defeat a substring scan

## What went, and what stayed

Removed: section banners (`device_map_planner.py` yielded exactly its six 3-line banners and nothing else), `Step 1` scaffolding, and comments restating the line directly below (`# Binary search on length`, `# Get untrained tokens`, `# Allocate one buffer per GPU`).

Kept: everything naming a failure mode, a model, a library version, a measured number, an errno, an upstream filename, an issue, or a security property. In these modules that is nearly all of them. `hf_xet_fallback.py` yields 1 line out of 467 because its short comments are wrapped continuations of measured-threshold comments. `dataset_utils.py` lines 1138 to 2262, about 400 comments, yield none: they record Fuyu `image_patches_indices`, Gemma 3 `num_crops`, Llama 3.2 Vision `num_tiles`, TRL 0.22.2 and costs like 94 seconds per 100k rows.

The editors also kept 22 lines triage had cleared. Several are the deletion-only rule working as intended: a two-line wrapped comment whose second line carries the fact is all or nothing, since salvaging half needs a reword. Others are semantic: `peft_utils.py:307` sits above a line that clears the hook dict, so it states the opposite of the following statement rather than restating it.

## Deletion only, as a rule

No surviving comment was reworded, rewrapped or re-indented. That makes two hazards unbreakable by construction rather than by care:

- **374 literals** that a test asserts against a module's source and that occur only inside a comment. Reflowing one of those blocks so the literal straddles a line break fails the test with every word retained.
- A package-wide scan in `tests/test_hf_xet_tuning.py` that asserts every *quoted* `HF_XET_*` name is a known variable. Deleting a comment can never trip it; rewording one so an env var name ends up in quotes would.

One specific line worth calling out: `empty_model.py:1154` is `# "model.multi_modal_projector.linear_2",` inside a set literal. It is commented-out data whose text a test asserts on, and a pass that treats commented-out code as deletable would take it. It is untouched.

## Verification against main

- **Codegen unchanged.** No rewriter in `compiler.py` changes its decision across 96 files. Unlike the earlier batches this is not a negative control: these files feed `create_new_function`'s `x in new_source` import filter, a plain substring test over text that includes comments, so deleting the last mention of a symbol would have dropped an import. It is stable over 2335 symbols.
- **All 8 `re.sub` replacement strings byte-identical**, and every comment sentinel intact (`_AITER_MARKER`, `_full_license_header`, the versioning header whose line count is parsed positionally, the `# LM Head` slice anchor and its 800 character window).
- **No code changes.** For all 14 files the non-comment token stream is byte-identical to main, which also proves nothing inside a string literal moved.
- **Comment-only diff** via the in-repo `scripts/verify_comment_only_diff.py`: 14/14 OK.
- **Licence headers** byte-identical in all 14, including the LGPL variant in `hf_cache.py`, which differs from its siblings and was left as is rather than normalised.
- **Lint**: the narrow CI gate reports the same 4 pre-existing errors at base and head (`pack_dataset` in the excluded function, two `__all__` names in `peft_utils.py`); `compileall` and both dynamic-execution linters clean.

## Tests

Base and head measured on the same host and compared file by file:

- all 51 mlx files, one process each

Identical in every partition. The 4 failures in `test_rl_replacements_compile_disable.py` reproduce on a clean main checkout and that file is untouched.

## What this leaves

The remaining comment volume in `unsloth_zoo/` sits in the codegen-critical files and `temporary_patches/`. Those need a different approach from a comment sweep, since in `temporary_patches/` the length of a comment can decide whether a rewriter fires, so I am not proposing them as a follow-up in this form.

### DIFF
--- a/unsloth_zoo/dataset_utils.py
+++ b/unsloth_zoo/dataset_utils.py
@@ -69,13 +69,11 @@ def _longest_common_sublist(lists):
         """Return (exists, sublist) for a common sublist of `length`."""
         common = set()
         first = lists[0]
-        # All sublists of `length` from the first list
         for i in range(len(first) - length + 1):
             sub = tuple(first[i:i + length])
             common.add(sub)
         pass
 
-        # Keep only sublists also present in every remaining list
         for lst in lists[1:]:
             current = set()
             for i in range(len(lst) - length + 1):
@@ -89,7 +87,6 @@ def _longest_common_sublist(lists):
         return True, list(common.pop())
     pass
 
-    # Binary search on length
     left, right = 1, min_len
     result = []
 
@@ -125,7 +122,6 @@ def _find_common_token_ids(component, tokenizer, force_match = False):
     elif component.startswith("\n"): left_text = "\n"
     stripped = component.strip()
     
-    # Add current pieces and also newlines
     all_input_ids = []
     if not force_match:
         for left in range(3):
@@ -224,7 +220,6 @@ def get_chat_template_parts(tokenizer):
     specials = sorted({str(s) for s in (set(getattr(tok, "all_special_tokens", []) or []) | added) if s}, key = len, reverse = True)
 
     def strip_lead(s, *prefixes):
-        # Remove any of the given leading strings, repeatedly
         changed = True
         while changed:
             changed = False
@@ -233,7 +228,6 @@ def get_chat_template_parts(tokenizer):
         return s
 
     def strip_shared(a, b):
-        # Drop leading special-tokens/whitespace common to both until roles differ
         while True:
             wa, wb = re.match(r"\s+", a), re.match(r"\s+", b)
             if wa and wb and wa.group() == wb.group():
@@ -247,7 +241,6 @@ def get_chat_template_parts(tokenizer):
         return a, b
 
     def gap_mode(from_ends, to_starts):
-        # Most common text between adjacent content blocks
         out = []
         for e in from_ends:
             nxt = [s for s in to_starts if s >= e]
@@ -712,7 +705,6 @@ def train_on_responses_only(
     # Keep the original object (may be a VLM processor) so auto-detect can read a
     # chat template that lives only on the processor; the matcher uses the inner one.
     processor = tokenizer
-    # Get non vision tokenizer
     if hasattr(tokenizer, "image_processor") or hasattr(tokenizer, "tokenizer"):
         tokenizer = tokenizer.tokenizer
     if  not hasattr(tokenizer, "_unsloth_input_part") or \
@@ -755,7 +747,6 @@ def train_on_responses_only(
         )
     pass
 
-    # Store some temporary stuff
     A_first = A_must[0]
     len_A_must = len(A_must)
     A_left_reversed = A_left[::-1]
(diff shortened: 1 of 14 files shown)


danielhanchen:
Idempotency evidence, run on a staging copy of this branch so the main CI queue is not used for it. All green there, including the full test suite.

This is the first batch where these checks are load-bearing rather than negative controls. The 14 files here are in the set `compiler.py` scans, so a comment deletion could in principle change what a rewriter emits. Four properties, each mutation calibrated before being trusted:

**I1, rewriter decisions unchanged.** 480 rewriter checks over 96 files, 12 firings, 0 decision changes. Calibrated by shortening the two-line comment in `temporary_patches/ministral.py` to one line, which moves the `value_states` cast inside `fix_attention_dtype_consistency`'s 200 character lookahead and flips it from inserting a cast to skipping. The AST is unchanged, the tests pass, and `verify_comment_only_diff.py` reports the file clean, so nothing else in the tree can see that.

**I3, rewriters are fixpoints.** `f(f(src)) == f(src)`, 0 non-fixpoint. Calibrated by making the marker emitted at `compiler.py:548` diverge from the `_AITER_MARKER` guard at `:507`: the rewriter re-wraps its own output, +49 lines on a second pass. Since that rewriter never fires on a non-AMD host, the check runs it against a fixture with the device probe stubbed and asserts the fixture was actually rewritten before drawing a conclusion.

**I4, import selection.** This is the one that matters most for this batch. `create_new_function` picks which names to import into a generated module with `x in new_source`, a plain substring test over text that includes comments. Deleting the last comment that mentions a symbol drops its import. Stable over 2335 symbols across all 14 files. Calibrated by adding a comment naming a symbol absent from a file, which produces exactly that spurious import.

**I5, `re.sub` replacement strings.** All 8 sites byte-identical. An introduced backslash in one corrupts every substitution that uses it. Calibrated by injecting one, 0 to 2 backslashes. The check compares each assignment's source segment rather than an evaluated value, because the 8 sites are five different AST node types and evaluating reached only the two plain constants while reporting a pass.

danielhanchen:
Closing: the remaining reduction in these files is small and not worth the review cost. Keeping the evidence here for reference.

PR #1160, opened 2026-09-06
Reduce comment volume in unsloth_zoo/compiler.py

danielhanchen:
## Summary

Removes 50 comment lines from `unsloth_zoo/compiler.py` (747 to 697 comments). What goes is restatements of the line below them (`# Try standard import`, `# Restore original sys.path if we modified it`), bare section labels (`# Check versioning`, `# Write function`, `# Quick exit`), and one trailing restatement. Everything that explains a rewrite, names a transformers or peft symbol, a version, or a failure mode stays.

## Why this needed more care than the earlier batches

This module rewrites Python source at runtime, so a comment is not automatically inert: several finders anchor on consecutive lines or read a fixed character window, comments inside string literals are emitted into generated modules, and a number of comment lines are asserted by tests against the module source. Before any edit, every comment in the codegen-critical set (compiler.py, saving_utils.py, temporary_patches/, 4,751 comments in total) was deleted one at a time, and then all at once per file, and all five compiler.py rewriters (higher precision softmax, higher precision layernorms, attention dtype consistency, residual stream, AMD aiter SDPA) were rerun and their decisions compared to the unmodified file. Exactly two comment lines change a rewriter decision, both in temporary_patches/ministral.py, and both are untouched.

Excluded from the pass, by construction:

- test-asserted literals whose only occurrence in the module is a comment line
- comments inside string literals (re.sub replacement bodies, emitted licence and versioning headers, generated helper source)
- the idempotency sentinels (`_AITER_MARKER`, `_full_license_header`) and the positionally parsed versioning header
- licence headers and the in-function `All Unsloth Zoo code licensed under LGPLv3` and `licensed under AGPL3` markers
- class bodies read back with `inspect.getsource` and compared to upstream, and verbatim upstream mirrors
- commented-out code, numbered step series that other comments refer to, and anything naming a model, version, symbol, number, or failure mode

Edits are deletion only: a whole comment line is removed, or a trailing comment is stripped leaving the code bytes unchanged. No surviving comment was reworded or rewrapped.

## Verification

| Check | Result |
|---|---|
| `scripts/verify_comment_only_diff.py --base origin/main` | OK, AST identical after docstring strip |
| Token stream (comments and layout tokens dropped) vs main | identical |
| Generated-module snapshot over 96 files | 0 differences |
| Rewriter decisions vs main, fixpoint on a second pass, import selection over 2,335 symbols, 8 re.sub replacement strings | 0 changes |
| Sentinels and frozen test anchors | 9 of 9 intact |
| Comments containing `def` above a def or class | 0 new |
| Licence headers | byte-identical |
| `tests/security` | 156 passed |
| Full CPU suite excluding mlx | 4 failed, 4877 passed, 105 skipped, the same 4 pre-existing failures as main |

The suite numbers above are for all three codegen-critical PRs applied together; the three touch disjoint files.

### DIFF
--- a/unsloth_zoo/compiler.py
+++ b/unsloth_zoo/compiler.py
@@ -82,7 +82,6 @@ UNSLOTH_COMPILE_USE_TEMP = False
 # Disable some compilations if old versions are seen
 OLD_TORCH_VERSION = Version(torch.__version__) < Version("2.5.0")
 
-# device capability
 major = None
 minor = None
 # Bound before the branches: DEVICE_TYPE == "cpu" takes none of them, and
@@ -106,7 +105,6 @@ pass
 
 OLD_TRITON_VERSION = Version(triton.__version__) < Version("3.0.0")
 
-# Check if Unsloth Studio is allowed
 import importlib.util
 
 if importlib.util.find_spec("unsloth_studio") is None:
@@ -116,7 +114,6 @@ else:
 pass
 
 
-# Ignore logging messages
 class HideLoggingMessage(logging.Filter):
     def __init__(self, text):
         self.text = text
@@ -282,7 +279,6 @@ _patch_functions = [
 ]
 
 
-# Empty causal mask
 def no_update_causal_mask(*args, **kwargs):
     return None
 
@@ -593,7 +589,6 @@ def _get_compile_folder(use_tempfile=False):
     else:
         location = UNSLOTH_COMPILE_LOCATION
         try:
-            # Try creating the directory
             os.makedirs(location, exist_ok=True)
             return location, UNSLOTH_COMPILE_USE_TEMP
         except Exception as e:
@@ -601,7 +596,6 @@ def _get_compile_folder(use_tempfile=False):
                 f"Unsloth: Failed to create directory `{UNSLOTH_COMPILE_LOCATION}` because {str(e)}"
             )
 
-            # Instead use a temporary location!
             location, UNSLOTH_COMPILE_USE_TEMP = _get_compile_folder(use_tempfile=True)
     return location, UNSLOTH_COMPILE_USE_TEMP
 
@@ -619,7 +613,6 @@ def get_compile_folder(use_tempfile=False):
 pass
 
 
-# Mask creation functions
 @functools.lru_cache(1)
 def get_mask_functions():
     try:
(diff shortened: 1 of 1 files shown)


danielhanchen:
Staging CI on an owned replica of this branch (staging commit 7e810a00, the PR head plus a staging-only workflow that reruns the compiler.py rewriters against main and checks they are fixed points):

| Workflow | Result |
|---|---|
| Tests CI | success (run 34013113029) |
| Wheel CI | success (run 34013112997) |
| Lint CI | success (run 34013113003) |
| Trim idempotency (staging only) | success (run 34013112991) |

Runs are on danielhanchen/unsloth-zoo-staging; the staging PR is closed, never merged.

danielhanchen:
End-to-end check on top of the static gates, run on the three codegen PRs applied together (they touch disjoint files):

Imported unsloth with `UNSLOTH_COMPILE_LOCATION` pointed at a fresh directory, imported the TRL trainers (so `rl.py` regenerated every `Unsloth*Trainer.py` from `rl_replacements.py`), and loaded tiny random gpt-oss, qwen3-moe, gemma-3, gemma-3n, qwen3-vl-moe and llama checkpoints through `FastModel` so the compiler and the temporary patches ran for real. Then hashed every generated file and the bytecode of every function on the patched transformers modeling modules. Same script on main and on the trimmed tree, and twice on the trimmed tree into the same cache.

| Check | Result |
|---|---|
| Generated cache files (44) main vs trimmed | 43 byte-identical; `moe_utils.py` is a verbatim copy of the trimmed source, code tokens identical |
| `Unsloth*Trainer.py` (16 files from `rl_replacements.py`) | byte-identical |
| Bytecode of patched transformers functions (881) | identical |
| Second run on the trimmed tree | same files rewritten as a second run on main (the dtype modules and the cache copy of moe_utils.py are always rewritten), contents unchanged |
| `unsloth_compiled_module_gemma3n.py` and `_gpt_oss.py` | one import line changes name order between any two runs, on main as well; same name set, rest of file identical |

`rl_replacements.py`, `patching_utils.py`, `vllm_utils.py` and `llama_cpp.py` are untouched by every comment PR in this series.

danielhanchen:
Closing: the remaining reduction in these files is small and not worth the review cost. Keeping the evidence here for reference.

PR #1163, opened 2026-09-06
Keep the forced-float32 Gemma4 RMSNorm through the compiler

danielhanchen:
## Summary

`patch_Gemma4RMSNorm` in `temporary_patches/gemma4_float32.py` replaced `Gemma4RMSNorm.forward` with a function that was itself named `forward`. `compiler.create_standalone_class` treats a replacement forward with that name as the original method: it emits the patched body as a standalone `Gemma4RMSNorm_forward`, then tries to splice a call to it over the original method's source text. The class attribute no longer holds that text, so the splice matches nothing and the generated class silently keeps the upstream forward.

The upstream forward returns the input dtype. Under `UNSLOTH_FORCE_FLOAT32=1` the fp32 residual stream therefore reached a bf16 `q_proj` unchanged, and a 4-bit gemma-4 load died on the first matmul:

```
RuntimeError: expected mat1 and mat2 to have the same dtype, but got: float != c10::BFloat16
```

A replacement with any other name takes the compiler's other path, which locates `def forward` in the class text and replaces it with a call to the new function. This PR renames the replacement to `_gemma4_forced_rmsnorm_forward` and leaves a comment explaining why the name is load bearing.

## Tests

`tests/test_gemma4_forced_rmsnorm_compile.py`, CPU only:

- Drives the real `create_standalone_class` against a fake modeling module whose forward was replaced under both spellings, and executes the generated code. The renamed replacement changes the output dtype to float16; the `forward`-named one is dropped and the output stays float32.
- The dropped case is kept as a negative control. If the compiler ever learns to handle that spelling, the test goes red and says the rename is no longer load bearing, rather than leaving an indirection nobody can justify.
- A third test parses the patch's own source and refuses a replacement spelled `forward`, so the fix cannot quietly revert.

Existing gemma4 and compiler suites still pass (244 passed, 18 skipped across the forced-float32 and compiler files).

## Verification on a GPU

`unsloth/gemma-4-E2B-it`, 4-bit, `UNSLOTH_FORCE_FLOAT32=1`, transformers 5.16.1, torch 2.13.0, one B200:

| install | result |
|---|---|
| unsloth 2026.9.2 + zoo 2026.9.1 (PyPI) | crash above, in `gemma4_float32.py` at `self.q_proj(hidden_states)` |
| this branch + the matching unsloth loader fix | loads and generates; 462 fp16, 397 fp32 and 129 uint8 parameters, one-at-a-time outputs identical to the bf16 baseline |

The crash needs both halves to reproduce on a bf16 GPU: this compiler path, and the `unsloth` loader discarding the initial `UNSLOTH_FORCE_FLOAT32` request. The loader half is a separate PR against `unslothai/unsloth` and links back here.

### DIFF
--- a/tests/test_gemma4_forced_rmsnorm_compile.py
+++ b/tests/test_gemma4_forced_rmsnorm_compile.py
@@ -1,0 +1,181 @@
+# Unsloth Zoo - Utilities for Unsloth
+# Copyright 2023-present Daniel Han-Chen, Michael Han-Chen & the Unsloth team. All rights reserved.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published
+# by the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+
+"""The forced-float32 Gemma4 RMSNorm replacement must survive the compiler.
+
+`patch_Gemma4RMSNorm` swaps `Gemma4RMSNorm.forward` for a wrapper that returns
+fp16 (the sub-layer dtype the forced-float32 design wants) instead of the input
+dtype. The compiler then regenerates every modeling class from source, and it
+has two ways of handling a class whose forward was replaced:
+
+  * replacement still named `forward`: emit the patched body as a standalone
+    `<Class>_forward`, then splice a call to it over the ORIGINAL method's
+    source text. That text is no longer what the class attribute holds, so the
+    splice matches nothing and the generated class keeps the upstream forward.
+  * replacement with any other name: locate `def forward` in the class text and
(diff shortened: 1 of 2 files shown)


danielhanchen:
The loader half is unslothai/unsloth#10368. On a bf16 GPU both are needed: without that one the initial UNSLOTH_FORCE_FLOAT32 request is discarded by the loader, and without this one the repaired path still returns the fp32 residual into bf16 projections.

danielhanchen:
Closing. The premise does not hold on the path unsloth actually uses. UNSLOTH_FORCE_FLOAT32 is internal: the loader sets it per load, so at import time it is "0" and the init phase installs nothing. The RMSNorm replacement is installed at pre_compile, and whatever the compiler does with the class after that, the post_compile phase re-applies the patch to the class the model holds.

Measured on the loader-driven path (an fp16 request for gemma-4-E2B-it in 4-bit on a B200, the same branch a T4 takes), with the unmodified zoo 2026.9.1:

| | value |
|---|---|
| Gemma4RMSNorm.forward on the loaded model | `patch_Gemma4RMSNorm.<locals>.forward` from `gemma4_float32.py` |
| output dtype for a float32 input | float16 |
| parameters | 462 fp16, 397 fp32, 129 uint8 |

So the forced-fp32 boundary is in place without this change. The drop described in the PR only happens when the flag is set by hand before import, which puts the patch on the class ahead of the compiler; that is not a supported way to use the flag. The compiler's handling of a replacement still named `forward` is real and the test here documents it, but it does not reach users through this patch, so there is nothing to ship.

PR #1165, opened 2026-09-06
Fix the flex-attention backward shared-memory gate missing every num_stages=1 GPU

danielhanchen:
The flex-attention backward shared-memory patch has been a silent no-op on Blackwell-class GPUs. On a GB10 (sm_121, 40 SMs) a default LoRA run on Gemma-3 (head_dim 256) dies on a cold Inductor cache with this patch loaded and active:

```
InductorError: RuntimeError: No valid triton configs. OutOfMemoryError: out of resource:
triton_tem_fused_flex_attention_backward Required: 114688 Hardware limit: 101376
```

## Why it did not fire

`CUDAConfigHeuristic.get_flex_attn_bwd_configs` returns exactly one config for `head_dim >= 128` on this part: block 64, `num_stages=1`. Three separate things then went wrong, and each on its own is enough to keep the gate shut.

**1. The estimate scales with `num_stages`.** It was fitted on an sm8x part, whose `head_dim >= 128` entry uses `num_stages=2`. `sm10x`, `sm11x` and `sm12x` all use `num_stages=1` there, which halves the estimate. head_dim 256 bf16 block 64 scored `1 * 64 * 256 * 2 * 2 + 10240 = 75776`, comfortably under the limit, so `needs_patch` stayed `False` and the failing config was returned untouched.

Measured on GB10 (torch 2.14.0+cu134, Triton 3.8.0) by forcing single configs and reading Triton's own `Required:`:

| block | num_stages | Required |
| --- | --- | --- |
| 64 | 1 | 114688 |
| 64 | 2 | 114688 |
| 128 | 1 | 229376 |

So on that part the requirement is linear in block size and independent of `num_stages`.

**2. The limit was the wrong one.** Triton refuses on `shared_memory_per_block_optin` (101376 here), not `shared_memory_per_multiprocessor` (102400). Those 1024 bytes are enough to bless a config Triton then rejects, and they are exactly the gap between the two numbers in the error above.

**3. The fallback ladder asked the same broken estimate** which blocks to offer, so a wrong estimate both failed to open the gate and then chose the replacements.

## The fix

The two formulas disagree and neither generalises. The original sm8x measurement (`141312` estimated against `140800` actual) implies a tile factor near 4.3; GB10 measures 3.5. Rather than pick one, take the maximum of both.

The direction matters and is not symmetric. An over-estimate offers a few extra candidates that the autotuner discards. An under-estimate is a silent failure to compile. A maximum is also never below the formula that ships today, so every GPU where this patch fires now still gets it, and the change can only add coverage.

The limit is now `min(shared_memory_per_block_optin, shared_memory_per_multiprocessor)`, tolerating either attribute being absent on older torch.

The estimate now decides only *whether* to intervene. Every block strictly below the offending one is offered, rather than only those the estimate blesses. Inductor's autotuner catches `OutOfResources` per candidate and keeps what compiles, so offering a config that does not fit costs one discarded candidate, while failing to offer one that does is the whole bug. Verified directly: a forced list of `[block 64 (OOMs), block 16 (fits)]` compiles cold and produces gradients matching the good-only run.

Finally, `patch_flex_attention_bwd_configs()` wrapped whatever was already installed, so calling it twice walked two copies of the ladder. It is now guarded.

## Verification

On GB10, cold Inductor cache, torch 2.14.0+cu134 / Triton 3.8.0:

- `flex_attention` backward at head_dim 256 bf16 **FAILS** with the shipped patch and **PASSES** with this one, producing real gradients.
- head_dim 32, 64 and 128 keep exactly one config, so nothing else pays autotune time.
- A large-shared-memory part never reaches the gate: 114688 is below A100's 166912 opt-in.
- LoRA on `unsloth/gemma-3-270m-it` trains on the default flex path at 4.4 steps/s, with the loss curve tracking the eager path to three digits.
- Existing suites: `test_temporary_patches_imports.py` and `test_temporary_patches_exhaustive.py`, 146 passed, 13 skipped.
- New `tests/test_flex_attention_bwd_shmem_gate.py`, 253 passed, GPU-free. (61 when this
  was first written; keeping the float8 rows a floor for every dtype widened the
  parametrisation.)

No-op behaviour is unchanged for torch < 2.10 (no `FlexBwDConfig`), for non-CUDA, and for ROCm classes.

## One note for anyone reproducing this

It looks intermittent and it is not. `unsloth_zoo/patching_utils.py` pops `TORCHINDUCTOR_CACHE_DIR` during `import unsloth`, so every run shares one cache directory and setting that variable beforehand does nothing. Once any process has compiled a working backward graph, later runs hit the FX-graph cache and never consult the config heuristic at all, so the same command passes or fails depending only on what some earlier run left behind. Set the cache directory *after* the import to get a genuinely cold run.

That popping is worth revisiting on its own, since it silently overrides an explicit user choice, but it is left alone here.

## Confirmed independently on a second Blackwell part

Re-run on a B200 (sm_100, torch 2.13.0+cu130) after merging this branch with current
`main`, which merges clean:

| block | num_stages | head_dim | old estimate | new estimate | Triton's measured requirement |
| --- | --- | --- | --- | --- | --- |
| 64 | 1 | 256 | 75,776 | 114,688 | 114,688 |
| 128 | 1 | 256 | 141,312 | 229,376 | 229,376 |
| 64 | 2 | 256 | 141,312 | 141,312 | - |

The new estimate reproduces the figures measured on GB10 exactly, on different silicon,
which is the part of the fix that had only one hardware witness before.

B200 reports 233,472 bytes per SM, so the gate stays shut for all of these and the configs
Inductor picked are returned untouched. That is the intended no-op: a part with headroom
pays no extra autotune time. It also means B200 cannot witness the failure itself, and this
does not claim otherwise; the GB10 evidence above is still the only end-to-end reproduction.

`tests/test_flex_attention_bwd_shmem_gate.py` passes 253 on that merged tree.

### DIFF
--- a/tests/test_flex_attention_bwd_shmem_gate.py
+++ b/tests/test_flex_attention_bwd_shmem_gate.py
@@ -38,20 +38,43 @@ flex = pytest.importorskip("unsloth_zoo.temporary_patches.flex_attention_bwd")
 torch = pytest.importorskip("torch")
 
 
-def _staged(block, stages, head_dim, dtype_size):
-    """The formula this module shipped with, kept here as the floor it must not go under."""
-    return stages * block * head_dim * dtype_size * 2 + flex._SHMEM_OVERHEAD
+def _dtypes():
+    """Every dtype the backward is reached with, including any float8 this torch has."""
+    out = [torch.bfloat16, torch.float16, torch.float32]
+    for name in ("float8_e4m3fn", "float8_e5m2"):
+        dt = getattr(torch, name, None)
+        if dt is not None:
+            out.append(dt)
+    return out
+
+
+def _staged(block, stages, head_dim, dtype):
+    """The formula this module shipped with, dtype rule included.
+
+    Its rule answers 4 for anything that is not bf16 or fp16, which is four times a
+    float8's real width. Reproduced faithfully rather than corrected, because this is
+    the floor the estimate must not drop below, not a model of the hardware.
+    """
+    size = 2 if dtype in (torch.bfloat16, torch.float16) else 4
+    return stages * block * head_dim * size * 2 + flex._SHMEM_OVERHEAD
 
 
 class TestTheEstimateIsNeverLowerThanItWas:
-    """Whatever else changes, no GPU may stop being covered."""
+    """Whatever else changes, no GPU may stop being covered.
+
+    The float8 rows are the ones that caught a real slip: correcting the element width
+    to its true 1 byte dropped the estimate for an fp8 kernel to well under what the
+    shipped formula produced, which would have closed the gate on a case it opens today.
+    Each term now keeps its own dtype rule so the maximum is a floor for every dtype.
+    """
 
     @pytest.mark.parametrize("block", [16, 32, 64, 128])
     @pytest.mark.parametrize("stages", [1, 2, 3])
     @pytest.mark.parametrize("head_dim", [64, 128, 192, 256])
-    def test_it_never_drops_below_the_original_formula(self, block, stages, head_dim):
-        got = flex._estimate_shmem(block, stages, head_dim, torch.bfloat16)
-        assert got >= _staged(block, stages, head_dim, 2), (
+    @pytest.mark.parametrize("dtype", _dtypes(), ids = lambda d: str(d).replace("torch.", ""))
+    def test_it_never_drops_below_the_original_formula(self, block, stages, head_dim, dtype):
+        got = flex._estimate_shmem(block, stages, head_dim, dtype)
+        assert got >= _staged(block, stages, head_dim, dtype), (
             "an estimate below the original one would close the gate on a GPU where it "
             "fires today, which is a silent failure to compile"
         )
@@ -75,7 +98,7 @@ class TestTheMeasuredRequirementIsCovered:
         """The exact config GB10 was handed, against the limit Triton enforces."""
         assert flex._estimate_shmem(64, 1, 256, torch.bfloat16) > 101376
         # And the old formula, which is why it slipped: 1 * 64 * 256 * 2 * 2 + 10240.
-        assert _staged(64, 1, 256, 2) == 75776
+        assert _staged(64, 1, 256, torch.bfloat16) == 75776
 
     def test_a_head_dim_that_fits_still_fits(self):
         """head_dim 128 compiles on GB10 today, so the gate must stay shut for it."""
--- a/tests/test_flex_attention_bwd_shmem_gate.py
+++ b/tests/test_flex_attention_bwd_shmem_gate.py
@@ -1,0 +1,168 @@
+# Unsloth Zoo - Utilities for Unsloth
+# Copyright 2023-present Daniel Han-Chen, Michael Han-Chen & the Unsloth team. All rights reserved.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+
+"""The shared-memory gate in `temporary_patches.flex_attention_bwd`.
+
+The patch appends smaller backward configs when the ones Inductor picked cannot fit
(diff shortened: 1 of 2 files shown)


danielhanchen:
Confirmed the estimate in unsloth_zoo/temporary_patches/flex_attention_bwd.py still scales with num_stages and compares against the per-SM figure, so the gate stays shut on the num_stages=1 parts at head_dim 256. Will get this reviewed.

danielhanchen:
Proof on the hardware this was written for, plus a measurement that shows why the released zoo does not fix it.

## The released zoo really does miss this case

`unsloth_zoo` 2026.9.2 ships `temporary_patches/flex_attention_bwd.py` and registers it, so the patch is live. It still does nothing at `head_dim=256` on GB10, because stock torch 2.14 offers exactly one backward config there:

```
FlexBwDConfig(block_m1=64, block_n1=64, block_m2=64, block_n2=64, num_stages=1, num_warps=4)
```

The released estimator is linear in `num_stages`, so at `num_stages=1` it computes `1*64*256*2*2 + 10240 = 75776`, compares that against the limit, decides nothing needs patching, and returns the config untouched. The kernel then asks for 114688 and the compile dies:

```
torch._inductor.exc.InductorError: RuntimeError: No valid triton configs.
OutOfMemoryError: out of resource: triton_tem_fused_flex_attention_backward_1
Required: 114688 Hardware limit:101376
```

That is a 38912-byte under-estimate, and it is silent.

## This branch predicts it exactly

Running this branch's `_estimate_shmem` on the device, against the requirements Triton itself reports:

```
device: NVIDIA RTX Spark N1X | shared_memory_per_block_optin: 101376 | per_multiprocessor: 102400

 block  stages    hd   estimate   measured   gate fires
    64       1     256    114688     114688         True
    64       2     256    141312     114688         True
   128       1     256    229376     229376         True
    64       2     128     75776          -        False
    64       1      64     28672          -        False

released formula, block=64 stages=1 hd=256: 75776 -> fires: False
```

The linear term lands on 114688 and 229376 exactly, and the maximum of the two formulas keeps the sm8x case that the staged formula was fitted to. It does not over-fire: `head_dim` 64 and 128 stay untouched, which matches the fact that those work fine on this part today.

The `shared_memory_per_block_optin` preference matters here too. That is 101376 on this device against 102400 per multiprocessor, so the per-SM figure is the wrong property by a 1024-byte window.

## End to end

`head_dim=256`, `S=1024`, bf16, `torch.compile(flex_attention)`, this branch's module in front of site-packages, on the real GPU:

```
module file: ...\zoo-src\unsloth_zoo\temporary_patches\flex_attention_bwd.py
has tile factor: True
[FWD] finite=True  t=17.3s
[BWD] qgrad_finite=True  t=24.6s
```

Autotune picked one of the fallback configs this branch generates:

```
best_kernel: triton_flex_attention_backward_9
  BLOCK_M1=16, BLOCK_M2=16, BLOCK_N1=16, BLOCK_N2=16,
  QK_HEAD_DIM=256, V_HEAD_DIM=256, num_stages=1, num_warps=8
  best_time 4.0125 ms  (8 choices, all Triton)
```

Against the released zoo the same script fails to compile at all.

## Why it matters

`head_dim=256` is what Gemma 3 uses, so flex-attention training of Gemma 3 on this class of part hits it. Models at `head_dim` 64 or 128, which is most of them, were never affected either way.

Verified on the same box as the rest of the Windows-on-ARM work: torch 2.14.0+cu134, Triton 3.8, sm_121, 40 SMs, `shared_memory_per_block_optin` 101376.

danielhanchen:
Confirmed the estimate in unsloth_zoo/temporary_patches/flex_attention_bwd.py still scales with num_stages and compares against shared_memory_per_multiprocessor, so the gate never fires on the sm_12x head_dim 256 case you measured. Will get this reviewed.

danielhanchen:
Confirmed the gate in unsloth_zoo/temporary_patches/flex_attention_bwd.py still scales with num_stages and compares against shared_memory_per_multiprocessor, so a block 64 / num_stages=1 config at head_dim 256 is returned untouched and the compile dies. Will get this reviewed.

PR #1167, opened 2026-09-07
Follow the dtype loop's rename, and stop the tests spelling the operand out

danielhanchen:
## Summary

Six source contracts in `tests/test_unreadable_modeling_source.py` have failed since #967, and they are the last of that fallout still on main. #1168 took the other three failures; this takes these, and then fixes the reason they broke rather than only the anchors.

No production code changes.

## What happened

#967 pinned the pristine forward into a local before any rewrite:

```python
original_forward = function.forward
try:
    source = inspect.getsource(original_forward).rstrip()
except (OSError, TypeError, tokenize.TokenError):
```

Same exception tuple, same wrap-rather-than-drop behaviour, and the pin is the improvement the commit was making, so a second pass cannot read its own generated output back. The tests were still searching for `inspect.getsource(function.forward)` and reported it as the guard having been removed:

```
FAILED test_the_nn_forward_patch_loop_is_guarded - AssertionError: the forward getsource is no longer inside a try
FAILED test_both_getsource_guards_are_present - AssertionError: inspect.getsource(function.forward) is unguarded, so an unreadable source is fatal again
FAILED test_the_compiler_config_check_is_kept - ValueError: substring not found
```

`ValueError: substring not found` names neither the anchor nor the production edit that moved it, which is most of why this took a while to read.

## Effect

This is red on `unslothai/unsloth`'s `Core (HF=default + TRL=default)`, `Core (HF=latest + TRL=latest)` and `Core (HF=4.57.6 + TRL<1)`. That job clones `unsloth_zoo@main` and runs this suite, so it fails on unsloth's `main` and on every open unsloth PR regardless of that PR's diff, including several that touch no Python at all.

## The fix, in two parts

**1. Follow the rename.** Four anchors move to `original_forward`. The other two `function.forward` references in the file, the `get_compiler_config` `hasattr` and the `__unsloth_dtype_wrapped__` `getattr`, still match `compiler.py` exactly and are left alone.

**2. Stop spelling the operand out.** A correct rename should not fail six tests, so `_nn_forward_read()` now finds the read by shape off the parse, an assignment to `source` of `inspect.getsource(<Name>).rstrip()`, and the anchors derive from what it finds.

Discovery on its own would weaken these tests, since a rename to something that is not the forward at all would still pass. So `test_the_forward_source_is_read_from_the_pinned_original` holds the discovered name to its contract: it is `function.forward`, or a local assigned from it before the read. That is #967's guarantee, pinned directly.

Bare `SRC.index` calls go through `_require_index`, which names the broken contract instead of raising `ValueError`.

## Verification

```
python -m pytest tests/test_unreadable_modeling_source.py tests/test_compiler_getsource_tokenerror.py -q
38 passed
```

Same two files on main: 6 failed, 31 passed.

Mutated `compiler.py` in three directions rather than assuming the new form still bites:

| mutation | expected | result |
| --- | --- | --- |
| rename `original_forward` to `pristine_forward` | pass | 32 passed |
| delete the `try`/`except` around the read | fail | 3 failed |
| point the pin at something other than `function.forward` | fail | 1 failed, the new test |

The full CI gate list from `Repo tests (CPU)`, 37 files, is 1508 passed / 4 skipped / 0 failed on this branch.

## Also: the MLX lane had the same defect

`consolidated-tests-ci.yml` re-stated the transformers cap as the literal `"5.5.0"` while its own comment said it meant whatever `pyproject.toml` pins. When #1144 raised that cap the gate failed reporting `transformers 5.15.1 exceeds the cap this package pins`, which was not true: the PR had raised the cap, so the install did match the pin.

It now asks the two questions separately.

**Does the resolution match what we declare?** Read off our own metadata, so a deliberate cap change no longer produces a phantom failure. It also checks the whole specifier set rather than only the ceiling, so a resolution landing on an excluded release such as 4.57.4 or 5.0.0 no longer passes silently, which the old `<= 5.5.0` never caught.

**Is the declared ceiling still the one we chose?** Metadata alone would let an automated bump raise the cap unnoticed, which is the one thing the literal was accidentally catching. So 5.5.0 stays as a separate, deliberate assertion with its reason written down: mlx-vlm >= 0.6.9 requires transformers >= 5.14, and raising the cap frees pip to resolve mlx-vlm 0.6.17 and take that jump as a side effect. Raising it is now a decision made in the same commit, not a consequence.

Behaviour of the new gate:

| scenario | result |
| --- | --- |
| main today, cap `<=5.5.0`, env 4.57.6 | pass |
| cap raised to `<=5.15.1` as in #1144 | fail, naming the ceiling and why |
| resolution drifts to 5.15.1, cap unchanged | fail |
| resolution lands on excluded 4.57.4 | fail, new coverage |
| transformers requirement removed entirely | fail, instead of passing vacuously |

## Note on the third failure

The earlier version of this description said `test_compiler_getsource_tokenerror.py` was left alone because it needed a decision about intent. #1168 has since made that decision and split it in two, keeping the TokenError handler genuinely exercised by stripping the markers first. Nothing is owed here. #1166, which covered the same ground, is closed as superseded.

### DIFF
--- a/.github/workflows/consolidated-tests-ci.yml
+++ b/.github/workflows/consolidated-tests-ci.yml
@@ -542,15 +542,61 @@ jobs:
           .lanes/venv-suites/bin/python -c "import mlx_vlm, mlx_lm; print('mlx-vlm', mlx_vlm.__version__)"
           # If the joint resolution missed the transformers cap, this job is
           # gating an environment the package forbids.
+          #
+          # Read the cap off our own installed metadata rather than repeating it
+          # as a literal. A literal here says "<=5.5.0" and means "whatever
+          # pyproject pins": the moment pyproject legitimately moves the cap the
+          # gate reports drift that does not exist, which is exactly how #1144
+          # went red. It also only ever checked the ceiling, so a resolution
+          # landing on one of the excluded releases (4.57.4, 5.0.0, ...) passed.
+          # The whole specifier set is the contract, so assert the whole thing.
           .lanes/venv-suites/bin/python - <<'PY'
-          from importlib.metadata import version
+          from importlib.metadata import requires, version
+          from packaging.requirements import Requirement
           from packaging.version import Version
-          transformers = Version(version("transformers"))
-          assert transformers <= Version("5.5.0"), (
-              f"transformers {transformers} exceeds the cap this package pins; "
-              "the MLX packages were resolved without it"
+
+          specifiers = {
+              str(req.specifier)
+              for raw in (requires("unsloth_zoo") or [])
+              for req in [Requirement(raw)]
+              if req.name.lower().replace("_", "-") == "transformers"
+          }
+          assert specifiers, (
+              "unsloth_zoo declares no transformers requirement, so this gate is "
+              "checking nothing. Did the dependency move out of pyproject.toml?"
           )
-          print("transformers", transformers, "within the pinned cap")
+          assert len(specifiers) == 1, (
+              f"unsloth_zoo pins transformers inconsistently across extras: "
+              f"{sorted(specifiers)}. Pick one before this gate can mean anything."
+          )
+          pinned = Requirement(f"transformers{specifiers.pop()}").specifier
+          transformers = Version(version("transformers"))
+          assert transformers in pinned, (
+              f"transformers {transformers} is outside the range this package "
+              f"pins ({pinned}); the MLX packages were resolved without it"
+          )
+          print("transformers", transformers, "within the pinned range", pinned)
+
+          # The ceiling is a deliberate policy choice, not an incidental number,
+          # so state it here as well and on purpose. mlx-vlm >= 0.6.9 requires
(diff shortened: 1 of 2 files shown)


danielhanchen:
The two remaining red checks here are not from this change, they are already failing on `main`:

```
Repo tests (CPU)                 094c0092a failure   4419c2487 failure   c9f0ae8d3 success
Core drift (3 upstream pins)     094c0092a failure   4419c2487 failure   c9f0ae8d3 success
```

`Repo tests (CPU)` is `test_compiled_cache_collective.py::test_direct_recovery_can_still_import_cache_helpers`:

```
AssertionError: the recovered module could not import moe_utils from the compiled cache, so a generated
MoE module would load with its backend names undefined and fail at the first forward.
```

That is #967's own suite, and it is worth a look on its own terms, since the assertion describes a real load failure rather than a stale anchor. I have not touched it here; this PR only moves the four source anchors the rename invalidated.

PR #1178, opened 2026-09-08
fix(mlx): complete text-model routing and cache-free training

Lyxot:
Text-only architectures available only in mlx-vlm still failed with `Model type ... not supported` after #1147. Several supported wrappers also failed during cache-free text training, and converted Moondream3 checkpoints failed strict loading because sanitization duplicated native parameter prefixes.

This change completes those loader contracts while preserving #1147's arity and causality checks:

- Route text architectures through mlx-vlm when their remapped mlx-lm module is absent and mlx-vlm resolves the model class. Unknown architectures retain the existing unsupported-model path; an installed but broken mlx-lm module is not treated as absent.
- Supply sequential three-axis positions for cache-free text prefill when the language model declares mRoPE. The hidden-state loss bypasses the language wrapper that normally supplies those positions. Explicit positions, cached calls, and embedded multimodal inputs retain their existing paths.
- Preserve already-native parameter names when sanitization only renames an unchanged tensor to a nonexistent parameter. Source-format conversions and tensor transformations still run. A descriptor preserves original class-level sanitizer calls and export inspection.
- Attach LoRA once per selected physical layer, retaining main's per-layer target selection. HRM exposes recurrent execution entries that refer to the same physical layers.

No model-type allow-list is added. Tokenizer/processor failures and unrelated image collator/save failures remain outside this change.

Validation used materialized tiny checkpoints with the real MLX load and training paths, including adapter movement checks:

- `glm4v`, `glm4v_moe`, `glm_ocr`, `paddleocr_vl`, `qwen2_vl`, and `moondream3` now pass text-only training. The complete regression sweep preserved all 36 previous text-only passes and five intentional refusals.
- All eight mlx-vlm-only text architectures reach the correct backend: Cohere2-MoE, DeepSeek-V4, LongCat-Flash-Sparse, and HRM-Text train; LLaDA2-MoE receives the existing bidirectional-model refusal. Laguna/Mellum still encounter tokenizer configuration errors, and MiniMax-M3 encounters the processor custom-code refusal with `trust_remote_code=False`.
- Audited installed language models' cache reads; the affected decoder paths share the mRoPE prefill contract. Other apparent unguarded reads are normalized or supplied by their outer callers.
- 45 selected tests passed before rebasing, including actual GGUF expert-export equivalence and #1147 probe regressions. Seven targeted production mutations failed their corresponding tests.
- After rebasing onto current main, eight focused Metal cases, 10 loader tests, and all 55 LoRA role-selection tests pass. Mutations removing either recurrent deduplication or per-layer selection fail the strengthened regression test. All six target text-only families pass again, and HRM passes the full load/inference/default-text/text-only step set.

Tested with MLX 0.32.2, mlx-lm 0.31.3, mlx-vlm 0.7.0rc0, and Transformers 5.5.0. Tiny random weights validate loading and training contracts, not output quality.

### DIFF
--- a/tests/test_mlx_distributed_loader.py
+++ b/tests/test_mlx_distributed_loader.py
@@ -109,6 +109,22 @@ class _VLMOnlyClass:
     """Stands in for a class only mlx-vlm ships; routing never calls it."""
 
 
+def test_vlm_only_text_routing_checks_remapped_module_presence(monkeypatch):
+    import unsloth_zoo.mlx.loader as loader
+
+    monkeypatch.setattr(loader, "_mlx_lm_module_name", lambda _: "remapped")
+    monkeypatch.setattr(loader, "_resolve_mlx_vlm_model_class", lambda _: _VLMOnlyClass)
+    seen = []
+    monkeypatch.setattr(loader.importlib.util, "find_spec", lambda name: seen.append(name))
+    assert loader._mlx_vlm_only_text_model({"model_type": "alias"}, "alias")
+    assert seen == ["mlx_lm.models.remapped"]
+    monkeypatch.setattr(loader.importlib.util, "find_spec", lambda _: object())
+    assert not loader._mlx_vlm_only_text_model({"model_type": "alias"}, "alias")
+    monkeypatch.setattr(loader.importlib.util, "find_spec", lambda _: None)
+    monkeypatch.setattr(loader, "_resolve_mlx_vlm_model_class", lambda _: None)
+    assert not loader._mlx_vlm_only_text_model({"model_type": "absent"}, "absent")
+
+
 def test_text_only_vlm_load_stays_on_vlm_path_when_mlx_lm_has_no_model(monkeypatch):
     import unsloth_zoo.mlx.loader as loader
 
@@ -354,6 +370,7 @@ def test_from_pretrained_distributed_vlm_forwards_normalized_override(monkeypatc
         "install_mlx_compile_patches", "_ensure_vlm_processor_inputs_patched",
         "_ensure_vlm_prompt_utils_patched", "_convert_mlx_dtype",
         "_patch_mixed_precision_set_dtype", "_ensure_minicpmo_mlx_sanitize",
+        "_ensure_native_vlm_weight_names",
         "_ensure_minicpmo_vision_dtype", "_fix_gemma4_kv_sharing",
         "_fix_gemma3_vision_post_layernorm_eps", "_fix_gemma3_vision_attention_fp32_sdpa",
         "_fix_gemma3_vision_encoder_fp32_layernorm", "_fix_gemma3_vision_post_layernorm_fp32",
--- a/tests/test_mlx_gguf_moe_export_metal.py
+++ b/tests/test_mlx_gguf_moe_export_metal.py
@@ -38,6 +38,52 @@ CONFIG = dict(
 )
 
 
+@metal_only
+def test_native_name_sanitizer_preserves_vlm_expert_export(monkeypatch, tmp_path):
+    import inspect
+    from types import SimpleNamespace
+    import mlx.nn as nn
+    from mlx_vlm.models.qwen3_5_moe.qwen3_5_moe import Model
+    from mlx_vlm.models.llava_onevision.llava_onevision import Model as StaticModel
+    from unsloth_zoo.mlx import loader, utils
+
+    model = Model.__new__(Model)
+    nn.Module.__init__(model)
+    model.config = SimpleNamespace(text_config=SimpleNamespace(
+        tie_word_embeddings=False, num_hidden_layers=2, num_experts=2,
+    ))
+    source = {}
+    for layer in range(2):
+        prefix = f"model.language_model.layers.{layer}.mlp.experts"
+        source[f"{prefix}.gate_up_proj"] = mx.arange(48).reshape(2, 6, 4).astype(mx.float32) + layer * 100
+        source[f"{prefix}.down_proj"] = mx.arange(24).reshape(2, 4, 3).astype(mx.float32) + layer * 100
+    staged = model.sanitize(dict(source))
+    static_model = StaticModel.__new__(StaticModel)
+    owners = [model, static_model]
+    vocabulary = set(utils._mlx_sanitizer_vocabulary(owners))
+    def exported(directory):
+        directory.mkdir()
+        mx.save_safetensors(str(directory / "model.safetensors"), staged)
+        assert utils._prepare_moe_gguf_export_directory(directory, model=model) > 0
+        return mx.load(str(directory / "model.safetensors"))
+
+    expected = exported(tmp_path / "before")
+    assert any(name.endswith("experts.gate_up_proj") for name in expected)
+
+    monkeypatch.setattr(Model, "sanitize", Model.sanitize)
+    monkeypatch.setattr(StaticModel, "sanitize", inspect.getattr_static(StaticModel, "sanitize"))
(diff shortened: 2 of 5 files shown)


Lyxot:
@codex review

danielhanchen:
Confirmed the routing gate in unsloth_zoo/mlx/loader.py still only fires for text_only=True, so the mlx-vlm-only text architectures do hit the unsupported path, and the cache-free mRoPE positions are genuinely missing in _forward_text_hidden_states. This branch also installs _ensure_native_vlm_weight_names on every mlx-vlm load, which changes sanitize for ordinary vision loads too, and it bundles the LoRA per-layer dedup on top; could you split the sanitizer and LoRA parts out and confirm existing vision loads are unaffected before I review?

danielhanchen:
Confirmed the routing gap in unsloth_zoo/mlx/loader.py where a text architecture with no mlx_lm.models module still fails, and #1096 covers that half already, so could you narrow this branch to the routing change and split out the mRoPE prefill positions, the sanitize name preservation and the LoRA layer dedup? The sanitize descriptor is installed on the mlx-vlm class for every load, which also affects normal VLM weight loading, and the branch no longer applies cleanly on current main, so a rebase would help too.

danielhanchen:
Confirmed the routing gap in unsloth_zoo/mlx/loader.py where an architecture only mlx-vlm ships still falls through, and the LoRA dedupe for repeated physical layers looks right. The sanitize descriptor in _ensure_native_vlm_weight_names now wraps every mlx-vlm model type on load, so could you confirm existing VLM checkpoints still load with identical weight names, and ideally split the mRoPE prefill and sanitizer changes into their own PRs?

danielhanchen:
Ran this on real Apple Silicon and on Linux with the real MLX CPU backend, base against head, and pushed two fixes to the branch. Summary first: all four sub-changes fix real, reachable defects, and three separate problems are in the way of merging.

**What each sub-change actually fixes**, reproduced on `5335e91f` and on head, against the installed mlx-vlm 0.6.4:

- Routing. Six real architectures qualify under these pins (`cohere2_moe`, `deepseek_v4`, `hrm_text`, `laguna`, `llada2_moe`, `minimax_m3`), enumerated from the installed package rather than assumed. Base: `ValueError: Model type cohere2_moe not supported.` Head: routed to mlx-vlm. All 42 unchanged routing cases behave identically, including an mlx-lm module that exists but raises at import (correctly not treated as absent) and the `_mlx_lm_module_name` remapping (`kimi_k2` still goes to mlx-lm through `deepseek_v3`).
- Cache-free mRoPE. Base raises `AttributeError: 'NoneType' object has no attribute 'offset'` on every qwen2_vl / qwen2_5_vl / qwen3_vl_moe / qwen3_omni_moe / glm_ocr text prefill; head matches the model's own `get_rope_index` to `maxabsdiff = 0.0` over seven batch and length combinations. The guard fires on a freshly constructed wrapper, because `_position_ids` is set in `LanguageModel.__init__`, so it is not dead code. Explicit positions, cached calls and non-mRoPE models are untouched, and `_position_ids` is never mutated.
- Native weight names. Of 45 instantiable mlx-vlm classes, exactly two have non-idempotent sanitizers (`moondream3` re-prefixes `text.` to `text.model.`, `paddleocr_vl` does an unanchored `key.replace("model", "language_model.model")`). Base fails a strict reload with `Received 644 parameters not in model: text.model.model.blocks...`; head loads. The important part for the question asked twice above: fed real source-format keys, which is what an ordinary vision load does, **all 45 classes produce identical output with and without the descriptor**, and class-level `getattr(cls, "sanitize")` still returns the backend's own function, so `_mlx_sanitizer_vocabulary` and `_mlx_sanitizer_writes_in_place` are unaffected.
- LoRA dedup. `mlx_vlm.models.hrm_text` really does list 2 physical layers over 8 entries. Base raises `ValueError: cannot convert unsupported module type mlx_lm.tuner.lora.LoRALinear to LoRA`; head attaches one adapter per physical layer and a real SGD step trains only the adapters.

**Pushed to this branch** (`d191d831`), both mutation-checked against head:

1. The dedup dropped later entries' targets. With `layer_keys = [["q_proj"], ["v_proj"]]` on a layer listed twice, head returned `attached=1` with `v_proj` unadapted, which neither the merge base nor `main` does. Now the per-entry target sets are unioned per physical layer. Two regression tests live in `tests/test_mlx_lora_group_selection.py`, so the Linux CPU lane gates this rather than only Apple Silicon.
2. `test_native_name_sanitizer_preserves_vlm_expert_export` imported `mlx_vlm.models.llava_onevision`, which does not exist before mlx-vlm 0.7.0. `pyproject.toml` caps mlx-vlm below 0.7.0 under the `transformers<=5.5.0` pin, so on a supported install the test failed with `ModuleNotFoundError` rather than skipping. It now resolves a staticmethod-sanitize family at runtime through the loader's own resolution (0.6.4 picks `granite_vision`, 0.7.x still picks `llava_onevision`). Verified passing on the CPU backend.

**Still outstanding, and these are yours to decide:**

- The branch does not merge into current `main`. The conflict is entirely in `linear_to_lora_layers`, which `main` rewrote after your merge base into `_layer_wanted` plus a `selected` build, a `_check_mlx_lora_base` sweep and a `dry_run` early return. Resolving it "prefer PR" produces code that does not even lint (`F821 Undefined name 'modules'`). The correct rebase deduplicates once in the `selected` loop so `dry_run` and the attach loop stay consistent; note `main` currently miscounts a shared layer (`dry_run=8` for `hrm_text`'s 2 physical layers).
- On Apple Silicon `tests/test_mlx_training_e2e_metal.py` passes all 67 tests and then the interpreter dies: `Segmentation fault: 11`, exit 139, which fails the job. The same file on staging `main` without this PR is 60 passed, exit 0, so this branch introduces it. A bisect job is running to name which new test leaves MLX unable to shut down; `tests/_run_then_exit_hard.py` exists for exactly this class of problem if the root cause turns out to be upstream.
- Two latent shape hazards, neither reachable with any shipped config I could find, listed in case you want them closed now: a backbone that declares `mrope_section` but accepts only 2-D positions (`minimax_m3_vl`, whose published config has `rope_scaling: null`) raises under the 3-axis broadcast; and the `inputs_embeds` path still crashes on base and head alike, since the guard requires `inputs_embeds is None`.

Given four independent fixes in one branch, the earlier request to split still looks right to me: the routing change and the LoRA dedup are self-contained and could land quickly, while the sanitizer descriptor is the one that deserves its own review because it patches every mlx-vlm class the loader touches.

danielhanchen:
@codex review

danielhanchen:
@codex security review

danielhanchen:
Pushed 3a6ca77ed: merged current main in and resolved the conflict, so this is MERGEABLE again.

linear_to_lora_layers had been rewritten on both sides. I kept main's structure (the DoRA preflight, the selected/root_modules split, the dry_run count) and applied this branch's physical-layer dedup to all three, so a recurrent stack that lists one layer object at several indices is counted once, preflighted once and adapted once, with every entry's layer_keys unioned rather than only the first entry's kept. Before, dry_run and the DoRA preflight still walked per entry, so the count disagreed with what was attached.

Each MLX suite run in its own process on the merged tree: test_mlx_lora_group_selection 76 passed, test_mlx_trainer_internals 226, test_mlx_generate 77, test_mlx_save_lora_adapters_filter 44, test_mlx_switch_lora 29, test_mlx_quantized_merge_save 21, test_mlx_distributed_loader 11. One process per file matters here: tests/mlx_simulation replaces sys.modules["mlx.core"] process-wide, so sharing a process reports failures that are not real.

danielhanchen:
@codex review

PR #1179, opened 2026-09-08
gpt-oss: stop paying for coordinate_descent_tuning to get combo kernels

danielhanchen:
`coordinate_descent_tuning` and `combo_kernels` were both driven by `use_combo_kernels`, so any card above the memory threshold got the tuning too. They cost very differently, so this splits them and defaults the expensive one off.

## Measurements

These options feed `moe_forward_inference`, which is reached under `not self.training and qlen == 1`, so decode is the path that matters. Measured there, 3 outer reps, noise floor taken from the combo-off reps of the same GPU:

| GPU | flag | compile overhead | decode tok/s |
|---|---|---|---|
| A100 39.49GB | combo off | 7.48 s | 11.73 |
| | combo on | 7.36 s | 11.87 (+1.19%, inside 2.47% noise) |
| | combo on + coord | **9.55 s (+28%)** | 11.84 (+0.94%, inside noise) |
| L4 22.03GB | combo off | 6.99 s | 12.06 |
| | combo on | 6.93 s | 12.08 (+0.17%, inside 1.49% noise) |
| | combo on + coord | **9.21 s (+32%)** | 12.08 (+0.17%, inside noise) |

Combo kernels are free to compile and make no measurable difference to throughput, on T4 16GB, L4 22GB, A100 40GB and B200 180GB, in both training and decode. The tuning adds 28-32% compile overhead and produced no gain on any of them, which is what its own `# Very slow!` comment already suspected.

Training compile time agrees, comparing warm reps so session cold start does not confound it: A100 39295 / 39243 / 38784 ms and L4 39953 / 39450 / 39309 ms for off / on / on+coord.

## Correctness

Greedy decode output is bit-identical with combo kernels on and off, on both A100 and L4, across every arm. Each arm is also self-consistent across repeats, printable ratio 1.0, longest repeated-token run 1. So this is not trading output quality for anything.

## What changed

`coordinate_descent_tuning` now defaults off and is opt-in via `UNSLOTH_COORDINATE_DESCENT_TUNING=1` for anyone who wants to re-measure it. `combo_kernels` keeps its existing gate and behaviour, since it is free and harmless.

## One thing recorded rather than changed

The gate is `device_memory/1024/1024/1024 > 40`, i.e. GiB. An A100-SXM4-40GB reports **39.49 GiB** and therefore falls below it, so combo kernels are off on a 40GB A100 and only switch on at 80GB and above. I left the threshold alone because combo kernels measured as no-effect on both sides of it, so moving it would change nothing measurable. It is now in a comment so the next reader does not have to rediscover it.

## Verification

```
UNSLOTH_COORDINATE_DESCENT_TUNING=0 -> combo=True coord=False
UNSLOTH_COORDINATE_DESCENT_TUNING=1 -> combo=True coord=True
```
on a 178GB card, with `no_combo_fused_torch_compile_options` keeping `combo_kernels=False` in both cases as before.

### DIFF
--- a/unsloth_zoo/temporary_patches/gpt_oss.py
+++ b/unsloth_zoo/temporary_patches/gpt_oss.py
@@ -1413,22 +1413,13 @@ else:
     device_memory = 0
 use_combo_kernels = False if device_memory/1024/1024/1024 <= 40 else True
 
-# coordinate_descent_tuning used to be driven by use_combo_kernels as well, so asking for
-# combo kernels also bought the tuning. They cost very differently and were measured apart:
+# coordinate_descent_tuning used to be driven by use_combo_kernels too, so asking for combo
+# kernels also bought the tuning. Measured apart on T4/L4/A100/B200: combo kernels are free
+# (decode compile 7.36s vs 7.48s off), the tuning costs +28% on A100 and +32% on L4 decode
+# compile for no measurable gain anywhere. Default it off, opt in to re-measure.
 #
-#   flag                        compile overhead (decode)      step time / decode tok/s
-#   combo_kernels               A100 7.36s vs 7.48s off        within noise on T4/L4/A100/B200
-#   coordinate_descent_tuning   A100 9.55s (+28%)              no measurable benefit anywhere
-#                               L4   9.21s (+32%)
-#
-# So the tuning is the expensive half and had no measured payoff, on either training or
-# decode, on any of T4 16GB, L4 22GB, A100 40GB or B200 180GB. Default it off and let it be
-# turned back on explicitly for anyone who wants to re-measure.
-#
-# Note also that the 40GB test is in GiB: an A100-SXM4-40GB reports 39.49, so it falls BELOW
-# this threshold and combo kernels are off there. Left as-is because combo kernels measured
-# as no-effect on both sides of the boundary, so moving it would change nothing; recorded so
-# the next reader does not assume a 40GB card takes the combo path.
+# The 40GB test above is in GiB, so an A100-SXM4-40GB reports 39.49 and falls BELOW it.
+# Left as-is: combo kernels measured as no-effect on both sides of the boundary.
 use_coordinate_descent = os.environ.get("UNSLOTH_COORDINATE_DESCENT_TUNING", "0") == "1"
 
 fused_torch_compile_options = get_torch_compile_options(
--- a/unsloth_zoo/temporary_patches/gpt_oss.py
+++ b/unsloth_zoo/temporary_patches/gpt_oss.py
@@ -1412,12 +1412,31 @@ elif DEVICE_TYPE in ("cuda", "hip") and torch.cuda.is_available():
 else:
     device_memory = 0
 use_combo_kernels = False if device_memory/1024/1024/1024 <= 40 else True
+
+# coordinate_descent_tuning used to be driven by use_combo_kernels as well, so asking for
+# combo kernels also bought the tuning. They cost very differently and were measured apart:
+#
+#   flag                        compile overhead (decode)      step time / decode tok/s
+#   combo_kernels               A100 7.36s vs 7.48s off        within noise on T4/L4/A100/B200
+#   coordinate_descent_tuning   A100 9.55s (+28%)              no measurable benefit anywhere
+#                               L4   9.21s (+32%)
+#
+# So the tuning is the expensive half and had no measured payoff, on either training or
+# decode, on any of T4 16GB, L4 22GB, A100 40GB or B200 180GB. Default it off and let it be
+# turned back on explicitly for anyone who wants to re-measure.
+#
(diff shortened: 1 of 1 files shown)


danielhanchen:
@codex review

danielhanchen:
@codex review

danielhanchen:
Verified on a real T4 (Kaggle, Tesla T4 14.56 GiB, torch 2.10.0+cu128, gpt-oss-20b 4bit).

A T4 is below the 40 GiB gate, so `use_combo_kernels` is already False there and
`coordinate_descent_tuning` was already off before this change. That makes T4 the regression
case for this PR rather than the interesting one: the bar is that nothing moves. Three arms,
each in its own process with its own Inductor cache.

| arm | unsloth_zoo | env | use_combo_kernels | use_coordinate_descent | fused dict | no_combo dict |
|---|---|---|---|---|---|---|
| main | main | none | False | attribute absent | combo=False coord=False | combo=False coord=False |
| PR | this branch | none | False | False | combo=False coord=False | combo=False coord=False |
| PR | this branch | `UNSLOTH_COORDINATE_DESCENT_TUNING=1` | False | True | combo=False **coord=True** | combo=False **coord=True** |

Results:

- Default on the PR branch is identical to main: the live option dicts match exactly, and a
  greedy 24-token decode returns byte-identical token ids. No behaviour change on T4.
- The opt-in reaches the compile options. This is the part worth checking rather than assuming,
  since `get_torch_compile_options` filters keys against `inspect.getsource(torch._inductor.config)`
  and drops unknown ones silently, so an env var that never landed in the dicts would look fine
  from the source alone. It lands in both dicts, and `combo_kernels` correctly stays False.
- All three arms generate the same coherent text.

The dicts are read live after import rather than inferred from the source, and every arm runs a
real decode, so a flag that compiles but breaks generation could not pass here.

Two things worth stating rather than leaving to be misread. The decode timings (24.8s for the
first arm against 10.0s for the other two) are first-call compile cost on a cold Inductor cache,
not an effect of the flag. And `model.embed_tokens.weight` sits on CPU in every arm, which is the
existing single-GPU embedding offload and is unrelated to this PR; I have a separate fix for the
`model.device` consequence of that in unslothai/unsloth#10569.

The compile-time case for the change was measured separately on four tiers:
`coordinate_descent_tuning` costs +26.9% compile on a single T4, +25.2% split across two,
+28% on A100, +32% on L4 and +24.7% on B200, with no throughput gain anywhere.

danielhanchen:
Two more T4 runs, both on Kaggle with torch pinned to 2.11.0+cu128 (Tesla T4 14.56 GiB, capability 7.5, single visible device, pin asserted at runtime). Same model, 512-token prompt, 64 greedy tokens, 3 outer by 3 inner reps, fresh Inductor cache per arm. `experts_qlen1_calls = 252` in all 18 runs, so the `qlen == 1` path the flag gates was genuinely exercised.

**Run 1, combo kernels forced on, this branch's opt-in off.** Answers whether torch 2.11 behaves like the 2.10 runs reported earlier.

| arm | tok/s (3 reps) | warm compile | cold first call | combo files |
|---|---|---|---|---|
| OFF | 3.76 / 3.70 / 3.65 | 4.98 s | 33.5 s | 0 |
| ON | 3.64 / 3.59 / 3.65 | 4.86 s | 27.1 s | 6, then 2 |
| ON_COORD | 3.59 / 3.58 / 3.61 | 5.89 s | 35.9 s | 6, then 4 |

Noise floor 2.97% from the OFF reps, ON is -1.62%, inside it. Compile is not worse with combo kernels and the cold first call is actually faster. All 9 runs byte-identical to the OFF reference, printable ratio 1.0, unique token ratio 0.688, coherent output. Same answer as torch 2.10, so this is not version-sensitive.

**Run 2, the shipped gate itself.** Rather than forcing the flag, this spoofs the memory reader inside `temporary_patches/gpt_oss.py` only, so `use_combo_kernels = device_memory > 40GiB` evaluates True and the module builds its own dicts. Live dicts confirmed at measure time: `fused_torch_compile_options` combo True, `no_combo_fused_torch_compile_options` combo False, coordinate descent False.

Result: **0 combo kernels formed, in all three reps, with combo True sitting in the fused options the whole time.** A B200 control says the same, 0 formed via the gate against 6 formed when both dicts are forced.

The six combo kernels in every measurement so far come from `no_combo_fused_torch_compile_options`, which is off on purpose with the reason recorded at line 1515 (`InductorError: AttributeError: 'NullKernelHandler' object has no attribute 'index_to_str'`). The gate only feeds `moe_forward_inference`, and nothing there forms a combo kernel.

That does not change this PR, it strengthens it. The old coupling meant asking for combo kernels also bought `coordinate_descent_tuning`, so the coupling was paying a real compile cost for a flag that turns out to be inert. Worth noting separately: with combo kernels absent, the tuning costs about +6% compile on this path rather than the +19 to +27% measured when they are forced on, which fits, since there is less for it to tune.

PR #1185, opened 2026-09-09
fix(mlx): honor processor tokenizer trust and clarify refusals

Lyxot:
The native mlx-vlm GOT processor hardcodes `trust_remote_code=True` for its tokenizer, so `FastMLXModel.from_pretrained(..., trust_remote_code=False)` can still execute the repository's tokenizer code. Image-processor recovery also omits the caller's trust flag. This follow-up carries the caller's choice through nested tokenizer construction and image-processor recovery, keeping the default `False`.

This is the standalone follow-up [requested during #1170's review](https://github.com/unslothai/unsloth-zoo/pull/1170#issuecomment-5583825157). It preserves the merged tokenizer selection, special-token handling and config-decoupling implementation. It does not restore the unproven tiktoken fast-file preference.

Remote-code refusals now name the required Python file and `trust_remote_code=True`. The loader retains an actual tokenizer refusal when mlx-vlm replaces it with a generic processor-resolution error, while still allowing a successful native processor fallback. Remote-only tokenizer metadata is retained so those diagnostics remain available. The existing thread-local tokenizer scope carries the flag and restores dispatch on exit; standard and distributed text paths preserve their existing explicit opt-in behavior.

Validation on real MLX with published tokenizer files and tiny random checkpoints, comparing current main with this change:

| Case | Main | This PR |
|---|---|---|
| GOT, explicit `False` | Loads remote tokenizer despite the flag | Refuses, naming `tokenization_qwen.py` |
| GOT, explicit `True` | Loads | Loads with identical token IDs |
| DBRX, explicit `True` | Loads | Loads with identical token IDs |
| LLM-jp VL, explicit `True` | Loads | Loads with identical token IDs |
| DBRX / LLM-jp VL, `False` | Repository-level refusal | Names `tiktoken.py` / `processing_llmjpvl.py` and the opt-in flag |

GOT, DBRX, LLM-jp VL and Phi-4 multimodal also passed load and generation checks, including image inference for the VLMs. No model weights were downloaded or dependency sources patched. These are separate real-runtime checks from the unit tests' MLX simulation shim.

- 31 focused loader tests passed on transformers 5.5.0; 16 relevant trust/refusal cases passed on 4.51.3.
- 12 isolated mutation checks failed their intended tests, covering both trust directions, nested scopes, direct/swallowed refusals, native fallback precedence, distributed loading and image-processor recovery.

```bash
python -m pytest tests/test_mlx_save_export_regressions.py tests/test_mlx_distributed_loader.py -q -k 'tokenizer or processor or distributed or strict_fallback'
```

### DIFF
--- a/tests/test_mlx_distributed_loader.py
+++ b/tests/test_mlx_distributed_loader.py
@@ -174,7 +174,9 @@ def test_apply_mlx_distributed_sharding_modes_and_guards():
         _apply_mlx_distributed_sharding(object(), tensor_group=tensor_group, model_name="fake")
 
 
-def test_load_mlx_lm_distributed_pipeline_filters_quant_shards(monkeypatch, tmp_path):
+@pytest.mark.parametrize("trust", [False, True])
+def test_load_mlx_lm_distributed_pipeline_filters_quant_shards(monkeypatch, tmp_path, trust):
+    from transformers import AutoTokenizer
     import mlx_lm.utils as mlx_lm_utils
     from unsloth_zoo.mlx.loader import _load_mlx_lm_distributed
 
@@ -227,12 +229,23 @@ def test_load_mlx_lm_distributed_pipeline_filters_quant_shards(monkeypatch, tmp_
 
     monkeypatch.setattr(mlx_lm_utils, "_download", _download)
     monkeypatch.setattr(mlx_lm_utils, "load_model", _load_model)
-    monkeypatch.setattr(mlx_lm_utils, "load_tokenizer", lambda *_a, **_k: types.SimpleNamespace(name="tok"))
+    (model_path / "tokenizer_config.json").write_text('{"tokenizer_class":"RemoteTokenizer"}')
+    tokenizer_calls = []
+
+    def tokenizer(path, **kwargs):
+        tokenizer_calls.append(kwargs)
+        return types.SimpleNamespace(name="tok")
+
+    monkeypatch.setattr(AutoTokenizer, "from_pretrained", staticmethod(tokenizer))
+    monkeypatch.setattr(
+        mlx_lm_utils, "load_tokenizer",
+        lambda path, *_a, **_k: AutoTokenizer.from_pretrained(path),
+    )
 
     model, tokenizer, config = _load_mlx_lm_distributed(
         "fake/repo",
         "llama",
-        {"return_config": True},
+        {"return_config": True, "tokenizer_config": {"trust_remote_code": trust}},
         pipeline_group=_FakeGroup(name="pipeline"),
     )
 
@@ -245,6 +258,7 @@ def test_load_mlx_lm_distributed_pipeline_filters_quant_shards(monkeypatch, tmp_
     assert ("download", local_shards) in events
     assert ("load", list(local_shards)) in events
     assert all(not path.exists() for path in load_paths)
+    assert tokenizer_calls[0]["trust_remote_code"] is trust
     assert config["eos_token_id"] == 3
     assert config["model_type"] == "llama"
     assert model._unsloth_mlx_distributed_parallel_mode == "pipeline"
--- a/tests/test_mlx_save_export_regressions.py
+++ b/tests/test_mlx_save_export_regressions.py
@@ -1788,9 +1788,11 @@ def test_vlm_sanitizer_replay_uses_real_model_instances():
     ) == {"visual.proj.weight": "tensor"}
 
 
+@pytest.mark.parametrize("trust", [False, True])
 def test_repair_degraded_vlm_processor_rebuilds_from_sidecar_configs(
     monkeypatch,
     tmp_path,
+    trust,
 ):
     import unsloth_zoo.mlx.loader as loader
 
@@ -1809,13 +1811,11 @@ def test_repair_degraded_vlm_processor_rebuilds_from_sidecar_configs(
     )
 
     image_processor = object()
-    monkeypatch.setattr(
-        loader,
-        "_build_vlm_image_processor_from_config",
-        lambda model_path, processor_config, preprocessor_config, model_type=None: (
-            image_processor
-        ),
-    )
+    def build_image_processor(*args, **kwargs):
+        assert kwargs["trust_remote_code"] is trust
+        return image_processor
+
+    monkeypatch.setattr(loader, "_build_vlm_image_processor_from_config", build_image_processor)
 
     (tmp_path / "processor_config.json").write_text(
         json.dumps({"processor_class": "FakeProcessor"}),
@@ -1839,6 +1839,7 @@ def test_repair_degraded_vlm_processor_rebuilds_from_sidecar_configs(
         degraded,
         tmp_path,
         "glm_ocr",
+        trust_remote_code=trust,
     )
 
     assert isinstance(repaired, FakeProcessor)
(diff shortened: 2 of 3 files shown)


Lyxot:
@codex review

Lyxot:
@codex review

danielhanchen:
Confirmed main only forwards trust_remote_code when it is True (unsloth_zoo/mlx/loader.py around the extra_kwargs block), so a nested processor tokenizer can still run repository code after an explicit False. This now makes repos like GOT refuse on the default path where they previously loaded; could you confirm the intended behaviour for existing users before I review?

danielhanchen:
Answering my own question above, with a simulation rather than an argument. I ran everything on Linux x86-64 with the real MLX runtime (`mlx` + `mlx-cpu`, not the torch shim), in isolated `venv`s, one fresh interpreter per case so a cached dynamic module could not hide an execution.

## Is this a real issue?

Yes, and it is wider than GOT. The fix is not "propagate a keyword": it is the only thing standing between an explicit `trust_remote_code=False` and arbitrary repository code.

`mlx-vlm` 0.6.4 ships 19 native processors that hardcode `trust_remote_code=True` for their nested tokenizer, and 11 of those `kwargs.pop("trust_remote_code", None)` first, i.e. they deliberately discard the caller's choice:

```
fastvlm  gemma4  gemma4_unified  kimi_vl  lfm2_vl  locateanything
molmo  molmo2  phi3_v  phi4mm  unlimited_ocr
```

Every fixture below carries a `tokenization_marker.py` that appends to a sentinel file when the module is imported, so the tables report whether the repository's Python *ran*, not whether a keyword was rewritten.

## Before vs after

172 cases per configuration (7 probes x 7 local repos x `trust_remote_code` in {absent, `False`, `True`, `None`} x cold/warm `HF_MODULES_CACHE` x whether the user answers `y` at the Transformers consent prompt), run against `origin/main` (`83dce16`) and this branch:

| Configuration | main: repo code executed without authorization | this PR |
|---|---|---|
| transformers 5.5.0, mlx-vlm 0.6.4, no torchvision | **8 / 172** | 0 / 172 |
| transformers 5.5.0, mlx-vlm 0.6.4, torchvision | **8 / 172** | 0 / 172 |
| transformers 4.51.3, mlx-vlm 0.1.27 (oldest supported) | 0 / 172 | 0 / 172 |

Every remaining execution on this branch is an explicit `trust_remote_code=True`, or bare Transformers with no Unsloth in the call path. The full production chain (`_bind_mlx_vlm_processor_loader` over the real `mlx_vlm.utils.load_processor`) on main:

| Repo shape | trust | main | this PR |
|---|---|---|---|
| remote tokenizer, `tokenizer.json` present | absent | **loads, marker fired** | refuses, names `tokenization_marker.py` |
| remote tokenizer, `tokenizer.json` present | `False` | **loads, marker fired** | refuses, names the file |
| remote tokenizer, no `tokenizer.json` | absent / `False` | **loads, marker fired** | refuses, names the file |
| same | `True` | loads | loads, same token ids |
| no remote code anywhere | all | loads | identical class, ids, eos/bos |

The image-processor half reproduces too, but only with torchvision installed (without it `AutoImageProcessor` is a torchvision-gated `DummyObject` that raises before any trust decision, so that leg is unreachable in the `Repo tests (CPU)` configuration):

| trust | main | this PR |
|---|---|---|
| absent, user answers `y` at the prompt | **`MarkerImageProcessor`, marker fired** | returns `None`, never prompts |
| absent / `False` / `None` | prompts on stdin | refuses without prompting |
| `True` | `TypeError` (no such keyword) | loads the remote processor |

I also stress-tested `_mlx_tokenizer_loading_scope` directly (patch/restore, restore after an exception, 8-deep nesting with mismatched flags unwound through an exception, a second thread during an active scope, two threads with opposite settings against a barrier, abandoned context manager). 8/8 pass here; 2 fail on main, and they are the two that matter:

```
main   override_direction        {'scope_false_inner_true': 'loaded (HOLE)', 'scope_true_inner_false': 'ValueError'}
PR     override_direction        {'scope_false_inner_true': '_MLXRemoteCodeError', 'scope_true_inner_false': 'loaded'}
main   concurrent_opposite_trust {'untrusted': 'loaded', 'trusted': 'ValueError'}
PR     concurrent_opposite_trust {'untrusted': '_MLXRemoteCodeError', 'trusted': 'loaded'}
```

main is wrong in both directions: it runs code a caller refused, and refuses code a caller authorized.

## One real defect, fixed in 8ba071b

`mlx_vlm/models/base.py` patches `AutoProcessor.from_pretrained` and wraps its native-processor branch in `except Exception: pass`, then chains to Transformers. So the refusal this PR raises inside `Phi4MMProcessor.from_pretrained` never reached the resolution-error handler: the call **succeeded** and returned the image-processor half.

```
[Phi4MMProcessor.from_pretrained] kwargs={'trust_remote_code': False}
[Phi4MMProcessor] RAISED _MLXRemoteCodeError: ... requires tokenization_marker.py ...
RESULT: CLIPImageProcessorPil | tokenizer: False
```

Security-wise that was still safe, but the caller got an object with no `.tokenizer` and no `.apply_chat_template` and found out via an `AttributeError` much later, instead of the named refusal this PR advertises. `refusals` was already being tracked; it just was not consulted on the success path. A recorded refusal now wins when the returned processor has no usable tokenizer, and a fallback that *did* produce one still takes precedence. Reverting only that check fails the new test; 225 tests pass with it.

## Behaviour change existing users will see

Real, and worth stating in the PR body. The `os.path.isfile(tokenizer.json)` guard means repos with a tokenizer `auto_map` and no serialized `tokenizer.json` now refuse on the default path where they used to load. Checking the published repos:

| Repo | `tokenizer.json` | tokenizer `auto_map` | affected |
|---|---|---|---|
| `stepfun-ai/GOT-OCR2_0`, `mlx-community/GOT-OCR2_0-4bit` | no | yes | **yes** |
| `mlx-community/dbrx-instruct-4bit` | no | yes (`tiktoken.py`) | **yes** (already documented in #1170) |
| `microsoft/Phi-4-multimodal-instruct` | yes | no | no |
| `allenai/Molmo-7B-D-0924`, `mlx-community/Molmo-7B-D-0924-4bit` | yes | yes | no |

What main did for those was not "load correctly": it stripped the `auto_map` and substituted a generic `TokenizersBackend`, so GOT's tiktoken-based tokenizer was silently replaced with one that produces different ids. Refusing and naming `tokenization_qwen.py` is the better of the two. It is still a default-path break, so it belongs in the PR description.

Studio is already set up for this: `_requires_trust_remote_code_for_model` (`studio/backend/routes/inference.py`) flags a raw `auto_map` and routes the load through the consent dialog, and both `mlx_inference.py` and `training/worker.py` forward the flag explicitly with a `False` default. Today Studio tells the user a repo needs approval and the MLX path loads it anyway; after this, the enforcement matches the UI.

## Blast radius elsewhere

`unsloth_zoo.mlx.loader` is not imported by `import unsloth_zoo` (verified: not in `sys.modules`, and no third-party `mlx` module loads), so Windows / Linux+NVIDIA / Linux+AMD / WSL / CPU-only are untouched. `mlx` does publish `win_amd64`, `win_arm64` and manylinux wheels now, so the Linux run above is a real platform, not a proxy for macOS. No browser or GPU surface is involved.

Not answerable here: Apple Silicon Metal execution. The three Apple lanes on this PR are SKIPPED and this host has no Apple hardware.

Beyond the PR's own 225: `test_mlx_save_export_edge_cases` 86, `test_mlx_moe_gguf_export_invariants` 33, `test_mlx_module_exports` 19, `test_mlx_switch_lora` 29, all green; `compileall`, `ruff --select E9,F63,F7,F82` and both exec-lint gates clean on the touched files.

Looks good to me otherwise. Could you add the GOT / DBRX default-path note to the description, and I will merge.

danielhanchen:
@codex review

danielhanchen:
@codex security review

danielhanchen:
@codex review

danielhanchen:
Cross-platform results, run off the org queue on an owned staging repo so the three OSes went in parallel.

| OS | Result |
|---|---|
| ubuntu-latest | 225 passed |
| macos-15 (Apple Silicon, arm64) | 225 passed |
| windows-latest | 224 passed, 1 failed |

Plus the repo's own suites on the staged branch: Lint, Wheel, Tests CI all green.

macOS matters most here and it is green: `macos-15` is arm64, which is the only hardware `is_mlx_available()` accepts (`platform.system() == "Darwin" and platform.machine() == "arm64"`), so this is the first run of these tests on the hardware the change actually ships to.

The single Windows failure is **pre-existing and unrelated** to this PR:

```
FAILED tests/test_mlx_save_export_regressions.py::test_trusted_dir_handles_a_root_trusted_path
  AssertionError: assert False is True
  where False = _trusted(..., env_value='\\')   # '\\' = os.sep
```

The test passes `env_value=os.sep`, which is `/` on POSIX (a real root) but `\` on Windows, where a root is `C:\`. Both the test body and `_trusted` are byte-identical between this PR's merge base and its head, and `unsloth_zoo/mlx/utils.py` is untouched, so nothing here caused it. It has simply never been observed because the repo's own `Repo tests (CPU)` lane is Linux-only. Worth a separate one-line fix; not this PR's to carry.

Two bugs in our staging harness surfaced first and are fixed, since both made the run report noise instead of an answer:

- The generated per-OS job never installed `unsloth`, and `unsloth_zoo/__init__.py` refuses to import without it. Its `ImportError` names no module, so the missing-dep forgiveness did not recognise it and 222 of 225 tests failed on all three OSes.
- The exit-2 branch forgave a collection error only when there was no `ERROR` line at all, but `pytest -q` prints exactly `ERROR <nodeid> - ...` for one, so it could never fire on the case it was written for. That is what reddened macOS on its own: this repo deliberately declares no torch on darwin arm64, so the torch-backed simulation shim is absent by design there.

No UI evidence to attach. The change ships no UI; its one user-visible surface is the refusal text, and that is reachable only behind `is_mlx_available()`, i.e. Apple Silicon, which this host is not. A before/after pair produced here would show two identical screens and prove nothing.

PR #1187, opened 2026-09-10
Unbreak main against transformers 5.17 and trl 1.13

danielhanchen:
`Core drift (3 upstream pins)` has been red on main since 2026-09-09, and it takes unsloth's `Core (HF=latest + TRL=latest)` down with it, since that job clones zoo main and runs this suite. Three failures, all from upstream releases published that afternoon.

`transformers` 5.17.0 went to PyPI at 15:39Z. Nothing we merged is involved: unsloth-zoo #1183 landed about four hours earlier and touches only the runtime eager-fallback wrapper, not the source-generation path.

## Gemma 3 stopped getting fused linear cross entropy

This is the one worth reading. 5.17 rewrote `Gemma3ForConditionalGeneration.forward` to delegate the loss:

```python
loss = self.loss_function(
    logits=logits, labels=labels, vocab_size=self.config.text_config.vocab_size, **lm_kwargs
)
```

That is `cross_entropy_find_2`'s shape, so the rewrite should still have fired. It did not, because `$KWARGS$` expanded to `(?:, \*\*(loss_kwargs|kwargs))?` and gemma3's is spelled `lm_kwargs`.

A finder that does not match is a silent no-op. `apply_fused_lm_head` returns the forward untouched, prints nothing outside `UNSLOTH_ENABLE_LOGGING=1`, and the model trains correctly while using the memory the fused path exists to save. The only thing that noticed was the drift test, and it reported the missing string rather than the consequence.

`$KWARGS$` now captures any identifier. The replacements splice the captured name in, so the next spelling upstream picks needs no change here. Measured on transformers 5.17.0:

| class | before | after |
| --- | --- | --- |
| Gemma3ForConditionalGeneration | applied=False | applied=True |
| Gemma3ForCausalLM | applied=True | applied=True |
| Qwen2_5_VLForConditionalGeneration | applied=True | applied=True |
| Gemma2ForCausalLM | applied=True | applied=True |
| LlavaForConditionalGeneration | applied=True | applied=True |
| PaliGemmaForConditionalGeneration | applied=True | applied=True |

One case flips, nothing regresses, and every applied rewrite still parses.

`test_compiler_cross_entropy_find_3_shift_logits_pattern` was asserting a literal string in upstream's source, which is why it reported drift instead of the no-op it caused. It is now two tests:

- `test_the_fused_cross_entropy_rewrite_still_reaches_gemma3` runs the real rewriter against the installed forward and requires it to apply and to `ast.parse`. Whatever shape upstream picks next, the contract is that some finder matches it.
- `test_compiler_cross_entropy_find_3_still_matches_its_own_shape` keeps `cross_entropy_find_3` covered against a frozen 4.57.6 fixture, so the pattern stays guarded even though no shipped model uses that shape any more.

Both fail when the fix is reverted.

The zoo-site line numbers in the drift messages are now resolved at runtime. They had gone stale twice, still citing `:1508` / `:1599` / `:1683` for finders that had moved past `:2500`, which sends the next reader to unrelated code.

## MiniMaxM3VL3DRotaryEmbedding is gone

5.17 renamed it to `MiniMaxM3VLVisionRotaryEmbedding` and lifted the `get_vision_position_ids` call out of it into `MiniMaxM3VLVisionModel.forward`.

Re-pointing the test at the new class would assert nothing. `MiniMaxM3VLVisionModel` is a `PreTrainedModel` subclass and is never compiled, so no `_forward` is emitted for it at all. I checked the rest of 5.17 too: of the 41 classes that call the helper, 40 are `PreTrainedModel` subclasses, and the single plain `nn.Module`, paddleocr_vl's encoder, defines the helper itself and so fails the test's own precondition. No shipped model presents this shape any more.

So the end to end proof moves to a synthetic fixture. It writes a real modeling package to disk (so `inspect.getsource` and `linecache` behave as they do upstream), hangs it off `transformers.models.__path__` so the compiler's own `importlib.import_module` finds it, and drives the real `unsloth_compile_transformers`. One leaf module calls an imported-only listed helper and must come out `fullgraph = False`; a second calls nothing listed and must stay `fullgraph = True`, so a blanket demotion fails it just as loudly. Verified by breaking `calls_disable_compile_function` in both directions.

The upstream case is kept and skips with a message naming the transformers version and the missing class, so it still runs on anything below 5.17.

## trl's `_tp_size`

29 failures in `test_trl_sft_logits_metrics.py`:

```
trl/trainer/sft_trainer.py:1872: self._total_train_tokens += num_tokens_in_batch // self._tp_size
AttributeError: 'SFTTrainer' object has no attribute '_tp_size'. Did you mean: 'get_tp_size'?
```

Not a rename, despite the hint. `get_tp_size` is an unrelated `transformers.Trainer` method. trl 1.13's `SFTTrainer.__init__` sets `_tp_size` itself, so real users are fine; the test builds its trainer with `object.__new__` and hand-seeds the instance state, and never picked the new attribute up. No unsloth or zoo code reads `_tp_size`, so there is nothing to migrate.

Seeded, and `test_the_stub_covers_what_trl_reads` now compares the attributes `compute_loss` reads against what the stub provides. The next attribute trl adds fails one test that names it, instead of 29 that do not.

## One more, uncovered by fixing that

`test_the_entropy_warning_is_not_shown_when_both_are_omitted` was failing for the `_tp_size` reason, and would have kept failing for a second one underneath it.

It read the warning through `caplog`, which installs its handler on the root logger. `misc.py` borrows a transformers logger, and once `unsloth` is installed and imported (which every lane does) that logger stops propagating to root. The warning still reached the user on stderr, so the test was reporting "not shown" about a message that was shown. It now attaches to the logger that emits it.

## Verification

Against transformers 5.17.0, trl 1.13.0, peft 0.20.0, torch 2.10 CPU, with `unsloth` installed as the lanes do:

```
616 passed, 57 skipped
```

across `test_compiler_dynamic_exec`, `test_compiler_rewriter_exhaustive`, `test_upstream_source_patterns`, `test_upstream_signatures`, `test_temporary_patches_exhaustive`, `test_zoo_source_upstream_refs`, `test_trl_sft_logits_metrics`, `test_disable_compile_functions_reach_imported_helpers`, `test_ci_lane_scripts_parse`, `test_upstream_pinned_symbols_transformers` and `test_extended_dep_api_pins`. Three consecutive runs, same result each time.

Merging this also clears `Core (HF=latest + TRL=latest)` on unslothai/unsloth, which is currently red for exactly these three reasons.

### DIFF
--- a/tests/test_disable_compile_functions_reach_imported_helpers.py
+++ b/tests/test_disable_compile_functions_reach_imported_helpers.py
@@ -18,12 +18,15 @@
 `called_functions` holds only names the modeling file both calls AND defines, so an
 imported-only helper is imported raw for Dynamo to inline, and the CALLER is demoted."""
 
+import importlib.util
 import inspect
+import json
 import os
 import subprocess
 import sys
 
 import pytest
+import transformers
 
 from unsloth_zoo import compiler as compiler_module
 from unsloth_zoo.compiler import (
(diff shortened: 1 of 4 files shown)


danielhanchen:
Confirmed the narrow KWARGS pattern at compiler.py:2878 means Gemma3ForConditionalGeneration silently falls off the fused linear cross entropy path on transformers 5.17, and that the drift job only reported the missing string rather than the memory cost. Will get this reviewed.

danielhanchen:
Compatibility evidence for this change, since it touches the source rewriter that every model goes through.

The whole production diff is one regex: `$KWARGS$` now captures any identifier instead of the literal pair `loss_kwargs|kwargs`.

### transformers, every release in range

Ran `apply_fused_lm_head` over every `*ForCausalLM` / `*ForConditionalGeneration` forward in 4.57.6 and in all 34 published 5.x releases, once on this branch and once on the merge base. 792 (version, class) pairs.

| Result | Count |
| --- | --- |
| Behaviour changed | 1 |
| Rewrites lost | 0 |
| New errors | 0 |
| Parse failures on generated source | 0 |

The single change is `5.17.0 Gemma3ForConditionalGeneration`, False to True, which is the bug this PR fixes.

The boolean sweep only sees a rewrite starting or stopping, so I also diffed the generated source text on transformers 4.57.6: 214 classes, byte identical between base and branch. Widening the capture emits the same code everywhere the old two name list already matched.

### TRL and PEFT

12 matrix points, all green: trl 0.22.2, 0.24.0, 0.25.1, 0.29.1 on transformers 4.57.6, then 1.0.0, 1.3.0, 1.6.0, 1.9.2, 1.13.0 against 5.0.0, 5.5.0, 5.10.0, 5.16.1, 5.17.0. peft 0.18.0, 0.19.1 and 0.20.0.

### PyTorch

2.8.0, 2.9.1, 2.10.0, 2.11.0, 2.12.1, 2.13.0 and 2.14.0 all pass 128 of 135 (7 skips). 2.6.0 and 2.7.1 fail identically on base and on this branch with `module 'torch.utils._pytree' has no attribute 'register_constant'`, so that is a pre-existing floor, not something this PR moves. Each point was built in its own fresh venv, since swapping torch inside a reused one leaves bitsandbytes ABI mismatched and measures the environment rather than the code.

### GPU, gemma-3-4b-it on B200, transformers 5.17.0

`Gemma3ForConditionalGeneration_forward` contains the fused cross entropy call only on this branch. Base has it for `Gemma3ForCausalLM` only, which is the silent regression.

Deterministic single forward and backward on fixed tensors, three repeats per leg: loss 16.963802337646484 on both legs every time. Grad norms base [34.3339, 34.3195, 34.3195] against branch [34.3331, 34.3601, 34.3601], so the between leg difference is smaller than the within leg spread.

I dropped the training loop version of that comparison. With 4 bit weights and bf16, base against base spread more than base against branch, so it had no resolving power.

### GRPO with fast_inference

The one contract change a GRPO run sees is that `UNSLOTH_RETURN_HIDDEN_STATES` now takes effect for this class, so `.logits` carries `[B, S, hidden]`. `rl_replacements` dispatches on that width and handles both.

```
base  logits=[2, 32, 262208]  hidden=[2, 32, 262208]  lm_head=[262208, 2560]
head  logits=[2, 32, 262208]  hidden=[2, 32, 2560]    lm_head=[262208, 2560]  recon_delta=0.0
```

`F.linear(hidden, lm_head)` reproduces the old logits exactly, so the hidden states handed to GRPO are the right ones.

Full GRPO run with `fast_inference = True` on vllm 0.11.2, transformers 4.57.6, torch 2.9.0: rewards bit identical across 10 runs. Step 2 loss is bimodal at -0.5169 and -0.5186, and both legs produce both values across repeated runs with isolated compile caches, so that is vllm side nondeterminism rather than a difference between the legs.

vllm above 0.11.2 is unaffected by construction: everything up to 0.15.1 pins `transformers<5`, where this change is a verified no-op across all 24 classes. 0.29.0 is the first release whose pin reaches transformers 5.x.

### CPU numerics

Bit identical loss and gradients, 0.000e+00, in single chunk fp32 across dense labels, `-100` padding, `num_items_in_batch` passed as int and as tensor through `**lm_kwargs`, fully masked rows, `UNSLOTH_RETURN_LOGITS=1`, multimodal with pixel_values, and `torch_compile=True`. With 4 chunks forced, deltas sit at the fp32 rounding floor near 1e-7 relative. A deliberately wrong negative control diverged by 1.5, so the harness can see a difference when there is one.

### Portability

Four files touched, one of them production, and the production change is a regex line. Nothing vllm, MLX, CUDA, ROCm, XPU or platform specific is touched, so there is no Linux, Mac, Windows or WSL exposure and no browser surface. All file IO in the touched test files names an encoding, checked by walking the AST rather than grepping, and paths go through `os.path.join` / `pathlib` with `os.pathsep` for PYTHONPATH.

### Separate finding, not part of this PR

vllm 0.29.0 removed `vllm.model_executor.layers.quantization.bitsandbytes` entirely, and `unsloth_zoo/vllm_utils.py:176` imports it at module scope with no guard. On main today that makes any `fast_inference = True` load fail with `ModuleNotFoundError` on vllm 0.29.0, in 16 bit as well as 4 bit, before any model code runs. The neighbouring `bitsandbytes_loader` patch a few lines above is already wrapped in try/except, so the unguarded import is the outlier. Worth a follow up on its own.

PR #1195, opened 2026-09-13
fix(fla): skip redundant autotune key work on each kernel launch

Souravrajvi0:
## Summary

- Stop building `AutotuneKey` on every Triton launch when `FLA_CACHE_MODE` is `disabled` (the default).
- After `compile_fla_no_autotune` installs `_ReuseBestCache`, launch kernels directly with the pinned config instead of going through `Autotuner.run` key construction on every call.

Fixes the CPU-side overhead reported in #10806 (Qwen3.5-9B LoRA on B200: ~2.5k FLA launches/step spending ~180 ms in `CachedAutotuner.run` despite configs already being pinned).

## Test plan

- [x] `pytest tests/test_fla_cached_autotuner.py`

Related: unslothai/unsloth#10744 (broader perf work); this PR is scoped to the vendored FLA autotuner launch path.


danielhanchen:
The autotuner change in unsloth_zoo/_vendored/fla/ops/utils/cache.py is a separate concern from the gpt_oss optional-import rework, and that second part now only applies the Mxfp4GptOssExperts patch when triton_kernels imports, where main applies it on the bf16 fallback path too. Could you split the gpt_oss changes into their own PR and confirm gpt-oss loading without triton_kernels still behaves the same?

Souravrajvi0:
Split per review: gpt_oss optional-import rework is now in #1198. This PR is FLA-only (autotuner fast path + Codex fixes for `FLA_CACHE_MODE=always` and test `sys.modules` isolation).

Souravrajvi0:
@danielhanchen Thanks for the review  split is done.

**This PR (#1195)** is FLA-only now: `cache.py` fast path, `compiler.py` `_ReuseBestCache`, and `tests/test_fla_cached_autotuner.py`. Also addressed Codex feedback (skip the reuse-best fast path when `FLA_CACHE_MODE=always`; module-scoped fixture so tests do not leave a `MagicMock` in `sys.modules["torch"]`).

**gpt_oss optional-import rework** is in **#1198** (`feat/gpt-oss-patch-cascade-19d8`). That branch is rebased onto current `main` (conflict in `gpt_oss.py` resolved by keeping the transformers 5.16 tensor-parallel guard from upstream while keeping `skip_patch` / `load_and_swizzle_ready`).

On **loading without `triton_kernels`**: `patch_gpt_oss()` still sets `Mxfp4HfQuantizer.is_trainable` and then **returns early** without touching `is_kernels_available`, so transformers can still set `dequantize=True` and use the dequantized bf16 path — same as `main`'s `else` branch when `HAS_TRITON_KERNELS` is false. Custom `Mxfp4GptOssExperts` / swizzle / load hooks only run when `triton_kernels` is present (and the mxfp4 expert replacement is gated on successful `matmul_ogs` import). Covered by `tests/test_gpt_oss_patch_cascade.py`.

Happy to adjust if you want different behavior when `triton_kernels` is installed but a sub-import fails.

danielhanchen:
The reuse-best fast path calls self.fn.run directly, so the reset_to_zero and restore_value pre/post hooks set up in CachedAutotuner.__init__ are skipped, and _runtime_autotune_tuple re-derives Triton's own key construction, which will drift across Triton versions. Could you confirm kernels declaring reset_to_zero still get their buffers zeroed on that path, and consider reusing Triton's key computation instead of duplicating it?

Souravrajvi0:
@danielhanchen Good catch on both points  addressed in 1386a83d on this branch.

**`reset_to_zero` / `restore_value` on the reuse-best fast path**

`_run_with_stored_config` no longer calls `self.fn.run` bare. It now follows the same hook sequence as Triton’s `Autotuner._bench` `kernel_call`: build `full_nargs`, run the config’s `pre_hook` (if any), then **`self.pre_hook(full_nargs)`**, launch the kernel, and **`self.post_hook(full_nargs, …)`** on success or failure. That includes the None-safe hooks installed in `CachedAutotuner.__init__`, so kernels that declare `reset_to_zero` still get those buffers zeroed on the fast path.

Added tests in `tests/test_fla_cached_autotuner.py` for pre/post hooks on reuse-best and for `reset_to_zero` via `CachedAutotuner.__init__`.

**Autotune key construction**

Removed the separate `_runtime_autotune_tuple` duplicate. Runtime keys now go through **`triton_runtime_autotune_key()`**, which delegates to **`Autotuner.make_autotune_key`** when the installed Triton exposes it, and otherwise uses the same key logic as **`Autotuner.run`**. **`AutotuneKey.build`** calls the same helper, so FLA cache lookup and Triton’s cache key stay aligned from one place.

Happy to adjust if you’d prefer the fast path to fall back to `super().run()` whenever `reset_to_zero` / `restore_value` is non-empty instead.

danielhanchen:
The reuse-best path in unsloth_zoo/_vendored/fla/ops/utils/cache.py now bypasses Autotuner.run for every FLA launch, so any behaviour Triton does there beyond the pre/post hooks you replicate is skipped on a path that works today. Could you confirm a linear-attention training run (loss curve and timing) matches main with FLA_CACHE_MODE unset before I review?

Souravrajvi0:
@danielhanchen Thanks for flagging this — you're right that the reuse-best shortcut was bypassing `Autotuner.run`, so anything Triton does there beyond the hooks we duplicated wasn't guaranteed to match `main` with `FLA_CACHE_MODE` unset.

I removed that fast path in **`4d606fa0`**: every launch goes through Triton's `Autotuner.run` again (same steady-state path as `main` with `compile_fla_no_autotune`'s `_ReuseBestCache`). The only remaining change when the FLA disk cache is off (default / unset) is skipping redundant per-launch autotune-key work before `super().run()`.

**CPU-side checks** (no GPU on the worker I used):
- `tests/test_fla_cached_autotuner.py` — 5 tests, including "warm cache still calls parent `Autotuner.run`" and "`FLA_CACHE_MODE` disabled → no key build"
- **`81c0dd47`**: `triton_runtime_autotune_key` matches `Autotuner.make_autotune_key` on Triton 3.6 (tensor + scalar cases)

I haven't run a linear-attention training A/B (loss + timing vs `main`) on GPU from that environment. If you still want that before merge, I can run it on a CUDA box; otherwise the launch path should now match `main` for unset `FLA_CACHE_MODE`, with the micro-optimization limited to avoiding unused key materialization when the disk cache is disabled.

danielhanchen:
Confirmed main builds the autotune key on every launch in `unsloth_zoo/_vendored/fla/ops/utils/cache.py` even when the FLA disk cache is off, so the guard is doing real work.

Could you still post the linear-attention loss and step-time comparison against main with `FLA_CACHE_MODE` unset, so I can size the win before reviewing?

Issue #1119, opened 2026-08-27
patch_gpt_oss abandons every later patch when one optional import fails


danielhanchen:
Found while sweeping transformers 5.0.0 through 5.16.1 for the MXFP4 load-path work (#1118).

`unsloth_zoo/temporary_patches/gpt_oss.py` does, at line 466, inside `patch_gpt_oss` (which spans roughly lines 114 to 577):

```python
return raise_error(...)
```

`raise_error` does not raise. It logs and returns, and only under `UNSLOTH_ENABLE_LOGGING`. So a single failed optional import silently abandons the remainder of a 460-line function, taking every patch defined below that point with it.

## Where it bites today

On transformers 5.10.0, `transformers.integrations.tensor_parallel` does not exist. Upstream moved it to `transformers/distributed/tensor_parallel.py` and deleted `shard_and_distribute_module` at that tag, then restored the `integrations` layout in 5.10.1 fourteen minutes later. PyPI has since yanked 5.10.0.

That import sits under a `< 5.16.0` gate, which 5.10.0 satisfies, so `patch_gpt_oss` bails at line 466 and loses **two** patches, not one:

- `load_and_swizzle_mxfp4`
- `replace_with_mxfp4_linear`

with no message unless logging is enabled.

The loss is conditional. `patch_gpt_oss` returns unconditionally at line 156 when `triton_kernels` is absent, so on installs without it nothing extra is lost. Traced: `last_line_executed=156`, `tensor_parallel_import_line=466`, `reached_tp_import=False`. The degradation is therefore real only on true GPT-OSS MXFP4 GPU setups.

## Why this is worth its own issue

The specific 5.10.0 symptom is being handled in #1118 (the test is `xfail`ed on exactly that version, with the consequence written out). This issue is about the pattern, which is not specific to mxfp4:

1. A mid-function `return raise_error(...)` reads like an error path but is a **silent** early return.
2. It couples unrelated patches: an optional import for feature A silently disables features B and C defined later in the same function.
3. It is invisible by default, because `raise_error` only speaks under `UNSLOTH_ENABLE_LOGGING`. That invisibility is precisely what let a real load-path corruption survive undetected across the whole transformers 5.x line until #1114.

## Possible directions

- Scope the optional import to the patch that needs it, rather than the whole function.
- Make each patch in `patch_gpt_oss` independently guarded so one failure cannot cascade.
- Make a skipped patch visible without requiring an env var, at least once per process.

[...]

PR #1198, opened 2026-09-13
fix(gpt_oss): skip silent cascade on optional import failures

Souravrajvi0:
## Summary

Split from #1195 per maintainer review.

When optional imports inside `patch_gpt_oss()` fail (e.g. `triton_kernels`, tensor parallel helpers), use `skip_patch()` instead of aborting the entire patch chain. This fixes the silent cascade reported in #1119.

## Changes

- Add `skip_patch()` in `temporary_patches/utils.py`
- Refactor `patch_gpt_oss()` optional-import blocks to continue after non-fatal failures
- Gate custom `Mxfp4GptOssExperts` replacement on successful `triton_kernels` / `matmul_ogs` import
- AST regression tests in `tests/test_gpt_oss_patch_cascade.py`

## Behavior note (for review)

Without `triton_kernels`, loading should still follow transformers' dequantized bf16 fallback (`is_kernels_available` left intact). Please confirm this matches `main` for your gpt-oss load paths.

## Test plan

- [x] `pytest tests/test_gpt_oss_patch_cascade.py`

Related: #1195 (FLA autotuner perf — now FLA-only)


Souravrajvi0:
Rebased onto latest `main` (83dce162). Resolved conflict in `gpt_oss.py` by keeping upstream's transformers 5.16 tensor-parallel guard while preserving `skip_patch` / `load_and_swizzle_ready` behavior.

Souravrajvi0:
@danielhanchen Following up here since this is the split-out gpt_oss PR from #1195.

**Scope:** optional-import cascade fix (#1119) — `skip_patch()` in `utils.py`, refactored `patch_gpt_oss()`, AST tests in `test_gpt_oss_patch_cascade.py`.

**No `triton_kernels`:** unchanged intent vs `main` — after `is_trainable` is patched we return without overriding `is_kernels_available`, so the bf16 dequant fallback path is unchanged.

**With `triton_kernels`:** custom MXFP4 expert / swizzle / loader patches apply only when imports succeed; partial failures log via `skip_patch` instead of aborting the whole chain. Rebased on latest `main`; merged your 5.16 TP loader guard with this structure.

Tests: `pytest tests/test_gpt_oss_patch_cascade.py` (3 passed locally after rebase).

danielhanchen:
Confirmed the early return raise_error(...) calls are still there in unsloth_zoo/temporary_patches/gpt_oss.py, so scoping each optional import is the right direction, and thanks for splitting it out of #1195. The diff is mostly a reindent of the Mxfp4GptOssExperts classes under the matmul_ogs_available guard though; could you narrow it to the guard changes, for instance by keeping the class definitions where they are and gating only the patch_function calls?

danielhanchen:
Confirmed the cascade is still there in unsloth_zoo/temporary_patches/gpt_oss.py, where the optional imports around lines 458 and 468 abandon load_and_swizzle_mxfp4 and replace_with_mxfp4_linear, and the split out of #1195 plus the rebase look right. Most of the diff is re-indenting the two expert classes under if matmul_ogs_available:, which makes the real change hard to see, so could you keep the class bodies where they are and gate them with an early guard instead?

Souravrajvi0:
Addressed the reindent review: the `Mxfp4GptOssExperts` / `_Training` class bodies stay at their original indent. Optional import failures still use `skip_patch()` instead of `return raise_error(...)`, and only the `patch_function` calls (plus routing / `load_and_swizzle_mxfp4` / `replace_with_mxfp4_linear`) are gated with early flags like `matmul_ogs_available`.

`gpt_oss.py` vs `main` is now ~+34/−20 instead of a class-body reindent. AST tests still pass, including one that the expert classes are not nested under the `matmul_ogs` guard.

danielhanchen:
Confirmed the early returns are still in unsloth_zoo/temporary_patches/gpt_oss.py around lines 577, 601, 611 and 694, so scoping each optional import with skip_patch is the right fix for #1119. Thanks for keeping the expert class bodies in place, the diff is much easier to read now and I will get this reviewed.

PR #1200, opened 2026-09-13
Honour the Inductor single compile worker forcing instead of undoing it

danielhanchen:
## Summary

`unsloth/_gpu_init.py` fingerprints the cgroup pinned GPU case (`docker --gpus '"device=N"'`) and sets `TORCHINDUCTOR_COMPILE_THREADS=1` together with the `UNSLOTH_FORCE_SINGLE_COMPILE_WORKER=1` sentinel. It has to: Inductor's compile worker subprocesses cannot enumerate a GPU that the cgroup pinned, so they raise

```
RuntimeError: Could not find an active GPU backend
```

from `torch/_inductor/runtime/triton_helpers.py::set_driver_to_gpu`.

That producer half is already on `unsloth` main. This repo is the consumer, and two places here silently undo it.

## The two fixes

**1. `patching_utils.py`, `patch_torch_compile`.** The non debug branch unconditionally did `os.environ.pop("TORCHINDUCTOR_COMPILE_THREADS", None)`, which lets Inductor respawn its compile worker pool. It now skips the pop when `UNSLOTH_FORCE_SINGLE_COMPILE_WORKER` is exactly `"1"`.

**2. `temporary_patches/common.py`, `determine_compile_threads`.** It returned the cpu count default. `get_torch_compile_options` puts that value into the Inductor options dict, and the options dict takes precedence over the environment variable, so the env var was overridden even when it survived step 1. It now returns `1` when `TORCHINDUCTOR_COMPILE_THREADS` is `"1"`.

The match on the sentinel is exact. `_gpu_init` writes the literal `"1"`, and the opt out knob's own default is `"auto"`, so `"auto"`, `"0"` and `""` all keep the pre-existing behaviour.

## Scope

Nothing in this repo sets either variable. On a host where nothing sets them, both code paths are byte for byte what they were: bare metal, `--gpus all`, Colab, Kaggle and Runpod are unaffected. The only behaviour that changes is on a host that has already asked for a single compile worker.

One thing worth naming: `determine_compile_threads` is `lru_cache`d. That is fine in production, since `_gpu_init` sets the variable at `import unsloth`, long before any compile options are built, but it does mean the value is fixed once per process.

## Tests

`tests/test_inductor_single_compile_worker.py`, 9 functions, CPU only, wired into the executing pytest step of the `Repo tests (CPU)` job.

| test | covers |
|---|---|
| `test_pop_is_skipped_when_single_worker_is_forced` | the env var survives `patch_torch_compile` |
| `test_pop_still_happens_without_the_sentinel` | control: the pre-existing cleanup is untouched on a bare metal host |
| `test_only_the_exact_sentinel_value_keeps_the_variable` | `"0"`, `""`, `"auto"` and `"true"` are not read as forcing |
| `test_determine_compile_threads_honours_the_env_var` | the options dict half |
| `test_determine_compile_threads_unchanged_when_unset` | control: PR #187's auto detection stands |
| `test_determine_compile_threads_ignores_other_values` | a request for 4 workers still gets the auto detected value, so the change is not widened |

Two of the nine fail against the code they fix. The tests set and clear both variables themselves, so they are valid on a host that has neither.

Split out of #694, which is being closed. The Gemma3 part of that PR is already fixed on main and better, so it is dropped rather than carried over.


danielhanchen:
## Verification update: a regression this branch introduced, now fixed

I ran this end to end rather than only through its unit tests, and the original version of the `determine_compile_threads` change was wrong in a way the tests could not see.

**vLLM sets `TORCHINDUCTOR_COMPILE_THREADS="1"` unconditionally at import**, in `vllm/env_override.py`. The first version of this branch read that variable directly, so on any machine with vLLM installed the first call cached `1` and every `torch.compile` in the process ran with a single Inductor worker, on an ordinary host where nothing had asked for that. `determine_compile_threads` is `lru_cache(1)`, so it stuck for the life of the process.

Measured with nothing forced, vLLM present, 192 core host:

| | `determine_compile_threads()` |
|---|---|
| `origin/main` | 32 |
| this branch, first version | **1** |
| this branch, after the fix | 32 |
| this branch, sentinel set | 1 |

The unit tests missed it because they clear the `lru_cache` and set the variable themselves, so they never observed the value a real import sequence produces. That is the general lesson here: the tests demonstrated internal consistency, not real behaviour.

**The fix** is to gate on `UNSLOTH_FORCE_SINGLE_COMPILE_WORKER`, which `_gpu_init` sets alongside `TORCHINDUCTOR_COMPILE_THREADS` and which nothing else writes. `patch_torch_compile` already gated on the sentinel, so the two halves now agree. The compile-threads variable is shared, and this is the proof that unrelated libraries write it.

Two tests added: one pinning that the env var **alone** does not force a single worker, and one checking the vLLM premise against the installed package rather than taking it on trust.

## What happens before and after

- **Before:** on a cgroup-pinned GPU (`docker --gpus '"device=N"'`), Inductor respawns its compile worker pool and dies with `RuntimeError: Could not find an active GPU backend`.
- **After:** the single-worker forcing set by `unsloth/_gpu_init.py` survives, and that crash does not occur.
- **On every other host:** byte-for-byte unchanged. Verified directly, including the vLLM case above, which is the one that nearly broke.

## Backwards compatibility

`unsloth/_gpu_init.py:175-212` already monkeypatches `determine_compile_threads` to `lambda: 1` and rewrites the zoo options dicts as a workaround for this exact problem. Both paths set the same value, so a mixed install in either direction is consistent:

| | old unsloth + new zoo | new unsloth + old zoo |
|---|---|---|
| result | safe, the old workaround is redundant but harmless | safe, old zoo pops and the workaround compensates |

Once this lands, that workaround can be retired separately.

danielhanchen:
@codex review

danielhanchen:
@codex security review

danielhanchen:
Confirmed the unconditional os.environ.pop("TORCHINDUCTOR_COMPILE_THREADS") is still there in patching_utils.py and that get_torch_compile_options would put the cpu count over it, and the sentinel gate keeps vLLM hosts on the auto detected count. Will get this reviewed.

danielhanchen:
Confirmed the unconditional pop is still there in unsloth_zoo/patching_utils.py:184 and that the options dict from determine_compile_threads outranks the env var, and the sentinel gate keeps the vLLM TORCHINDUCTOR_COMPILE_THREADS=1 case at the auto detected count. Will get this reviewed.

PR #1204, opened 2026-09-14
Do not compile Gemma3's vision mask builder, whatever the call looks like

danielhanchen:
## Summary

Gemma3 vision inference crashes on the first `generate` whenever `torch.compile` is enabled. It fails identically for `dtype=torch.float16` and `dtype=torch.bfloat16`, so it is not a dtype problem:

```
torch._dynamo.exc.Unsupported: Data-dependent branching
  transformers/masking_utils.py:685 in flex_attention_mask
    if attention_mask is not None and not fast_all(attention_mask):
  from user code:
    unsloth_compiled_cache/unsloth_compiled_module_gemma3.py in create_masks_for_vision_model
      -> create_causal_mask(...)
```

`fast_all()` returns a 0-dim tensor used in a Python `if`, which `fullgraph = True` cannot capture.

## Root cause

Mask builders were already meant to be emitted uncompiled, but the rule was a source substring in `DISABLED_KEYWORDS`:

```python
"create_causal_mask(**mask_kwargs)",  # Gemma3 create_masks_for_generate
```

transformers has since added a keyword argument to that exact call in Gemma3:

```python
full_mask = create_causal_mask(**mask_kwargs, block_sequence_ids=block_sequence_ids)
```

The closing paren moved, the literal stopped matching, and `create_masks_for_vision_model` started getting `@torch_compile_with_fallback(fullgraph = True, ...)`. gemma4 and gemma4_unified, whose spelling was untouched, still match and still work, which is why only Gemma3 broke.

| model | old literal matches | before this PR |
|---|---|---|
| `gemma3` | no, extra kwarg | compiled fullgraph, crashes |
| `gemma4` | yes | not compiled, works |
| `gemma4_unified` | yes | not compiled, works |

## Fix

Match the call rather than its spelling. `calls_mask_creation_function()` reuses two helpers that already exist in `compiler.py`:

- `calls_disable_compile_function()`, the existing `re.search(r"[^\w.]" + name + r"\s*\(")` detector, which is argument-agnostic and already ignores attribute calls like `self.create_causal_mask(`
- `get_mask_functions()`, which reads the `create*` names off the installed `transformers.masking_utils`

Reading the names off the installed transformers means no version gate is needed: a release that renames, adds or drops a factory is tracked automatically. Applied at both standalone-function emit sites.

The drifted literal is removed from `DISABLED_KEYWORDS` and replaced with a comment explaining why a spelling-sensitive entry must not go back.

## Blast radius, measured

AST scan of all 495 modeling files in transformers 5.17, restricted to module-level functions, which are the only ones this loop sees:

```
Module-level functions calling a masking_utils create* factory:
  create_masks_for_vision_model   n=3   models=[gemma3, gemma4, gemma4_unified]
```

That is the entire set. The 65 files containing `create_causal_mask(**mask_kwargs)` are class methods, which this loop never reaches. `rotate_half`, `apply_rotary_pos_emb` and `eager_attention_forward` are unaffected and still compiled.

## Verification

Bisected first by deleting that one decorator from the generated cache under `UNSLOTH_COMPILE_OVERWRITE=0`, which fixed it with compile otherwise fully enabled, then fixing the compiler and regenerating.

B200, transformers 5.17.0, torch 2.13.0, gemma-3-4b-it, compile enabled:

| check | before | after |
|---|---|---|
| vision float16, 2 passes for cache reuse | `Unsupported` | correct caption |
| vision bfloat16, 2 passes | `Unsupported` | correct caption |
| text-only bfloat16 | ok | unchanged |
| gemma-3n audio | ok | unchanged |
| Gemma3 LoRA training, 10 steps | ok | loss 3.666, 14 functions still compiled |

The training log confirms both halves of the intended behaviour:

```
Unsloth: Cannot compile function create_masks_for_vision_model since it builds attention masks via create_causal_mask.
```

Cross-version: on transformers 4.57.6 no modeling file defines `create_masks_for_vision_model`, so the rule is a no-op and the new tests skip. On 5.17 all three models are excluded.

Tests: 336 passed, 0 failed on the compiler lane including the `tests/security` hard gate. The wider fallback lane shows 10 failures in the `torch.compile` eager-fallback machinery; these reproduce identically on unpatched main in the same environment and are unrelated to this change.

## Tests

Adds `tests/test_mask_builders_not_fullgraph.py`, modelled on `tests/test_vision_utils_grid_helpers_eager.py`. Its central test reproduces the production error on CPU in under a second by really compiling the installed Gemma3 builder with `fullgraph=True`, so this cannot regress unnoticed on a suite with no GPU runner. It also pins the argument-shape independence the old literal lacked: `create_causal_mask(**mask_kwargs, block_sequence_ids=x)`, the plain form and a split-line form are all caught, while `self.create_causal_mask(`, `utils.create_causal_mask(` and a docstring mention are not.

`tests/test_disable_compile_functions_reach_imported_helpers.py` counts the detector's call sites; updated for the one the new helper adds.

## Not included

A blanket `except torch._dynamo.exc.Unsupported -> eager` fallback was considered and rejected. `_fall_back_to_eager_on_recompile_limit` catches `Unsupported` only for cache exhaustion and our own disabled-hook graph break, and that contract is asserted by `test_recompile_limit_fallback.py`, `test_eager_fallback_latch.py` and `test_disabled_hook_graph_break.py`, which all require a data-dependent `Unsupported` to propagate. Broadening it would break those tests, silently de-optimise unrelated regions, and re-run functions that may mutate cache or RNG state.

## Follow-up, not in this PR

Standalone lifted functions never get the `.item()` / `.tolist()` / `.nonzero()` data-dependent scan, the `dynamic_rope_update` demotion, or the vision-encoder check that nn.Module forwards get. Separately, `create_new_function(add_torch_compile=True)` hardcodes `fullgraph = True` with no checks at all; nothing passes `True` today, so it is latent rather than live. Both are behaviour changes across many models and want their own measurement.


danielhanchen:
## Simulation results: what changes, across 30 transformers releases

Ran the before/after classification over every module-level function in every modeling file, for 30 transformers releases from 4.57.0 to 5.17.0. Source-level analysis with `ast`, no imports, so the whole matrix runs in one pass.

For each function: `OLD` = excluded by pre-PR `DISABLED_KEYWORDS`. `NEW` = excluded by post-PR `DISABLED_KEYWORDS` or by `calls_mask_creation_function`.

```
versions scanned                    30  (4.57.0 -> 5.17.0)
total function classifications   40497
newly_compiled (crash risk)          0    <- nothing previously excluded is now compiled
newly_excluded (perf risk)           1    <- create_masks_for_vision_model, gemma3 only
versions where the PR is a no-op    21
first version affected           5.13.0
```

So the entire behavioural delta of this PR is one function in one model, and only on transformers 5.13.0 and later. On 4.57.x through 5.12.1 it changes nothing at all.

`newly_compiled = 0` is the important column: no function that was previously excluded becomes compiled, so the PR cannot introduce a new compile crash anywhere.

### Where the regression started

5.13.0 is the release that introduced `create_masks_for_vision_model`, and gemma3 shipped it with the extra kwarg from day one:

```
5.12.1  defines create_masks_for_vision_model = False
          create_causal_mask(**mask_kwargs),
5.13.0  defines create_masks_for_vision_model = True
          create_causal_mask(**mask_kwargs),
          create_causal_mask(**mask_kwargs, block_sequence_ids=block_sequence_ids)   <- escapes the literal
```

gemma4 and gemma4_unified use `create_causal_mask(**mask_kwargs)` unchanged, which is why only Gemma3 broke.

An earlier version of this analysis also flagged `create_causal_mask_mapping` on 5.0.0 to 5.8.1. That was an artefact of comparing against a single literal: `create_causal_mask_mapping` is itself a `DISABLED_KEYWORDS` entry, so those functions were already excluded before this PR. Comparing against the full list, as above, shows no change there.

## Upgrade path for an existing install

A user upgrading from a broken version has a stale `unsloth_compiled_cache/unsloth_compiled_module_gemma3.py` containing the old decorator. Simulated by injecting the pre-fix decorator into a generated cache file and re-running with default settings:

```
before run:  @torch_compile_with_fallback(fullgraph = True, ...) 
             def create_masks_for_vision_model(
after run:   def create_masks_for_vision_model(
             === VISION FP16 OK ===
```

The cache is regenerated on the next run and the stale decorator is gone, so no manual cache clearing is needed. The one exception is a user who has explicitly set `UNSLOTH_COMPILE_OVERWRITE=0`, which means "do not regenerate"; they keep the old file until the cache is deleted. That is inherent to that flag and applies to any compiler fix, not something this PR changes.

## Ordering bug this simulation found, now fixed

In the copy loop the mask rule set `bad = True`, which skips the whole decorator block including the branch that emits `@torch.compiler.disable(recursive = False)` for a name in `DISABLE_COMPILE_FUNCTIONS`. That decorator is stronger than emitting bare: it also stops Dynamo inlining the function into a compiled caller. No name on that list builds masks today so there was no live effect, but the precedence is now explicit and pinned by a test. The signature-fixup loop already had the correct order.

## Platform and hardware

The diff adds no imports and contains no reference to cuda, hip, xpu, mps, rocm, platform, `sys.platform`, `os.name`, device, MLX or Studio. It is string and regex work in the code generator, so there is no Linux / macOS / Windows / WSL divergence to test and no NVIDIA / AMD / Intel / CPU path that can behave differently. Nothing under `unsloth_zoo/` outside `compiler.py` references the symbols touched. No browser surface is involved.

## Tests

375 passed, 0 failed on the compiler, security and history-regression lanes, including `tests/test_zoo_history_regressions_deep.py::test_compiler_disabled_keywords_includes_5x_gemma3_mask`, which guards the `create_causal_mask_mapping` entry that PR #488 added and this PR deliberately keeps.

Cross-version test behaviour: on 5.17 the new tests run in full, on 4.57.6 they skip cleanly because no modeling file defines the builder.

## Upstream context

This is a known upstream limitation rather than something specific to unsloth: huggingface/transformers#42950 tracks `masking_utils` mask creation being untraceable under `fullgraph=True` because of the tensor-valued branch, and the workarounds noted there are to accept the graph break at mask construction or to keep the mask builder out of the compiled region. That is what this PR does, and it is also what the pre-existing `DISABLED_KEYWORDS` entry was already doing for gemma4 before the spelling drifted.

danielhanchen:
## Cross-platform runs and mask edge cases

### Windows and macOS, on real runners

Run on `danielhanchen/unsloth-zoo-staging`, branch `pr-1204-xplat-ci`, with CPU torch and transformers actually installed so the lanes are real gates rather than skips:

| runner | torch | transformers | result |
|---|---|---|---|
| macos-15 | 2.14.0 | 5.17.0 | 15 passed, 2 skipped |
| windows-latest | 2.14.0 | 5.17.0 | 15 passed, 2 skipped |
| ubuntu-latest | | | queued |

Each lane also asserts the classification directly on that OS, so it checks behaviour rather than just importing:

```
factories: ['create_bidirectional_mask', 'create_bidirectional_sliding_window_mask',
            'create_block_mask', 'create_causal_mask', 'create_chunked_causal_mask',
            'create_masks_for_generate', 'create_recurrent_attention_mask',
            'create_sliding_window_causal_mask']
  gemma3: ['create_causal_mask']
  gemma4: ['create_causal_mask']
  gemma4_unified: ['create_causal_mask']
OK: identical classification on this platform
```

The lanes fail closed: if the dependency install fails, or if pytest collects nothing, the job goes red rather than reporting a pass that proves nothing.

```bash
grep -qE "[1-9][0-9]* passed" out.txt || { echo "NO TESTS RAN -- lane proves nothing"; exit 1; }
```

These runs also caught a portability defect in the test added by this PR, fixed in d3abc3e1. `create_causal_mask` on the flex path compiles `create_block_mask` internally, so on an image with no triton the eager reference call in the probe died inside inductor before the probe could do anything:

```
NotImplementedError: 'triton.runtime.cache.triton_key' ... triton is stubbed out
```

The probe now asks for `backend="eager"`, since data-dependent branching is a Dynamo tracing failure raised before any backend runs, and skips if the platform cannot build a flex block mask at all. It still raises `Unsupported: Data-dependent branching` with the eager backend, so the probe is unchanged in what it proves. The 15 source-analysis tests were already passing on both platforms.

### Mask edge cases: identical output to the uncompiled reference

Every runtime check until now used a single image, which barely exercises the mask composition this PR touches. This drives the cases where the block ids actually vary, comparing compile-enabled-with-fix against `UNSLOTH_COMPILE_DISABLE=1` as the known-good reference, on gemma-3-4b-it:

| case | padded mask | tokens match | text match |
|---|---|---|---|
| text_only | False | yes | yes |
| one_image | False | yes | yes |
| two_images | False | yes | yes |
| batch_padded | **True** | yes | yes |

Token for token identical on all four. `batch_padded` is the important one: a padded batch makes `fast_all(attention_mask)` False, which is the exact branch inside `flex_attention_mask` that could not be captured under `fullgraph = True`. `two_images` exercises the multi-block path where images must not attend to future images.

So emitting the builder uncompiled changes what is compiled, not what is computed.

danielhanchen:
@codex review

danielhanchen:
@codex security review

danielhanchen:
@codex review

PR #1205, opened 2026-09-14
MLX Linux CPU lane: take the skip verdict from the first pytest run

danielhanchen:
## Why

The `MLX (Linux CPU: numerics + suites)` job runs five gated files and then, for each of them, a `Fail if ... skipped` step that **runs the same pytest file a second time** and greps its output:

```
python -m pytest tests/x.py -q -rs | tee out.txt
if grep -qiE '[0-9]+ skipped' out.txt; then ... exit 1; fi
```

The first run already knew the answer. Measured on two runs from 09-13, those five second executions cost **51s and 62s**, about **18% of a ~290s job**, to re-learn it.

## The change

Each gated run writes `--junitxml` to its own path under `$RUNNER_TEMP`, and the gate step reads that report instead of re-running pytest. Same five files, same one process each (`tests/mlx_simulation/` replaces `sys.modules["mlx.core"]` process-wide, so they cannot share one), same `-v -rs` in the log, and the pytest step still decides on its own exit status.

## The policies, which are not interchangeable

I read all five old steps before touching them. There are exactly two policies, and merging them would break three gates:

| Gated file | Old assertion | Policy |
| --- | --- | --- |
| `test_mlx_quantized_optimizer_state.py` | `grep -qiE '[0-9]+ skipped'` fails | any skip forbidden |
| `test_mlx_neftune_quant_map.py` | `! grep -qiE '[1-9][0-9]* passed'` fails | at least one pass |
| `test_mlx_rng_rewind.py` | `grep -qiE '[0-9]+ skipped'` fails | any skip forbidden |
| `test_mlx_quantized_merge_save.py` | `! grep -qiE '[1-9][0-9]* passed'` fails | at least one pass |
| `test_mlx_gated_delta_batch_grad.py` | `! grep -qiE '[1-9][0-9]* passed'` fails | at least one pass |

`no-skips` and `one-passed` are separate `--require` values in `scripts/pytest_junit_gate.py`, and the neftune step carries a comment saying why it is the looser one: its `metal_only` half is expected to skip on this lane.

## A report that cannot be trusted is a failure, not a pass

That is the whole risk of reading a report instead of re-running, so the helper fails loudly on: a missing file, a non-file, an empty or whitespace-only file, unparseable XML, the wrong root element, no `<testsuite>`, a missing or non-numeric `tests` attribute, a `tests` attribute that disagrees with the `<testcase>` count (a truncated or foreign report), zero testcases, and any failure or error.

The case these gates exist for is the whole module skipping at import. That is written as one `<testcase>` carrying `<skipped message="collection skipped">`, so both policies see it. xfail is bucketed separately and satisfies neither, which is what the old greps did, since the summary line calls it `xfailed`.

## Checks

- `tests/test_pytest_junit_gate.py`: **31 passed**. It generates real reports by running pytest on generated files (clean, one skip, module skipped at import, collection error, genuine failure, no tests, xfail), adds hand-written malformed reports, and asserts the workflow still has five gate steps, five distinct report paths, and no `-m pytest` left in a gate step.
- **Verdict-for-verdict against the old greps**, on the same real reports: `all_pass` 0/0, `half_skip` 1 for no-skips / 0 for one-passed, `allskip_marks` 1/1, `xfail` 0/0, `many_skip_one_pass` 1/0. Identical in every case.
- **End to end on the real YAML**: each step pair parsed out of the workflow and run as shell with a stand-in file. Clean file, pytest 0 and gate 0. All-skipped, pytest 0 and gate 1. Half-skipped, gate 1 under no-skips and 0 under one-passed. Genuine failure, pytest 1 and the gate never runs, as today. Module-level skip, pytest exit 5 and the job red, as today.
- `ruff check --select E9,F63,F7,F82` clean; `scripts/lint_exec_literals.py`, `scripts/lint_dynamic_exec.py`, `scripts/lint_workflow_triggers.py` clean; `tests/test_ci_lane_scripts_parse.py` and `tests/test_conftest_module_leak_gate.py`, 46 passed.

Not verified here: the MLX suites themselves need the real runtime, which is the thing they gate on, so the reports above come from stand-in files using the same skip mechanisms. `actionlint` is not installed on this box, so the workflow was checked by a PyYAML parse plus the repo's own lane-shell test.


danielhanchen:
Staging CI, on an owned staging repo (fresh upstream main + this PR head): `shimmyshimmer/unsloth-zoo-staging-2#132`.

**All four jobs green**: `Tests CI` (which carries this change), `Repo tests (CPU)`, `Core drift (3 upstream pins)`, `(Python 3.10-3.13 collect)`, plus `Lint CI` and `Wheel CI`. `MLX CI on Mac M1` skipped, as it is label-gated.

All five gate steps ran and passed, and the log shows them reading a report rather than re-running pytest:

```
Run .lanes/venv-numerics/bin/python -m pytest tests/test_mlx_quantized_optimizer_state.py -v -rs \
  --junitxml="$RUNNER_TEMP/mlx-reports/quantized_optimizer_state.xml"
...
Run python3 scripts/pytest_junit_gate.py --require no-skips ...
gate 'no-skips' satisfied by /home/runner/work/_temp/mlx-reports/quantized_optimizer_state.xml:
  39 test(s): 39 passed, 0 skipped, 0 failed, 0 errored, 0 xfailed
```

That is the thing worth checking on a real lane rather than on a stand-in: the report is written by the gated run itself, the gate parses it, and it reports the counts it decided on instead of a bare exit code.

`MLX (Linux CPU: numerics + suites)` came in at **4m**, against a 5.6m median for that job on `main`. One run on a staging runner is not a benchmark, and the runner is not the same machine, so read that as consistent with removing the 51-62s of re-execution rather than as a measurement of it.

PR #1207, opened 2026-09-14
fix(mlx): detect QQLinear when deciding whether merged_4bit must quantize

danielhanchen:
## Why

`save_merged_model` decides whether `merged_4bit` still has work to do by asking whether anything in the fused model is already quantized. `_model_has_quantized_module` answered that with a small class list plus a `Quantized*` name prefix:

```python
quantized_types = [nn.QuantizedLinear, nn.QuantizedEmbedding, QuantizedSwitchLinear]
...
return type(module).__name__.startswith("Quantized")
```

That covers `QuantizedLinear`, `QuantizedEmbedding`, `QuantizedSwitchLinear`, MLA's `QuantizedMultiLinear` and the two distributed sharded variants. Enumerating every `nn.Module` subclass reachable from `mlx.nn` and `mlx_lm` that carries a packed width turns up exactly one it misses:

```
mlx.nn.layers.distributed.QuantizedAllToShardedLinear   name=True
mlx.nn.layers.distributed.QuantizedShardedToAllLinear   name=True
mlx.nn.layers.quantized.QQLinear                        name=False   <-- missed
mlx.nn.layers.quantized.QuantizedEmbedding              name=True
mlx.nn.layers.quantized.QuantizedLinear                 name=True
mlx_lm.models.mla.QuantizedMultiLinear                  name=True
mlx_lm.models.switch_layers.QuantizedSwitchLinear       name=True
```

`QQLinear` is the activation-quantized linear. It subclasses `Module` directly, not `QuantizedLinear`, and it is not `Quantized*`-named, so both halves of the check miss it. An already-packed `QQLinear` model reads as unquantized and gets packed a second time on a `merged_4bit` save.

This is not reachable from unsloth_zoo today: `qat_scheme` is still refused for MLX LoRA (`loader.py`), and nothing in the tree constructs a `QQLinear`. It is the branch QAT lands on, so it is worth closing before something routes onto it rather than after.

## The change

Detection also accepts a module that carries `bits` and `group_size` **and** holds a `uint32` weight.

The weight check is the part that matters. Keying on the two attributes alone is the opposite bug: a dense module that happens to carry them reads as quantized, the merge is skipped, and `merged_4bit` silently writes full precision again, which is the exact failure this path exists to prevent. Asking for a packed weight separates the two.

`QQLinear` stores weights packed in eval mode and in higher precision in train mode. `save_merged_model` calls `model.eval()` before it looks, so the packed form is what detection sees.

## Verification

Real `mlx` 0.32.2 and `mlx-lm` 0.31.3 on Linux CPU. MLX quantization does not require Metal, so this is genuine packing, not a shim:

```
tests/test_mlx_quantized_merge_save.py    19 passed -> 21 passed
```

Both new tests are in the existing non-Metal file, which already runs in the `MLX (Linux CPU: numerics + suites)` lane and in `mlx-ci.yml`, so they gate rather than skip.

Mutation-checked. With the production change reverted and the tests kept:

```
1 failed, 20 passed
FAILED test_qqlinear_is_detected
```

`test_unpacked_bits_attributes_do_not_count_as_quantized` passes both ways by design. It pins the false-positive direction so the detector cannot later be loosened to bare attribute matching.

A 14-scenario before/after harness over the whole `merged_4bit` contract (dense base, 16-bit control, 8-bit base width preservation, caller-model mutation, float32 casting, ineligible dimensions, default-precision save, adapter fusion, reload round trip, `save_method="lora"`) produces byte-identical artifacts on this branch and on `main`, so nothing about existing behaviour moves:

| scenario | main | this branch |
|---|---|---|
| dense base, `merged_4bit` | 14 packed, 251136 B, bf16 + uint32 | identical |
| dense base, `merged_16bit` | 0 packed, 1444352 B | identical |
| 8-bit base keeps its own width | bits 8 preserved | identical |
| live model after save | unquantized, restored | identical |

Pre-existing failures in `test_mlx_save_export_edge_cases.py` (15) and `test_mlx_save_export_regressions.py` (78) reproduce identically on a clean `main` checkout with the same environment. They are llama.cpp / GGUF script-path tests and are unrelated.

## Credit

Found while reviewing #1186, which argued correctly that the quantized types share no base class. Its own fix for the underlying `merged_4bit` bug is superseded by #971, already merged, but that observation was right and this is the part of it worth keeping.


danielhanchen:
Folded into #1186, which is the original PR for this and is now rebased onto main carrying exactly this change. Closing here to keep it in one place.

PR #1208, opened 2026-09-14
gpt-oss MXFP4: run natively on Blackwell, and accept both triton_kernels layout APIs

danielhanchen:
## What breaks today

`nb/GPT_OSS_MXFP4_(20B)-Inference.ipynb` passes on Kaggle T4x2 and fails on a B200. `unsloth/gpt-oss-20b` loads natively at 13.46 GiB, weights swizzled, and then the first `generate` dies:

```
triton.compiler.errors.CompileTimeAssertionFailure: at 38:8:
  tl.static_assert(SWIZZLE_MX_VALUE == "HOPPER_VALUE" or SWIZZLE_MX_VALUE is None,
                   "Only Hopper swizzling is supported for values")
```

At the `triton_kernels` commit the notebooks pin (`0add6826`), `make_default_matmul_mxfp4_w_layout` returns `BlackwellMXValueLayout` for `cuda_capability_geq(10)` while that build's `_matmul_ogs` accepts only Hopper swizzling or none. The same kernel does accept `SWIZZLE_MX_SCALE == "BLACKWELL_SCALE"`, so it is specifically the value swizzle that is unimplemented, and upstream leaves `# return StridedLayout, dict()` commented out directly above the selection. This is not only the notebook pin: vLLM vendors a `triton_kernels` under the real module name carrying the same restriction.

A second defect sits in the same three lines. `triton_kernels` changed the selector's return contract from `(layout_class, ctor_kwargs)` to a layout **instance**, and zoo unpacks unconditionally:

```python
value_layout, value_layout_opts = layout.make_default_matmul_mxfp4_w_layout(mx_axis=1)
```

which raises `TypeError: cannot unpack non-iterable ... object` on any newer build.

## The fix

Both live in `swizzle_mxfp4`, the function zoo already owns, so one hook covers transformers 4.x (via `load_and_swizzle_mxfp4`) and 5.x (via `swizzle_mxfp4_convertops`, which calls the same module-level name).

- `_normalize_mxfp4_value_layout` accepts either contract, and the scales are handed `StridedLayout` in whichever form the installed `convert_layout` expects.
- `_blackwell_value_swizzle_unsupported` reads the installed kernel's own source and matches that exact `tl.static_assert` as an AST node, so the same words in a comment or an error string do not count. Absence of the assert is not treated as proof of Blackwell support, only as a reason to leave the build alone.
- The guard fires only when the weight is on CUDA, the weight's own device is sm_100+, the build chose `BlackwellMXValueLayout`, and the probe matched. `UNSLOTH_MXFP4_VALUE_LAYOUT=strided|default` forces either side.

Nothing touches `Mxfp4Config.dequantize`, `should_dequantize_mxfp4`, `convert_moe_packed_tensors` or the untransposed convention `saving_utils._mxfp4_base_returns_transposed` depends on.

## Measured

B200 (sm_100), torch 2.12.1+cu130, transformers 4.56.2, triton_kernels `0add6826`, the notebook's own pins.

Kernel level, model-free, one small MXFP4 weight against a dequantized reference:

| value layout | scales | result |
|---|---|---|
| build default (Blackwell) | strided | raises `Only Hopper swizzling is supported for values` |
| **strided (this PR)** | strided | **mean abs err 3.03, rel 0.0014, corr 0.999999** |
| Blackwell | Blackwell | `NotImplementedError: Must use persistent kernel and be TMA-compliant` |

Model level, `unsloth/gpt-oss-20b`, 163 tokens of fixed prose, against stock transformers `Mxfp4Config(dequantize=True)` bf16 on the same host:

| | perplexity | next-token top-1 | memory |
|---|---|---|---|
| bf16 reference | 4.487 | 63.2% | 38.96 GiB |
| **native MXFP4, this PR** | **4.507** | **64.4%** | **12.86 GiB** |

top-1 agreement with the reference 97.5%, mean KL 0.0012.

Non-Blackwell is untouched: Kaggle T4x2 (sm_75) runs the notebook 5/5 cells with correct answers, and the guard does not fire there because that build already selects `StridedLayout` below sm_90.

## Tests

`tests/test_gpt_oss_mxfp4_blackwell_value_layout.py`, 29 CPU-only tests with `triton_kernels` faked and capability faked, covering: the probe against the real pinned kernel source and against a Blackwell-capable one; comment-only and string-only lookalikes; unreadable and absent sources; both selector contracts; the guard firing on sm_100 and staying out of the way on sm_75/80/90, on capable builds, on non-CUDA weights, and when capability lookup raises; and both directions of the env override.

Green: the new file plus `test_mxfp4_load_path_layout`, `test_mxfp4_convert_no_double_transpose`, `test_mxfp4_transpose_convention`, `test_mxfp4_import_resilient`, `test_moe_quant_cleanup`, `test_temporary_patches_imports`, `test_gptoss_grouped_gate`, `test_gpt_oss_attention_mask` (90 passed), and `test_temporary_patches_exhaustive` (103 passed, 16 skipped).

## Separate bug found while measuring, not addressed here

On the same B200, `torch.compile` corrupts gpt-oss output regardless of quantization, which is why the numbers above are quoted with `UNSLOTH_COMPILE_DISABLE=1`:

| path | compile off | compile on |
|---|---|---|
| native MXFP4 | ppl 4.507 | ppl 642.7 |
| unsloth-bnb-4bit | ppl 4.936 | ppl 73.3 |

Per-module hooks put the first divergence at `model.layers.0.self_attn`, rel 0.63, with a fresh compiled cache, so it is not staleness and not MXFP4-specific. It is pre-existing and was simply unobservable on Blackwell while the crash above masked it. Filing separately.


danielhanchen:
Cross-platform check, run on a staging repo rather than the org queue, so the platforms this change does not target are covered too. The new test file is CUDA-free by design: it fakes `triton_kernels` and the device capability, so it is exactly what should run on a machine with no NVIDIA GPU at all.

| runner | result |
|---|---|
| windows-latest | 29 passed in 0.25s |
| macos-15 (Apple Silicon) | 29 passed, 2 warnings in 1.09s |
| ubuntu-latest | still queued on the staging repo; the same 29 pass locally on Linux, alongside the 90-test gpt-oss/mxfp4 set and the 103-test drift suite |

The macOS run matters specifically for `test_guard_is_a_no_op_without_cuda`: on a host with no CUDA the guard returns before touching `torch.cuda`, so MLX and Apple Silicon paths cannot be perturbed by it. Windows is the other side of the same claim, and `triton_kernels` is skipped outright by the Windows installer, so the probe there always reports "nothing recognised" and the build is left alone.

danielhanchen:
Colab regression matrix, running the unchanged `nb/GPT_OSS_MXFP4_(20B)-Inference.ipynb` with `unsloth_zoo` force-installed from this branch.

| Colab tier | GPU | capability | guard fired | notebook |
|---|---|---|---|---|
| T4 (high-RAM request) | Tesla T4, 14.6 GiB, host 176.9 GiB | sm_75 | **False** | all model cells ran, coherent output |
| A100 | A100-SXM4-40GB | sm_80 | **False** | all model cells ran, coherent output |
| G4 | **RTX PRO 6000 Blackwell Server Edition**, 95.0 GiB | **sm_120** | **True** | 7/7 cells clean, coherent output |

Two things worth calling out.

**G4 is consumer/workstation Blackwell, sm_120, not Ada.** `cuda_capability_geq(10)` is true there, so that host selects `BlackwellMXValueLayout` and hits the same Hopper-only assert as the B200. The guard fired and the notebook passed, so this PR fixes RTX PRO 6000 as well as B200. That is hardware I do not own, reached by renting it.

**T4 and A100 confirm the no-op.** Below sm_100 the build already selects `StridedLayout`, the probe never gets consulted, and `GUARD FIRED` stays False.

On the high-RAM point: `notebook_cloud_run.py` passes `--gpu T4` to the Colab CLI for `T4-HIGHRAM`, so the GPU shape is a standard 14.6 GiB T4. What the run did get is the high-RAM host, 176.9 GiB, and the log shows unsloth using it (`Offloading embeddings to RAM to save 1.08 GB`). So this covers the high-RAM host configuration, not a larger T4.

The first T4 and A100 runs reported 0/1 only because a probe cell I added imported `unsloth_zoo` before `unsloth`, which zoo refuses; the notebook's own cells ran fine in both. Fixed for the G4 run, which is why that one reports 7/7.

danielhanchen:
@codex review

danielhanchen:
@codex security review

danielhanchen:
Confirmed main still unpacks the selector blind at unsloth_zoo/temporary_patches/gpt_oss.py:173 and that the guard is inert below sm_100, and the T4/A100/G4 matrix covers the no-op claim. Will get this reviewed.

danielhanchen:
@codex review

PR #1213, opened 2026-09-14
Follow the transformers 5.17 vision rotary signature change when rebuilding from vLLM

danielhanchen:
## What happened

transformers **5.17.0** changed the vision rotary constructor for the whole Qwen VL family:

| transformers | `Qwen2_5_VLVisionRotaryEmbedding` | `Qwen3VLVisionRotaryEmbedding` | `Qwen3_5VisionRotaryEmbedding` |
|---|---|---|---|
| 4.57.6 | `(dim, theta=10000.0)` | `(dim, theta=10000.0)` | model absent |
| 5.0.0 | `(dim, theta=10000.0)` | `(dim, theta=10000.0)` | model absent |
| 5.2.0 through 5.16.0 | `(dim, theta=10000.0)` | `(dim, theta=10000.0)` | `(dim, theta=10000.0)` |
| **5.17.0** | **`(config, device=None)`** | **`(config, device=None)`** | **`(config, device=None)`** |

`finalize_huggingface_model` rebuilds the HF model from vLLM's weights and reinitialises `rotary_pos_emb` with `rotary_class(head_dim//2)`. On 5.17 that int becomes `self.config`, and the ctor immediately does `self.config.rope_parameters["rope_type"]`.

So this is not only a Qwen3.5 gap. On transformers 5.17, `fast_inference = True` is broken on main for **Qwen2.5-VL and Qwen3-VL as well**:

```
$ python probe.py   # unsloth/Qwen2.5-VL-3B-Instruct, fast_inference=True, transformers 5.17.0, main
AttributeError: 'int' object has no attribute 'rope_parameters'
```

## The change

Dispatch on `inspect.signature` rather than trying one form and catching. Guessing is not safe here: a config ctor that stores its argument before touching it accepts an int silently, and we would ship a corrupt rotary instead of an error.

`device` is `@deprecate_kwarg(version="5.18")`, so it is not passed at all. The module is constructed and then moved with `.to(target_device)`, which keeps working once the kwarg is removed. An unrecognised signature warns and leaves the load standing, matching the `rotary_emb` branch above.

## Before and after

Both trees, transformers 5.17.0, one B200, `load_in_4bit = False`:

| zoo | unsloth | model | before/after |
|---|---|---|---|
| main | main | Qwen2.5-VL-3B-Instruct | `AttributeError: 'int' object has no attribute 'rope_parameters'` |
| PR | PR | Qwen2.5-VL-3B-Instruct | loads, `inv_freq` on cuda:0 shape (20,), generates ` Paris. The capital of the United States is Washington, D` |
| main | main | Qwen3.5-2B | rejected at the vLLM VLM gate |
| main | PR | Qwen3.5-2B | `AttributeError: 'int' object has no attribute 'rope_parameters'` |
| PR | PR | Qwen3.5-2B | loads, `inv_freq` on cuda:0 shape (16,), generates |

The rotary is read back off the loaded model rather than inferred from the load succeeding, because a skipped reinit is silent.

## Version sweep

The shipped dispatch block run against the real rotary classes of each version, with the result compared to the natively constructed one:

```
tf           qwen2_5_vl     qwen3_vl      qwen3_5
4.57.6              dim          dim model-absent
5.0.0               dim          dim model-absent
5.2.0               dim          dim          dim
5.5.0               dim          dim          dim
5.8.0               dim          dim          dim
5.10.0              dim          dim          dim
5.12.0              dim          dim          dim
5.14.0              dim          dim          dim
5.16.0              dim          dim          dim
5.17.0           config       config       config
```

10 of 10 PASS, no failures. On 4.57.6 through 5.16.0 the dim form is selected, so those versions get byte-identical behaviour to main.

## Simulation

The dispatch is extracted from the installed source and executed against synthetic rotary classes, so the test fails if code and intent drift:

- transformers 4 positional dim: dim ctor, `dim == 40`, lands on target device
- transformers 5.17 `(config, device=None)`: config ctor, lands on target device
- **simulated 5.18 with `device` removed**: still rebuilt. The earlier revision of this PR passed `device =` and silently skipped the reinit here
- lazy config ctor that accepts an int without complaint: signature dispatch avoids it, where try-and-catch would not
- unknown signature, and a ctor that raises: exactly one warning, load continues

## Real loads on older transformers

The sweep above is signature-level. These are full `fast_inference` loads on one B200, PR trees, `load_in_4bit = False`, reading `inv_freq` back off the loaded model:

| transformers | rotary class picked | Qwen2.5-VL-3B | Qwen3.5-2B |
|---|---|---|---|
| 5.10.0 | `Qwen2_5_VisionRotaryEmbedding` (dim) | OK | fails, see below |
| 5.12.0 | dim | not run | OK |
| 5.14.0 | dim | not run | OK |
| 5.16.0 | dim | OK | OK |
| 5.17.0 | `Qwen2_5_VLVisionRotaryEmbedding` (config) | OK | OK |

The rotary is numerically identical across the split. Qwen2.5-VL `inv_freq` sums to `2.7094428539276123` on 5.10.0 and 5.16.0 via the dim form and on 5.17.0 via the config form; Qwen3.5 sums to `2.2846570014953613` on 5.12.0, 5.14.0, 5.16.0 and 5.17.0. So the new branch reproduces the old one exactly where the old one worked.

Qwen3.5 on 5.10.0 fails with `TypeError: empty(): argument 'dtype' must be torch.dtype, not str`, which is a config dtype problem during model construction, before the rotary reinit, and Qwen2.5-VL loads fine on that same version. It is unrelated to this change and not something this change regresses: Qwen3.5 could not reach vLLM at all before. The practical floor for Qwen3.5 + `fast_inference` is somewhere in 5.11 or 5.12.

## Scope of the device kwarg

Only the vision rotaries carry `@deprecate_kwarg("device", version="5.18")`. The text rotaries do not, so the pre-existing `rotary_emb` branch above is not affected and is deliberately left alone.

## Risk

`finalize_huggingface_model` has a single caller, `vllm_utils.py:1650`, on the vLLM to HF rebuild path. Nothing outside `fast_inference` reaches it, and no Unsloth Studio code references it.

Tests: `test_vllm_to_hf_conversion.py` + `test_vlm_layer_counts.py` give 17 failed / 68 passed, against 17 failed / 65 passed on main. The failure sets are identical and pre-existing in this environment (the installed vllm has no `vllm.lora.models`, which falls into the Apple Silicon triton stub); the three extra passes are the tests added here.

Cross-platform CI on ubuntu-latest, macos-15 (Apple Silicon), windows-latest, ubuntu-24.04-arm and macos-15-intel.

Pairs with unslothai/unsloth#10891. This one should land first: with the allowlist opened but an older zoo installed, Qwen3.5 reaches the rebuild and fails there, as the mixed row above shows.


danielhanchen:
Confirmed this hits unsloth_zoo/empty_model.py:898, where the positional dim is still passed to the transformers 5.17 config constructor for the whole Qwen VL family. Will get this reviewed alongside unslothai/unsloth#10891.

PR #1218, opened 2026-09-14
Pass the RNG flag to the offloaded checkpoint function

vineethsaivs:
The offloaded checkpoint wrapper raises an ambiguous tensor boolean error before running a block.

**Cause:** UnslothCheckpointFunction expects preserve_rng_state before the tensor inputs, but the wrapper omits it and binds the first activation to that flag.

**Fix:** Extract preserve_rng_state with its default True and pass it before the activation arguments.

**Test:** `UNSLOTH_ALLOW_CPU=1 python -m pytest tests/test_gc_rewriter_bound_keywords.py -q`. Three regressions fail before. The checkpoint suite passes 35 tests with one GPU-only skip after. Outputs, input gradients and final RNG state match PyTorch checkpointing for default, True and False RNG preservation. CPU recomputation path with the existing CPU fixture, tested on Transformers 5.5.0 and 5.17.0. GPU transfer and memory use were not measured. Changed-file lint passes.


danielhanchen:
Confirmed this against gradient_checkpointing.py:1297, where the activation is bound to the preserve_rng_state slot that UnslothCheckpointFunction.forward expects second, so the offloaded path matches the shim at line 1104 after your change. Will get this reviewed.

danielhanchen:
Reviewed and validated this on a B200, plus a version matrix in isolated venvs. The diagnosis is correct, the fix is correct, and I pushed a follow-up to your branch (a54f8d69) because a real end-to-end run dies one line above your change.

## Confirming the root cause

`UnslothCheckpointFunction.forward` has taken `preserve_rng_state` as its second positional since 2024 (gradient_checkpointing.py:712), and its `backward` already returns `(None, None) + grads` (:1035) on that assumption. The offloaded wrapper never filled that slot, so the first activation landed in it. Reproduced verbatim on current main:

```
if preserve_rng_state:
E   RuntimeError: Boolean value of Tensor with more than one value is ambiguous
unsloth_zoo/gradient_checkpointing.py:721: RuntimeError
FAILED test_offloaded_wrapper_matches_torch_checkpoint[None]
FAILED test_offloaded_wrapper_matches_torch_checkpoint[True]
FAILED test_offloaded_wrapper_matches_torch_checkpoint[False]
```

Green on your head, and the fix is complete: `:1104` and `:1298` are the only two `UnslothCheckpointFunction.apply` call sites in the repo, and after this they agree.

Worth recording why this survived so long: `patch_unsloth_gradient_checkpointing()` is the only thing that installs this shim, and nothing in unsloth or unsloth-zoo calls it. Every `use_gradient_checkpointing` value routes through `unsloth/models/_utils.py:353-366` to `patch_unsloth_smart_gradient_checkpointing` instead. It is still public API re-exported from `unsloth.models._utils`, so this is worth fixing, but no shipped configuration was hitting it.

## What I pushed on top

**1. The next crash, one line up.** The lazy init guard was `len(CPU_BUFFERS) == 0`, but `unpatch_unsloth_smart_gradient_checkpointing` sets `CPU_BUFFERS` to `None` rather than emptying it (:1169), and `prepare_model_for_training` calls that unpatch for every `use_gradient_checkpointing` other than `"unsloth"` (training_utils.py:447-449). Since installing this shim at all means doing it after model setup, `None` is the state it normally finds. A real LoRA run with `patch_unsloth_gradient_checkpointing()` installed died with:

```
File "unsloth_zoo/gradient_checkpointing.py", line 1294, in unsloth_offloaded_gradient_checkpoint
    if len(CPU_BUFFERS) == 0:
TypeError: object of type 'NoneType' has no len()
```

before it could ever reach your fix. Changed to `not CPU_BUFFERS`, which covers both the initial `[]` and the `None`.

**2. Your test cannot catch offload regressions.** The `cpu_offload_globals` fixture pins `MINIMUM_SIZE = 1 << 40`, which keeps the pinned-buffer D2H/H2D branch off, so what it verifies is the plain recompute path. Added `tests/test_offloaded_checkpoint_rng.py`, GPU-only, which sizes activations above `MINIMUM_SIZE` so the offload branch is the one under test and asserts `CPU_INDEX` advanced, i.e. that offload actually happened rather than trusting that it did.

**3. The file never runs in CI.** `tests/test_gc_rewriter_bound_keywords.py` appears in no pytest invocation in `consolidated-tests-ci.yml`; the only job touching it is `python-version-collect`, which runs `--collect-only`. Your three regressions would have gone green by absence. Added it to the `repo-tests-cpu` gate.

**4. `tests/test_offloaded_checkpoint_simulation.py`**, 32 tests, CPU-only so it runs on the Windows and macOS runners too. 28 of them fail without the fix. It pins the `.apply` calling convention directly (flag in slot 2, activations intact and in order), backward arity, RNG parity against pristine torch for absent/True/False, checkpoint-machinery keywords dropped rather than bound onto the block, tensor and constant keyword binding, the caller's kwargs dict left unmutated, the `functools.partial` shape transformers actually uses, buffer lifecycle, tuple and nested outputs, and public signature stability.

## Evidence

Version matrix, each on CUDA and CPU-only, isolated venvs:

| Stack | CUDA | CPU-only |
|---|---|---|
| py3.12 / torch 2.6.0 / tf 4.57.6 / TRL 0.22.2 / PEFT 0.18.0 | 111 passed, 3 skipped | 111 passed, 3 skipped |
| py3.10 / torch 2.10.0 / tf 5.0.0 / TRL 1.0.0 / PEFT 0.19.0 | 112 passed, 2 skipped | 112 passed, 2 skipped |
| py3.12 / torch 2.14.0 / tf 5.17.0 / TRL 1.13.0 / PEFT 0.20.0 | 112 passed, 2 skipped | 112 passed, 2 skipped |
| py3.13 / torch 2.13.0 / tf 5.17.0 / TRL 0.24.0 | 112 passed, 2 skipped | 112 passed, 2 skipped |

Llama-3.2-1B LoRA, 61 steps, identical batches across arms:

| Arm | Peak memory | Losses | Grad-norms |
|---|---|---|---|
| Naive HF checkpointing | 1.922 GB | 1.55, 1.846, 1.174 ... 1.457, 1.315 | 1.858, 3.403, 1.657 ... 3.213, 2.112 |
| Offloaded, this PR + follow-up | 1.792 GB | 1.55, 1.846, 1.174 ... 1.457, 1.314 | 1.858, 3.403, 1.657 ... 3.209, 2.115 |
| Unsloth smart GC (live path) | 1.792 GB | 1.55, 1.846, 1.174 ... 1.458, 1.312 | 1.858, 3.403, 1.657 ... 3.208, 2.106 |
| Offloaded on unfixed main | crashes | | |

The offload branch genuinely engaged ("Will smartly offload gradients to save VRAM!") and memory matched the smart path exactly. An older `unsloth==2025.12.1` paired with this patched unsloth_zoo also trains correctly on both the smart and offloaded paths, which is the pairing a user gets when only unsloth_zoo is upgraded.

## Two notes, neither blocking

`unsloth_gradient_checkpoint` is now the only shim that ignores `preserve_rng_state` - it routes through `Unsloth_Gradient_Checkpointer`, which has no such slot, so its recompute draws a fresh dropout mask and the input gradient does not match torch. The comment above `_TORCH_CHECKPOINT_KEYWORDS_LITERAL` calls that deliberate, and its installer `patch_gradient_checkpointing()` is equally uncalled, so I left it alone and added a test pinning the current behaviour so it fails loudly if someone changes it.

On a CPU-only host this function still cannot work - `initialize_unsloth_gradient_checkpointing` allocates pinned memory at :643 - so the follow-up converts one exception into another there rather than fixing it. That is unchanged in substance from before.

Nice catch on this one.

danielhanchen:
@codex review

danielhanchen:
@codex security review

danielhanchen:
@codex review

danielhanchen:
Converged. Codex approved head de2d74039 with no items in either round, and staging CI on that head is green: Lint, Wheel, and Tests (1735 passed, 27 skipped).

Two things landed on your branch after my earlier comment, both mine rather than yours:

- The two new test files were seeded from a neighbouring file that still carries the legacy LGPL header. unsloth_zoo wants the canonical AGPL-3.0 block on new files, which is also what the rest of tests/ uses, so `prepush_gate.py` blocked the push until it was corrected.
- A comment pass over the files this PR adds, in two sweeps. Comment lines added by the PR, excluding the two licence headers, went from 108 to 29; the source change itself is two lines of comment instead of six. AST-gated, so no code moved.

Cross-platform, run off the org queue on staging:

| Platform | Result |
|---|---|
| Linux x64 | 68 passed, 7 skipped |
| Linux ARM64 | 68 passed, 7 skipped |
| macOS 15 Apple Silicon | 68 passed, 7 skipped |
| Windows x64 | 32 passed, 6 skipped |

Two notes on the gaps there, neither caused by this PR. The Windows number is scoped to the two files this PR adds: the pre-existing `test_gc_rewriter_bound_keywords.py` imports `unsloth_zoo.compiler`, which imports triton, and triton ships no Windows wheels, so a collection error there aborts the whole invocation. Worth knowing that file can never run in a Windows lane. And macOS Intel cannot install unsloth_zoo at all, since torch stopped publishing Intel-mac wheels after 2.2.2 while the package requires torch>=2.4.0 there.

Also confirmed the CI line is doing its job: `test_gc_rewriter_bound_keywords.py` now runs inside the behavioral gates step. Before it, that file was reached only by `--collect-only`, so the three regressions you added would never have executed.

PR #1222, opened 2026-09-15
Let the security audit report on every PR, without scanning on every PR

danielhanchen:
Follow-up to making checks required. The security audit cannot currently be one, and this is what stands in the way.

## Why it cannot be required today

A workflow whose `pull_request` trigger is path-filtered reports **no check at all** when the filter does not match. Not "skipped", not "success" -- nothing. So a required check that is path-filtered blocks every PR that does not match the filter, permanently.

The audit is filtered to `pyproject.toml`, the two scanner scripts, `tests/security/**` and `security-audit.yml`. Verified against the last two PRs merged here:

```
PR 1219: NO security checks reported
PR 1220: NO security checks reported
```

Both would have been unmergeable had the audit been required. So it stays advisory, and a red audit on a dependency bump does not stop that bump landing.

## The change

The filter becomes a `changes` job, which is the same decision made somewhere that can still report:

- `on.pull_request` loses its `paths` filter, so the workflow starts on every PR
- `changes` lists the PR files and decides whether anything the scans read has changed. `push`, `schedule` and `workflow_dispatch` are forced to `true`, exactly as they behave today
- the four scan jobs gain `if: needs.changes.outputs.relevant == true`, so they run on precisely the PRs they ran on before, as a **skip** rather than a no-run
- a `security audit` gate job with `if: always()` reduces the four results to one context

The gate treats `skipped` as a pass, because a skip is the detector saying this PR cannot affect that scan. `failure` and `cancelled` are not passes.

`security audit` is then the context to require. This PR does not require it; that is a settings change once this lands.

## Cost

Two short jobs on a PR that changes nothing the audit reads, where previously the workflow did not start. The scans themselves run exactly as often as they do now, which is the expensive part and the part the original filter was protecting.

## Checks

Detection, 14 cases including near-misses:

| input | verdict |
|---|---|
| `pyproject.toml` | run |
| `scripts/scan_packages.py` | run |
| `tests/security/test_scan_packages.py` | run |
| `.github/workflows/security-audit.yml` | run |
| `unsloth_zoo/vllm_utils.py` + `pyproject.toml` | run |
| `unsloth_zoo/vllm_utils.py` | skip |
| `.github/workflows/lint-ci.yml` | skip |
| `vendor/pyproject.toml` | skip |
| `tests/security_helpers/foo.py` | skip |
| `docs/scripts/scan_packages.py` | skip |

Anchored so its only inexact direction is running a scan that was not needed, never skipping one that was.

Gate verdict, 8 scenarios:

| scenario | result |
|---|---|
| irrelevant PR, scans skipped | pass |
| dependency PR, all scans pass | pass |
| advisory audit fails | fail |
| scan-packages fails | fail |
| security tests fail | fail |
| trigger lint fails | fail |
| detector itself fails | fail |
| a scan is cancelled | fail |

Fail-closed on the detector, so a broken detector cannot wave everything through.

This PR edits `security-audit.yml`, which is in the detect list, so it exercises the relevant=true path on itself. The relevant=false path is the one to watch on the next unrelated PR.


danielhanchen:
Closing. This only existed to make the security audit a requireable check, and we are not requiring checks, so it buys nothing on its own: on the PRs the filter excludes the scans still skip and the gate would just report green for "nothing to scan", and on the PRs that do touch dependencies the workflow already runs and already shows red.

The gap it was paired with is fixed separately in #1221, which does not depend on any of this.

Leaving the detect-and-gate approach written down here in case required checks come back on the table: the pattern list moves into a changes job verbatim, the scans keep their condition as a skip rather than a no-run, and an if: always() gate turns the results into one always-reporting context.

PR #1224, opened 2026-09-15
Do not treat the zoo's own bitsandbytes stub as a sys.modules leak

danielhanchen:
## What

The sys.modules leak gate in `tests/conftest.py` counts unsloth_zoo's own bitsandbytes stub as pollution, so on any machine without a real bitsandbytes wheel the first test that touches a bitsandbytes name fails at teardown through no fault of its own.

Surfaced on a macOS Apple Silicon run of the suite:

```
84 passed, 1 error
ERROR tests/test_loss_normalization_contract.py::test_guard_is_present_and_mirrors_the_stock_predicate
AssertionError: ... replaced ['bitsandbytes.nn'] in sys.modules and did not put it back
(bitsandbytes.nn from None)
```

## Why it happens

`unsloth_zoo/stubs/bitsandbytes_stub.py` installs a meta path finder when no real wheel exists, which is every Apple Silicon machine. The finder mints submodules on first import, so `bitsandbytes.nn` appears in sys.modules during whichever test first imports something that reaches it. The gate sees a name that was absent, is now a bare `ModuleType`, and reports a swap.

It is not a swap. The stub is deliberate, process wide and idempotent, and "putting it back" would mean deleting a module the rest of the session expects to find.

## The change

One condition in the teardown gate, plus the predicate behind it. Every stub module carries `IS_UNSLOTH_STUB`, finder-minted ones included (`_make_module` sets it precisely so callers cannot be fooled by the permissive `__getattr__`), so the exemption keys on that rather than on the name.

The gate still catches what it exists for: a test that drops its own hand rolled `bitsandbytes.functional` in and leaves it there is still reported, and so is a test that replaces a real wheel with anything.

## Verification

End to end, same two tests (one leaving a marked stub, one leaving an unmarked substitute), run under an interpreter with no bitsandbytes installed:

| conftest | result |
| --- | --- |
| main | 2 passed, 2 errors |
| this branch | 2 passed, 1 error |

The remaining error is the unmarked substitute, which is correct. Under an interpreter that does have a real bitsandbytes, both are still reported, since replacing a real module is a swap either way.

Four unit tests added to `tests/test_conftest_module_leak_gate.py` covering the marked stub, an unmarked substitute, a real wheel and `None`. `tests/test_conftest_module_leak_gate.py` and `tests/test_loss_normalization_contract.py` pass together (38 passed, 1 skipped).

Found while running unslothai/unsloth-zoo#1217 across Linux, macOS and Windows. It is unrelated to that PR: the same test file on main hits it identically, since the trigger is the first import of a bitsandbytes name in the process.


danielhanchen:
@codex review

danielhanchen:
@codex security review

PR #1237, opened 2026-09-15
Keep negative embedding rows classified as trained

vineethsaivs:
All-negative embedding rows are treated as untrained, producing NaN new-token initializations or false errors for frozen embeddings.

The checks compare only each row's maximum with epsilon, so negative trained weights pass as near zero. Check both bounds around zero in token averaging and validation, without allocating a full absolute-value copy.

Tests: `UNSLOTH_ZOO_DISABLE_GPU_INIT=1 python -m pytest tests/test_add_new_tokens_padded_vocab.py -q`. Two regressions fail before. Seven tests pass after, using a tiny Llama, a real tokenizer, a finite forward loss and zero/near-zero controls. The repository's pinned Ruff gate and Python compilation checks pass. Native CPU tests, no model download or GPU run.

The regression suite now runs in CPU CI.


danielhanchen:
Confirmed this hits unsloth_zoo/tokenizer_utils.py where the untrained check still compares only the row maximum against eps, so an all-negative row is counted as untrained; the added lower bound looks right and I will get this reviewed.

danielhanchen:
Taking this over and reviewing it on its merits, not just its CI.

The classification is the point: "untrained" is meant to be "this row was never updated", which is an all-zero row, and `amax(row) <= eps` alone says yes to any row whose largest element is negative. A row of all -0.5 is trained and was being counted as untrained.

Checked which rows change verdict under the new test, and it is exactly the defect and nothing else:

| row | before | after |
| --- | --- | --- |
| all zeros | untrained | untrained |
| all negative (-0.5) | untrained | **trained** |
| all positive | trained | trained |
| mixed sign | trained | trained |
| all -1e-20 | untrained | untrained |

All three call sites are covered, `mean_of_trained_tokens` and both indicators in `fix_untrained_tokens`, and there is no fourth site with the one-sided test left behind. Two reductions rather than an absolute-value copy, so no extra allocation.

The tests are not vacuous: reverting the source change fails exactly `test_add_new_token_keeps_negative_trained_rows_in_the_mean` and `test_frozen_token_validation_uses_magnitude[-3.0-False]`, and 7 pass with it. Wiring the file into `Repo tests (CPU)` is right, since it had never run in CI.

The `Torch not compiled with CUDA enabled` failure that was on this PR was never yours. It was a zoo bug your new CI step surfaced, fixed in #1238, and CI is green here now.

@codex review

PR #1239, opened 2026-09-16
Treat safe_serialization = None as the safetensors default, not a pickle

danielhanchen:
## Summary

`safe_serialization = None` is what our own warning, and the troubleshooting page it points at, tell
a caller to pass in order to force safetensors. `None` is falsy to peft and to transformers, so a
save that forwards it verbatim selects the torch pickle writer and produces `adapter_model.bin` or
`pytorch_model.bin`: the documented remedy for `.bin` files produces `.bin` files. This normalises
the value at the save entry point and pins it with a test that also gates the rest of the package
against the same mistake.

## Root cause

`merge_and_dequantize_lora` forwards `safe_serialization` straight to a `save_pretrained`, on both of
its writers, without looking at it. `save_pretrained` treats the argument as a boolean, so `None`
means "no", not "default". Measured on a two-step Llama-3.2-1B LoRA:
`model.save_pretrained(directory, safe_serialization = None)` wrote a 22.6 MB `adapter_model.bin`,
while the same call with the argument omitted wrote `adapter_model.safetensors`.

## Changes

`unsloth_zoo/saving_utils.py` only.

- `normalize_safe_serialization` returns `True` for `None` and returns `True` and `False` unchanged,
  so an explicit `safe_serialization = False` still writes a pickle for a caller who wants one. It is
  exported, because the same normalisation is needed on the unsloth side of the same call chain.
- `merge_and_dequantize_lora` normalises the argument once, before either writer can see it.

## Compatibility

| corner | python | torch | transformers | trl | peft | base | branch |
|---|---|---|---|---|---|---|---|
| mA | 3.10.19 | 2.6.0+cu124 | 4.57.6 | 0.22.2 | 0.18.0 | fail (new test absent) | pass |
| mC | 3.13.12 | 2.11.0+cu130 | 5.5.0 | 1.0.0 | 0.20.0 | fail (new test absent) | pass |
| mD | 3.13.12 | 2.14.0+cu130 | 5.17.0 | 1.13.0 | 0.20.0 | fail (new test absent) | pass |

The change is a three-line pure-Python normalisation with no version-dependent behaviour, which is
why the oldest corner and the newest agree.

## Tests

`tests/test_safe_serialization_none.py`, 7 tests, new. It pins the three values of the argument, that
`merge_and_dequantize_lora` normalises before either writer is reached, and that the symbol is
exported. It also carries a drift gate that walks the package for any other save entry point which
accepts `safe_serialization` and forwards it unnormalised, so a future writer cannot reintroduce the
same defect quietly. CPU-only, no network, no model download. On main the module does not import,
because the symbol it pins does not exist yet.

The file is added to the existing gated pytest step in `.github/workflows/consolidated-tests-ci.yml`
rather than left to collect-only, since it is stdlib-only and costs about seven seconds.

## Overlap

`unslothai/unsloth-zoo#1232`, `#1234` and `#1236` also touch `unsloth_zoo/saving_utils.py`, in the MoE
merge layout, the partial-merge refusal and the GGUF export preflight respectively. None of them
touch `merge_and_dequantize_lora`'s argument handling, so there is no textual overlap with this
change.

## Known limitations

This covers the zoo side of the call chain. The unsloth side, including the `model.save_pretrained`
wrapper that was the call actually named in the report, is `unslothai/unsloth#11067`.
Either one alone fixes a real path; together they cover every save entry point that accepts the
argument.

See unslothai/unsloth#1792.


danielhanchen:
@codex review

danielhanchen:
@codex security review

danielhanchen:
@codex review

PR #1240, opened 2026-09-16
Use expm1 for the GRPO reverse-KL term

vineethsaivs:
Nearby policy and reference log probabilities make the reverse-KL term zero, or negative, in float32.

Subtracting the linear term and one from `exp(delta)` loses the second-order term through cancellation. At delta = 1e-6 it returns -5.96e-08, and a reverse KL is non-negative by construction, so with beta > 0 that term was subtracted from the loss instead of added. Use `expm1(delta) - delta`.

This repairs the KL value: the logged `mean_kl`, the scalar loss, and the impossible negative case. Training dynamics are unchanged, since both spellings differentiate to `exp(delta) - 1` and autograd evaluates that identically in float32.

Tests: `UNSLOTH_ZOO_DISABLE_GPU_INIT=1 PYTHONPATH=. python -m pytest tests/test_rl_replacements_cpu.py -q`. Four regressions fail before. All 88 CPU tests pass after, including a Taylor reference and gradients. Ruff and compilation checks pass. CPU only.


danielhanchen:
Confirmed this hits rl_replacements.py where exp(delta) - delta - 1 cancels away the small reverse-KL term in float32, and expm1 keeps it with the same gradient. Will get this reviewed.

danielhanchen:
Confirmed the defect and the fix, and measured two things worth recording before this goes in.

The cancellation is worse than "zero or inaccurate". In float32, with delta = ref - new:

| delta | `exp(d) - d - 1` | `expm1(d) - d` | exact |
| --- | --- | --- | --- |
| 1e-4 | 0.0 | 4.999e-09 | 5.000e-09 |
| 1e-6 | **-5.96e-08** | 4.547e-13 | 5.000e-13 |

At 1e-6 the old form returns a negative reverse KL, which is not merely imprecise: a KL penalty is non-negative by construction, so with beta > 0 that term was subtracted from the loss instead of added.

The gradient, however, does not change. Both spellings differentiate to `exp(d) - 1`, and autograd evaluates that identically in float32 for both:

| delta | grad, old | grad, new | exact |
| --- | --- | --- | --- |
| 1e-6 | 9.536743e-07 | 9.536743e-07 | 1.000000e-06 |
| 1e-7 | 1.192093e-07 | 1.192093e-07 | 1.000000e-07 |

So what this repairs is the KL value: the logged `mean_kl`, the scalar loss, and the impossible negative case. Training dynamics are essentially unchanged, because `1 + d` is representable in float32 well below where `d^2/2` is lost. Worth saying plainly in the description, since "preserve small reverse-KL penalties" reads as a change in what the penalty does to the policy.

Two notes for whoever merges.

It moves us off TRL's spelling. `trl/losses/grpo_loss.py:33` `k3_loss_fn` uses `torch.exp(log_p - log_q) - (log_p - log_q) - 1.0`, and verl and vime document and implement the same naive form. Our k3 currently matches TRL to 0.0 absolute difference, and this breaks that exact match in the direction of being more correct. That is the right trade, and it is worth sending upstream to TRL rather than carrying alone.

I file-searched for other copies. `unsloth_zoo/rl_replacements.py:556` is the only source site, so nothing else needs the same edit. But the generated `unsloth_compiled_cache/UnslothGRPOTrainer.py` embeds this expression inline (twice per generated file), so an existing cache keeps the old form until it is regenerated.

Test looks right, in particular asserting the gradient and using a Taylor reference rather than re-deriving the same cancellation in the expected value.

vineethsaivs:
Retitled and rewrote the description to say the KL value is what changes and that the gradient is identical, so it no longer reads as a change in policy behaviour. Good catch on the negative value at 1e-6; that is a stronger statement than the one I made and it is now the lede. Agreed on the compiled cache needing regeneration, and I will send the same change to TRL.

vineethsaivs:
Correction on the TRL half. Before sending it I measured the reachability and it does not hold up: in float32 `exp(d) - d - 1` returns exactly 0 only for `|d| <= 3.06e-4`, and a GRPO step at the default 1e-6 learning rate leaves per-token deltas around 1e-1, three orders above that. Over eight steps the logged mean KL was wrong by 0.00%.

So this stays what you described, a value-correctness fix for the negative case and the per-token zeros, and I am not filing it upstream as a training-impact bug. My earlier TRL report was closed for exactly that gap and I would rather not repeat it.

PR #1241, opened 2026-09-16
Keep DiscoPOP loss finite for large negative margins

vineethsaivs:
Large negative preference margins make the DiscoPOP loss and gradient NaN despite a finite mathematical result.

The exponential overflows while its sigmoid multiplier approaches zero, producing infinity times zero. Compute the exponential-sigmoid product from its combined log value.

Tests: `UNSLOTH_ZOO_DISABLE_GPU_INIT=1 PYTHONPATH=. python -m pytest tests/test_mlx_preference.py -q`. Three extreme-margin regressions fail before. All 139 preference tests pass after. The same three new loss/gradient cases also pass on native MLX, against an independent float64 reference. The repository's pinned Ruff gate and Python compilation checks pass. The repository CPU suite uses its Torch-backed MLX shim; native MLX checks also ran on Apple M2 Pro. No full language-model training run.


danielhanchen:
Confirmed this hits unsloth_zoo/mlx/preference.py where exp(-scaled) overflows while the sigmoid multiplier underflows, so the DiscoPOP loss and gradient go NaN, and the log-space form looks equivalent. Will get this reviewed.

PR #1243, opened 2026-09-16
Scale the fused CE per-chunk budget with VRAM, behind an opt-in flag

danielhanchen:
## What

The fused cross entropy loss chunks the lm_head matmul plus softmax to bound peak memory. The per-chunk budget is `min(free_vram * 0.5, 4.0)` GiB.

The `4.0` constant is correct on a small card, where half the free pool would otherwise round down to a single chunk and materialize the full float32 logits. It does not scale up. On a 183GB B200 at vocab 248320, hidden 2048, 2048 tokens the entire transient is only about 7.6 GiB, and the constant still forces 4 chunks. Every extra chunk re-reads the whole lm_head weight and relaunches the matmul, softmax and backward chain.

Setting `UNSLOTH_CE_VRAM_AWARE_CAP=1` replaces the constant with

```
cap_gb = clamp(total_vram_gb / 6.0, 4.0, 16.0)
```

The divisor is chosen so the cap is exactly `4.0` at 24 GiB.

| card | cap, flag off | cap, flag on |
| --- | --- | --- |
| T4 16GB | 4.00 | 4.00 |
| P100 16GB | 4.00 | 4.00 |
| L4 22.5GB | 4.00 | 4.00 |
| RTX 4090 24GB | 4.00 | 4.00 |
| A100 40GB | 4.00 | 6.60 |
| A100 / H100 80GB | 4.00 | 13.33 |
| B200 183GB | 4.00 | 16.00 |

Every card at or below 24 GiB reduces to the old formula bit-for-bit even with the flag on. The ceiling keeps the cap bounded, so a large card never degenerates into one unbounded chunk, which is the property the original constant was protecting. `free_vram * 0.5` remains the other half of the `min()`, so a large but nearly full card still chunks hard.

## Why it is opt-in

It is not a universal win. Chunking cost is dominated by the lm_head weight gradient, so an ordinary LoRA run with a frozen lm_head pays more memory for much less speed. Measured at vocab 248320, hidden 2048, 2048 tokens on a B200:

| | 4 chunks (stock) | 1 chunk (opted in) |
| --- | --- | --- |
| lm_head trained | 13.50 ms / 3.13 GiB | 8.81 ms / 3.84 GiB |
| lm_head frozen | 6.99 ms / 1.47 GiB | 4.99 ms / 2.89 GiB |

So the 1.5x is a trainable lm_head result: full finetuning, vocab extended runs, or anything training the head. Plain LoRA gets about 2 ms for an extra 1.4 GiB.

On a 40 GiB A100 at vocab 151936, hidden 896 with a frozen lm_head it is worse than that. The single chunk landing measured 325.9 ms/step against 322.4 ms/step stock, at 2.42 GiB peak against 1.29 GiB, because the optimum there is 2 chunks rather than 1 and the policy jumps straight past it.

Until the policy targets the knee of that curve rather than the fewest chunks that fit, the default stays exactly where it is. This PR is the mechanism plus the evidence, not a default change.

## Two supporting changes

`_get_chunk_multiplier` is memoized on `(vocab_size, target_gb, fixed_gb)`. Passing the `None` sentinel into it pinned the first free VRAM observation for the life of the process, so the chunk count could never react to memory pressure that appeared later. `get_chunk_size` now resolves `None` before the memoized call, which puts the live reading into the cache key. Since a live float key is unbounded, the memo is an LRU with a bounded size instead of `functools.cache`. It still exposes `cache_clear` and `cache_info`.

`get_chunk_size` gains an optional trailing `device_index`, and the call site in `UnslothFusedLoss.forward` passes the lm_head device. With `device_map="balanced"` the lm_head can sit on `cuda:3` while the budget was always read from `cuda:0`. That was harmless while the cap was a constant, but not once it is derived from that card's own size and free pool.

## Safety

- Default off means the policy is the pre-change formula on every card. There is a test that pins this.
- The flag is read live rather than captured at import, so it can be toggled inside a process.
- When `mem_get_info` is missing or raises, on CPU, MPS and driverless CI, the budget falls back to the historical `4.0` constant.
- XPU keeps using `torch.xpu.mem_get_info`.
- `UNSLOTH_CE_LOSS_N_CHUNKS` and `UNSLOTH_CE_LOSS_TARGET_GB` overrides are unaffected.

## Testing

Loss is bit-identical across vocab and hidden size combinations, with gradient deltas at bf16 accumulation noise:

| vocab / hidden / tokens | chunks off -> on | abs loss diff | rel d_hidden | rel d_W |
| 248320 / 2048 / 2048 | 4 -> 1 | 0.0 | 1.11e-3 | 2.38e-4 |
| 248320 / 2048 / 4096 | 8 -> 4 | 0.0 | 5.71e-4 | 1.80e-4 |
| 151936 / 2048 / 2048 | 4 -> 1 | 0.0 | 8.82e-4 | 3.72e-4 |
| 32000 / 4096 / 2048 | 1 -> 1 | 0.0 | 0.0 | 0.0 |

Hardware: B200 locally, plus Kaggle T4x2 and Colab A100 through real notebook runs. The T4x2 run confirms the chunk count is identical to stock and there are no OOMs across 1, 2, 4, 8 and 16 chunks with the lm_head both trained and frozen.

202 tests pass, including the 9 pre-existing chunk cap tests unmodified, plus 28 new ones covering the default being off, small card equivalence, the bounded ceiling, the memo not freezing the live reading, XPU, multi GPU device selection, and graceful degradation.

## Not verified

- The 24 to 96 GiB band is exercised by the policy function and by the A100 40GB notebook run, but the only card where the changed path ran a full finetune is the B200 and that A100.
- XPU and ROCm are covered by unit tests with a faked `mem_get_info`, not on real hardware.
- Windows and macOS are untested. The code is platform agnostic Python, which is an argument rather than a measurement.


danielhanchen:
Context on where the 4 GiB constant came from, for anyone reviewing this.

The cap this PR makes configurable was introduced deliberately in #857 ("Cap the fused cross entropy chunk target so chunking stays active on large GPUs"). That PR's point was that `min(free_gb * 0.5, target)` with no ceiling means a large card with lots of free memory picks a target big enough that chunking effectively stops, and peak memory goes up for no throughput gain. It measured Qwen3-30B-A3B and a gemma MoE at 32k tokens: identical tok/s, peak memory down 6.4 GB and 21.4 GB.

So this PR is not a claim that #857 was wrong. The 4 GiB floor stays the default and is what every existing run keeps getting. `UNSLOTH_CE_VRAM_AWARE_CAP=1` is strictly opt-in, and the scaled value is clamped into [4, 16] GiB, so the worst case under the flag is still a hard ceiling rather than the unbounded pre-#857 behaviour that motivated the cap in the first place.

The reason to expose it at all: 4 GiB was chosen against 32k-token MoE runs on large cards, where it is clearly right. On a card with a lot of headroom and a short sequence, the same constant produces more chunks than the memory situation calls for and the per-chunk launch overhead shows up. Making the ceiling scale with total VRAM lets those cases pick a larger chunk without reintroducing the regression #857 fixed.

Two things a reviewer should push on:
- whether 16 GiB is the right ceiling. It is 4x the value #857 landed on, and I picked it as "large enough to matter on a 180 GB card, small enough to stay bounded", not from a sweep across the #857 workloads.
- whether the flag should exist at all versus just leaving the constant alone. If the answer is that the constant is fine everywhere, this PR should be closed rather than merged with a knob nobody sets.

Default-off behaviour is covered by tests that assert the historical constant and chunk count are unchanged with the flag unset, and that the flag is read live rather than captured at import.

danielhanchen:
Closing this. The premise held up under measurement, but the knob does not earn its keep.

Rechecked in the regime these runs actually use. LoRA freezes the lm_head, and both Unsloth's and the upstream chunked implementations gate the [vocab, hidden] grad_weight accumulation on weight.requires_grad, so measuring with a trainable lm_head scores work that never happens. With it frozen and torch_compile=True, which is what the call site passes, the fused CE runs at 11.47 ms and 2.02 GB at 4096 tokens x hidden 2048 x vocab 248320 on a B200. torch 2.13's F.linear_cross_entropy chunked path is 58.55 ms there, so our implementation is about 5x faster and that side of the argument is settled.

What does not hold up is the 16 GiB ceiling. It is 4x the value #857 measured its way to, I picked it as a round number rather than from a sweep across the workloads #857 used, and the region it governs is roughly 1 percent of a training step. A knob nobody sets, guarding a constant that was chosen with evidence, is maintenance with no return.

The more useful finding from the same work, recorded here so it is not lost: the fused loss sits about 4x above the achievable gradient floor. Against an fp64 reference at a realistic loss magnitude, relative RMS on dX is 6.7e-03, which is identical to plain eager bf16 (6.69e-03), while the floor for a bf16 output is 1.64e-03. The cause is that cross_entropy_loss.py computes logits via F.linear in bf16 and only upcasts afterwards with .float(), so the upcast lands after the rounding. It buys an excellent loss scalar (|dloss| 2.1e-05 against upstream's 4.1e-02) but cannot recover what the logits already lost. Computing that GEMM with out_dtype=torch.float32 would close it, at 2x the logits chunk memory, which is precisely the memory the chunking exists to bound. That is a real tradeoff worth a separate PR with its own measurements, not a rider on this one.

PR #1248, opened 2026-09-16
Count MoE tokens per expert without synchronising the device

danielhanchen:
torch.bincount's output size depends on the data even when minlength is given, so
on CUDA it D2H-copies to learn max(input) and blocks the host. Every MoE forward
paid that: 2 syncs per block, ~58us of host time per call.

Replace it with a scatter_add_ onto a pre-zeroed tensor, behind one shared helper
(count_tokens_per_expert) so every backend gets the same counter. Measured on a
B200 at E=256 with 16384 routed rows: 2 sync warnings under
torch.cuda.set_sync_debug_mode("warn") go to 0, and host time per call goes from
~58us to ~19us. On a 40-layer Qwen3.5/3.6-35B-A3B that removes 80 pipeline drains
per step.

Same treatment for the gpt-oss expert bias expansion, which used
repeat_interleave with tensor repeats and no output_size=, so it synced to learn
the output row count. sorted_indices already permutes rows into expert order, so
index_select by each row's expert id is the same result with no sync: 74us to 8us
host at E=128 x 16384 rows. Qwen3.5/3.6 have no expert biases, so that path is
gpt-oss only.

Two call sites in gpt_oss.py keep bincount on purpose: they .tolist() the result
immediately for a Python per-expert loop, so a sync-free counter buys nothing
there.

Correctness

scatter_add_ accumulates with atomics so the order is nondeterministic, but every
addend is exactly 1 and integer addition is associative and commutative, so the
sum is bit-exact and run-to-run deterministic. Verified bit-identical to bincount
over 719 cases: random shapes, all-tokens-to-one-expert, 90% skew, zero tokens,
single token, num_experts > num_tokens, indivisible splits and real top-k
routings. It also does not trip torch.use_deterministic_algorithms(True).

LoRA on the experts is the case that matters, since a wrong count silently
mis-segments grouped_mm and corrupts per-expert gradients without erroring. With
adapters on mlp.experts.gate_up_proj and mlp.experts.down_proj and every lora_B
perturbed to nonzero, old vs new gradients differ by at most 3.6e-12 across nine
routing regimes, against a new-vs-new control that differs by the same amount.
Counts were bit-identical in every case. Identical under UNSLOTH_MOE_LORA_MERGED=1.

Performance

No measurable end-to-end throughput change, and this is not filed as a speedup.
Six independent processes per arm per model put Qwen3.5-35B-A3B at +5.0% and
Qwen3.6-35B-A3B at -0.4%, both inside a +/-7-8% noise floor (p=0.30, p=0.91).
Removing 80 drains from a ~350ms step is below what that design can resolve.
Peak VRAM is identical.

What is consistent, in both models and in the right direction, is step-time
jitter: within-run relative stdev drops from 6.38% to 4.97% and from 6.55% to
4.52%. Draining the pipeline 80 times per step couples host jitter into step time.

It also makes the MoE routing prologue CUDA-graph capturable. bincount fails
capture outright with "Cannot copy between CPU and CUDA tensors during CUDA graph
capture" at every shape tested, decode and prefill; the sync-free counter captures
and replays correctly against fresh routing data. Nothing on main captures over
this path today, so that is an unblock rather than a fix.

UNSLOTH_MOE_COUNT_DEBUG=1 wires a call counter for tests. It is read once at
import and never per call: incrementing a module global inside the counter is a
Python side effect Dynamo cannot keep in the graph, and with it always on a
compiled MoE block re-traced every step and ran 572ms instead of 3.6ms.

545 tests pass across the MoE, gpt-oss and Qwen MoE suites, identical to main.

Related: #608 made the same bincount substitution but was closed unmerged. Its case rested on CUDA-graph decode speedups that only existed via unslothai/unsloth#5123, which is still open, and it carried nine other commits across vLLM and Gemma 4. This is the counting change on its own, justified by training evidence with no cross-repo dependency, and it does not touch the `@torch.compiler.disable` decorator (see below).

### On removing `@torch.compiler.disable`

#608 also proposed dropping that decorator on the grouped_mm path. I tried and could not justify it, so this PR only adds a comment at `qwen3_moe.py` recording why it stays.

`bincount` is not what blocks compilation there. Graph breaks are identical at 8 with `bincount` and with the sync-free counter, and 6 with no LoRA at all. All 8 come from one cause, a `torch._C.Generator` constructed inside the compiled region, which has nothing to do with the routing prologue. `fullgraph=True` does not even reach that: Dynamo crashes while building guards on a global that does not exist, `module 'transformers.models.qwen3_5_moe.modeling_qwen3_5_moe' has no attribute 'Qwen3_5MoeMLP_forward'`, which looks like a separate patching bug worth its own issue.

Note also that #608's justification for dropping it described the pre-merged LoRA path. On the default separated-LoRA path `forward_native_grouped_mm` contains exactly the things that argument says are absent: per-forward `_extract_lora_weights`, a `try/except RuntimeError` around the grouped mm, and an `offsets.cpu().tolist()` fallback loop over experts.


danielhanchen:
@codex review

danielhanchen:
@codex security review

danielhanchen:
@codex review

danielhanchen:
@codex review

PR #1252, opened 2026-09-16
perf(mlx): fuse MoE expert routing

Lyxot:
MoE blocks pick experts for every token through a chain of small operations. Qwen3.5-family blocks run a `precise=True` softmax across all experts, an `argpartition` across all experts, a slice, a gather, a sum and a divide; Gemma 4 routers run an `argpartition`, a gather, a softmax over the selected experts and a per-expert scale. At decode each launch costs about as much as the work it does, so the chain is a fixed tax paid once per token per MoE layer — 40 layers on Qwen3.6-35B-A3B, 30 on gemma-4-26B-A4B.

This fuses that chain into one Metal kernel. A simdgroup holds a row in registers in the layout the native softmax reads, so the normalizer sums in the native order; K rounds of `simd_max` with ties resolved toward the higher index reproduce the stable `argpartition(kth=-K)[..., -K:]` order; the Qwen denominator accumulates sequentially in the input dtype like `mx.sum` over at most eight values, and the Gemma softmax emulates the non-precise two-thread reduction, bf16 partials, float combine and rounded reciprocal. NaN sorts last and poisons the weights, as it does natively. Outputs are bit-identical, which is the design constraint on the kernel rather than a happy result: it reproduces the native chain's rounding instead of computing the mathematically nicer answer.

The scope recognizes routing bodies by the AST fingerprint of the upstream source, the same contract mechanism the MoE gate/up fusion already uses, and restores the original module classes on exit. It uses native operations for unsupported shapes and dtypes, training, sharded, distributed, drifted and target-verify calls. The kernel supports float16, bfloat16 and float32, expert counts divisible by 32 and at most 256 in Qwen mode and at most 1024 in Gemma mode, and top-k up to 8. Because the rounding emulation depends on the MLX build, the first use of each dtype and shape combination checks the kernel against the native chain on three probe batches, one of them integer-valued so top-k boundary ties are always exercised, and a mismatch keeps the native chain with a warning. GLM5-next and DeepSeek-V4 routers, which use compiled module-level selection with a different scoring tail, are out of scope.

The Gemma router's normalization weight is `scale * hidden_size ** -0.5`, constant while `scale` is, yet it was rebuilt on every call: one graph node per MoE layer per token. The scope now builds it once on entry, holds it on the module and drops it when the outermost scope closes. `scale` must therefore stay fixed while the scope is open, which is what `fused_moe_gate_up` already requires of the weights it packs and what every call site satisfies by entering both scopes in the same statement around a serialized generation loop. Replacing `scale` is detected and takes the native call; editing it in place is not detected, and edits between scopes are always picked up.

Enable the scope alongside the existing MoE gate/up, decode-convolution and residual-normalization fusions in both `generation_mode` and the loader's text and VLM `generate()` wrappers. No dependency changes.

### Performance

Measurements on 2026-09-16 compare this branch against its `main` baseline through the same Studio revision for both arms, over Studio's streaming HTTP API. Hardware: Apple M3 Max, 128 GiB unified memory. Runtime: MLX 0.32.2, MLX-VLM 0.7.1, Transformers 5.5.0. That Studio revision also enters `fused_residual_norm`, which neither arm exports, so both no-op it and expert routing is the only difference between them.

| Model | Prefill main → branch (tok/s) | Change | Decode main → branch (tok/s) | Change | Decode ABBA gain range |
|---|---:|---:|---:|---:|---:|
| Qwen3.6-35B-A3B (8-bit, 256 experts, 40 MoE layers) | 1,332.4 → 1,336.0 | +0.27% | 63.99 → 65.07 | **+1.70%** | +1.01% to +2.19% |
| Qwen3.6-35B-A3B (UD 4-bit, 256 experts, 40 MoE layers) | 1,462.8 → 1,462.2 | −0.04% | 70.67 → 72.71 | **+2.90%** | +2.45% to +3.49% |
| gemma-4-26B-A4B-it (8-bit, 128 experts, 30 MoE layers) | 1,190.8 → 1,191.0 | +0.01% | 61.35 → 62.29 | **+1.54%** | +0.96% to +1.95% |

Values are geometric means of the server-reported generator rates: three ABBA blocks per model, six timed samples per variant. Each block uses identical prompts and settings: 4,088–4,089 input tokens, 512 output tokens, greedy decoding, seed 42, thinking disabled, native unquantized KV cache, and a requested 8,192-token context. Prefix snapshots are cleared before each request, with zero reported cached tokens. Kernel instrumentation is disabled during timing. Each timed request follows at least 30 seconds at nominal thermal state with no detected test workload; all reported samples stayed at nominal thermal state.

All three decode results pass the predefined repeatability checks: within-variant coefficient of variation at most 3% (observed 0.38% to 1.02%) and relative spread between ABBA ratios at most 3% (observed 0.97% to 1.16%). The gain ranges above are observed block ranges, not confidence intervals. Prefill passes the same checks on all three models and establishes no change, which is the expected result at these prompt lengths: the routing chain runs one row per prompt token during prefill, so its dispatch overhead is amortized rather than paid per launch.

The two Qwen rows are the same architecture at two weight widths, and the difference between them is the more useful number in this table. The fusion removes the same work per layer per token either way, so its share grows as the rest of the decode step gets cheaper: 4-bit decodes about 10% faster than 8-bit and gains about 70% more from the fusion. Gemma gains least because its native chain is cheapest to remove — it argpartitions 128 raw scores and softmaxes only the 8 it selects, where Qwen runs a precise fp32 softmax and an argpartition across all 256 — and because it has 30 MoE layers rather than 40.

Activation was confirmed per model in untimed warmups, which distinguishes a small real gain from a silent fallback. In a 64-token audit the branch made zero native `mx.argpartition` calls against the baseline's 2,680 on either Qwen model and 2,010 on Gemma, while its fused kernel ran exactly that many times; the branch patched 40 of 40 and 30 of 30 MoE blocks, and the baseline patched none. Generated text matched across every timed ABBA block, and separate 64-token ID audits matched for all three models.

Checkpoints: `mlx-community/Qwen3.6-35B-A3B-8bit`, `unsloth/Qwen3.6-35B-A3B-UD-MLX-4bit` and `mlx-community/gemma-4-26b-a4b-it-8bit`. The Gemma figure comes from a second complete series: the first failed the variability check with a within-variant CV of 16.7% and a block spread of 22.0%, because that checkpoint loaded while the previous model's weights were still being released and its first block ramped from 44 to 59 tok/s. The rerun started from a quiescent machine and changed only the warmup count, four per variant instead of two. Every sample of the failed series is retained; none of it is mixed into the reported figure.

### Validation

Real Metal tests cover bitwise numerics against the native chain on both routing modes, the verification probe, boundary ties, shape and dtype eligibility, native fallbacks for training and drifted bodies, nested scopes, and cleanup after normal or exceptional generator exits.

```bash
python -m pytest tests/test_mlx_moe_metal.py tests/test_mlx_generate.py -q
```

Mutation-checked, 11 mutants, all killed: allowing 1024 experts in Qwen mode, raising the top-k cap to 10, accepting rank-1 logits, dropping the kernel's NaN-to-`+INFINITY` mapping, ignoring a drifted Gemma body, recomputing the Gemma normalization weight per call, never building it, keeping it after the scope closes, dropping the `root_size` factor from it, dropping the staleness guard on `scale`, and making that guard always true.

On the current `main` base, `tests/test_mlx_moe_metal.py tests/test_mlx_generate.py tests/test_mlx_generate_metal.py` gives 230 passed, and 229 at the first commit alone. Two VLM tests in `tests/test_mlx_generate_metal.py` fail identically on `main` in the local environment because a cached remote-code checkpoint imports `timm`, which is not installed there; they are unrelated to this branch.


Lyxot:
@codex review

PR #1254, opened 2026-09-16
Read labels off a batch only when the batch is a mapping

danielhanchen:
## Summary

Every GRPO run on `main` dies on its first step with `AttributeError: 'list' object has no attribute 'get'`. TRL's `GRPOTrainer` collates with `identity`, so a batch is the sampler's LIST of dataset rows, and `_unsloth_get_batch_samples` calls `.get` on it. This asks for a `Mapping` before reading the key.

One line of behaviour change, plus the comment above it, which reasoned about `labels` being a non-tensor but not about the sample itself not being a mapping.

## The traceback users see

Qwen2.5-0.5B-Instruct, `GRPOTrainer` with `per_device_train_batch_size=4`, `num_generations=4`, unsloth `2026.9.4`, unsloth-zoo at `b27e629` (current `main`):

```
  File "/.../transformers/trainer.py", line 1809, in _run_epoch
    batch_samples, num_items_in_batch = self.get_batch_samples(epoch_iterator, num_batches, self.args.device)
                                        ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/.../unsloth_zoo/loss_utils.py", line 402, in _unsloth_get_batch_samples
    batch_samples[0].get("labels") if len(batch_samples) > 0 else None, "ndim", None,
    ^^^^^^^^^^^^^^^^^^^^
AttributeError: 'list' object has no attribute 'get'
```

Byte for byte the same failure on TRL 0.22.2 with transformers 4.57.6, and on TRL 1.13.0 with transformers 5.17.0. Both reach it from `trainer.train()`, at step 0, before a single optimizer step. Nothing in GRPO recovers from it: `get_batch_samples` is the first thing the epoch loop calls.

## Blame

`git log -L 402,402:unsloth_zoo/loss_utils.py` lands on `e2e81b7a`, PR #1217, "Count tokens for custom losses without model keyword arguments".

Before #1217:

```python
if has_kwargs and len(batch_samples) > 0 and "labels" in batch_samples[0]:
```

After #1217:

```python
labels_are_countable = getattr(
    batch_samples[0].get("labels") if len(batch_samples) > 0 else None, "ndim", None,
) is not None
```

`in` is list safe by accident and `.get` is not. `"labels" in [{...}, {...}]` asks whether any ELEMENT equals the string `"labels"`, which for a list of dicts is simply `False`, so the whole counting body was skipped and GRPO trained. `.get` is a `dict` method, so the same input is an `AttributeError` before any guard can run. The `len(batch_samples) > 0` check that came along was never the protection here; the shape of the element was.

The rewrite was not gratuitous, and this PR does not undo it. `in` cannot tell a tensor `labels` from a non-tensor one, so a mapping carrying list labels used to enter the `try`, blow up on `labels[..., 1:]` and come back out as `RuntimeError`. The `.get` plus the `.ndim` probe declines to count that instead, which is what stock transformers does with the same input. That is preserved exactly.

## The fix

```python
labels_are_countable = getattr(
    batch_samples[0].get("labels") if len(batch_samples) > 0
    and isinstance(batch_samples[0], Mapping) else None, "ndim", None,
) is not None
```

`Mapping`, not `dict`: transformers hands collated batches back as `BatchEncoding`, which is a `UserDict` and therefore fails `isinstance(x, dict)`. Spelling the guard as `dict` would silently stop counting every `DataCollatorForLanguageModeling` batch, turning a crash into wrong gradients. A test pins that.

The loop below does `labels = x["labels"]`, which fails on a list in the same way, so it was reachable too. It is not reached now: the gate above it is the same gate the pre-#1217 code used, read off `batch_samples[0]`, so nothing that is not a mapping of tensors gets into the body. A parametrised test covers the list shapes that would otherwise arrive there, including a list whose rows do carry a `labels` key.

## Tests

`tests/test_grpo_identity_collator_batch_samples.py`, 9 tests, CPU only, no model downloads.

| Test | What it pins |
|---|---|
| `test_grpo_identity_collator_batch_is_a_list_of_dicts` | The shape, derived from TRL itself: asserts `data_collator=identity` is still in `GRPOTrainer.__init__`, then runs a real `DataLoader` with that collator and checks the batch is a list of dicts |
| `test_a_list_batch_is_not_countable_and_does_not_raise` | The regression: a list batch returns `None` and the samples untouched |
| `test_a_list_batch_is_not_countable_on_the_widened_route_either` | #1217 widened who is eligible to `compute_loss_func` without `**kwargs`; both routes read `batch_samples[0]` through the same expression, so both crashed |
| `test_no_list_batch_shape_reaches_the_counting_body` (x3) | Empty list, rows with no labels, and rows that DO carry `labels`, so `x["labels"]` in the loop is covered too |
| `test_a_mapping_batch_still_counts_exactly_what_it_counted_before` | Control: a normal causal batch still counts, to the same number (7 shifted targets) |
| `test_a_mapping_with_non_tensor_labels_declines_instead_of_raising` | Control: what #1217 bought |
| `test_a_mapping_that_is_not_a_dict_subclass_still_counts` | Control: `BatchEncoding` must still be counted |

### Can it fail

Not argued, measured. All runs below are the same two test files against the same venvs, changing only `loss_utils.py`.

**Against `main` (`b27e629`), test file only:** 5 failed, 4 passed, on TRL 0.22.2 and on TRL 1.13.0 alike. Every failure is `AttributeError: 'list' object has no attribute 'get'` at `loss_utils.py:402`. The 4 that pass are the shape probe and the three controls, which must be green on both sides, and are.

**Against this branch:** 9 passed on both.

**Negative control, stated then executed.** The control that matters is `test_a_mapping_with_non_tensor_labels_declines_instead_of_raising`, and the change that must turn it red is respelling the guard as the pre-#1217 membership test, `len(batch_samples) > 0 and "labels" in batch_samples[0]`. That spelling is list safe, so the regression tests would go green with it, and only the control can tell the two apart. Applied:

```
E   RuntimeError: 'list' object has no attribute 'shape'
/.../unsloth_zoo/loss_utils.py:519: RuntimeError
FAILED tests/test_grpo_identity_collator_batch_samples.py::test_a_mapping_with_non_tensor_labels_declines_instead_of_raising
1 failed, 8 passed
```

One test red, the exact one named, and #1217's gain is what it measures.

**Second control.** Weakening `Mapping` to `dict` must turn `test_a_mapping_that_is_not_a_dict_subclass_still_counts` red. Applied:

```
FAILED tests/test_grpo_identity_collator_batch_samples.py::test_a_mapping_that_is_not_a_dict_subclass_still_counts
1 failed, 8 passed
```

That is the shape the local shim carried during PR #11031's verification, which is why it is pinned rather than left to review.

### Real GRPO runs

Same script, same seed, unsloth `2026.9.4`, Qwen2.5-0.5B-Instruct on a B200, 3 steps.

| | TRL 0.22.2 / transformers 4.57.6 / torch 2.9.0 | TRL 1.13.0 / transformers 5.17.0 / torch 2.13.0 |
|---|---|---|
| `main` | `AttributeError` at step 0 | `AttributeError` at step 0 |
| this branch | trains, losses `[0.0, 0.0, 0.0003]` | trains, losses `[6.99e-06, 1.397e-05, 0.00014533]` |

Run with a fresh `UNSLOTH_COMPILE_LOCATION`, so `UnslothGRPOTrainer.py` was regenerated from scratch in each case. The generated trainer is unaffected: this function is patched onto `transformers.Trainer`, not copied into the generated source, and the traceback on `main` shows the generated trainer calling straight through to the installed `loss_utils`.

## Compatibility

`collections.abc.Mapping` and `isinstance` only. No transformers, TRL or torch API is involved.

`tests/test_grpo_identity_collator_batch_samples.py` plus the existing `tests/test_loss_normalization_contract.py`, on this branch:

| torch | transformers | TRL | Result |
|---|---|---|---|
| 2.6.0 | 4.57.6 | 0.22.2 | 92 passed, 1 skipped |
| 2.8.0 | 5.0.0 | 0.27.2 | 92 passed, 1 skipped |
| 2.9.0 | 4.57.6 | 0.22.2 | 92 passed, 1 skipped |
| 2.11.0 | 5.5.0 | 1.0.0 | 92 passed, 1 skipped |
| 2.11.0 | 4.57.6 | 1.13.0 | 93 passed |
| 2.13.0 | 5.17.0 | 1.13.0 | 93 passed |
| 2.14.0 | 5.17.0 | 1.13.0 | 93 passed |

The single skip is pre-existing, `this TRL has no chunked CE`, and is not one of the tests added here. `tests/test_zoo_history_regressions_deep.py`, `tests/test_undefined_name_gate.py`, `tests/test_relative_imports_resolve.py` and `tests/test_python39_compatibility.py` are green too, the last because `collections.abc.Mapping` predates 3.9.

## The same class of bug elsewhere

Swept `loss_utils.py` and the rest of the repo for accesses that assume a dict where an identity collator can hand back a list.

- `unsloth_zoo/loss_utils.py:402`, the one fixed here, is the only `batch_samples[0].` access in the repo.
- `unsloth_zoo/loss_utils.py:419`, `labels = x["labels"]` in the counting loop, has the same failure mode on a list but is unreachable once the gate above is correct. Left as is rather than guarded twice, so there is exactly one place that decides.
- `unsloth_zoo/training_utils.py:721`, `torch.count_nonzero(x["labels"][..., 1:] != -100) for x in batches`, is the same expression on the same kind of value. It is inside `unsloth_train`, which builds its own `DataLoader` and then calls `model(input_ids=..., labels=...)` on each batch, so a batch that was not a mapping of tensors would not survive the next four lines either. Not reachable from a GRPO collator. Reported, not changed.
- `unsloth_zoo/mlx/preference.py:298,307`, `completion[0].get("role")`, is indexing a chat message list, where element 0 is a message dict by construction. Different shape, not this bug.

No other `[0].get(` exists in `unsloth_zoo/`.

## Relationship to open PRs

This blocks unslothai/unsloth#11031 and #1233. Both are GRPO changes, so neither can be verified against `main` at all: the run dies in `get_batch_samples` before reaching the code they touch. #11031's verification had to carry an uncommitted local shim in its zoo checkout, spelled `isinstance(batch_samples[0], dict)`, to get a GRPO run to start. Merging this removes the need for that shim, and replaces its `dict` with `Mapping`, which is the spelling `BatchEncoding` requires.


danielhanchen:
@codex review

danielhanchen:
@codex security review

PR #1257, opened 2026-09-16
Fall back to pageable host memory when pinned memory runs out

danielhanchen:
## The defect

`use_gradient_checkpointing = "unsloth"` stages every offloaded activation through a page-locked host buffer. Pinned memory is a **third budget**, separate from both VRAM and system RAM. Exhausting it raises `cudaErrorMemoryAllocation`, which torch reports as a bare `RuntimeError: CUDA error: out of memory` carrying no allocator summary. So the run dies claiming OOM while the card is nearly empty, which is why these reports look impossible.

WSL2 is where it bites. It caps a process at roughly 1-2GB of pinned memory and refuses single pinned allocations past a few MB ([documented limitation](https://docs.nvidia.com/cuda/wsl-user-guide/index.html#known-limitations-for-linux-cuda-applications)).

Four open issues are this one defect. All on WSL, all with `"unsloth"` checkpointing, all with most of the card free, and every one of them worked again with `use_gradient_checkpointing = True`, which does not offload:

| Issue | Card | Reporter's own finding |
| --- | --- | --- |
| #338 | RTX 3090 24GB | `True` fine, `"unsloth"` OOMs; `non_blocking = False` on the host copy fixed it |
| #1552 | Tesla M40 24GB | traced to the same place via #1744 |
| #1744 | RTX 2080 Ti 22GB | bisected to `pin_memory=True` then `x.resize_(new_size)`, and named WSL's pinned-memory limit |
| #1797 | RTX 3090 Ti 24GB | 8.7GB of 24GB in use at the crash; keeping the tensor on GPU instead of CPU fixed it |

## Why the growth path is the one that matters

Three host allocations had no fallback:

1. the 200 startup buffers, so a host that refuses pinning outright failed at `get_peft_model` time
2. the growth allocation for a new buffer slot
3. `resize_` on an already pinned buffer

The third is the one reporters on #1744 and #1797 bisected to. `resize_` reallocates through the tensor's **own** allocator, so growing a pinned buffer issues a fresh, larger `cudaHostAlloc`, and that is the call WSL refuses once a real activation arrives. Confirmed directly:

```
pinned before resize: True  ptr 139710269751296
pinned after  resize: True  ptr 139709615439872   <- reallocated, still pinned
```

The startup buffers are only 128KB and pin fine under a per-allocation cap, so guarding the first allocation alone does not save the run. That is why there is a separate test for each shape.

## The change

Pageable host memory serves every one of these copies and costs only overlap, so each site degrades to it and warns once rather than aborting. `rl_replacements.py` already guards its GRPO offload exactly this way; this brings the gradient-checkpointing path to the same contract.

The `non_blocking` flag now follows the buffer. Only a pinned destination can genuinely overlap, so asking for an async copy into pageable memory just hides a synchronising transfer behind a misleading flag.

`UNSLOTH_DISABLE_PINNED_MEMORY=1` skips pinning up front, for hosts already known to cap it. It follows the existing `UNSLOTH_DISABLE_DOUBLE_BUFFER` precedent in this file.

The OOM check matches on message text and re-raises anything else, so an illegal memory access or a device-side assert still surfaces instead of being quietly swallowed into a pageable fallback.

## Reproduction

Reporters have 16-24GB cards, so nothing reproduces on a 183GB B200 without constraining it. Two independent constraints, both verified to actually refuse an over-budget allocation before any result was trusted:

- **device budget**: `set_per_process_memory_fraction`, set explicitly per device, emulating 24GB
- **pinned host budget**: `cudaHostAlloc` failure injected via `LD_PRELOAD` at WSL's documented limits (2GB cumulative, 8MB per allocation), returning `cudaErrorMemoryAllocation`

Llama-3.1-8B 4-bit, seq 2048, batch 2, 12 steps, seed 3407, identical token budget per config.

| gradient checkpointing | pinned budget | main | peak allocated |
| --- | --- | --- | --- |
| `"unsloth"` | uncapped | ok | 6.740 GB |
| `True` | uncapped | ok | 7.495 GB |
| `False` | uncapped | ok | 7.495 GB |
| `"unsloth"` | WSL limits | **CUDA error: out of memory** | died at 5.80 GB of a 24GB budget |
| `True` | WSL limits | ok | 7.495 GB |
| `False` | WSL limits | ok | 7.495 GB |

On main the failure lands exactly where the reporters said:

```
File "unsloth_zoo/gradient_checkpointing.py", line 810, in forward
    if new_size > x.numel(): x.resize_(new_size)
torch.AcceleratorError: CUDA error: out of memory
```

18.2GB of the emulated 24GB card was free at that moment. This is a **peak**, not a leak: allocated and reserved were logged every step and are flat across steps in every run that survives.

Worth stating plainly: `"unsloth"` is not a memory regression against `True`. It uses **0.755GB less**. The offloading does what it claims; the bug is purely the unguarded host allocation.

With this change, the `"unsloth"` + WSL-limits cell completes all 12 steps.

## Non-regression

Measured, not assumed. Four base runs and both fixed runs on the unaffected configuration:

- **peak memory identical to three decimals**: 6.740GB allocated / 6.793GB reserved, every run
- **loss**: base-against-base spread is 0.0015; base-against-fix is 0.0011, i.e. inside the noise floor measured on this hardware

## Tests

`tests/test_pinned_host_buffer_fallback.py`, 18 tests. All 18 fail on main and pass here.

Each source change is pinned by a named test, verified by reverting them one at a time:

| Reverted change | Tests that go red | Others |
| --- | --- | --- |
| growth-path fallback in `_grow_host_buffer` | `test_grow_host_buffer_falls_back_when_pinned_resize_refuses`, `test_checkpointed_block_survives_a_per_allocation_pinned_cap` | 16 stay green |
| init loop back to raw pinned | `test_initialisation_survives_a_platform_that_refuses_all_pinning` and 2 more | 15 stay green |
| `non_blocking` back to unconditional `True` | `test_pageable_buffers_are_not_copied_with_a_non_blocking_claim` | 17 stay green |

`test_checkpointed_block_survives_a_per_allocation_pinned_cap` is the one shaped like WSL: small pinned allocations succeed, large ones do not, so it isolates the growth path specifically. It asserts the startup buffers did pin, so it cannot pass for the wrong reason.

`test_pageable_buffers_are_not_copied_with_a_non_blocking_claim` asserts the offload branch actually ran before checking anything, so it cannot pass vacuously.

Existing suite: `test_checkpoint_modes.py`, `test_checkpoint_compile_mix.py`, `test_offloaded_checkpoint_simulation.py`, `test_offloaded_checkpoint_rng.py`, `test_disabled_hook_checkpoint_guard.py`, `test_hip_bf16_offload_dtype.py` all unchanged. The 4 failures in `test_checkpoint_compile_mix.py` are present identically on main, before this branch.

## What the emulation cannot answer

Emulating 24GB on a 183GB B200 is not the same as a real 24GB card. Fragmentation behaves differently against a much larger real heap, and `set_per_process_memory_fraction` bounds the caching allocator rather than the physical device, so driver-side and non-PyTorch allocations sit outside the cap. The pinned-memory limit is injected at the CUDA API boundary rather than imposed by WSL's virtual PCIe layer, so it reproduces the failing call and its exact error, not WSL's allocator behaviour around it. This host has no WSL, so the fix is verified against WSL's documented limits rather than on WSL itself.

I am not closing any issue on that basis, but this removes the failing call that every one of those reporters independently landed on.


danielhanchen:
Additional verification.

**Version range.** Tests plus a training run under the emulated WSL pinned limits, at both ends of the supported range:

| | transformers | TRL | torch | PEFT | tests | training |
| --- | --- | --- | --- | --- | --- | --- |
| ceiling | 5.17.0 | 1.14.0.dev0 | 2.14.0+cu130 | 0.21.0 | 18/18 pass | 8/8 steps, peak 6.740 GB |
| floor | 4.57.6 | 0.22.2 | 2.6.0+cu124 | 0.18.0 | 15/18 pass | see below |

The 3 floor failures and the floor training run all die with `CUDA error: no kernel image is available for execution on the device`. torch 2.6.0+cu124 ships no sm_100 kernels, so it cannot run on this Blackwell host at all; the failure is the host, not this branch. The 15 tests that do not launch a kernel pass on torch 2.6.0, which is what establishes that the API surface used here (`is_pinned`, `resize_`, `pin_memory=`, and `AcceleratorError` subclassing `RuntimeError`) is valid at the floor. **End-to-end compute on torch 2.6.0 is not answerable on this hardware** and I am not claiming it.

**Wiped compile cache.** Re-run with `UNSLOTH_COMPILE_LOCATION` deleted beforehand and overwrite left on, so nothing was inherited from an earlier build. Cache regenerated to 17 modules, and the result is unchanged: 12/12 steps under the WSL pinned limits, peak 6.740 GB allocated / 6.793 GB reserved, identical to base, loss deviation 0.0012 against the 0.0015 base-against-base noise floor.

`UNSLOTH_COMPILE_DISABLE` was used only for diagnosis and is not load-bearing for any number quoted here.

danielhanchen:
@codex review

danielhanchen:
@codex security review

danielhanchen:
Confirmed the failing call is still unguarded on main, where `x.resize_(new_size)` on a pinned buffer in gradient_checkpointing.py re-issues a larger cudaHostAlloc, and the fallback keeps `non_blocking` tied to the buffer. Will get this reviewed.

danielhanchen:
I benchmarked this end to end. Short version: **the normal path is free, the `is_pinned()` call is not free in isolation but is 0.15% of a step so it does not matter, and the fallback path costs 51% of throughput, which is a lot more than "somewhat slower".**

Every number below is measured. Setup, arms and what I could not test are at the bottom.

## Summary

| Question | Answer |
| --- | --- |
| Is `x.is_pinned()` cheap? | No, 5.19 us per call inside the training loop (0.51 us idle), because it is a `cudaPointerGetAttributes` driver query, not a cached flag |
| Does that matter? | No. 120 calls per step, 0.62 ms per step, **0.154% of step time** |
| Normal path (pinned works) overhead | **+1.36%** median, versus **+1.84%** for the negative control. Inside the noise band, so no measurable regression |
| Allocation overhead of `_new_host_buffer` | +0.65 us per call over bare `torch.empty(pin_memory = True)`, paid 200 times at init. ~0.13 ms once per process |
| Fallback path cost | **+51.3% seconds per step**, 2.664 to 1.770 steps/sec |
| Peak memory | Identical to 4 decimal places across all arms |
| New test file | 18 passed on head, 18 fail on base, and the store-back line is covered by a mutation test |

---

## 1. Is `x.is_pinned()` cheap? No, but it does not matter here

`is_pinned()` is not a cached flag. It calls `cudaPointerGetAttributes` on the pointer. I confirmed that by measuring the same call in the same process before and after a CUDA context exists:

| CUDA context | `is_pinned()` on a pageable CPU tensor | n |
| --- | --- | --- |
| not initialised | 204 ns (min 194) | 9 x 100k |
| initialised | 680 ns (min 649) | 9 x 100k |

A 3.3x jump the moment a context exists is the driver round trip. Isolated cost, torch 2.14.0+cu130, B200:

| tensor | numel | median | min | max | n |
| --- | --- | --- | --- | --- | --- |
| pinned | 128Ki | 507 ns | 501 | 550 | 9 x 200k |
| pageable | 128Ki | 628 ns | 602 | 713 | 9 x 200k |
| pinned | 4Mi | 511 ns | 488 | 538 | 9 x 200k |
| pageable | 4Mi | 623 ns | 608 | 679 | 9 x 200k |
| reference: `Tensor.numel()` | | 73 ns | | | 9 x 200k |
| reference: `Tensor.is_cuda` | | 47 ns | | | 9 x 200k |

So it is 7 to 13x a real cached attribute read, and the `False` answer is the slower one (the pageable path has to clear `cudaErrorInvalidValue`). Cost does not scale with buffer size, as expected for a pointer query.

### Real call count, instrumented rather than estimated

I built a counting twin of the head branch that wraps both call sites and records count and time. Llama-3.2-1B (16 layers, 15 checkpointed), batch 2, seq 2048, grad accum 4, 30 steps:

| run | steps | offload calls | reload calls | calls/step | ns/call in situ | ms/step | step (s) | % of step |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| 0 | 30 | 1801 | 1801 | 120.1 | 5262 | 0.632 | 0.3946 | 0.160% |
| 1 | 30 | 1801 | 1801 | 120.1 | 5509 | 0.661 | 0.4245 | 0.156% |
| 2 | 30 | 1801 | 1801 | 120.1 | 5109 | 0.613 | 0.4075 | 0.151% |
| 3 | 30 | 1801 | 1801 | 120.1 | 4964 | 0.596 | 0.3900 | 0.153% |
| **median** | | | | **120.1** | **5186** | **0.623** | | **0.154%** |

Two things worth flagging. The count is exactly what the PR description implies, 60 offload plus 60 reload per step (4 micro batches x 15 layers x 2). And the in-situ cost is 5.19 us, ten times the 0.51 us I measure on an idle process, because the driver query contends with the kernel launches and async copies already in flight. That is the honest number to use, not the microbenchmark.

0.154% of a step. It is real, it is measurable, and it is far below the run-to-run noise on any shared box.

### I measured the hoist fix anyway

I built the obvious fix as a fourth arm: a `CPU_BUFFERS_PINNED` list parallel to `CPU_BUFFERS`, written at allocation and at grow, read by index at both copy sites. Both sites already have `CPU_INDEX` in hand, so it is a clean change:

```python
x.copy_(arg, non_blocking = CPU_BUFFERS_PINNED[CPU_INDEX])
...
buffer.copy_(x, non_blocking = CPU_BUFFERS_PINNED[CPU_INDEX])
```

End to end it buys nothing measurable (`hoist` +1.51% vs `head` +1.36% vs base, both inside a +1.84% control band). That is the expected result given 0.154%. I would still take it, since it is strictly cheaper and removes a driver call from the hot path for free, but it is a tidiness change, not a perf fix. Not a blocker either way.

Also worth noting the enqueue measurement, which is why the end-to-end effect is invisible: at these sizes the copy stream is the bottleneck, so the extra CPU microsecond is absorbed.

| size | `non_blocking = True` | `= x.is_pinned()` | `= hoisted flag` | n |
| 128Ki D2H enqueue | 4028 ns | 4033 ns | 4032 ns | 9 x 2000 |
| 4Mi D2H enqueue | 73421 ns | 73399 ns | 73403 ns | 9 x 2000 |

---

## 2. Normal path overhead: none I can measure

Real LoRA fine-tune, `use_gradient_checkpointing = "unsloth"`, Llama-3.2-1B-Instruct 4bit, seq 2048, batch 2, grad accum 4, 30 steps per run with the first 10 discarded. Fixed-length synthetic corpus so every step does byte-identical work, which cut step-time spread from 13% to 3%.

Five arms, 6 rounds, arm order rotated every round, never blocked. `ctl` is a second independent checkout of the merge base through the identical harness.

| arm | runs | steps | median s/step | per-run medians | spread | steps/s | peak alloc GB |
| --- | --- | --- | --- | --- | --- | --- | --- |
| base | 6 | 120 | 0.3754 | 0.3676 - 0.3851 | 4.78% | 2.664 | 1.660 |
| ctl (base again) | 6 | 120 | 0.3843 | 0.3750 - 0.4168 | 11.14% | 2.602 | 1.660 |
| head | 6 | 120 | 0.3816 | 0.3659 - 0.3905 | 6.71% | 2.621 | 1.660 |
| hoist | 6 | 120 | 0.3797 | 0.3677 - 0.3907 | 6.25% | 2.633 | 1.660 |
| head, `UNSLOTH_DISABLE_PINNED_MEMORY=1` | 6 | 120 | 0.5650 | 0.5601 - 0.5868 | 4.78% | 1.770 | 1.660 |

Per-round paired deltas against base, same round meaning same box conditions:

| round | ctl | head | hoist | fallback |
| 0 | +1.40% | +1.34% | +2.13% | +53.42% |
| 1 | +5.02% | +3.34% | +4.15% | +53.05% |
| 2 | -0.68% | +1.51% | +1.56% | +48.34% |
| 3 | +1.08% | +0.64% | +0.55% | +51.40% |
| 4 | +2.28% | +1.38% | +1.45% | +46.05% |
| 5 | +11.10% | -2.45% | -1.97% | +51.16% |
| **median** | **+1.84%** | **+1.36%** | **+1.51%** | **+51.28%** |

**head is +1.36% against a control that is itself +1.84%.** The PR arm is closer to base than a byte-identical copy of base is. There is no normal-path regression to find at this resolution. Round 5 shows why the control matters: `ctl` drifted +11.10% that round purely from neighbour load, and a block design would have pinned that on whichever arm happened to run then.

Peak allocated memory 1.6601 GB and peak reserved 1.6836 GB, identical across all five arms. Final loss agrees to 5 decimal places (10.97284 to 10.97286), so nothing numerically moved.

### Allocation-time overhead

`_new_host_buffer` versus the bare `torch.empty(..., pin_memory = True)` it replaces, n = 15 each:

| numel | variant | median | min |
| --- | --- | --- | --- |
| 128Ki | `torch.empty(pin_memory = True)` | 1.09 us | 0.82 |
| 128Ki | `_new_host_buffer` | 1.74 us | 1.49 |
| 128Ki | `torch.empty` pageable | 0.83 us | 0.73 |
| 8Mi | `torch.empty(pin_memory = True)` | 1.13 us | 0.88 |
| 8Mi | `_new_host_buffer` | 1.69 us | 1.51 |
| 8Mi | `torch.empty` pageable | 1.07 us | 0.96 |
| 128Ki | alloc + `resize_` (base) | 6.39 us | 6.12 |
| 128Ki | `_new_host_buffer` + `_grow_host_buffer` | 7.21 us | 7.00 |

+0.65 us per allocation, mostly the `os.environ.get` in `_pinned_memory_disabled()` plus the try block. `initialize_unsloth_gradient_checkpointing` does 200 of these, so 0.13 ms once per process. If anyone cares, reading the env var once at import instead of per allocation removes most of it, but this is not worth a round trip.

---

## 3. The fallback path: +51%, and it is mostly lost overlap, not lost bandwidth

Raw copy bandwidth on this box, bf16, n = 11 x 20 per cell:

| bytes | dir | pinned GB/s | pageable GB/s | ratio |
| 0.25 MiB | D2H | 30.14 | 6.64 | 4.54x |
| 0.25 MiB | H2D | 29.77 | 8.33 | 3.57x |
| 2 MiB | D2H | 51.20 | 13.25 | 3.86x |
| 2 MiB | H2D | 49.63 | 13.94 | 3.56x |
| **16 MiB** | **D2H** | **56.34** | **16.62** | **3.39x** |
| **16 MiB** | **H2D** | **54.63** | **16.77** | **3.26x** |
| 64 MiB | D2H | 56.95 | 18.08 | 3.15x |
| 64 MiB | H2D | 55.32 | 17.34 | 3.19x |

I instrumented the offload site to record the real shapes rather than guessing. This config offloads exactly one tensor size, `(2, 2048, 2048)` bf16 = 8,388,608 elements = **16 MiB**, 360 times per 6 steps. So the 16 MiB row is the one that applies, and the host buffers do get grown from the 128Ki initial size to 8Mi during training (they are shrunk back by the teardown path, which is why a naive post-run inspection shows 131072).

### Which dominates

Per step: 60 x 16 MiB D2H + 60 x 16 MiB H2D = 1.0066 GB in each direction.

| | pinned | pageable | delta |
| --- | --- | --- | --- |
| DMA time per step, both directions | 36.3 ms | 120.6 ms | **84.3 ms** |
| measured step time | 375.4 ms | 565.0 ms | **189.6 ms** |

Raw bandwidth accounts for 84.3 of the 189.6 ms, so **44% bandwidth, 56% lost overlap**. Losing the overlap dominates. That makes sense: with `non_blocking = False` the host thread blocks until each copy completes, so the offload leaves `EXTRA_STREAM` and lands squarely in the critical path, and the CPU can no longer run ahead to queue the next layer's kernels. In the pinned case those same 36.3 ms are almost entirely hidden behind compute.

So the PR's warning text is directionally right but understates it. I would change "somewhat slower" to a real number. Suggested wording: "each step is roughly 1.5x slower". That is what a user who hits this will actually see, and it is much better than the alternative, which today is a bare `CUDA error: out of memory` on an empty GPU.

---

## 4. Correctness findings

Reporting these, not fixing them.

**(a) The `PINNED_MEMORY_AVAILABLE` latch across two models in one process.** The re-probe is fine, but `_WARNED_ABOUT_PINNED_MEMORY` is never reset alongside it, so the second model's fallback is silent:

```
model A: latch after fallback = False | warned = True
model B: latch after fallback = False | warned AGAIN = False
```

A user loading a second model in the same process gets the slow path with no explanation, and if they only saw the warning once they will reasonably assume it applied only to model A. Resetting `_WARNED_ABOUT_PINNED_MEMORY = False` next to `PINNED_MEMORY_AVAILABLE = True` in `initialize_unsloth_gradient_checkpointing` would fix it. Minor, but it is a one-line fix in a block you are already touching.

**(b) Threads.** 8 threads x 200 concurrent `_new_host_buffer` calls with pinning failing: no exceptions, all 1600 results consistently pageable, warning printed exactly once, latch settled on `False`. The globals are written without a lock, but every write is an idempotent store of the same value and the GIL makes the `if _WARNED...: return` check atomic enough in practice. I could not make it misbehave. Not a concern.

**(c) The store-back is genuinely covered.** I built a mutant of the head branch with `CPU_BUFFERS[CPU_INDEX] = x` deleted and ran the new test file against it:

```
1 failed, 17 passed
FAILED test_checkpointed_block_survives_a_per_allocation_pinned_cap
```

Exactly one test fails, and it is the right one. That is the test you want. Separately I verified on the live slot that `_grow_host_buffer` returns a different object on fallback (`newb is not b0` = True, pinned = False, numel 8192), and the backward path reads `CPU_BUFFERS[CPU_INDEX]` fresh rather than closing over the old reference, so the store-back is load bearing and it works.

**(d) Mixed pinned and pageable buffers after a mid-run fallback.** Three pinned slots then three pageable slots in one `CPU_BUFFERS` list, `[True, True, True, False, False, False]`, each copied with its own `non_blocking` and compared after a sync. All six round trip bit-exact. No issue.

---

## 5. The new test file

`tests/test_pinned_host_buffer_fallback.py` on head: **18 passed** in 0.41s. Same file against the merge base: **18 failed**, three of which are the behavioural ones rather than import failures (`test_checkpointed_block_survives_pinned_exhaustion_and_keeps_gradients`, `test_checkpointed_block_survives_a_per_allocation_pinned_cap`, `test_pageable_buffers_are_not_copied_with_a_non_blocking_claim`). Plus the mutation result in 4(c). The tests discriminate.

---

## Method and limits

Three independent worktrees off one clone: head `ada26ed`, base `2dcbe9c` (the merge base), and a second independent checkout of `2dcbe9c` as a negative control through the identical harness. Plus two instrumented twins of head (counting, hoisted flag) and one mutant. Arms were interleaved in 6 rounds with the order rotated each round, never run as blocks, because this box is shared and loaded.

Box: NVIDIA B200, torch 2.14.0+cu130, CUDA 13.0, transformers 4.57.6, trl 0.24.0, unsloth 2026.9.4, Linux. One GPU for the training arms, a second for the microbenchmarks.

What I could not measure:

- **WSL2 itself, which is the environment this PR exists for.** This host is native Linux. I could never trigger a genuine `cudaHostAlloc` failure, so every fallback measurement here is driven either by `UNSLOTH_DISABLE_PINNED_MEMORY=1` or by injected allocator failures. The recovery logic is right and the tests cover it, but nobody in this thread has yet seen the real WSL2 allocator take the real path. That is worth one confirmation from an affected reporter before merge.
- **The `_is_host_alloc_oom` string matching against a real WSL2 error message.** The tests use the four message shapes we expect. If WSL2 phrases it differently the fallback will not engage and the user still gets the OOM. Same ask, one real repro.
- Multi-GPU, larger models, and longer sequences. Everything above is single GPU on a 1B model. The `is_pinned()` cost scales linearly with layers x grad accum, but so does step time, so I would expect the 0.154% fraction to hold.

danielhanchen:
@codex review

danielhanchen:
## Follow-up: the driver query is gone from the hot path

`5fa130e2c6f3d6d7deb95c50db6b1a8a026f537b`. This supersedes the `is_pinned()` section of my
previous comment, which measured the cost as live. It no longer is.

### What changed

`non_blocking = x.is_pinned()` ran on every offload and every reload copy. That is not a cached
flag, it is a `cudaPointerGetAttributes` driver query: 554.8 ns on a pinned tensor, 670.0 ns on a
pageable one, against 68.8 ns for `is_cuda` as a reference. Counted rather than estimated, it ran
112 times per step in a 15 layer, 4 micro batch loop.

The pinned-ness of a slot only changes where the buffer is created or grown, so it is now cached
once at those two points and read in the hot path.

| | calls to `is_pinned()` per step |
|---|---|
| previous head | 112.0 |
| second independent checkout of that head, as a control | 112.0 |
| this head | **0.0** |

| read | ns |
|---|---|
| `is_pinned()` on a pinned tensor | 554.8 |
| `is_pinned()` on a pageable tensor | 670.0 |
| the cached attribute | **54.1** |

### Why an attribute on the tensor rather than a list beside CPU_BUFFERS

The obvious implementation is a `CPU_BUFFERS_PINNED` list indexed by `CPU_INDEX`, and both call
sites already have that index. It was rejected, because the failure that matters here is a flag
reading pinned while the tensor is pageable, which issues an async copy from pageable memory. That
failure is only reachable with a parallel list: an index-keyed list can drift in length or content
relative to the list it describes, while an attribute travels with the object and cannot come to
describe a different slot. The default also fails in the safe direction, since any tensor that
reaches `CPU_BUFFERS` without going through the two helpers reads False and gets a synchronous copy,
which is slower and never wrong. Nine `CPU_BUFFERS` mutation sites were enumerated and checked
against that rule rather than trusting the two obvious ones.

Two things had to be established by experiment rather than assumed. A view does not carry the
attribute, since `x[:new_size].view(shape)` is a fresh tensor object, so the flag is read off the
slot before the slice. And in place `resize_` does preserve object identity and pinned-ness across
both grow and shrink, so the attribute survives the reset path; it is re-marked there anyway rather
than relying on that.

Dropping the parentheses instead, `non_blocking = x.is_pinned`, is not a shortcut to the same place.
`Tensor.is_pinned` is a method, so that expression is a bound method object and always truthy. torch
happens to reject it with `TypeError: copy_(): argument 'non_blocking' must be bool, not
builtin_function_or_method` rather than silently going async, but a test now pins the type either way.

### Controls

Three mutations, each red for its own reason:

- the fallback return not re-marked: `AttributeError: 'Tensor' object has no attribute '_unsloth_is_pinned'`
- a pageable replacement inheriting a stale True, which is the dangerous direction: `assert True is False`
- the bound-method slip above, which takes out four tests

Plus an in-test control that deliberately desyncs slot 0 and asserts the guard fires.

### End to end

Not resolvable, and I am not claiming it. Three arms, 6 rounds, order rotated, 30 timed steps per
run: 26.319 ms, 25.312 ms and 25.986 ms for previous head, this head, and the control. The two
identical head checkouts differ from each other by 0.333 ms and per-arm stdev is about 1.2 ms, so the
apparent 0.67 to 1.01 ms gain sits inside the noise.

An amplification fit is what puts a number on it: injecting 8 extra `is_pinned()` calls per copy site
costs 0.765 ms/step, so about 853 ns per call in situ on this harness, and about 0.096 ms/step for
the 112 calls the previous head paid, or 0.38% of a 25.3 ms step. On the real model the in situ cost
measured higher, 5186 ns per call, because the driver query contends with in-flight launches, so the
saving there is larger than 0.38%, still well under 1%. The honest summary is that this removes 112
driver round trips per step and the benefit is real but below what this box can resolve end to end.

### Also fixed

`initialize_unsloth_gradient_checkpointing` re-probed `PINNED_MEMORY_AVAILABLE = True` without
clearing `_WARNED_ABOUT_PINNED_MEMORY`, so a second model in the same process fell back to pageable
host memory with no message at all. One line, covered by
`test_a_second_model_in_the_same_process_still_warns`.

### Tests

`tests/test_pinned_host_buffer_fallback.py` 18 to 26 passed. Gradient checkpointing sweep, 8 files:
130 passed, 2 skipped. `tests/test_checkpoint_compile_mix.py` has 4 failures that reproduce
identically on an untouched checkout of `ada26ed`, so they are pre-existing on this torch 2.14 box.

### Still not answered

WSL2, which is the whole point of the PR. This host is native Linux, so `_is_host_alloc_oom` has
never been matched against a real WSL2 allocator error string. If WSL2 phrases it differently the
fallback will not engage and the user still gets the OOM. That is worth one confirmation from an
affected reporter before merge.

PR #1259, opened 2026-09-16
Stop the compiled trainers' empty-logits sentinel from claiming protocol dunders

danielhanchen:
Twin of unslothai/unsloth#11130, for the copy of the sentinel that lives in `compiler._cross_entropy_code` and is written into `unsloth_compiled_cache`. Either copy alone leaves the failure reachable through the other.

`EmptyLogits.__getattr__` answers every name with a callable, so `hasattr()` on the sentinel was true for every name too. `torch.distributed.utils._apply_to_tensors` does this:

```python
elif hasattr(x, "__dataclass_fields__"):
    dc = dataclasses.replace(x)
```

FSDP2's mixed-precision output cast (`_fsdp_state._cast_output_dtype`) walks the model output with that helper on every post-forward hook, so the sentinel said yes and the step died with `TypeError: replace() should be called on dataclass instances`.

## Measured, 2 GPUs

`accelerate launch --config_file <fsdp2.yaml>` on a Qwen2.5-0.5B LoRA SFT, 4 steps: before, every step died in that TypeError; after, both ranks log `[3.5781, 3.9453, 3.5156, 3.4297]`.

## What is deliberately unchanged

- `.to(device)` is still the no-op accelerate needs, special-cased first.
- `EMPTY_LOGITS.shape` still returns the callable that raises the `UNSLOTH_RETURN_LOGITS` message.
- Tensor dunders are bound as real instance attributes by the loop under `EMPTY_LOGITS`, so they never reach `__getattr__`.
- `__eq__`, `__repr__` and `__reduce__` are found by the normal lookup, which runs before `__getattr__`.

## Test

`tests/test_empty_logits_protocol_probes.py` execs the emitted definition the way a generated module does, then makes the exact probe torch makes. Confirmed red by restoring the one-line `__getattr__` (7 failures), green with the fix. No GPU, no launcher.


danielhanchen:
@codex review

danielhanchen:
@codex security review

danielhanchen:
Confirmed the sentinel in unsloth_zoo/compiler.py:2463 still answers hasattr for __dataclass_fields__, which is exactly what torch.distributed's output cast probes before calling dataclasses.replace. Will get this reviewed alongside the unsloth-side twin.

danielhanchen:
@codex review

PR #1260, opened 2026-09-16
Lift the TRL ceiling to 1.13.0 so the declared window can resolve its own guards

danielhanchen:
## Summary

This repository's `pyproject.toml` caps TRL at `<=0.24.0` at lines 31 and 108. That number was set on 2025-11-03 in 72262f3, when 0.24.0 was the newest release on PyPI. PyPI's latest today is 1.13.0, thirty releases later.

**The cap makes a real guard unreachable, not merely dead code.** `unsloth/models/rl_replacements.py` gates `vllm_generation_init_patch` on:

```python
if Version(importlib_version("trl")) < Version("0.28.0"):
    return
```

That patch is what stops a second vLLM instance being built when `fast_inference` already has one. Under a `<=0.24.0` window, 0.28.0 is not resolvable, so on **every configuration a user can actually install** the guard returns immediately and the second instance gets built. The sibling gate on `openenv_vllm_reload_weights` (`>= 0.26.0`, `rl_replacements.py:2888`) is unreachable in exactly the same way.

`unsloth` depends on `unsloth_zoo`, and pip resolves the intersection of the two windows. So this file, not unsloth's, is what decides which TRL a user gets, and lifting the cap in unsloth alone changes nothing about what gets installed.

## Pairs with unslothai/unsloth#11137

**These two must land together.** unslothai/unsloth#11137 lifts the same cap on the unsloth side and carries the evidence table, the full cross-repo site list and a new `tests/test_trl_cap_sites.py` whose load-bearing assertion is precisely "no runtime version guard is unreachable through the declared window". Either PR alone leaves the resolved intersection at 0.24.0 and changes only what the package advertises.

## Evidence

Five TRL versions, one venv each, CPU torch 2.10.0+cpu, transformers 5.17.0, Python 3.12, `PYTHONPATH=.` from the repo root with `UNSLOTH_IS_PRESENT=1`, `UNSLOTH_COMPILE_DISABLE=1`, `TORCHDYNAMO_DISABLE=1`, `TORCH_COMPILE_DISABLE=1`. Suite is this repo's TRL-touching CPU gates: `test_loss_normalization_contract.py`, `test_upstream_pinned_symbols_trl_vllm.py`, `test_upstream_import_fixes_drift.py`, `test_zoo_history_regressions_deep.py`, `test_upstream_pinned_symbols_accelerator.py`.

| TRL | resolves | result | named failure |
|---|---|---|---|
| 0.22.2 | yes | pass (149 passed, 25 skipped) | none |
| 0.24.0 | yes | pass (149 passed, 25 skipped) | none |
| 1.0.0 | yes | pass (149 passed, 25 skipped) | none |
| 1.7.0 | yes | pass (150 passed, 24 skipped) | none |
| 1.13.0 | yes | pass (150 passed, 24 skipped) | none |

Separately, `test_pypi_version_sync.py`, `test_wheel_top_level_packages.py`, `test_python39_compatibility.py`, `test_trl_entropy_metric.py`, `test_trl_sft_logits_metrics.py` and `test_pip_cache_actions.py` on TRL 1.13.0: 125 passed, 3 skipped.

**The extra test is the point.** The only difference between the 0.24.0 and 1.13.0 skip sets is one line:

```
SKIPPED [1] tests/test_loss_normalization_contract.py:115: this TRL has no chunked CE
```

Under the shipped cap that test is skipped on every pull request, because no TRL the window admits has chunked cross-entropy. Under the lift it runs and passes. Reverting the cap is therefore its own mutation test: the coverage this repository already wrote goes from exercised back to skipped.

### Coverage limits

Five points were measured, not a range. The window admits 0.25.0 through 1.13.0; twenty-five of those releases were not individually run. `NEWLY_ADMITTED` in unslothai/unsloth#11137's new `tests/test_trl_cap_sites.py` names only the five 1.x releases that were actually swept there (0.29.1, 1.0.0, 1.6.0, 1.7.0, 1.13.0), so a later specifier rewrite cannot quietly claim more than was measured. A per-minor pass is not "every patch release passes".

## Changes

`pyproject.toml` lines 31 and 108: `<=0.24.0` to `<=1.13.0`. `!=0.19.0` kept at both, and the `sys_platform`/`platform_machine` marker on line 31 is untouched.

`.github/workflows/consolidated-tests-ci.yml`: one comment. It read "`.[core]` resolves trl 0.24.0 here, which masks `labels[position_ids == 0]`", as the reason the `test_loss_normalization_contract` boundary count is exercised on every PR. This commit moves the number that comment quoted, so it now states the mechanism instead: `.[core]` resolves a TRL that masks the document boundary itself, which TRL added in 0.23.1 and every release since keeps. Verified by the skip-set diff above, which shows nothing in that file became skipped at 1.13.0.

## What this PR does not change

`.github/workflows/consolidated-tests-ci.yml:947-948` carries `trl>=0.18.2,<1.0.0` and `trl>=1,<2` as a deliberate two-lane split across both majors. Neither is a cap that drifted, and both are left alone.

The `CEILING = Version("5.5.0")` policy gate at line 607 of the same file is the transformers ceiling, not a TRL mirror, and is out of scope here.

No new test file is added on this side. The cross-repo cap-site gate lives in unslothai/unsloth#11137, which is where the unreachable-guard assertion reads `rl_replacements.py`.

## A second, independent cap fix in this same file

unslothai/unsloth-zoo#1263 raises the transformers FLOOR in this pyproject, at lines 55 and 100, for a different root cause: peft declares a bare `transformers` requirement with no floor of its own, so `transformers>=4.51.3` admits a resolve where `peft/tuners/lora/model.py` imports `transformers.modeling_layers`, which does not exist before 4.52.0. That pair satisfies every declared constraint and then fails at import.

The two changes touch different lines of the same file and are kept apart on purpose: this one is the trl ceiling at lines 31 and 108, #1263 is the transformers floor at 55 and 100. Whichever lands second rebases. They are independent and neither blocks the other.


danielhanchen:
@codex review

PR #1262, opened 2026-09-16
Keep GRPO KL metrics finite for fully masked completions

vineethsaivs:
Fully masked completions make the mean KL metric NaN, even when the training loss and gradients are finite.

The metric divides by each completion mask count without the zero-count guard already used for the loss. Use the same clamped denominator for the per-completion KL metric, so fully masked rows contribute zero.

Test: `UNSLOTH_ZOO_DISABLE_GPU_INIT=1 PYTHONPATH=. python -m pytest tests/test_rl_replacements_cpu.py -q`. Four regressions fail before. All 87 CPU tests pass after, including mixed and fully masked batches with KL enabled and disabled. Pinned Ruff gate and Python compilation checks pass. Native PyTorch CPU loss and gradient checks; no full language-model training or GPU run.


danielhanchen:
Confirmed the metric path in unsloth_zoo/rl_replacements.py still divides by the raw per-row mask count while the loss above it already clamps, so a fully masked completion logs NaN KL. Will get this reviewed.

vineethsaivs:
Rebased onto main in 76f6d13; it had drifted into a conflict on the CPU test file where both branches appended tests. Suite is 96 passed, 1 skipped.

PR #1263, opened 2026-09-16
Raise the transformers floor to 4.52.4, the oldest one peft 0.18.0 can import

danielhanchen:
## Summary

`pyproject.toml` declared `transformers>=4.51.3` alongside `peft>=0.18.0`, and that pair
cannot import. The floor rises to 4.52.4, which is the oldest transformers that actually
works with the peft floor declared in the same file.

## Why the two bounds contradict each other

* peft declares **no transformers floor of its own**: its metadata on 0.18.0 through 0.21.0
  requires a bare `transformers`, unbounded. peft delegates that floor to whoever depends
  on it, and this file never supplied one.
* peft 0.18.0 imports `GradientCheckpointingLayer` from `transformers.modeling_layers` at
  `peft/tuners/lora/model.py:26`, unconditionally.
* `src/transformers/modeling_layers.py` does not exist before transformers 4.52.0 (404 at
  `v4.51.3`, present at `v4.52.0`).

So a resolve could satisfy every declared constraint with transformers 4.51.3 and then fail
at import:

```
unsloth_zoo/saving_utils.py:96: from peft import PeftModelForCausalLM, PeftModel
  -> peft/tuners/lora/model.py:26: from transformers.modeling_layers import GradientCheckpointingLayer
ModuleNotFoundError: No module named 'transformers.modeling_layers'
```

`unsloth_zoo`'s own use of that module (`temporary_patches/misc.py`) is inside a
`try/except`, so it is not involved. The break is the declared bound.

## Measured, not inferred

In an environment holding the floor pin set (python 3.10, torch 2.6.0, peft 0.18.0,
trl 0.18.2, triton 3.2.0), with only transformers varied:

| transformers | zoo/unsloth import set |
| --- | --- |
| 4.51.3 | 5 failed, 21 passed |
| 4.52.0 | 34 passed |
| 4.52.4 | 34 passed |
| 4.53.1 | 34 passed |
| 4.54.1 | 34 passed |

Everything passes from 4.52.0 on, so `modeling_layers` was the only missing piece rather
than the first of several. The floor is 4.52.4 and not 4.52.0 because 4.52.0 through 4.52.3
were already rejected by name in this same window and were never candidates; those four
exclusions are now dominated by the floor and drop out of the specifier. The ceiling,
`<=5.5.0`, is untouched.

## Related

* unslothai/unsloth#11016 makes the identical change to unsloth's own `pyproject.toml` and
  adds the drift gate that pins the floor to the CI lanes that run it. That gate is what
  found this: the lane failed on its first run.
* unslothai/unsloth-zoo#1260 moves the **trl** bound in this same file (lines 31 and 108).
  This PR touches only the transformers requirement (lines 55 and 100), so the two conflict
  trivially at worst. If #1260 lands first, this branch rebases onto it.


danielhanchen:
@codex review

danielhanchen:
@codex security review

PR #1264, opened 2026-09-17
perf(mlx): reduce CCE backward temporary storage

Lyxot:
The runtime CCE backward held two temporaries wider than the values they carry. A bfloat16 logits chunk got a float32 `d_logits` output, and a trainable head's gradient accumulated into a full-vocabulary float32 buffer that was cast to the weight dtype once at the end. Under a compiled training step both sit next to the logits chunk, so for a half-precision model they set more of the CCE backward peak than the values need.

### What changed

- `refactor(mlx): template CCE dlogits kernel output type`: the dlogits Metal kernel assigned float expressions straight into its output buffer, so it only compiled for float16 and float32 outputs. Each write now casts to a templated output type `O`, passed at both call sites. Output dtypes are unchanged in this commit, so it compiles to the same kernels.
- `perf(mlx): reduce CCE backward temporary storage`: `d_logits` is stored in the logits dtype, and the classifier gradient is built in the weight's own dtype. Each chunk's GEMM still runs in float32 (`d_logits.T @ hidden` upcast) and is rounded once when written into its disjoint vocabulary slice, which is exactly the rounding the final cast used to do, just per chunk. The hidden-gradient path already cast `d_logits` to the hidden dtype before its GEMM, so hidden gradients are bitwise unchanged. The float32 hidden copy is also hoisted out of the chunk loop.

### Impact and risks

- Quantized heads (QLoRA): the half-width `d_logits` halves that temporary, which lowers the step peak of real 4-bit Qwen3.5 models by 20–26% (below). With a frozen bf16 head the narrowed buffer is not what sets the step peak, and memory is unchanged.
- Trainable heads with bf16 weights (bf16 full fine-tuning): the float32 vocabulary buffer is gone. The default MLX full fine-tuning upcasts trainable weights to float32 (`float32_mixed_precision=True`), where logits and weights are already float32 and nothing changes.
- Numerics: hidden gradients are bitwise identical; the bf16 weight gradient is rounded per chunk rather than after accumulation, which is the same value because chunks do not overlap. `d_logits` itself is now bf16 for bf16 logits before being upcast for the weight GEMM, which is the only place precision can differ; loss curves below track main within run-to-run noise.
- No API, config or dependency changes.

### Performance

All runs use gradient checkpointing and a compiled training step, AdamW, on an Apple M3 Max (128 GB), MLX 0.32.2, MLX-LM 0.31.3, MLX-VLM 0.6.17. Baseline is `main` at the merge base.

**Real models**: `mlx-community` checkpoints through `FastMLXModel` + `MLXTrainer` with trainer defaults (`use_cce`, gradient checkpointing, compile), batch 2, 2048 tokens, LoRA r=16. `main` and the branch ran back to back for each model with a 45 s cool-down before every run and no thermal or performance warning recorded. Step time is the median over 6 measured steps after 3 warm-up steps; peak is `mx.get_peak_memory()` over the measured steps, which repeated exactly across runs. The 4-bit Qwen3.5 rows are the median of two runs, the others a single run.

| Model | Peak main → branch | Step time main → branch |
|---|---:|---:|
| Qwen3.5-2B-4bit, QLoRA (4-bit tied head, 248k vocab) | 8.06 → 6.00 GB (**−26%**) | 5.19 → 5.16 s |
| Qwen3.5-4B-4bit, QLoRA (4-bit tied head, 248k vocab) | 10.14 → 8.08 GB (**−20%**) | 12.22 → 12.37 s |
| Qwen3.5-2B-bf16, LoRA (frozen bf16 head) | 10.55 → 10.55 GB | 4.58 → 4.55 s |
| Qwen3.5-0.8B-bf16, bf16 full fine-tuning (`float32_mixed_precision=False`), 1024 tokens | 9.73 → 9.24 GB (**−5%**) | 1.43 → 1.42 s |
| gemma-4-e4b-it-4bit, QLoRA (4-bit tied head, softcap 30, 262k vocab) | 14.86 → 14.86 GB | 13.81 → 12.78 s (noise, see below) |

Loss curves match main within its own run-to-run spread (Qwen3.5-2B-4bit final step loss: main 4.1497 / 4.1520, branch 4.1495 / 4.1500; Qwen3.5-0.8B full fine-tuning final step: main 3.6121, branch 3.6094). gemma-4-e4b's losses are bitwise identical to main, so that run does not reach the narrowed buffers and its time difference is run-to-run noise.

**Simulated** head-dominated steps: an `mlx_lm` llama (hidden 2048, vocab 151,936, 4 rows × 1025 tokens, bf16) with a compiled `value_and_grad` + AdamW step, one fresh process per measurement, 3 rounds with rotated level order, minimum peak and minimum of per-process medians. Ratios are branch / main.

| Scenario | Peak main → branch | Memory | Time |
|---|---:|---:|---:|
| LoRA, 4-bit head, 2 layers | 3440 → 2269 MB | **0.66x** | 0.98x |
| LoRA, 4-bit head, 16 layers | 3664 → 2493 MB | **0.68x** | 1.00x |
| VLM LoRA, 4-bit head, 6% supervised | 3572 → 2385 MB | **0.67x** | 0.98x |
| Full fine-tuning, bf16 head, 1 layer | 5818 → 5170 MB | **0.89x** | 1.00x |
| LoRA, dense frozen head | 3456 → 3456 MB | 1.00x | 0.99x |
| DPO, text evaluation (paths not using the CCE backward) | unchanged | 1.00x | 0.96x / 1.00x |

DPO and text evaluation first read 1.06x and 1.08x time in the 3-round sweep; a 5-round rerun of just those two gave 0.96x and 1.00x, as expected for paths this change does not reach.

### Validation

```bash
python -m pytest tests/test_mlx_runtime_cce_compile.py -q
```

24 passed on Metal. The new `test_runtime_cce_backward_peak_memory` bounds the compiled backward peak for a 4-bit head (hidden gradient only) and a trainable bf16 head (hidden and weight gradients) over 128 rows × 8192 vocabulary with ignored targets. Both cases fail on `main` (13.26 MiB and 93.76 MiB) and pass here; the existing tests in that file cover losses and gradients across dense, quantized, softcapped, label-smoothed and ignored-target cases.


Lyxot:
@codex review

Lyxot:
@codex review

Lyxot:
Codex review: no major issues found on `6099f8a6f` (run on a staging copy of this PR, since the Codex connector is not responding in this repo).

danielhanchen:
Checked this against the backward in unsloth_zoo/mlx/cce/runtime_cce.py, where d_logits was widened to float32 for bfloat16 logits and the classifier gradient held a full-vocabulary float32 buffer, so the saving on 4-bit heads looks real. Will get this reviewed.

danielhanchen:
Reviewed this. The non Metal half of the change I could verify here directly, on the Linux CPU build of MLX, and it holds up.

Gradients are bitwise identical between the merge base and this head across four configurations, including a ragged vocabulary (8195 with chunk 2048) so the final partial chunk is exercised: dense bf16, dense f32, dense bf16 ragged, and f32 hidden with a bf16 weight. Loss is identical too. So moving the `astype(weight.dtype)` inside the chunk loop is exact, as you say: `get_chunk_plan` builds `range(0, vocab_size, chunk)` and the loop clamps `v_end`, so the slices are disjoint and covering and each `grad_weight` row is written by exactly one `slice_update`, never accumulated into.

Backward peak at the new test's own shapes, `mx.get_peak_memory()` on the CPU allocator, five repeats per arm:

| leg | merge base | this head | the test's budget |
|---|---|---|---|
| dense | 90.79 to 91.79 MiB | 75.29 to 75.79 MiB | 88 MiB |
| quantized | 6.02 to 7.29 MiB | 6.02 to 6.29 MiB | 12 MiB |

Reverting only the production hunk and keeping the test puts the dense leg back to 425.25 MiB on my larger probe, so the saving is the hunk and not something else.

Three things worth your attention.

The description says `d_logits` is stored in the logits dtype. That is true on the quantized branch, where the expression is now `dlogits_out_dtype = logits.dtype`. On the dense trainable branch the expression is unchanged and still reads `mx.float32 if logits.dtype == mx.bfloat16 else logits.dtype`, and that branch is handed `template=[("O", mx.float32)]`, which is a valid but no op narrowing. So on the dense path the memory win comes only from the `grad_weight` dtype and the hoisted `hidden_f32`. It looks deliberate, since on that branch `d_logits` has a second consumer in an fp32 GEMM, but the description reads as though both branches narrowed. Either narrowing the dense branch too or saying why it cannot would settle it.

The quantized leg of the new memory test passes on both arms in my measurements, so as written it is a ceiling rather than a regression guard, and it is within its own effect size of the allocator noise: five identical runs of the same code gave 6.02, 6.02, 6.02, 6.79 and 7.29 MiB, which is 1.27 MiB of jitter against a claimed saving of about 1.3 MiB. The dense leg is the one that discriminates, and it does so cleanly. Consider taking the minimum of N for the quantized leg, or dropping it.

Separately, and not caused by this PR: `tests/test_mlx_runtime_cce_compile.py` is named in no workflow. `grep -rn runtime_cce .github/workflows/` returns nothing, so the whole file, including this new test, runs nowhere in CI, and the only job that touches it is `pytest tests/ --collect-only`. The `mlx` label I added gets the Apple Silicon job running, but that job's explicit file list does not include this one. Adding it there would be worth a small separate change.

I could not execute the templated Metal kernel or the dtype change itself, since this host has no Apple GPU and `dlogits_kernel` is None without Metal. That part is reviewed at source level only.

PR #1265, opened 2026-09-17
perf(mlx): compact supervised CCE rows for finite text, VLM and preference plans

Lyxot:
Most instruction, completion-only, vision and preference corpora supervise a small fraction of the positions in a padded batch, yet the classifier projection ran over every position: prompt tokens, image placeholder spans and padding included. With a 150k–260k vocabulary that projection and its backward are a large part of a training step, and evaluation paid the same dense cost.

This gathers the supervised positions of finite batch plans into a fixed capacity before the projection, for text, VLM and preference (DPO/ORPO) training and for their evaluation. The gather is exact and denominators are still counted over the full mask, so losses and metrics match the dense path up to floating-point rounding.

### What changed

- **Text CCE** (`FiniteTextBatchPlan.configure_cce_compaction` / `prepare_cce_batch`): the plan resolves one capacity per admitted (family, width) shape from the largest supervised count it schedules, rounded up to a multiple of 256, and admits it only when it at least halves the target rows. `prepare_cce_batch` appends flat indices padded with `-1`; `make_cce_loss_fn` accepts them as `cce_indices`, gathers hidden rows and targets and marks padding slots with the ignore index. Because capacity is fixed per shape, the compiled step keeps a single trace.
- **VLM CCE**: widths are not fixed the way a text plan's are, so a batch carries row/column pairs resolved against its label grid. A batch whose supervision overflows the capacity uses the dense projection and is remembered, so stochastic rebuilds of the same slot do not oscillate. Compaction is enabled only when the doubled shape catalog fits the compiled-variant cap, and only single-process. For a quantized head a second 256-row capacity is admitted when the cap carries it, sized so it shares vocabulary chunks with the half capacity.
- **DPO**: a compact projection with a custom backward gathers response rows and projects only those; the backward scatters the compact cotangent back to the full grid before the native projection VJP, so gradient shape and tiling match the dense path. Per-token metrics are scattered back into the label grid. The LoRA reference policy scores through the same compact forward.
- **ORPO**: its NLL covers the whole chosen sequence, so the plan takes the objective kind, starts the chosen branch at the first predicted position and admits a shape only when its capacity is at most a quarter of the padded grid; shapes not that sparse (the common ORPO case) stay dense.
- **Evaluation**: text evaluation batches (including those prepared by `train_on_responses_only`) are built as a finite plan when `use_cce` is set, preference evaluation shares the compact forward, and VLM evaluation compacts each eager batch directly from its labels. Evaluation runs eager, so compacted shapes cost no compiled variants.
- **Trainer wiring**: with `use_cce=True`, DPO and ORPO build `make_dpo_cce_loss_fn` / `make_orpo_cce_loss_fn`. DPO keeps the existing loss when the head is not eligible (vocabulary under 8192, head scale or softcap, a trainable quantized head, no hidden-state backbone); ORPO compacts its native cross-entropy for any head with a vocabulary of at least 8192. Training batches go through `prepare_cce_batch` for all three plan types, and loss calls unpack the batch tuple so the optional indices reach the loss.

### Impact and risks

- Text and preference training and evaluation compact only finite plans; lazy streaming batches keep the dense projection. VLM training compacts plans, VLM evaluation compacts its eager batches.
- Compiled variants: text adds no variants (the indexed shape replaces the dense one per admitted shape). VLM compaction reserves a compacted twin of every planned signature (and a 256-row twin for quantized heads) and is skipped when that exceeds `compile_max_variants`.
- Preference losses scored through the compact projection are computed from gathered rows; the scattered metrics and losses match the dense path to float tolerance (see validation), not bitwise.
- Behavior with `use_cce=False` and heads with a vocabulary under 8192 is unchanged, and multi-process VLM training keeps the dense projection.

### Performance

All runs use gradient checkpointing and a compiled training step, AdamW, on an Apple M3 Max (128 GB), MLX 0.32.2, MLX-LM 0.31.3, MLX-VLM 0.6.17. Baseline is `main` at the merge base.

**Real models**: `mlx-community` checkpoints through `FastMLXModel` + `MLXTrainer` with trainer defaults (`use_cce`, gradient checkpointing, compile), batch 2, 2048 tokens, LoRA r=16. `main` and the branch ran back to back for each model with a 45 s cool-down before every run and no thermal or performance warning recorded. Step time is the median over 6 measured steps after 3 warm-up steps; peak is `mx.get_peak_memory()` over the measured steps, which repeated exactly across runs. Each model is a single run per level. Rows are prompt/completion pairs with about 9% of tokens supervised (389 of 4094 for Qwen3.5, 374 of 4094 for Gemma 4); evaluation is 4 batches of 2 rows, timed on a second pass.

| Model | Train peak main → branch | Step time main → branch | Eval time main → branch | Eval peak |
|---|---:|---:|---:|---:|
| Qwen3.5-2B-4bit, QLoRA (4-bit tied head, 248k vocab) | 7.78 → 4.55 GB (**−42%**) | 5.02 → 4.07 s (**−19%**) | 6.91 → 5.25 s (**−24%**) | 3.28 → 3.28 GB |
| Qwen3.5-4B-4bit, QLoRA (4-bit tied head, 248k vocab) | 9.81 → 7.20 GB (**−27%**) | 12.58 → 11.80 s (**−6%**) | 17.78 → 15.98 s (**−10%**) | 4.82 → 4.81 GB |
| gemma-4-e4b-it-4bit, QLoRA (4-bit tied head, softcap 30, 262k vocab) | 13.47 → 13.47 GB | 11.58 → 11.22 s (−3%) | 15.92 → 15.31 s (−4%) | 7.03 → 7.03 GB |

Gemma 4's training peak is set by its backbone, not the head: on gemma-4-e2b-it-4bit a loss on the same hidden states that never touches the vocabulary peaks at the same 13.09 GB as the CCE loss, so compaction can only save time there. Gemma's bf16 loss curves are deterministic on each level and offset slightly between them (gemma-4-e4b step 2: 11.0333 main, 11.0231 branch; final eval loss 11.2374 on both). A direct gradient check on real models shows this is bf16 rounding, not compaction: in float32 the compacted and dense gradients agree to 1.1e-4 relative error (Gemma 4 E2B) and 4.2e-6 (Qwen3.5-2B), while in bf16 the dense and compacted gradients are the same distance from the float32 gradient (Gemma 0.444 vs 0.446, Qwen3.5 0.0196 vs 0.0190) and differ from each other by less than that (0.008, 0.003). Qwen3.5's loss curves agree within main's own run-to-run spread.

**Simulated**: an `mlx_lm` llama (hidden 2048, vocab 151,936, 2 layers, bf16) with a compiled `value_and_grad` + AdamW step (eager for evaluation, as the trainer runs it), one fresh process per measurement, 3 rounds with rotated level order, minimum peak and minimum of per-process medians. Ratios are branch / main.

| Scenario | Peak main → branch | Memory | Time |
|---|---:|---:|---:|
| Text LoRA, 4 × 1025 tokens, 10% supervised | 3456 → 1057 MB | **0.31x** | **0.38x** |
| VLM LoRA, 4-bit head, 2 × 2049 tokens, 6% supervised | 3572 → 1580 MB | **0.44x** | **0.36x** |
| DPO, 4-bit head, 4 pairs × 1024, half of each in the prompt | 10574 → 8102 MB | **0.77x** | **0.90x** |
| ORPO, 4-bit head, one 1024-token pair padding three 128-token pairs | 10574 → 7222 MB | **0.68x** | 1.00x |
| Text evaluation, 10% supervised | 905 → 791 MB | **0.87x** | **0.29x** |
| VLM evaluation, 4-bit head, 6% supervised | 959 → 738 MB | **0.77x** | **0.27x** |
| Fully supervised text LoRA (dense, 4-bit, 16-layer 4-bit), full fine-tuning, text evaluation | unchanged | 1.00x | 0.99–1.00x |

### Validation

```bash
python -m pytest tests/test_mlx_training_e2e_metal.py -k "compact or pair_ownership or small_capacity" -q
python -m pytest tests/test_mlx_trainer_internals.py::test_response_only_eval_batches_stay_a_finite_plan tests/test_mlx_preference.py::test_the_config_carries_every_objective_field_into_the_run -q
```

35 and 3 passed on Metal. The training tests compare compacted against dense losses and gradients under `mx.compile`, the evaluation tests compare eager totals, and both count projected rows to prove the compact shape is the one that runs:

- text training with one trace across batches, dense and 4-bit heads;
- DPO and ORPO training across dense/4-bit heads, float32/bf16, with and without a LoRA reference policy, including the reference adapter scale being restored;
- VLM training with row/column alignment against a trimmed hidden sequence, and the small-capacity admission, remembered dense fallback and variant-budget rules;
- text, preference (unequal final batch) and VLM evaluation totals and token counts.

The trainer tests cover response-only evaluation batches staying a finite plan and the DPO config reaching `make_dpo_cce_loss_fn` with `use_cce=True`.


Lyxot:
@codex review

Lyxot:
@codex review

Lyxot:
@codex review

Lyxot:
@codex review

Lyxot:
@codex review

Lyxot:
@codex review

Lyxot:
Codex review: no major issues found on `8c41995ad` (run on a staging copy of this PR, since the Codex connector is not responding in this repo).

danielhanchen:
Confirmed main projects every position even for completion-only batches, so the compaction in `unsloth_zoo/mlx/utils.py` lands on the default `use_cce` path, and the reported memory and step-time numbers are the part I want to reproduce. Could you split the preference work (the DPO custom backward and the ORPO admission rule) into its own branch so the text and VLM paths can go in first?

danielhanchen:
Reviewed this on real MLX. The optimization is real and the parity holds; I have three questions about things the description does not claim.

What I verified here, on the Linux CPU build of MLX. All 35 of the compaction tests pass, and reverting the three production files to the merge base while keeping the tests makes all 35 fail, so they are falsifiable. Losses and gradients are bit identical between dense and compacted, not merely close, and `ntoks` is unchanged, so the denominator is unaffected. Memory and time at roughly 10 percent supervision, fresh process per repeat:

| shape | base peak | head peak | base time | head time |
|---|---|---|---|---|
| 4 x 513, d 64, V 8192 | 268.6 MB | 34.9 MB | 2.80 s | 0.31 s |
| 4 x 1025, d 256, V 32768 | 1086.2 MB | 207.6 MB | 71.1 s | 8.73 s |
| 8 x 2049, d 1024, V 32768 | 4560.8 MB | 879.2 MB | 1103.4 s | 121.0 s |

Peak repeated exactly across repeats; the base against itself gives a time noise band of about 10 percent, and these ratios sit far outside it. Running the head with indices deliberately not attached reproduces the base exactly, so the win is the compaction and the dense fallback costs nothing. This is the pure MLX fallback path rather than the Metal kernel, so it confirms direction and rough magnitude, not your specific Metal figures.

I also went after the invariant the denominator depends on, since `ntoks` is counted on the full mask before compaction and that is only correct because the compact set is a superset of the supervised set. I could not break it: 0 violations across 290 executed cases against the real `_vlm_cce_forward`, including labels wider than the input ids, hidden shorter than the label grid so the `[:, :h_seq]` truncation fires, image token masking, prompt masking, and an all padding batch. It rests on every masking step in that branch being monotone, only ever adding the ignore index, plus the truncation being a prefix. That contract is not written down anywhere, and the next person to add a fifth masking step has nothing telling them it must only add. One line of comment at `_compact_vlm_cce_batch` would fix that.

The three questions.

First, `trainer.py` now passes `return_plan=True` unconditionally inside `_prepare_response_labeled_eval_batches`. There is no `use_cce` condition in that function, unlike the eval batch factory further up which reads `batch_factory = _create_text_batch_plan if args.use_cce else create_batches`. The base has three `return_plan` occurrences and this head has four. So a `train_on_responses_only` user running with `use_cce=False` now gets a `FiniteTextBatchPlan` where they used to get a materialized list. Values and iteration order are identical and `len()` still works, so callbacks are fine, but the object handed to `eval_dataloader` is no longer a list and materialization moves into the eval loop. Is that intended? It is the one change here that reaches users who never enable CCE.

Second, the ORPO half. The admission rule counts the chosen branch from step 1 rather than from the prompt boundary, which is right given `_orpo_scores` scores that branch over the wider NLL mask, but it means `count` is about half the padded grid while the ORPO limit is a quarter of it. Across 36 uniform length configurations, pairs in 1, 2, 4 by sequence in 512, 1024, 2048 by prompt fraction in 0.5 to 0.95, DPO admitted 36 and ORPO admitted 0. ORPO only admits on ragged batches, one long row among short ones, which is the shape your own test uses. Since `create_preference_batch_plan` length sorts before batching on the default `dataset_order`, the default path tends to produce exactly the uniform batches that can never admit.

Third, and related: `_preference_forward` still does `logits = _model_logits(model(batch[:, :-1]))` unconditionally, before the `cce_indices` branch, so on the ORPO path the vocabulary wide logits are still materialized and compaction only shrinks the `cross_entropy` input. Measured at V 32768, 4 x 1024: 2080.2 MB dense against 1569.2 MB compacted, so 0.75x, against 0.13x to 0.19x for the text path. That is one logits sized temporary, not the projection. The test asserting `projected == [768]` is true about the cross entropy input rather than about the head projection, so it passes correctly while measuring a narrower quantity than the name suggests. #1278 replaces this path entirely, which is consistent with your plan to rebase, and is also why splitting the preference work out of this PR would leave the text and VLM halves cleanly reviewable.

Two smaller notes. `trainer.py` reads `loss_fn._unsloth_cce_kind` with no default, and only `make_preference_eval_fn` sets that attribute, while the training factories set only `_unsloth_cce_compaction`; every other cce attribute on that line uses a default. And `_unsloth_cce_small_capacity_limit` hand duplicates `max(2048, ceil(V/16))` rather than calling `_resolve_chunk_size`, which already disagrees with it today: the real resolver also clamps to the budget, aligns to 256, takes `min(chunk, vocab)` and uses two bytes per element for bf16 hidden where the copy hardcodes four. Every disagreement I found is conservative, so the failure mode is a missed optimization rather than a wrong result, but on a small memory budget with a large vocabulary the duplicate falls below 256 and the 256 row path can never fire.

I could not execute anything on Metal here, so the `@metal_only` tests are reviewed by running them against the CPU runtime, and the distributed paths are reasoned rather than run.

PR #1268, opened 2026-09-17
Adopt the larger driver total on integrated CUDA devices

danielhanchen:
Defensive hardening for unified-memory CUDA hosts. Framing up front, because it matters for review: this is **not** a fix for an observed bug in this repo. I went looking for one and it is not here.

## What I measured

On an NVIDIA RTX Spark N1X (Blackwell, unified memory, Windows ARM64) with torch 2.14.0+cu134:

```
props.total_memory : 48735117312 = 45.39 GiB
mem_get_info(0)[1] : 48735117312 = 45.39 GiB
is_integrated      : 1
```

Both agree, and both are already the full CUDA budget rather than the 8128 MiB dedicated carve-out that `nvidia-smi` reports on this machine. A grep for `nvml`, `nvidia-smi` and `nvidia_smi` across `unsloth_zoo/` returns zero hits, so this repo never reads the carve-out in the first place.

So there is no before-and-after number to show here. unsloth-zoo was already correct on this hardware.

## Why land it anyway

`props.total_memory` is not guaranteed to be the driver budget on every torch build and driver combination. Where it is the carve-out instead, any code treating it as a VRAM budget silently under-sizes by several multiples. This repo already encodes the correct rule for the ROCm side: compare, and adopt only a LARGER driver total, never adopt it outright. This applies the same rule to the CUDA path.

## The rule

```python
if total <= 0 or not cuda_props_are_integrated(props):
    return total
driver_total = _driver_total_memory(index)
if driver_total is None:
    return total
return max(total, driver_total)
```

Only ever adopts a larger value. Never shrinks a reported total.

## Not regressing anything else

- **Discrete NVIDIA**: gated behind `is_integrated`, so a discrete card returns `props.total_memory` untouched and never even reaches `mem_get_info`. Two tests pin that.
- **ROCm**: excluded by name. clr left `is_integrated` unassigned before ROCm 6.2, so False there means "unknown", not "discrete", and the existing AMD handling is left alone.
- **XPU**: untouched.
- **No new CUDA context**. `mem_get_info` is only reached after `get_device_properties`, which has already flipped `torch.cuda.is_initialized()`. Discrete devices never reach it at all.

## Call sites changed

Two, both of which read `get_device_properties(...).total_memory` as a VRAM budget:

- `temporary_patches/gpt_oss.py:1554`, which drives the 40 GB combo-kernel threshold. A carve-out reading turns combo kernels off on a machine that qualifies.
- `flex_attention/utils.py:56`, which drives the 16 and 24 GB kernel-option thresholds. A carve-out reading picks 32x32 blocks on a 45 GiB device.

In `flex_attention/utils.py` I kept the `min()` over `range(device_count())` rather than switching to a helper that returns 0 on a GPU-less host. `min()` over an empty range raises, and that ValueError is what drives such a host into the `except` at line 353 and sets `HAS_FLEX_ATTENTION = False`. Returning 0 would silently flip a CPU-only host to flex-attention-enabled with 32x32 kernel options.

Deliberately left alone, since they already read the driver directly via `mem_get_info` and are correct: `gradient_checkpointing.py:785`, `device_map_planner.py:978`, `rl_replacements.py:355`, `saving_utils.py:6158`, `tiled_mlp.py:73`, `vllm_utils.py:104-107`. Also left alone because they are not memory budgets: the `gcnArchName` string reads, and `flex_attention_bwd.py:100` which is `shared_memory_per_multiprocessor`, an SM limit.

## Tests

New `tests/test_cuda_unified_total_memory.py`, 19 tests: driver larger adopts (shaped on the real 8128 to 46477 MiB case), driver equal is a no-op, driver smaller does not shrink, driver zero discarded, `mem_get_info` raising degrades to props, `mem_get_info` returning `(None, None)` degrades to props, discrete returns props untouched and asserts `mem_get_info` is never called, a wheel with no `is_integrated` field behaves the same, a HIP build with the flag set is untouched, and unreadable devices answer 0.

Related suites `test_tegra_expandable_segments.py`, `test_temporary_patches_imports.py`, `test_alloc_conf_tegra_matrix.py`: 67 passed, 15 skipped.

Full suite run twice, on this branch and on a pristine `origin/main` worktree, with the failure lists diffed:

```
origin/main : 322 failed, 7053 passed, 539 skipped
branch      : 320 failed, 7076 passed, 539 skipped
only in BASELINE: 2 flaky gloo distributed cases
only in BRANCH  : none
```

Zero new failures. The pre-existing 320 are unrelated (gguf quant validation, MLX trainer internals, merge e2e, SSRF guard, upstream drift checks).

## Related

The user-visible symptom that started this, a `usable_gb=0.3` reading on a machine with a 45.39 GiB budget, does not originate here. It comes from Studio's `nvidia-smi` path, which reports the dedicated carve-out. That is tracked separately against the unsloth repo.


danielhanchen:
## Simulation report

`#1268 | issue: NOT PRESENT IN THIS REPO (hardening) | PR as submitted: DEFECTIVE, 1 fixed by me | merge: YES`

I took this one over and ran it through the full simulation. Pushed one fix to the branch as `362c1ac3`; everything else stands.

### Before and after

There is no before and after, and the PR says so itself: on the reporter's hardware the two figures already agree and this repo never reads NVML. So the question I can actually answer is the one that decides the merge, which is whether this changes anything where it should not.

On the local discrete NVIDIA B200:

```
props.total_memory                191505498112, is_integrated 0
cuda_props_are_integrated(props)  False
cuda_total_memory(0)              191505498112  (== props.total_memory)
flex_attention.utils imports      HAS_FLEX_ATTENTION True, kernel_options None
```

No new import cycle, no behaviour change, `mem_get_info` never reached.

### Enumeration

Full cross product of what `cuda_total_memory` reads: props total (3) x `is_integrated` (absent / 0 / 1) x driver total (unreadable / zero / smaller / equal / larger / not an int) x HIP or CUDA (2) x properties readable (2) = **216 cells**.

| outcome | cells |
| --- | --- |
| props total, unchanged | 107 |
| raised to the driver total | 1 |
| zero, device unreadable | 108 |
| shrank below the props total | 0 |

`mem_get_info` was consulted in **12 of 216** cells, every one of them integrated, non-HIP, with a positive props total. That is the shape a hardening change should have: the context-pinning call is reached rarely, and the answer differs from today's in exactly one cell.

### The one defect, now fixed

`cuda_total_memory` answers 0 for a device whose properties cannot be read. That is right for a caller comparing budgets, and wrong for one of the two call sites this PR changes: `flex_attention/utils.py` reads the value as a VRAM size, and 0 GiB lands in the "16GB or less" arm.

Measured directly, with `get_device_properties` made to raise on a host that reports one device:

```
main    -> RAISES RuntimeError  => lands in the except, HAS_FLEX_ATTENTION = False
branch  -> 0.0                  => HAS_FLEX_ATTENTION stays True, 32x32 kernel options
```

That is precisely the silent flip this PR's own body says it avoided for a GPU-less host. The empty-range case was preserved; the unreadable-device case was not.

`362c1ac3` reads `get_device_properties` at the call site and hands it in through the function's existing `props` parameter, so the failure propagates as it did on main. Three tests lift the real `vram_of_gpu` assignment out of the module source with `ast` and evaluate it, so they exercise the shipped expression rather than a copy and need no GPU: unreadable device raises, no devices raises, a readable discrete device still reports its properties total. **On the PR as submitted 2 of those 3 fail.** Whole file: 22 passed.

I deliberately left `gpt_oss.py` alone. There, main would raise during `import unsloth_zoo.temporary_patches.gpt_oss` and take the import down with it, so degrading to `device_memory = 0` and combo kernels off is strictly better, not a silent flip.

### Mutation

7 mutants, 6 killed. The survivor is an equivalent mutant rather than a coverage gap, and it is worth naming:

```
SURVIVED  a driver total of zero is adopted
```

`_driver_total_memory` returning 0 instead of None still yields `max(total, 0) == total`, because the caller has already returned for `total <= 0`. No test can distinguish it. The PR body lists "driver zero discarded" among the pinned behaviours; that test cannot fail.

Killed: driver total adopted outright, HIP no longer excluded by name, discrete cards paying for the primary context, the plain `integrated` spelling dropped, a caller-supplied `props` ignored, and a non-integer driver total accepted.

### What I could not prove

- I have no unified-memory NVIDIA part. Every integrated case here is a stub with `is_integrated` set. Whether a real RTX Spark or Jetson reports a larger driver total on some torch build is the thing this PR already says it cannot demonstrate.
- The 8128MiB against 46477MiB figure is quoted from the reporter's machine, not reproduced here.
- Nothing ran on Windows ARM64, which is the host the report came from, and there is no staging repo configured for unsloth-zoo, so there is no cross-platform leg for this one. The changed code has no `os.name` or `sys.platform` branch, so platform dependence would have to come through torch itself.
- No training run. The combo-kernel and kernel-option thresholds were exercised as expressions, not as a workload.

danielhanchen:
@codex review

danielhanchen:
@codex review

PR #1270, opened 2026-09-17
fix(data): stop response-only labels at message boundaries

taking-lying-flat:
`train_on_responses_only` currently copies labels from an assistant marker until the next user marker or the end of the sequence. For tool conversations, this can supervise tool output and later role headers, and `last_response_only` can retain an earlier assistant message. Applying the mask after concatenation can also supervise the next sample's system prompt.

Stop each response span at EOS (keeping the EOS target), BOS, a shared special-token message opener, or the next assistant marker. Revisit message openers so subsequent assistant messages form separate spans. Preserve existing label masks and avoid treating a plain-text shared prefix as a message boundary.

The change is confined to `unsloth_zoo/dataset_utils.py` (24 insertions, 2 deletions). Custom templates without recognizable message or sample boundaries still require role-based labels.

Validation:

- 101 local regression checks passed, covering ChatML, Llama, Gemma and Phi-style layouts; tools, consecutive assistants, last-response-only masking, concatenated samples, existing labels, automatic marker detection, and list/tensor inputs. The validation scripts remain local.
- Related existing tests after repairing the local environment: **285 passed, 6 skipped, 0 failed**; skips require an uncached tokenizer fixture. The four original `VideoReader` import errors were resolved by backporting [huggingface/datasets#8087](https://github.com/huggingface/datasets/pull/8087) into local `datasets==4.3.0+tvcompat1`, preserving Unsloth's version constraint. All 101 local regressions passed again. The dependency patch and validation scripts are not included in this PR.


danielhanchen:
Confirmed the span loop in unsloth_zoo/dataset_utils.py runs from the assistant marker to the next user marker, so tool output and following headers are supervised; could you say which templates you checked the shared-special-token opener against, since it is derived from the Q/A markers? Will get this reviewed.

taking-lying-flat:
Thanks @danielhanchen! The shared-opener checks covered **ChatML, Llama 3-style, and Gemma-style layouts**. Phi-style was a separate EOS-boundary control. Here are the exact fixtures and what they establish for the current head, `8c1c544`.

First, a clarification on scope: the 101 local regressions use synthetic `PreTrainedTokenizerFast`/`WordLevel` tokenizers with `WhitespaceSplit`, registered special tokens, and matching Jinja chat templates. The template-family names in the PR refer to those controlled layouts. Validation with the corresponding pretrained tokenizers is still outstanding. These checks use the function's default `force_match=True`.

The instruction/response markers and derived openers are:

- **ChatML:** `<|im_start|>user\n` / `<|im_start|>assistant\n`. The tokenized cores start with the same `<|im_start|>` ID, followed by different role IDs, so the opener is exactly `[im_start_id]`. The fixture uses `<|im_end|>` as the turn terminator and `</s>` as tokenizer EOS.
- **Llama 3-style:** `<|start_header_id|>user<|end_header_id|>\n\n` / `<|start_header_id|>assistant<|end_header_id|>\n\n`. The opener is exactly `[start_header_id]`; the common `<|end_header_id|>` suffix is not part of it because the prefix comparison stops at the differing role IDs. The fixture uses `<|eot_id|>` as the turn terminator and `<|end_of_text|>` as tokenizer EOS.
- **Gemma-style:** `<start_of_turn>user\n` / `<start_of_turn>model\n`. The opener is exactly `[start_of_turn_id]`. The assistant role is serialized as `model`. The fixture uses `<end_of_turn>` as the turn terminator and `<eos>` as tokenizer EOS.

In all three fixtures, the turn terminator deliberately differs from `eos_token_id`. Thus, in an assistant → tool → assistant sequence, the first response stops at the tool message's shared opener; the EOS check cannot explain that result. The preceding assistant's turn terminator remains supervised, while the next opener, role/header tokens, and tool body stay masked. I also inspected the explicit and auto-detected markers: they produce the same tokenized cores and openers in all three fixtures.

**Phi-style:** the markers are `<|user|>\n` / `<|assistant|>\n`, with each role token registered atomically. Their textual `<|` prefix does not correspond to a shared token ID, so `message_start` is empty. This fixture sets `eos_token_id` to `<|end|>`; each response stops there and retains that EOS target. It verifies the EOS path, and does **not** establish coverage for Phi variants where `<|end|>` differs from tokenizer EOS.

Regarding how the opener is inferred: the implementation compares the prefix of `Q_must` and `A_must` returned by `_find_common_token_ids`, stopping at the first unequal ID. It retains that complete prefix only if **at least one** ID in it belongs to `tokenizer.all_special_ids`. It neither compares raw string prefixes nor searches for an arbitrary shared subsequence. A separate character-tokenizer control uses `\n### User:\n` / `\n### Assistant:\n`, with no special IDs, and checks that `Answer\n### Heading\nStill answer` remains fully supervised.

For a concrete example, the ChatML fixture encodes:

```text
<s>
<|im_start|>user\nQ1<|im_end|>
<|im_start|>assistant\nCALL<|im_end|>
<|im_start|>tool\nWEATHER<|im_end|>
<|im_start|>assistant\nFINAL<|im_end|>
</s>
```

With `last_response_only=False`, the supervised tokens are `CALL`, `<|im_end|>`, `FINAL`, `<|im_end|>`, `</s>`. With `True`, they are only `FINAL`, `<|im_end|>`, `</s>`. All role headers and `WEATHER` have label `-100`. The boundary is revisited by the outer scan so the later assistant becomes a separate span; this also handles consecutive assistant messages.

The **101 passing cases** break down as follows. Counts are per family, across all four synthetic layouts:

| Check | Cases per family |
| --- | ---: |
| Tool chain: existing labels on/off × last-response-only on/off × list/tensor inputs | 8 |
| Consecutive assistants: existing labels on/off × last-response-only on/off | 4 |
| Non-assistant tail: system, developer, or tool | 3 |
| Concatenated samples with system prompts: existing labels on/off × list/tensor inputs | 4 |
| Concatenation without appending a separate final EOS to the first sample | 1 |
| Ordinary user/assistant multi-turn: last-response-only on/off | 2 |
| Preserve an additional `-100` inside an existing assistant label span | 1 |
| Auto-detected markers on the tool chain: last-response-only on/off | 2 |

That is `25 × 4 + 1` plain-text-prefix control. Each case compares the complete label sequence against expected role-based labels. For the no-separate-final-EOS case, Phi still ends its last turn with `<|end|>` = EOS, so that particular Phi case does not isolate the BOS path.

The remaining limits are worth making explicit for review. `WhitespaceSplit` does not exercise real tokenizer newline/whitespace merging, and this 101-case suite does not cover `force_match=False`. The synthetic role streams also do not establish that each model's native chat template accepts every tested role. The shared-opener rule depends on the supplied/detected Q/A cores retaining that opener and on its special-token registration; the metadata check is a heuristic, not a general proof of a message boundary. Role-only markers can remove that signal. Without another recognized boundary, custom layouts still require role-based labels. Existing labels are preserved inside retained spans, but cannot repair incorrect span discovery for `last_response_only`.

The local suite was rerun against `8c1c544`: **101 passed**. The validation harness remains local and is not included in the PR. Thanks for getting this reviewed!

shimmyshimmer:
Review record for unslothai/unsloth-zoo#1270, carried over from the mirror it was reviewed on: https://github.com/danielhanchen/unsloth-zoo-staging-review/pull/9.

Findings are quoted as posted and attributed to the account that posted them. The reactions shown are the triage recorded on the mirror; none were re-applied here. Commits are named as the mirror's, with this PR's equivalent where one was found.

### Round 1 — reviewed mirror `8ea48b6fb0` (this PR's `98c4bdad6`)

> **chatgpt-codex-connector[bot]** ([on the mirror](https://github.com/danielhanchen/unsloth-zoo-staging-review/pull/9#issuecomment-5757858003))
>
> Codex Review: Didn't find any major issues. Keep them coming!
>
> **Reviewed commit:** `8ea48b6fb0`

No triage reaction was recorded on the mirror

### Verdict

The round converged at mirror `8ea48b6fb0` (this PR's `98c4bdad6`).

This is the review as it stood at that commit. Anything pushed to this PR afterwards was not part of it.

PR #1273, opened 2026-09-17
Stop the per-layer device tests asserting a device a CPU-only build cannot name

danielhanchen:
Two tests added in #1229 fail on every CPU-only runner. That is what unsloth's `Core` and `Core zoo` jobs use, so they have been red on **every open unsloth PR** since #1229 landed, including ones that touch nothing related:

```
FAILED tests/test_per_layer_device.py::test_a_meta_layer_publishes_the_accelerate_execution_device
FAILED tests/test_per_layer_device.py::test_a_chained_hook_is_unwrapped_to_the_alignment_device
AttributeError: '_FakeLayer' object has no attribute '_per_layer_device'
```

That message names neither the device nor the cause, which is most of why this took a while to pin down.

### Cause

A bare integer `execution_device` is the index of the **default accelerator**, so it needs one to exist. Measured on the exact wheel CI installs:

```
torch 2.10.0+cpu
cuda available: False
torch.device(2)         RuntimeError: Cannot access accelerator device when none is available.
torch.device("cpu")     cpu
torch.device("cuda:3")  cuda:3
```

`_execution_device_for_meta_layer` catches that `RuntimeError`, treats the hook as naming no device it can use, and moves on; `verify_and_set_device` then publishes nothing. That behaviour is correct. There is no device 2 on that machine to send activations to, and publishing one would be the `or 0` mistake the function's own comment warns about. The tests were wrong to assert otherwise on a host that has no accelerator.

Note the third line. The **string** spellings construct on any build, which is why `"cuda:3"` and `"xpu:1"` elsewhere in this file were always fine and only the two integer cases broke. `test_a_nested_chain_is_followed_and_the_outermost_real_device_wins` also passes an integer, and is unaffected, because the search returns `"cpu"` from the inner chain before it ever reaches it.

Worth recording: hiding the GPUs with `CUDA_VISIBLE_DEVICES=""` does **not** reproduce this. The wheel still has the backend compiled in, so `torch.device(2)` still succeeds. That cost me two wrong hypotheses before I measured the build itself rather than the host.

### The change

`test_a_meta_layer_publishes_the_accelerate_execution_device` keeps only the `"cpu"` case, which constructs on every build.

The integer case keeps its coverage instead of being deleted. accelerate types `execution_device` as `int | str | torch.device | None`, so the bare index is a real shape to accept, not just a spelling of the string form. It moves to its own test gated on whether an index can be named here.

The CPU-only half of the contract gets a test, `test_an_integer_execution_device_publishes_nothing_without_an_accelerator`. It had no coverage at all, and it is the half CI actually runs, so without it the integer path was silently untested exactly where it is exercised.

`test_a_chained_hook_is_unwrapped_to_the_alignment_device` switches to the string spelling rather than being gated. What it is about is the search descending into accelerate's `SequentialHook` wrapper, where the device spelling is incidental, so it now runs everywhere instead of skipping on precisely the runner that matters.

### Verification

- CUDA build (this box): `24 passed, 1 skipped`, the skip being the new CPU-only test.
- CPU-only build (`torch 2.10.0+cpu`, matching CI): both previously failing cases verified green by driving the same code path directly, since the full zoo import chain needs deps that venv does not carry.
- Non-vacuity: replacing the refusal with a device fallback makes the new test fail, so it is pinning the behaviour rather than the absence of an attribute.


danielhanchen:
@codex review

danielhanchen:
@codex security review

PR #1274, opened 2026-09-18
fix(mlx): release KV-shared attention cotangents under gradient checkpointing

Lyxot:
## Summary

LoRA training of Gemma 4 E2B/E4B with MLX gradient checkpointing kept a T x T attention cotangent alive for every KV-shared layer, so peak memory grew quadratically with sequence length across the shared tail of the stack. This change releases those buffers inside each borrowing layer's backward. Forward values, loss, and gradients are unchanged (bitwise on real checkpoints).

## Root cause

KV-shared layers receive the source layer's K/V as an argument (`shared_kv=` on mlx-vlm >= 0.5.0, `_SharedKVSlot.borrow()` on older mlx-vlm), so it crosses the `mx.checkpoint` boundary as a primal. Its cotangent is consumed only by the source layer's backward, many layers earlier, and `mx.eval` defers computing it until there. In training MLX always takes the unfused SDPA path, so each borrowing layer's dK/dV depends on its B x H x T x T attention weights and score cotangents, which stay alive until the source layer runs: roughly 150 MB per shared layer at T=2048 on E2B.

Diagnostics on real E2B (35 layers, T=2048, eager backbone loss): stopping the gradient into shared K/V drops the peak from 4.81 to 2.19 GB, while rewiring the layer loop so each shared layer borrows from the previous one changes nothing. The retention is a scheduling effect, not a topology one.

## Change

Before `mx.checkpoint`, the hidden state and borrowed K/V pass through an identity `mx.custom_function` whose VJP returns `mx.depends(d_hidden, [dK, dV])`. The hidden-state cotangent, which the previous layer needs immediately, now depends on dK/dV, so they are computed inside the borrowing layer's backward and the T x T buffers are freed right away. Nothing is copied; the forward and decode paths only gain an identity node. Both the native `shared_kv` path and the legacy `_SharedKVSlot` path are covered.

## Validation

Layer-truncation peak above resident (eager value_and_grad of a backbone-only loss w.r.t. LoRA params, gradient checkpointing, batch 1):

| model | layers | T=1024 before -> after (GB) | T=2048 before -> after (GB) |
|---|---|---|---|
| gemma-4-e2b-it-4bit | 5 / 10 / 15 (no shared layers) | unchanged | unchanged |
| gemma-4-e2b-it-4bit | 20 | 1.444 -> 1.239 | 2.612 -> 2.108 |
| gemma-4-e2b-it-4bit | 35 | 2.097 -> 1.316 | 4.812 -> 2.228 |
| gemma-4-e4b-it-4bit | 5 / 12 / 24 (no shared layers) | unchanged | unchanged |
| gemma-4-e4b-it-4bit | 42 | 2.094 -> 1.425 | 4.582 -> 2.275 |

Real `MLXTrainer` runs (defaults: compiled step, CCE, gradient checkpointing; batch 2 x 2048):

| model | peak before | peak after | losses |
|---|---|---|---|
| gemma-4-e2b-it-4bit | 13.09 GB | 11.09 GB | identical for 10 steps; final LoRA adapters bitwise equal (410/410) |
| gemma-4-e4b-it-4bit | 14.86 GB | 13.44 GB | identical for 5 steps |

Numerics with a real next-token cross-entropy loss and non-zero LoRA B: gradients with checkpointing after this change, before it, and without checkpointing are bitwise equal (E2B T=1024: 410/410; E4B T=1024: 516/516; E2B batch 2: 410/410), and inference logits through the patched layers are bitwise equal to the unpatched ones. The comparison is sensitive to the changed path: blocking the shared-K/V gradient changes the source layers' K/V adapter gradients by ~100% relative.

Legacy path: the mlx-vlm v0.4.4 Gemma 4 classes driven through `_SharedKVSlot` go from 360.6 to 224.0 MB (an equivalent unshared stack is 201.9 MB), with gradients matching the uncheckpointed run.

New test `test_checkpointed_kv_shared_layers_hold_what_unshared_layers_hold` (Metal) builds a tiny mlx-vlm `Gemma4TextModel` and asserts that a checkpointed stack with KV sharing peaks within 1.25x of the same stack without sharing, with gradients matching the uncheckpointed model. Removing the tie or the `mx.depends` fails the memory assertion (357 MB vs 212 MB); zeroing the K/V cotangent fails the gradient assertion.

```bash
python -m pytest tests/test_mlx_training_e2e_metal.py -q -k kv_shared_layers
python -m pytest tests/test_mlx_generate_metal.py tests/test_mlx_vlm_label_masks.py -q -k "shared_kv or slot or checkpoint"
```

## Notes

- The committed test covers the native `shared_kv` path only; the legacy branch cannot be imported with the pinned mlx-vlm and was validated out of tree as described above.
- Separate, pre-existing issue found while validating (not addressed here): on the text-only path, `_fix_gemma4_kv_sharing` builds `_TrainingKVStore` caches, which the checkpoint wrapper does not recognise, so gemma3n text-only training with gradient checkpointing drops the gradient from shared layers into the source layers' K/V. Swapping in `_SharedKVSlot` makes gradients bitwise equal to the uncheckpointed run, after which this change also applies (126 -> 28 MB per shared layer at T=2048).


Lyxot:
Review and CI staging for this PR ran on a fork mirror (same base `aaee6ba57`, same head `47bbfc4b7`), since the Codex review connector is not currently responding in this repository.

**Codex review:** "Didn't find any major issues" for `47bbfc4b7a`: https://github.com/Lyxot/unsloth-zoo/pull/14#issuecomment-5725588569. No review threads were opened, so there was nothing to address.

**CI on `47bbfc4b7`:**

| Workflow | Result |
|---|---|
| Tests CI (repo tests CPU, Python 3.10-3.13 collect, MLX Linux CPU numerics + suites, core drift) | [success](https://github.com/Lyxot/unsloth-zoo/actions/runs/35310493752) |
| MLX CI on Mac M1 (install + import smoke, Apple Silicon) | [success](https://github.com/Lyxot/unsloth-zoo/actions/runs/35310493806) |
| Lint CI | [success](https://github.com/Lyxot/unsloth-zoo/actions/runs/35310493595) |
| Wheel CI | [success](https://github.com/Lyxot/unsloth-zoo/actions/runs/35310493631) |

The Gemma4 audio probe is label-gated and was skipped. The staging PR has been closed without merging.

danielhanchen:
Confirmed the borrowed K/V does cross the mx.checkpoint boundary in unsloth_zoo/mlx/utils.py without its cotangent being consumed locally, and the depends-based tie keeps gradients matching the uncheckpointed run. Will get this reviewed.

PR #1275, opened 2026-09-18
fix(mlx): keep KV-shared gradients on the cacheless text path under gradient checkpointing

Lyxot:
Depends on #1274, and this branch includes its commit. Only the second commit (`fix(mlx): keep KV-shared gradients on the cacheless text path under gradient checkpointing`) is new here; once #1274 merges, the diff shrinks to that commit.

## Bug

Some stacks don't thread shared K/V natively: gemma3n on current mlx-vlm, and Gemma 4 on mlx-vlm < 0.5.0. For these, text-only training with gradient checkpointing (the trainer default) silently drops part of the gradient. The KV-shared layers stop sending gradient into the source layers' K/V, and through them into everything upstream. The forward pass and the loss are unchanged, so nothing looks wrong.

On gemma-3n-E2B-it-4bit (LoRA r=16 with non-zero `lora_b`, CCE loss, T=512, B=2), only 150 of 420 LoRA gradients matched bitwise between the GC and no-GC runs. The `k_proj` LoRA gradients of source layers 18/19 had 89–97% relative error.

## Root cause

When the backbone is called with `cache=None`, `_fix_gemma4_kv_sharing` gives it `_TrainingKVStore` caches. The gradient-checkpointing wrapper (`_patch_layer_class_for_gc`) only passes K/V across `mx.checkpoint` as a traced argument and result when the cache is a `_SharedKVSlot`. With any other cache, the shared layers read `cache.state` inside the checkpointed region. MLX treats that read as a captured constant, so the recomputed VJP has no path back to the source layers. The multimodal loss paths already build `_SharedKVSlot` caches through `_build_shared_kv_caches`, so they were not affected. The text-only path was, including `make_cce_loss_fn`.

## Change

- The cacheless shim now builds `_SharedKVSlot` caches, and `_TrainingKVStore` is removed. Both expose the same cache interface (offset 0, `state`, `update_and_fetch`), so forward values do not change. `_SharedKVSlot` also follows the borrow/install/recorded contract that the checkpoint wrapper relies on.
- New real-Metal regression test in `tests/test_mlx_training_e2e_metal.py`: a tiny gemma3n backbone runs through `_fix_gemma4_kv_sharing`, and every parameter gradient must match between GC and no-GC. Before the fix, 70 of 155 gradients mismatch and the test fails.

Once gradients flow through the shared layers, those layers depend on the cotangent tie from #1274 to release their attention cotangents. That is why this PR stacks on it.

## Validation

mlx 0.32.2, mlx-vlm 0.6.17, Apple Silicon.

```text
gemma-3n-E2B-it-4bit, GC vs no-GC gradients (CCE, T=512, B=2, LoRA r=16, non-zero B)
  before: 150/420 bitwise equal (layers 18/19 k_proj: 89-97% rel. error)
  after:  420/420 bitwise equal

Legacy Gemma 4 (mlx-vlm v0.4.4 layer sources), tiny config, GC vs no-GC
  before: 91/175 mismatched (all source layers + embeddings)
  after:  0/175 mismatched

gemma-3n-E2B-it-4bit, per-layer peak for shared layers, CCE, T=2048
  with #1274 tie: 28.4 MB/layer (30 layers: 3.08 GB)
  without tie:    126.4 MB/layer (30 layers: 4.06 GB)

MLXTrainer, 10 steps, 2 x 2048, compiled step + CCE + GC
  main:  loss 18.84 -> 13.54, peak 13.68 GB
  after: loss 18.84 -> 13.25, peak 13.84 GB  (step-1 loss identical; forward unchanged)
```

The slight rise in trainer peak memory and step time comes from the source layers now receiving the backward work that was previously skipped.

```bash
python -m pytest tests/test_mlx_training_e2e_metal.py -q -k kv_shared
```


Lyxot:
Review completed on an exact-diff staging PR (based on the #1274 head, so it contains only this change): [Codex's normal review comment](https://github.com/Lyxot/unsloth-zoo/pull/15#issuecomment-5725705598) says "Didn't find any major issues" for `4c80b21e42`, the current head of this PR.

This PR has no review threads, so nothing needed fixing or resolving.

CI on the staging PR, all passing:

| Check | Result |
| --- | --- |
| [Repo tests (CPU)](https://github.com/Lyxot/unsloth-zoo/actions/runs/35311463749/job/105494208320) | pass |
| [MLX (Linux CPU: numerics + suites)](https://github.com/Lyxot/unsloth-zoo/actions/runs/35311463749/job/105494208118) | pass |
| [Python 3.10-3.13 collect](https://github.com/Lyxot/unsloth-zoo/actions/runs/35311463749/job/105494208089) | pass |
| [Core drift (3 upstream pins)](https://github.com/Lyxot/unsloth-zoo/actions/runs/35311463749/job/105494207879) | pass |
| [Source lint (Python + YAML + JSON)](https://github.com/Lyxot/unsloth-zoo/actions/runs/35311463774/job/105494208329) | pass |
| [Wheel build + content sanity + import smoke](https://github.com/Lyxot/unsloth-zoo/actions/runs/35311463783/job/105494208333) | pass |
| MLX install + import smoke (Apple Silicon) | skipped by workflow condition |

The new regression test needs real Metal, so CI skips it. It passes locally on Apple Silicon, and it fails against the pre-fix loader. The staging PR is closed without merging.

danielhanchen:
Confirmed the cacheless path in unsloth_zoo/mlx/loader.py still hands out a cache the checkpoint wrapper does not recognise, so gemma3n text-only training with checkpointing loses the gradient into the source layers K/V. Will get this reviewed once the base lands.

danielhanchen:
Reviewed this on real MLX. The defect reproduces and the fix is correct.

I could not run Metal here, but none of the mechanism is Metal-specific, so I reproduced it on the Linux CPU build of MLX (`mlx` + `mlx-cpu`, mlx 0.32.2, mlx-vlm 0.7.1) with a tiny gemma3n backbone:

| arm | cache the shim hands out | gradients matching GC vs no-GC |
|---|---|---|
| merge base `fb0c99e2` | `_TrainingKVStore` | 85 of 155 (70 mismatched, worst 79.8% relative on `layers.0.altup.correction_coefs.weight`) |
| #1274 head `47bbfc4b` | `_TrainingKVStore` | 85 of 155 |
| this PR `4c80b21e` | `_SharedKVSlot` | 155 of 155 |

That matches your 70-of-155 figure. The corrupted set is exactly the source layers and everything upstream of them, never the borrowing layers, which is the signature you would expect when the cotangent through borrowed K/V is dropped. Forward loss is bit-identical on every arm, so only gradients move.

Reverting only `loader.py` on your head, keeping the test, makes exactly the new test fail and leaves #1274's passing. So the test is falsifiable and the fix is what carries it.

Two things worth noting beyond the change itself.

The defect is live on current mlx-vlm, not only on old versions. `_gemma4_has_native_shared_kv` returns False for gemma3n on mlx-vlm 0.7.0 and 0.7.1, since `models/gemma3n/language.py` declares neither `previous_kvs` nor `shared_kv_sink`, so the shim fires on every shipped version. gemma4 is unaffected because it does thread shared K/V natively.

I pushed one commit to this branch. The test's `finally` did `del Gemma3Model._kv_sharing_patched` unconditionally, but `_fix_gemma4_kv_sharing` has seven early returns that never set that flag. On any of them the delete raises `AttributeError` out of the `finally`, and that AttributeError is the only thing the one line pytest summary reports, so a real failure would be hidden behind it. I reproduced that directly. The commit deletes the flag only when this test's own call set it, and guards on the class `__dict__` rather than `hasattr`, because `hasattr` also sees a flag inherited from an earlier patch and consuming that one disarms the shim's idempotence guard so the next load double wraps `__call__`. It also asserts the patch installed, since an early return previously let the test pass having proved nothing. After the change: the test still passes, reverting `loader.py` now fails on the real assertion rather than on an AttributeError, and forcing an early return reports "the shim did not install".

One note for #1274 rather than this PR: `_tie_to_hidden_state` calls `tuple(shared_kv)`, so a caller passing a list now sees a tuple inside the layer. Harmless for both shipped consumers, but it is an observable type change on the kwarg branch.

Worth stating plainly: neither new test has ever executed in CI. Both are `@metal_only` and the Apple Silicon job is opt in behind the `mlx` label, so every run on this branch was skipped. I have added the label and run the file on Apple Silicon separately.

danielhanchen:
Correction to my previous comment: I wrote that I had run the test file on Apple Silicon. That was ahead of the evidence and is not yet true.

Adding the `mlx` label did start the Apple Silicon job, but every run of it on these branches has been cancelled before finishing by an unrelated CI sweep, three times now, so no completed Metal run exists yet. The same is true of the runs on #1264, #1265, #1274 and #1278. For the same reason, the red checks currently shown on this PR are cancellations rather than failures: the annotation on each is "The run was canceled", including on two jobs that were green before my commit.

My Apple Silicon jobs on an owned staging repo are still queued. I will post the real result when one completes, and say so plainly if none does.

Everything else in that comment stands: the gradient reproduction, the revert test and the teardown fix were all run on the Linux CPU build of MLX here, and none of that needs Metal.

PR #1276, opened 2026-09-18
Stop the live-tensor walk dying on somebody else's dead weakref

danielhanchen:
`Core zoo (HF=default + TRL=default)` went red on unsloth `main` with:

```
FAILED tests/test_moe_lora_grouped_mm_capability.py::test_no_grad_decoding_retains_nothing
ReferenceError: weakly-referenced object no longer exists
```

A message about a weakref, raised from a test about tensor retention, in a file that never creates one.

### Cause

`gc.get_objects()` returns every tracked object in the **process**, including dead `weakref.proxy` objects other modules leave behind. PEFT's parametrization makes them. `torch.is_tensor(o)` is `isinstance(o, torch.Tensor)`, and isinstance on a dead weakproxy **raises** rather than answering False, so the walk dies instead of counting.

Reproduced deterministically rather than inferred:

```python
referent = _Referent()
proxy = weakref.proxy(referent)
del referent
gc.collect()
for o in gc.get_objects():
    torch.is_tensor(o)      # ReferenceError: weakly-referenced object no longer exists
```

One dead proxy anywhere in the process is enough. That is why this only failed under xdist, where other modules share a worker, and never when the file ran on its own: it passes 3/3 here in isolation.

### The change

The count skips an object it cannot read. That is correct on its own terms - a dead weakproxy is not a live tensor, and live tensors are the only thing being counted - so the invariant the test exists for is untouched.

`tests/test_moe_fused_lora_merge_layout.py` already clears its own proxies for exactly this reason, with a comment naming this hazard, and it keeps doing so. That is worth having, but it cannot be the whole answer: it asks every producer to remember, and any test in any file may leave one behind. So the consumer is made safe too.

### Verification

A/B against a plugin that leaks a dead proxy the way a neighbouring module would:

| | result |
|---|---|
| without the guard (`main`) | `FAILED ... ReferenceError`, identical to CI |
| with the guard | `1 passed` |

The file goes from 31 to 32 passed, the extra being `test_the_live_tensor_walk_survives_a_dead_weakproxy`, which pins the behaviour so the guard cannot be dropped silently. It asserts the raw `isinstance` still raises on a dead proxy, so it fails if the guard is removed rather than quietly testing nothing.


danielhanchen:
@codex review

danielhanchen:
@codex security review

PR #1278, opened 2026-09-18
perf(mlx): score preference objectives through runtime CCE

Lyxot:
DPO and ORPO still build vocabulary-wide logits for every scored position, so a preference step holds batch × sequence × vocabulary logits (and their bf16 copy) at its peak. With 248k–262k vocabularies that dominates the memory of LoRA preference training, even after #1265 compacts the projection to the supervised rows.

This scores policy and LoRA reference log-probabilities directly from hidden states with the existing per-token runtime CCE kernel, which never materializes vocabulary-wide logits and reduces in float32.

> [!NOTE]
> Stacked on #1265 (compact supervised CCE rows). The first eight commits are that PR, rebased unchanged on the current `main`; only the last commit, `perf(mlx): score preference objectives through runtime CCE`, is new here. Review that commit alone; I will rebase once #1265 lands.

### What changed

- **Scorer** (`_make_preference_cce_scorer`): builds a runtime CCE kernel for the model's head and scores token cross-entropy from the hidden states, with optional compact row indices. It accepts unwrapped, bias-free dense heads and frozen quantized heads (head scale and logit softcap included, e.g. Gemma 4's softcap 30) and returns `None` otherwise, in which case the existing logits path is used unchanged.
- **Logit-sum metrics**: the `logits/chosen` / `logits/rejected` sums are chunked over the vocabulary without an N × V buffer.
- **DPO / ORPO wiring**: `make_dpo_cce_loss_fn` / `make_orpo_cce_loss_fn` pass the scorer into `make_dpo_loss_fn` / `make_orpo_loss_fn`; the LoRA reference pass goes through the same scorer. Preference evaluation (`make_preference_eval_fn(model=...)`) shares it.
- **Compaction admission**: with the scorer there is no dense projection to replace, so compaction is admitted whenever the supervised capacity is below the padded grid (ORPO previously needed a quarter of the grid or less), and the trainer reports CCE for any preference loss that scores through it.

### Impact and risks

- Only `use_cce=True` preference training and evaluation change; SFT, VLM and `use_cce=False` are untouched.
- Log-probabilities are computed with different rounding than the bf16 logits path, so losses and reward metrics shift slightly; they are not bitwise equal to `main`. Against a float32 reference both paths are equally close (table below). On Gemma 4, whose DPO loss swings strongly between steps on this synthetic data, that rounding difference is visible in the loss curve after a few optimizer steps; the curve is deterministic on each branch.
- Heads the scorer does not accept (trainable quantized heads, wrapped or biased heads) keep the previous path.

### Performance

Apple M3 Max (128 GB), MLX 0.32.2, gradient checkpointing and a compiled training step throughout. `main` is the current `main`; "compaction" is #1265 alone; "this PR" is the full stack.

**Real models**: `mlx-community` checkpoints through `FastMLXModel` + `MLXDPOTrainer` / `MLXORPOTrainer` with trainer defaults (`use_cce`, gradient checkpointing, compile), LoRA r=16, batch 2, `max_length` 2048, `max_prompt_length` 1024, DPO with a LoRA reference policy. Each batch holds one long pair (about 960 / 680 / 510 prompt / chosen / rejected words, prompt capped at 1024 tokens) and one short pair (about 510 / 260 / 200 words), so every batch is padded while the compiled step traces once; about 1,740 response tokens are scored per step. Step time is the median of 5 measured steps after 2 warm-up steps; peak is `mx.get_peak_memory()` over the measured steps. Two rounds with the level order reversed in the second, a 60 s cool-down before every run, nothing else running on the GPU, and no thermal or performance warning recorded by `pmset -g therm` in any run. Peaks repeated to 0.01 GB across rounds; times are the mean of the two rounds.

| Model / objective | main | compaction (#1265) | this PR | vs main |
|---|---:|---:|---:|---:|
| Qwen3.5-4B-4bit, DPO (4-bit tied head, 248k vocab) | 19.20 GB, 27.8 s | 12.99 GB, 25.9 s | **10.07 GB, 24.0 s** | **0.52x memory, 0.86x time** |
| Qwen3.5-4B-4bit, ORPO | 14.01 GB, 20.7 s | 14.01 GB, 20.7 s | **10.21 GB, 19.5 s** | **0.73x memory, 0.94x time** |
| gemma-4-e4b-it-4bit, DPO (4-bit tied head, softcap 30, 262k vocab) | 26.73 GB, 30.4 s | 26.73 GB, 29.8 s | **19.11 GB, 26.8 s** | **0.71x memory, 0.88x time** |

ORPO batches here are too dense for #1265's ORPO compaction, and Gemma's softcapped head is not eligible for #1265's DPO projection, so those rows show this PR's effect alone.

Log-probability accuracy on the same real models (4 responses of 450–700 tokens; absolute error of the summed response log-probability against float32 hidden states × a float32 dequantized head, softcap in float32):

| Model | bf16 logits path (main) | runtime CCE scorer (this PR) |
|---|---:|---:|
| Qwen3.5-4B-4bit | 0.04 – 0.39 nats | 0.21 – 0.64 nats |
| gemma-4-e4b-it-4bit | 0.10 – 0.83 nats | 0.11 – 0.47 nats |

**Simulated**: an `mlx_lm` llama (hidden 2048, vocab 151,936, 2 layers, bf16, 4-bit head) with a compiled `value_and_grad` + AdamW step and gradient checkpointing, one fresh process per measurement, rounds interleaved across levels, measured against the merge base before the rebase (which measured identical to the `main` of that time).

| Scenario | main | compaction (#1265) | this PR | vs main |
|---|---:|---:|---:|---:|
| DPO, 4 pairs × 1024 tokens, half of each in the prompt | 10574 MB, 1866 ms | 8102 MB, 1996 ms | **3673 MB, 884 ms** | **0.35x memory, 0.47x time** |
| ORPO, 4 pairs × 1024 tokens, three quarters in the prompt | 10574 MB, 2079 ms | 10574 MB, 2128 ms | **4486 MB, 975 ms** | **0.42x memory, 0.47x time** |
| SFT LoRA (dense / 4-bit head), full fine-tuning | unchanged | unchanged | unchanged | 1.00x |

### Related MLX CCE branches

All are independent of each other except where stacked; each is measured against `main`.

```text
main
├─ #1264  perf/mlx-cce-backward-storage   CCE backward temporaries      (−11% to −34% memory)
├─ #1265  perf/mlx-cce-compaction         compact supervised rows       (text −69%, VLM −56%, DPO −23% memory)
│  └─ this PR  perf/mlx-cce-preference-fused  preference via runtime CCE (DPO 0.52x, ORPO 0.73x memory on Qwen3.5-4B)
└─ #1277  perf/mlx-cce-label-smoothing    fused label-smoothing kernels (fixes a compile crash; Qwen3.5-4B 0.68x memory vs main eager)
```

### Validation

```bash
python -m pytest tests/test_mlx_training_e2e_metal.py -q
```

144 passed on Metal after the rebase. The new and updated tests compare the runtime CCE scorer against the logits path for DPO and ORPO losses and gradients across dense, 4-bit and softcapped heads, float32 and bf16, low and high prompt shares, with and without a LoRA reference policy; check the chunked logit sums against a float32 reference; and cover preference evaluation with an unequal final batch and compaction preserving pair ownership.


danielhanchen:
Checked the scorer in unsloth_zoo/mlx/preference.py and the fallback for ineligible heads, and the memory drop on the default use_cce DPO and ORPO path looks like it reaches every MLX preference run. Ping me once the base lands and this is rebased onto main and I will review the final commit.

Lyxot:
Review completed on an exact-diff staging PR (based on this PR's compaction commits from #1265, so it contains only the new preference commit): [Codex's normal review comment](https://github.com/Lyxot/unsloth-zoo/pull/17#issuecomment-5728300232) says "Didn't find any major issues" for `8dfd0e7273`, the current head of this PR.

This PR has no review threads, so nothing needed fixing or resolving.

CI on the staging PR, all passing (upstream CI on this head also passed):

| Check | Result |
| --- | --- |
| [Repo tests (CPU)](https://github.com/Lyxot/unsloth-zoo/actions/runs/35331297378/job/105556004494) | pass |
| [MLX (Linux CPU: numerics + suites)](https://github.com/Lyxot/unsloth-zoo/actions/runs/35331297378/job/105556004313) | pass |
| [Python 3.10-3.13 collect](https://github.com/Lyxot/unsloth-zoo/actions/runs/35331297378/job/105556004596) | pass |
| [Core drift (3 upstream pins)](https://github.com/Lyxot/unsloth-zoo/actions/runs/35331297378/job/105556004179) | pass |
| [Source lint (Python + YAML + JSON)](https://github.com/Lyxot/unsloth-zoo/actions/runs/35331297262/job/105556003524) | pass |
| [Wheel build + content sanity + import smoke](https://github.com/Lyxot/unsloth-zoo/actions/runs/35331297292/job/105556003796) | pass |
| MLX install + import smoke (Apple Silicon) | skipped by workflow condition |

The preference scorer tests need real Metal, so CI skips them; they pass locally on Apple Silicon. The staging PR is closed without merging.

danielhanchen:
Reviewed and merged. I rebased it for you onto the merged #1265 rather than asking you to: the branch is now main plus your single preference commit, and I verified the resolution is exact, the delta against main is byte identical to commit `8dfd0e72` itself (3 files, +209/-172). No force push was needed.

What I checked here, on the Linux CPU build of MLX. Reverting `preference.py` and `trainer.py` to main while keeping your tests makes every selected test fail with `AttributeError: module 'unsloth_zoo.mlx.preference' has no attribute '_head_logit_sums'`, so the tests are falsifiable and the code carries them. 46 of 50 of the preference tests pass; the 4 that fail are a tolerance rather than a defect, and worth a change on your side.

The failures are `test_preference_cce_scores_hidden_states_like_the_logits[dpo-float32-{low,high}-False-softcap]` and the two assertions in `test_preference_logit_sums_keep_float32_precision`, all comparing a chunked float32 reduction against a non chunked one with `atol=1e-3, rtol=0`. Measuring the actual error across vocabulary sizes:

| vocab | max abs sum | max abs error | relative | atol 1e-3 holds |
|---|---|---|---|---|
| 2048 | 63.7 | 0 | 0 | yes |
| 8192 | 69.7 | 9.5e-05 | 1.4e-06 | yes |
| 32768 | 134.7 | 4.4e-04 | 3.3e-06 | yes |
| 131072 | 456.6 | 3.7e-03 | 8.1e-06 | no |

Relative error is pinned between 1e-6 and 8e-6 everywhere, which is float32 epsilon accumulated over thousands of summands, and it stays flat as the row count grows from 4 to 256, so there is no systematic bias. Only the absolute error grows, because the sums themselves grow with the vocabulary. An `rtol=0` tolerance on a quantity whose magnitude scales with V cannot hold as V grows.

More to the point, scoring both arms against a float64 reference shows your path is the more accurate one:

| case | dense path | chunked path (this PR) |
|---|---|---|
| V 2053, softcap 0.5 | 6.06e-07 | 6.06e-07 |
| V 2053, uncapped | 1.24e-06 | 5.03e-07 |
| V 32768, softcap 0.5 | 3.38e-06 | 9.11e-07 |

So neither arm is wrong and yours is closer to the truth; the test is simply asserting an absolute tolerance where a relative one belongs. Worth switching those two assertions to a relative tolerance so the file is not tied to one vocabulary size and one backend's reduction order.

On the Apple Silicon job being red: that failure is not yours. It is `test_vlm_evaluation_compacts_sparse_batches[True]` failing with `assert [1024, 1024] == [256, 256]`, which is a #1265 test that is now on main and that your delta does not touch. The 256 row capacity is admitted only when `_unsloth_cce_small_capacity_limit` clears it, and that value is `budget // (base_chunk * 4)` with `budget` coming from `_get_memory_budget()`, so it depends on how much memory the machine has. It holds on a 128 GB M3 Max and does not on a hosted runner. I am fixing that separately against main.

Codex reviewed the rebased delta on a staging copy and found no issues.

Not proven by me: nothing in this PR was executed on Metal here. The four failures above are characterised rather than reproduced on Apple Silicon, and the real model numbers in your description are not something I could reproduce.

PR #1279, opened 2026-09-18
fix(data): preserve padding-free response-only collators

taking-lying-flat:
With `packing=False` and padding-free batching enabled, `train_on_responses_only` replaces TRL's text collator with `DataCollatorForSeq2Seq`. Two samples of 128 and 32 tokens change from `[1, 160]` to `[2, 128]`, losing `position_ids` and Unsloth's `packed_seq_lengths` wrapper while both padding-free flags remain enabled.

Preserve the original TRL `DataCollatorForLanguageModeling` instance when its `padding_free` mode is active. It already consumes the generated labels, so this retains flattening and the attached metadata wrapper. Scope the exemption to the TRL collator hierarchy; existing processor-backed repairs and replacement of unrelated collators retain their current behavior.

Only `unsloth_zoo/dataset_utils.py` is changed: **8 insertions, 1 deletion**.

Validation:

- **13 local regression checks passed** with real TRL 0.24.0 `SFTTrainer`/collators on CPU: wrapped/unwrapped padding-free batching, existing labels, `last_response_only`, repeated masking, dataloader output, TRL subclasses, stale config flags, and unrelated label-rebuilding collators.
- Related existing tests: **368 passed, 1 skipped** (`test_pretokenized_bypass_collator.py`, `test_train_on_responses_list_labels.py`, `test_vlm_collator_masking.py`, `test_dataset_num_proc.py`).
- Python compilation, the repository's narrow Ruff gate, exec-literal/dynamic-execution lint, and `git diff --check` passed.

Validation scripts remain local; this PR contains only production code. The CPU checks exercise Unsloth's metadata helper and response-only masking, without running the full Unsloth GPU patch stack.


danielhanchen:
Confirmed this hits dataset_utils.py where the collator swap discards the padding-free instance that trainer.py just wrapped with packed_seq_lengths, so the auto-enabled padding-free path quietly stops applying. Could you confirm the TRL collator consumes the labels column written by the masking pass rather than rebuilding them from input_ids, and I will get this reviewed.

taking-lying-flat:
@danielhanchen Confirmed: TRL's text `DataCollatorForLanguageModeling` consumes the `labels` written by the masking pass.

- In [TRL 0.24.0](https://github.com/huggingface/trl/blob/v0.24.0/trl/trainer/sft_trainer.py#L157-L164), `torch_call` reads each example's `labels` when the column is present; it falls back to `input_ids` only when labels are absent.
- [TRL 1.0.0](https://github.com/huggingface/trl/blob/v1.0.0/trl/trainer/sft_trainer.py#L173-L177) does the same with `example.get("labels", example["input_ids"])`.

The padding-free path flattens those labels alongside the inputs. Its sequence-boundary masking and optional completion/assistant masks can add `-100` entries; they do not restore targets that response-only masking excluded.

I also reran the masking function from this PR's head (`97e8a06`) followed by the real TRL collator on CPU: 4 cases on each version, 8 passed. These cover both values of `last_response_only`, a pre-existing masked token inside a response, repeated masking, and additional completion/assistant masks. The collated labels matched the expected masks, with flattened inputs and reset position IDs intact.

So keeping this collator preserves the response-only labels. Thank you for checking this interaction before review.

danielhanchen:
Reviewed on a B200 plus CPU differentials. Verdict:

`issue: REAL | PR as submitted: CORRECT, no changes needed | merge: YES`

**ELI18.** Unsloth turns padding-free batching on by default for `SFTTrainer`, and wraps the TRL collator so every batch also carries `packed_seq_lengths`. `train_on_responses_only` then threw that wrapped collator away and put a plain padding collator in its place, so the padding-free path silently stopped running even though `args.padding_free` stayed `True`. This PR exempts exactly that collator from the swap, and nothing else changes.

### Setup

- base `14db8d46` (main), head `97e8a06`, merge base `a285ff9c`. Head is behind main, but reverting the diff hunk on head leaves `dataset_utils.py` byte-identical to main, so there is no conflict in this file.
- trl 0.25.1, transformers 5.17.0, torch 2.13.0+cu130, datasets 3.6.0, peft 0.21.0. Note this is not the 0.24.0 quoted in the description; 0.25.1 has no `return_position_ids` field, so `enable_padding_free_metadata`'s `hasattr` guard is a no-op there and position ids come out unconditionally under `padding_free`.
- Third arm: head with the hunk reverted, to confirm the hunk is what moves the behaviour.

### Before / after, real captured output

Two chat rows of 147 and 44 tokens, real `SFTTrainer`, real TRL collator, real `enable_padding_free_metadata`, then `train_on_responses_only`, reading the first batch off `get_train_dataloader()`:

| | base (main) | head-with-hunk-reverted | head |
|---|---|---|---|
| collator is the same object | `False` | `False` | `True` |
| collator class | `DataCollatorForSeq2Seq` | `DataCollatorForSeq2Seq` | `DataCollatorForLanguageModeling` |
| `_unsloth_padding_free_lengths_wrapped` | lost | lost | `True` |
| batch keys | `input_ids, attention_mask, labels` | same | `input_ids, labels, position_ids, packed_seq_lengths` |
| `input_ids.shape` | `[2, 147]` | `[2, 147]` | `[1, 191]` |
| sequence starts in `position_ids` | absent | absent | `[0, 147]` |
| `packed_seq_lengths` | absent | absent | `[147, 44]` (int32) |
| supervised tokens | 25 | 25 | 25 |

The supervised `(input_id, label)` pairs are byte-identical between base and head, so response-only masking is preserved exactly; head only adds TRL's own `labels[position_ids == 0] = -100` boundary mask, which is what padding-free already does without `train_on_responses_only`.

### Real GPU run

Llama-3.2-1B-Instruct, bf16 LoRA r=16, alpaca-cleaned rendered through the chat template, `train_on_responses_only`, `max_steps=61, bsz=2, grad_accum=3, seed=3407`. Same 67208 real tokens consumed in every arm. Three repeats for base and head.

| Arm | collator after | first batch | real tok/s (median of 3, range) | peak memory | final train_loss | losses 1,2,3,60,61 | grad norms 1,2,3,60,61 |
|---|---|---|---|---|---|---|---|
| base, `padding_free=True` | `DataCollatorForSeq2Seq` | `[2, 108]`, no `packed_seq_lengths` | 3112 (2958-3142) | 2.9369 GB | 1.233564 | 1.628, 1.573, 1.181, 0.945, 1.228 | 4.633, 2.715, 2.759, 1.271, 0.777 |
| head, `padding_free=True` | `DataCollatorForLanguageModeling` | `[1, 179]`, `packed_seq_lengths=[108, 71]` | 3094 (3084-3138) | 2.8525 GB | 1.233701 | 1.628, 1.572, 1.177, 0.946, 1.226 | 4.633, 2.713, 2.744, 1.272, 0.778 |
| base, `padding_free=False` (control) | `DataCollatorForSeq2Seq` | `[2, 108]` | 2922 | 2.9369 GB | 1.233690 | 1.628, 1.573, 1.179, 0.945, 1.225 | 4.633, 2.714, 2.774, 1.278, 0.775 |
| head, `padding_free=False` (control) | `DataCollatorForSeq2Seq` | `[2, 108]` | 3080 | 2.9369 GB | 1.233639 | 1.628, 1.573, 1.182, 0.947, 1.227 | 4.633, 2.714, 2.796, 1.272, 0.776 |

Losses agree to 1.4e-4 across all four arms and grad norms track to three decimals, so the flattened batches train correctly through the real attention path. Memory drops 2.9 percent. Throughput is flat: at 1B params with roughly 110-token rows there is only about 25 percent padding waste, and the run-to-run spread is wider than the difference, so this run does not show a speedup and I am not claiming one. The point of the fix is that the padding-free contract is applied at all.

### No regression on the other branches

Head arm, same harness:

| Collator handed to `train_on_responses_only` | head result |
|---|---|
| TRL `DataCollatorForLanguageModeling`, `padding_free=True` | preserved |
| ordinary subclass of it | preserved via the MRO walk |
| unrelated class carrying `padding_free = True` | replaced, correctly not exempted |
| transformers `DataCollatorForLanguageModeling` with the flag forced on | replaced, correctly not exempted (wrong module) |
| `packing=True` | unchanged on both arms |
| `padding_free=False` | replaced on both arms, `[2, 67]`, 16 supervised tokens, identical |

`last_response_only` both ways gives 22 and 9 supervised tokens on both arms.

Full suite: `tests/test_pretokenized_bypass_collator.py` is 264 passed / 4 failed both with and without the hunk, and those 4 (`test_a_formatted_numeric_column_still_reaches_masking[torch|numpy]`, `test_a_numpy_formatted_raw_split_reaches_the_tokenizer`, `test_a_numpy_formatted_pretokenized_split_is_masked`) also fail on plain main, so they are pre-existing and unrelated. The six tests that assert the collator IS replaced all still pass. Whole `tests/` run: 9 pre-existing failures on each arm, none shared with this code path.

### What I did not prove

- Only trl 0.25.1 was exercised. The 0.24.0 and 1.0.0 reasoning in the thread is from source reading, not from a run.
- A collator that wraps TRL's by composition rather than subclassing is not exempted and will still be replaced. Acceptable limitation, worth a follow-up only if someone hits it.
- Skipping the swap also skips its column cleanup. In practice TRL's collator reads only `input_ids`, `labels`, `seq_lengths`, `completion_mask` and `assistant_masks` and ignores the rest, so extra `text` / `messages` / string id / ragged `prompt_ids` columns pass through harmlessly on head, while the same fixture raises `Unable to create tensor ... prompt_ids` on base. A column such as `token_type_ids` is dropped from the batch on head, but that is TRL's padding-free behaviour regardless of this PR, not something the PR introduces.
- CUDA only. No ROCm, MLX or CPU-training coverage.

Nothing to change. Good catch on the interaction.

danielhanchen:
Full simulation pass on top of the earlier review. Pushed one commit to this branch: `tests/test_padding_free_collator_survives.py`, a regression test. No production code touched.

`#1279 | issue: REAL | PR as submitted: CORRECT | merge: YES`

**ELI18.** Unsloth switches on padding-free batching by default for SFT, then hangs a wrapper on the batch builder so each batch also reports where one example ends and the next begins. Turning on response-only masking threw that batch builder away and put a plain padding one back, so padding-free silently stopped happening even though everything still reported it as on. Anyone combining response-only masking with the default trainer gets flattened batches and the sequence-length metadata back, at the same loss and about three percent less memory.

### Before / after

| # | what a user gets today | what they get after the PR | on main | on branch | verdict |
|---|---|---|---|---|---|
| 1 | collator replaced: `DataCollatorForSeq2Seq`, batch `[2, 147]` with `attention_mask` | instance survives: `DataCollatorForLanguageModeling`, `[1, 191]`, `position_ids` restarting at `[0, 147]` | reproduced | proven | REAL |
| 2 | `packed_seq_lengths` absent, `_unsloth_padding_free_lengths_wrapped` lost with the instance | `packed_seq_lengths = [147, 44]` int32, wrapper flag still True | reproduced | proven | REAL |
| 3 | response-only supervision: 25 tokens | 25 tokens, byte-identical `(input_id, label)` pairs | unchanged | proven | correctly unchanged |
| 4 | GPU LoRA 61 steps: loss 1.233564, peak 2.9369 GB | loss 1.233701, peak 2.8525 GB | reproduced | proven | REAL |
| 5 | `packing=True` and `padding_free=False`: replaced, `[2, 67]`, 16 supervised | identical on both arms | n/a | proven | correctly unchanged |
| 6 | user supplies their own padding-free TRL collator: silently replaced | preserved | reproduced | proven | REAL, see finding 1 |

### Gate A

1. **Executed-line proof.** `sys.settrace` over `dataset_utils.py` during the repro: lines 2048/2049/2050 hit 3/1/2 times, `_padding_free_collator` bound to `True`. The `padding_free=False` control hits 2048 only (short circuit) and binds `False`. Not a symbol-absent green.
2. **Revert.** Reverting only the hunk leaves `dataset_utils.py` byte-identical to main and reproduces base exactly. The 4 new tests that pass on head fail on that arm.
3. **Prior fix.** One high-confidence match, #836 (merged, same block, different scope). No padding-free exemption in the history of these lines.
4. **Proportionality.** 8 insertions, one predicate, exactly 1 cell moves out of 64.

### Gate B

- **B1 inference: N/A, proven.** `_padding_free_collator` occurs twice, both inside `train_on_responses_only`, nowhere else in either repo. The hunk is in no loop and no forward. Training throughput measured anyway, median of 3: base 3112 real tok/s (2958-3142), head 3094 (3084-3138), flat inside a band wider than the difference; peak memory 2.9369 to 2.8525 GB.
- **B2 GPU detection: N/A.** No `cuda`/`device`/`gpu`/`rocm`/`hip`/`xpu`/`mps`/`capability`/`torch.version` token in the added lines.
- **B3 prompt: N/A.** No `system`/`prompt`/`chat_template`/`jinja`/`apply_chat` token in the added lines.

### The inverse question

`ast` diff: 79 defs on each side, 0 added, 0 removed, exactly 1 changed (`train_on_responses_only`), no nested def inside it changed.

Enumerating the predicate over `hasattr` x `_processor_backed` x `_bypassed_vision_collator` x `isinstance(Seq2Seq)` x `packing_enabled` x `_padding_free_collator`:

```
total cells 64 | identical 63 | differing 1 | violations with F false 0 | new replacements 0
the one differing cell: H=1 P=0 B=0 S=0 K=0 F=1, base replaces, head preserves
reachable subspace (F implies not S, not P, not B): 36 cells | 35 identical | 1 differing
```

Byte-identical routing for every user whose collator is not a padding-free TRL one.

### Tests that can fail

`tests/test_padding_free_collator_survives.py`, 7 cases, now on the branch:

| arm | result |
|---|---|
| head | 7 passed |
| base (main) | 4 failed, 3 passed |
| head with the hunk reverted | 4 failed, 3 passed, identical to base |
| guard widened to any truthy `padding_free` | 2 failed, both scope controls |
| `padding_free` requirement dropped | 1 failed, the mode control |

Every case is falsifiable by a specific mutation; no control passes unconditionally. Built against the real TRL collator, not a stand-in.

### Version corners

Declared range is `trl>=0.18.2,!=0.19.0,<=1.13.0`, read from `pyproject.toml` at the merge base. Static scan of all 39 stable in-range releases: the class sits at `trl.trainer.sft_trainer` in every one, so the module-prefix match never goes stale in range. `padding_free` is absent in 0.18.2 and present from 0.19.1 on.

| trl | base | head |
|---|---|---|
| 0.18.2 (floor) | 2 passed, 5 skipped | 2 passed, 5 skipped, guard inert and fail-safe |
| 0.19.1 | 4 failed, 3 passed | 7 passed |
| 0.24.0 | 4 failed, 3 passed | 7 passed |
| 0.25.1 (installed) | 4 failed, 3 passed | 7 passed |
| 1.13.0 (ceiling, latest) | 4 failed, 3 passed | 7 passed |

The floor is the interesting corner: no padding-free collator exists there, so `getattr(_collator, "padding_free", False)` is False and the guard degrades to the old behaviour rather than breaking. Two signature changes had to be handled in the test to reach the corners at all: `completion_only_loss` is gone by 1.13.0, and position ids moved from the `attention_mask` key to `position_ids` in 0.20.

### Blast radius

One new name, local to the function, two occurrences, zero readers elsewhere. Full suite on identical test trees: head 9 failed / 8283 passed, head-with-hunk-reverted 10 failed / 8284 passed. Failure sets identical except `test_rmsnorm_recompile_guards::test_each_kernel_reports_its_own_fallback_state`, which fails on the **reverted** arm and passes on head; in isolation that file is 26 passed on both, so it is the dynamo-cache ordering flake, not a PR effect. The six tests asserting the collator IS replaced all still pass.

### Cross-platform

Staging run, counts read from the run logs rather than the ticks:

| runner | result |
|---|---|
| ubuntu-latest | 7 passed |
| windows-latest | 7 passed |
| macos-15 (Apple Silicon) | 7 passed |
| ubuntu-24.04-arm | 7 passed |

Wheel, Lint and Tests CI all green; MLX skipped. Staged from fresh upstream main plus this head, no merge conflict.

### Findings, none blocking

1. The guard routes on `collator.padding_free`, not `args.padding_free`. That is reachable: passing your own collator makes `blocked` true in `unsloth/trainer.py`, which sets `args.padding_free` False while your collator keeps its own flag, so head preserves and flattens where base padded. I think that is the more faithful behaviour, since base was silently discarding an explicitly supplied collator, but it is a second behaviour change worth knowing about. Unsloth's own blocklist path is unaffected: `blocked` sets `padding_free` False before TRL builds the collator, and that cell measures identical on both arms.
2. A composition wrapper around TRL's collator has no qualifying ancestor and is still replaced. Known limitation.
3. Skipping the swap also skips its column cleanup, and `enable_padding_free_metadata` forces `remove_unused_columns=False`. TRL's collator reads only `input_ids`, `labels`, `seq_lengths`, `completion_mask`, `assistant_masks`, so extra columns pass through harmlessly on head, while the same fixture raises `Unable to create tensor ... prompt_ids` on base. `token_type_ids` is dropped from the batch on head, but that is TRL's padding-free behaviour regardless of this PR.

### What I could not prove

- The 34 in-range trl releases between the tested corners were checked statically only.
- CUDA only, one B200. No ROCm, Intel, MPS, multi-GPU or CPU training.
- The staging legs ran head only; the base arm is the local one, so there is no base arm on macOS, Windows or ARM64.
- Throughput was measured at 1B params with roughly 110-token rows, where padding waste is about 25 percent. Whether padding-free wins throughput at longer sequences is unmeasured and unclaimed.
- `windows-11-arm`, `macos-26` and Intel macOS did not run.

shimmyshimmer:
Review record for unslothai/unsloth-zoo#1279, carried over from the mirror it was reviewed on: https://github.com/danielhanchen/unsloth-zoo-staging-review/pull/11.

Findings are quoted as posted and attributed to the account that posted them. The reactions shown are the triage recorded on the mirror; none were re-applied here. Commits are named as the mirror's, with this PR's equivalent where one was found.

### Round 1 — reviewed mirror `1ba75d7489` (this PR's `3e00e0163`)

> **chatgpt-codex-connector[bot]** ([on the mirror](https://github.com/danielhanchen/unsloth-zoo-staging-review/pull/11#issuecomment-5760199880))
>
> Codex Review: Didn't find any major issues. Another round soon, please!
>
> **Reviewed commit:** `1ba75d7489`

No triage reaction was recorded on the mirror

### Verdict

The round converged at mirror `1ba75d7489` (this PR's `3e00e0163`).

This is the review as it stood at that commit. Anything pushed to this PR afterwards was not part of it.

PR #1280, opened 2026-09-18
Compile a real cold model at the MoE stash probe, rather than simulating one

shimmyshimmer:
The cold-start tests around the MoE stash probe all reach it by monkeypatching `torch.compiler.is_compiling` to return True while the model runs eagerly. That gets the probe to execute, but `torch.compile` never traces it, so the compiled path the probe exists to serve had no coverage at all.

That gap is not theoretical. Two separate defects have now lived in exactly that spot:

- `id(forward)` on a bound method. The probe reads `experts_module.forward` off a module Dynamo is tracing, and the forwards are installed on the class, so CPython allocates a fresh bound method on every attribute access. Taking `id()` of it made Dynamo install `___check_obj_id(module.forward, <address>)`, which cannot hold by the time it is checked, so Dynamo raised `AssertionError: Guard failed on the same frame it was created` under both fullgraph settings, with no eager fallback.
- `id()` on a code object. The fix for the above was untraceable on some torch versions, which reject it with `Unsupported: id() with unsupported args`.

Both were found by hand. Neither was caught by the suite.

This adds one test that compiles a real model cold, parametrized over both expert families and both `fullgraph` settings. The reference output is built from a separate model so the subject's first call is genuinely the compiled one, and the test asserts no verdict has been recorded before it runs, so an eagerly measured answer cannot stand in for the probe. Both families are covered deliberately: the probe runs before the answer is known, so a guard installed there breaks the supported families too, not only the rerouted ones.

Verified in both directions rather than only the passing one. All 4 cases pass on main. Reintroducing the `{id(forward)}` seed fails 3 of the 4 with the original `Guard failed on the same frame it was created`, so the test pins the defect rather than merely running near it.

Tests only, no behavior change. The full file is 53 passed.


shimmyshimmer:
@codex review

PR #1284, opened 2026-09-18
Pin the CCE chunk budget in the VLM small-capacity evaluation test

danielhanchen:
`test_vlm_evaluation_compacts_sparse_batches[True]` fails on the hosted Apple Silicon runner with `assert [1024, 1024] == [256, 256]`. It is currently red on main.

## Cause

The test asserts that a quantized VLM head projects 256 rows. That 256-row capacity is admitted only when it shares vocabulary chunks with the half capacity:

```python
loss_fn._unsloth_cce_small_capacity_limit = budget // (base_chunk * 4)
```

with `budget = runtime_cce._CHUNK_BUDGET or runtime_cce._get_memory_budget()`, and `_get_memory_budget()` returning 0.1% of the device's recommended working set, floored at 4 MB and capped at 128 MB.

At the test's 2053 vocabulary, `base_chunk` is 2048, so the limit is `budget // 8192`. The capacity under test is 1024, and `_compact_vlm_cce_batch` takes the 256-row branch only when `256 < capacity <= limit`:

| device | budget | limit | 1024 <= limit | result |
| 128 GB M3 Max | 128 MB | 16384 | yes | `[256, 256]`, passes |
| 8 GB M1 (hosted runner) | 6 MB | 768 | no | `[1024, 1024]`, fails |

So the assertion depends on how much memory the runner has, not on the admission rule it is meant to cover. Nothing is wrong with the production behaviour: declining the 256-row capacity on a small device is what the rule is for.

## Change

Pin `_CHUNK_BUDGET` for the duration of the test.

## Validation

On the Linux CPU MLX runtime, which reaches this arithmetic because the limit is computed before any Metal kernel:

```text
before, budget pinned to 6 MB   1 failed, 1 passed   assert [1024, 1024] == [256, 256]
after,  budget pinned to 6 MB   2 passed
after,  budget pinned to 128 MB 2 passed
after,  budget pinned to 4 MB   2 passed
```

The 6 MB figure is what an 8 GB M1 reports, which is the hosted `macos-15` runner.


danielhanchen:
@codex review

danielhanchen:
@codex security review

PR #1285, opened 2026-09-18
Spell the mlx cce imports absolutely

danielhanchen:
## The failure

`tests/test_relative_imports_resolve.py` is red on `main` (`f5254a8e`), and it takes unsloth's `Core zoo` job down with it on every open PR there:

```
mlx/preference.py: transformers would try to open mlx/cce.runtime_cce.py during custom_object_save and fail. Spell these absolutely: ['cce.runtime_cce']
mlx/utils.py: transformers would try to open mlx/cce.py during custom_object_save and fail. Spell these absolutely: ['cce']
```

Reproduced locally on clean `main`: `2 failed, 153 passed`.

## Cause

Two dotted relative imports crept back in:

| File | Line | From |
|---|---|---|
| `unsloth_zoo/mlx/preference.py` | 13 | #1278 |
| `unsloth_zoo/mlx/utils.py` | 4310 | #1265 |

transformers' `dynamic_module_utils` finds imports with a regex and joins the raw capture onto the directory, with nothing converting dots to separators:

```
from .cce.runtime_cce import x  ->  unsloth_zoo/mlx/cce.runtime_cce.py
from .cce import runtime_cce    ->  unsloth_zoo/mlx/cce.py
```

`cce` is a package, so neither file exists, and `custom_object_save` walks this graph recursively: one unresolvable import anywhere crashes the save.

## The fix

Both are spelled the way the rest of the package already spells them. `utils.py:68-69` has carried absolute cce imports all along, so these two were drift from the file's own convention rather than a deliberate choice. Each site gets a short note saying why, since this is the second and third time the relative form has come back.

`from .utils import ...` in `preference.py` is deliberately left alone: a sibling module resolves to a real file, which is exactly the case the guard permits. Rewriting it would be noise.

No new test: the existing guard is what caught this, and it caught it correctly. It already covers both import shapes and checks resolvability rather than dottedness.

## Verification

- `tests/test_relative_imports_resolve.py`: **155 passed** (was `2 failed, 153 passed`).
- Both files compile.

What this host cannot check is the mlx paths themselves, since `mlx` is not installed here and those suites skip. The change is import spelling only, and inside a package the relative and absolute forms name the same module, so the runtime behaviour is unchanged by construction; the MLX CI lane is what confirms it.


danielhanchen:
@codex review

danielhanchen:
@codex security review

PR #1290, opened 2026-09-20
Size TiledMLP tiles over the whole batch, not one sequence

vineethsaivs:
With any batch size above 1 the default arctic tiling stops bounding memory: a batch of 8 by 64 tokens at chunk size 8 is tiled as eight tiles of 8 rows plus one of 448.

`tiled_forward_arctic_size` counts the shards with `divmod(S, chunk_size)`, but `TiledMLP.forward` flattens to `(B*S, hd)` and appends whatever `chunk_size * n_shards` leaves over as one final tile. That leftover is `(B-1)*S` rows, so the peak the tiling exists to cap grows with the batch. The `target_gb` sibling already counts over `bsz*qlen`. Count the shards over `B*S`, the axis the forward actually splits.

Test: `UNSLOTH_ZOO_DISABLE_GPU_INIT=1 python -m pytest tests/test_tiled_mlp_batch_shards.py -q`. Three regressions fail before, at batch sizes 2, 4 and 8; batch size 1 passed both ways, which is why this went unseen. Each asserts no tile exceeds the chunk size, that the tiles cover the flattened batch exactly once, and that the output matches the untiled forward. Eight CPU tests pass after, with the existing target_gb file. The pinned Ruff gate and compileall pass. Executed on Apple M2 Pro CPU; no GPU or distributed run. On a Mac with mlx installed the tiled forward raises before reaching any of this; that is a separate one-line fix I am opening alongside, and it does not affect Linux CI.


danielhanchen:
Confirmed this against tiled_mlp.py, where the arctic path derives n_shards from S but the forward splits the flattened B*S rows, so the trailing tile absorbs the rest of the batch while target_gb already counts over bsz*qlen. Will get this reviewed.

PR #1291, opened 2026-09-20
Give TiledMLP autocast a device type torch understands

vineethsaivs:
On Apple silicon with mlx installed, every tiled MLP forward raises "Expected one of cpu, cuda, ... device type at start of device string: mlx" from inside the autocast decorator.

`tiled_mlp.py` builds its `torch.amp.custom_fwd` and `custom_bwd` decorators from `DEVICE_TYPE`, the package's own label, which is "mlx" there; torch parses that argument as a device on the first decorated call. `DEVICE_TYPE_TORCH` exists for exactly this and maps "mlx" to "mps" and "hip" to "cuda". These two lines are the only place in the package handing a torch API the untranslated label. Use `DEVICE_TYPE_TORCH` for both decorators.

Test: `UNSLOTH_ZOO_DISABLE_GPU_INIT=1 python -m pytest tests/test_tiled_mlp_device_type.py -q`. Two regressions fail before on this machine with the RuntimeError above: one parses the decorator argument as torch does, one pins the source so an mlx or hip host is covered from a Linux runner too. Six CPU tests pass after, including the existing target_gb file. The pinned Ruff gate and compileall pass. Executed on Apple M2 Pro CPU; no GPU or distributed run. Reproduced on Apple silicon with mlx installed; the hip case is by inspection.


danielhanchen:
Confirmed tiled_mlp.py still builds both autocast decorators from DEVICE_TYPE while DEVICE_TYPE_TORCH is what maps mlx to mps and hip to cuda, so this looks right; will get it reviewed.

Issue #1251, opened 2026-09-16
MXFP4 gpt-oss: dead dequantize import, inert env flag, and an unreachable training path


danielhanchen:
Three separate problems in the MXFP4 gpt-oss path, found while investigating whether `openai/gpt-oss-20b` could be LoRA-trained in MXFP4 instead of substituting `unsloth/gpt-oss-20b-unsloth-bnb-4bit`. Filing rather than sending a PR because validating a fix needs an actual MXFP4 model load, and the shared env here cannot do one (transformers 5.17 pins `kernels<0.17`, 0.17.0 is installed), so I could not test a patch.

Environment: transformers 5.17.0, torch 2.13.0+cu130, B200.

### 1. `transformers.integrations.mxfp4.dequantize` no longer exists, and the import is not soft

`unsloth_zoo/temporary_patches/gpt_oss.py:484` and `:518` both do:

```python
from transformers.integrations.mxfp4 import dequantize
```

That symbol was deleted upstream in 5.16 (PR #47579, the DTensor TP rewrite). Confirmed on 5.17.0:

```
transformers 5.17.0 | has dequantize: False
has Mxfp4Dequantize: True
```

The call is inside a `try`, but the `except` re-raises as `RuntimeError("Failed to dequantize MXFP4 gate_up_proj: ...")`, so this is a hard failure at load rather than a graceful fallback.

`unsloth_zoo/temporary_patches/mxfp4.py` already documents this exact deletion around lines 164-171 and handles it for `dequantize_convertops`, so the compat pattern exists; these two sites were missed. The `Mxfp4Dequantize` ConversionOps path is the 5.x equivalent.

### 2. `UNSLOTH_MXFP4_NO_DEQUANTIZE` is dead code

`mxfp4.py:54` `should_dequantize_mxfp4()` and `mxfp4.py:73` `get_mxfp4_config_for_training()` read that env var and build an `Mxfp4Config(dequantize=...)`, but nothing calls `get_mxfp4_config_for_training()`. Its only apparent call site is inside its own docstring, under `Usage:`. So the flag reads like a supported switch and does nothing.

Either wire it up or drop it, but as written it is a trap for anyone who finds it and assumes MXFP4 stays quantized when they set it.

### 3. `Mxfp4GptOssExperts_Training` is unreachable

[...]

PR #1294, opened 2026-09-20
fix(mxfp4): GPT-OSS expert property dequant without removed dequantize

Souravrajvi0:
## Summary

Fixes **#1251 (part 1)** — `Mxfp4GptOssExperts.gate_up_proj` / `down_proj` properties imported `transformers.integrations.mxfp4.dequantize`, which was removed in transformers 5.16 (PR #47579). On 5.16+ that import fails and the `except` re-raises as `RuntimeError`, breaking load when those properties run.

## Changes

- Add `dequantize_mxfp4_moe_blocks_scales()` in `mxfp4.py`, using the same live hooks as the loader path (`dequantize_convertops` when present, else `convert_moe_packed_tensors` + `transpose(1, 2)`).
- Point the GPT-OSS expert property getters at that helper instead of the deleted `dequantize` symbol.
- Regression tests: source guard + numeric parity with patched `dequantize_convertops`.

## Out of scope (follow-up PRs)

- **#1251 §2:** wire `UNSLOTH_MXFP4_NO_DEQUANTIZE` / `get_mxfp4_config_for_training()`
- **#1251 §3:** unreachable `Mxfp4GptOssExperts_Training` vs `forward_mxfp4_gpt_oss_with_lora`

## Testing

- `tests/test_gpt_oss_mxfp4_expert_property_dequantize.py`
- Existing MXFP4 layout tests should remain green (CI)

Full MXFP4 `gpt-oss-20b` load on B200 was not run in this environment; the helper mirrors the already-patched `dequantize_convertops` path documented in `mxfp4.py`.


danielhanchen:
Confirmed the expert getters in unsloth_zoo/temporary_patches/gpt_oss.py still import the module-level dequantize symbol, and the new helper matches the convertops dispatch already in mxfp4.py. Could you note which transformers version you tested the fallback branch against before I review?

Souravrajvi0:
Addressed the review note: added testing notes in `tests/test_gpt_oss_mxfp4_expert_property_dequantize.py` — `dequantize_convertops` parity runs when that symbol exists (transformers 5.0+); the `convert_moe_packed_tensors` fallback is the same path covered by existing MXFP4 layout tests on CPU CI.

Branch rebased on current `main`.

